PR #1, opened 2024-07-04
wigsill examples app & switch to pnpm

iwoplaza:


apps/wigsill-examples/.eslintrc.cjs
iwoplaza, line 1: Generated by `pnpm create vite`, seem to be good defaults.

apps/wigsill-examples/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "dev": "vite",
+    "build": "tsc -b && vite build",
+    "lint": "eslint . --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
+    "preview": "vite preview"
+  },
+  "dependencies": {
+    "jotai": "^2.8.4",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
+    "wigsill": "workspace:*"
```
iwoplaza, line 16: The `workspace:` prefix makes `pnpm` get the module from within the workspace (packages/wigsill in this case), not from the [npm registry](https://npmjs.com). `*` just means any version that exists.

apps/wigsill-examples/src/App.tsx
```diff
@@ -0,0 +1,51 @@
...
+const sampleShaderAtom = atom(async (get) => {
+  const runtime = await get(runtimeAtom);
+
+  const program = new ProgramBuilder(runtime, sampleShader).build({
+    bindingGroup: 0,
+    shaderStage: GPUShaderStage.COMPUTE,
+  });
+
+  return program.code;
+});
```
iwoplaza, line 25: `jotai` allows us to create async (or not) resources that depend on each other, in this case `sampleShaderAtom` depends on `runtimeAtom`. The value of `sampleShaderAtom` will be recalculated each time `runtimeAtom` is changed. `runtimeAtom` has no dependencies, so only computes once.

apps/wigsill-examples/src/App.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+function ShaderCodeView() {
+  const resolvedCode = useAtomValue(sampleShaderAtom);
```
iwoplaza, line 28: Since the atom is asynchronous, React will suspend rendering this component until the promise resolves. It will show the nearest parent <Suspense />'s fallback.

apps/wigsill-examples/src/common/useEvent.ts
iwoplaza, line 1: Just a helper React hook for creating stable event handler functions.

PR #8, opened 2024-07-08
#3 Refactor SimpleWGSLDataType to accept a record of parameters

### LINKED ISSUE #3: Refactor `SimpleWGSLDataType` to accept a record of parameters instead of a list for better readability.
I think passing in a record of parameters will be more readable than listing them out.

This is how they look currently:
```ts
export const vec3f = new SimpleWGSLDataType(
  TB.tupleOf([TB.f32, TB.f32, TB.f32]),
  16,
  'vec3f',
);
```

This is how I think they would look more readable:
```ts
export const vec3f = new SimpleWGSLDataType({
  schema: tupleOf([f32, f32, f32]),
  byteAlignment: 16,
  code: 'vec3f',
});
```

The change from `TB.tupleOf` to `tupleOf` and `TB.f32` to `f32` helps too. To avoid conflicts between built-in schemas of `typed-binary` and our own, we could split the `std140.ts` file into `numeric.ts` (bool, u32, f32, ..., vec2f, ..., mat4f) and house the `struct`, `arrayOf` and `dynamicArrayOf` in the files that define their classes, so `structDataType.ts` (rename to `struct.ts`), `array.ts` (new file with just the function), and `dynamicArraySchema.ts` (rename to `dynamicArray.ts`).

mhawryluk:
closes #3

### DIFF
--- a/packages/wigsill/src/std140/array.ts
+++ b/packages/wigsill/src/std140/array.ts
@@ -1,4 +1,4 @@
-import { arrayOf as tbArrayOf } from 'typed-binary';
+import * as TB from 'typed-binary';
 import { code } from '../wgslCode';
 import { SimpleWGSLDataType } from './std140';
 import { AnyWGSLDataType } from './types';
@@ -8,7 +8,7 @@ export const arrayOf = <TSchema extends AnyWGSLDataType>(
   size: number,
 ) =>
   new SimpleWGSLDataType({
-    schema: tbArrayOf(elementType, size),
+    schema: TB.arrayOf(elementType, size),
     byteAlignment: elementType.byteAlignment,
     code: code`array<${elementType}, ${size}>`,
   });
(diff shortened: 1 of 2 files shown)


packages/wigsill/src/std140/array.ts
```diff
@@ -0,0 +1,14 @@
+import { arrayOf as tbArrayOf } from 'typed-binary';
```
mhawryluk: is there a way to avoid this aliasing?

iwoplaza: I think we have to live with it ;-;

packages/wigsill/src/std140/struct.ts
```diff
@@ -74,5 +74,8 @@ class StructDataType<TProps extends Record<string, AnyWGSLDataType>>
     return ctx.resolve(this._identifier);
   }
 }
+export const struct = <P extends Record<string, AnyWGSLDataType>>(
```
iwoplaza: A style preference to keep at least one empty line between classes/functions.

```suggestion
}

export const struct = <P extends Record<string, AnyWGSLDataType>>(
```

PR #9, opened 2024-07-08
Running tests as part of CI

iwoplaza:


package.json
```diff
@@ -8,7 +8,9 @@
     "build": "tsup",
     "prettier": "prettier '*.{js,json,md}' '{src,tests,benchmarks,docs}/**/*.{ts,tsx,md,mdx}' --write",
     "eslint": "eslint --fix --no-eslintrc --c .eslintrc.json '*.{js,json,ts}' '{packages/apps}/**/*.{ts,tsx}'",
-    "test": "vitest --ui --coverage"
+    "test": "vitest",
+    "coverage": "vitest coverage",
```
mhawryluk: seems not to work. i think that should be `vitest --coverage`

iwoplaza: Good catch! Changing now.

PR #10, opened 2024-07-08
#5 add a repeat macro to wigsill

### LINKED ISSUE #5: Implement a `repeat` macro that executes a statement N times.
### Example usage 1:

```ts
const program = wgsl.fn`() -> {
  var value: u32 = 1;
  ${repeat('value = value * 2;', 5)}
}`;
```

Resulting code:

```ts
fn item_1() -> {
  var value: u32 = 1;
  value = value * 2;
  value = value * 2;
  value = value * 2;
  value = value * 2;
  value = value * 2;
}
```

### Example usage 2:

```ts
const program = wgsl.fn`() -> {
  var value: u32 = 1;
  ${repeat((idx) => `value += value * ${idx + 1};`, 5)}
}`;
```

Resulting code:

```ts
fn item_1() -> {
  var value: u32 = 1;
  value += value * 1;
  value += value * 2;
  value += value * 3;
  value += value * 4;
  value += value * 5;
}
```

reczkok:
closes #5

### DIFF
--- a/packages/wigsill/tests/macro.test.ts
+++ b/packages/wigsill/tests/macro.test.ts
@@ -3,12 +3,12 @@ import { repeat, wgsl } from 'wigsill';
 
 describe('repeat', () => {
   it('repeats a string', () => {
-    const result = repeat('a', 3);
+    const result = repeat(3, 'a');
     expect(result).toEqual(wgsl`${['a', 'a', 'a']}`);
   });
 
   it('repeats a computed string', () => {
-    const result = repeat((idx) => `a${idx}`, 3);
+    const result = repeat(3, (idx) => `a${idx}`);
     expect(result).toEqual(wgsl`${['a0', 'a1', 'a2']}`);
   });
 });


packages/wigsill/tests/macro.test.ts
```diff
@@ -0,0 +1,14 @@
+import { describe, expect, it } from 'vitest';
+import { repeat, wgsl } from 'wigsill';
+
+describe('repeat', () => {
+  it('repeats a string', () => {
+    const result = repeat('a', 3);
+    expect(result).toEqual(wgsl`${['a', 'a', 'a']}`);
+  });
+
+  it('repeats a string with a function', () => {
```
iwoplaza: Small nitpick, imo might be more descriptive.

```suggestion
  it('repeats a computed string', () => {
```

reczkok: It looks much better - 'string with a function' makes no sense

PR #12, opened 2024-07-08
Examples can be made as React components, and have exposed parameters through dat.gui

iwoplaza:
#6

### DIFF
--- a/apps/wigsill-examples/src/App.tsx
+++ b/apps/wigsill-examples/src/App.tsx
@@ -32,7 +32,7 @@ function App() {
         <aside className="px-4 min-w-64">
           <header className="pb-6">
             <h1
-              className="mx-auto px-4 text-2xl py-4 font-outfit"
+              className="mx-auto p-4 text-2xl font-outfit cursor-pointer"
               onClick={() => setCurrentExample(RESET)}>
               <strong>wigsill</strong> - examples
             </h1>
--- a/apps/wigsill-examples/src/common/ExampleLink.tsx
+++ b/apps/wigsill-examples/src/common/ExampleLink.tsx
@@ -24,7 +24,7 @@ export function ExampleLink(props: {
   return (
     <a
       // Even though we prevent the default behavior of this link
-      // or is good to have this set semantically.
+      // it is good to have this set semantically.
       href={`#example=${exampleKey}`}
       onClick={handleClick}
       className={cs(


apps/wigsill-examples/src/App.tsx
```diff
@@ -1,50 +1,57 @@
...
-      <Suspense fallback={'Loading...'}>
-        <ShaderCodeView />
-      </Suspense>
-    </main>
+    <>
+      <div className="flex h-screen">
+        <aside className="px-4 min-w-64">
+          <header className="pb-6">
+            <h1
+              className="mx-auto px-4 text-2xl py-4 font-outfit"
```
mhawryluk: +`cursor-pointer`
`px-4 py-4` => `p-4`

apps/wigsill-examples/src/common/ExampleLink.tsx
```diff
@@ -0,0 +1,39 @@
...
+    e.preventDefault();
+    setCurrentExample(exampleKey ?? RESET);
+  });
+
+  const active = currentExample === exampleKey;
+
+  return (
+    <a
+      // Even though we prevent the default behavior of this link
+      // or is good to have this set semantically.
```
mhawryluk:
```suggestion
      // it is good to have this set semantically.
```

apps/wigsill-examples/src/common/useExampleWithCanvas.ts
```diff
@@ -0,0 +1,44 @@
...
+
+    return () => {
+      exampleRef.current?.dispose();
+      gui.destroy();
+      cancelled = true;
+    };
+  }, [initExampleFn]);
+
+  return {
+    canvasRef,
```
mhawryluk: why are we returning an object here, instead of just the value? is that a convention?

apps/wigsill-examples/src/App.tsx
iwoplaza:
```suggestion
              className="mx-auto p-4 text-2xl font-outfit cursor-pointer"
```

apps/wigsill-examples/src/common/useExampleWithCanvas.ts
iwoplaza: It was to be more explicit about what we are returning, but I guess if its a single value, it should be fine and more readable 👍.

PR #37, opened 2024-07-11
Custom WGSL parser.

iwoplaza:
Implements #35

### DIFF
--- a/packages/wigsill-parser/tests/floatLiteral.spec.ts
+++ b/packages/wigsill-parser/tests/floatLiteral.spec.ts
@@ -22,10 +22,10 @@ describe('float_literal', () => {
   ];
 
   it('parses float', () => {
-    for (const example_float of EXAMPLE_FLOAT_LITERALS) {
-      expect(parse(example_float)).toEqual({
+    for (const exampleFloat of EXAMPLE_FLOAT_LITERALS) {
+      expect(parse(exampleFloat)).toEqual({
         type: 'float_literal',
-        value: example_float,
+        value: exampleFloat,
       } satisfies FloatLiteral);
     }
   });
--- a/packages/wigsill-parser/tests/functionDecl.spec.ts
+++ b/packages/wigsill-parser/tests/functionDecl.spec.ts
@@ -13,11 +13,11 @@ describe('function_decl', () => {
       type: 'translation_unit',
       declarations: [
         {
-          type: 'function_decl' as const,
+          type: 'function_decl',
           attrs: [],
           header: {
-            type: 'function_header' as const,
-            identifier: 'example' as const,
+            type: 'function_header',
+            identifier: 'example',
           },
           body: [],
         },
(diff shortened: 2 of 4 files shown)


packages/wigsill-parser/tests/floatLiteral.spec.ts
```diff
@@ -0,0 +1,32 @@
...
+    '0xa.fp+2',
+    '0x1P+4f',
+    '0X.3',
+    '0x3p+2h',
+    '0X1.fp-4',
+    '0x3.2p+2h',
+  ];
+
+  it('parses float', () => {
+    for (const example_float of EXAMPLE_FLOAT_LITERALS) {
```
mhawryluk: -> exampleFloat

packages/wigsill-parser/tests/functionDecl.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+    const code = `
+      fn example() {
+      }
+    `;
+
+    const expected = {
+      type: 'translation_unit',
+      declarations: [
+        {
+          type: 'function_decl' as const,
```
mhawryluk: are all these `as const` necessary? it seems to work the same without them

packages/wigsill/tests/function.test.ts
```diff
@@ -0,0 +1,102 @@
+import { parse } from '@wigsill/parser';
+import { describe, expect, it } from 'vitest';
+import { StrictNameRegistry, type WGSLSegment, wgsl } from '../src';
+import { ResolutionCtxImpl } from '../src/programBuilder';
+
+function parseWGSL(segment: WGSLSegment) {
```
mhawryluk: this could be moved to a separate commons file, as it will be used by multiple test files

packages/wigsill/tests/function.test.ts
```diff
@@ -0,0 +1,102 @@
...
+  const ctx = new ResolutionCtxImpl({ names: new StrictNameRegistry() });
+
+  const resolved = ctx.resolve(segment);
+
+  const resolvedWithDependencies = `${ctx.dependencies.map((d) => ctx.resolve(d)).join('\n')}\n${resolved}`;
```
mhawryluk: ctx could have a resolveWithDependencies method, that way we wouldn't need to repeat this line's logic in multiple places

packages/wigsill-parser/tests/functionDecl.spec.ts
reczkok: Isn't the type of `type` 'string' then? I think having the type as `'function_decl'` is beneficial

iwoplaza: If it works without it, I will remove it. I think it was left behind from a time where I was testing if the types match with a different mechanism from `satisfies`;

packages/wigsill/tests/function.test.ts
iwoplaza: This should be solved anyways with the upcoming granular bindings rework 👍

PR #40, opened 2024-07-12
Create StrictNameRegistry for testing purposes

### LINKED ISSUE #17: Add an ability to inject a custom identifier generator into ResolutionCtx, one that generates predictable names for unit tests.
Right now, `NameRegistry` considers only one implementation and is created inside `ResolutionCtx` by default. I would propose renaming the current NameRegistry to `RandomNameRegistry` and creating an interface `NameRegistry` that exposes the method `nameFor`. Then, create a `StrictNameRegistry` that returns an item's `debugLabel` as the name, and throws an error if there is none.

mhawryluk:
closes #17

### DIFF
--- a/packages/wigsill/src/programBuilder.ts
+++ b/packages/wigsill/src/programBuilder.ts
@@ -4,11 +4,7 @@ import {
   NotAllocatedMemoryError,
 } from './errors';
 import { MemoryArena } from './memoryArena';
-import {
-  NameRegistry,
-  RandomNameRegistry,
-  StrictNameRegistry,
-} from './nameRegistry';
+import { NameRegistry, RandomNameRegistry } from './nameRegistry';
 import {
   IResolutionCtx,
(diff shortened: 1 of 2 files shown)


packages/wigsill/src/programBuilder.ts
```diff
@@ -151,6 +132,7 @@ type BuildOptions = {
   shaderStage: number;
   bindingGroup: number;
   arenas?: MemoryArena[];
+  nameRegistryType?: 'random' | 'strict';
```
iwoplaza: In order to allow for custom name registries, in case somebody needs them (like migrating to wigsill from another solution, or integrating with another solution), I would propose optionally passing the NameRegistry implementation as a parameter, instead of the type of registry we expect. Then, if not provided, default to `RandomNameRegistry`. In this case the implementation would be passed in through `BuildOptions.

PR #41, opened 2024-07-15
Make the WGSLRuntime request a device, instead of passing a device to it.

### LINKED ISSUE #7: Make the WGSLRuntime request a device, instead of passing a device to it.
From this:
```ts
const adapter = await navigator.gpu.requestAdapter(adapterOptions);
const device = await adapter!.requestDevice(deviceOptions);

const runtime = new WGSLRuntime(device);
```

To this:
```ts
const runtime = await createRuntime({
  adapter: adapterOptions, // can be specified if necessary.
  device: deviceOptions, // preexisting device, or options to request a new one. Requests one by default.
});
```

In most cases, users will just call:
```ts
const runtime = await createRuntime();
```

reczkok:
Closes #7

### DIFF
--- a/apps/wigsill-examples/examples/gradient-tiles.js
+++ b/apps/wigsill-examples/examples/gradient-tiles.js
@@ -12,7 +12,7 @@ import {
   onFrame,
 } from '@wigsill/example-toolkit';
 
-const runtime = createRuntime;
+const runtime = await createRuntime();
 const device = runtime.device;
 
 const xSpanData = wgsl.memory(u32).alias('x-span');


packages/wigsill/src/wgslRuntime.ts
```diff
@@ -55,4 +55,37 @@ class WGSLRuntime {
...
+    | {
+        adapter: GPURequestAdapterOptions | undefined;
+        device: GPUDeviceDescriptor | undefined;
+      }
+    | GPUAdapter
+    | GPUDevice,
+) {
+  let adapter: GPUAdapter | null = null;
+  let device: GPUDevice | null = null;
+
```
iwoplaza, line 68: To avoid assertions about the existence of `navigator.gpu`, we can throw an error early.

```suggestion

if (!navigator.gpu) {
  throw new Error('WebGPU is not supported by this browser.');
}

```

packages/wigsill/src/wgslRuntime.ts
```diff
@@ -55,4 +55,37 @@ class WGSLRuntime {
...
+  if (!options) {
+    adapter = await navigator.gpu.requestAdapter();
+    device = await adapter!.requestDevice();
```
iwoplaza: Lets throw a descriptive error in case a compatible adapter could not be found.

```suggestion
    if (!adapter) {
      throw new Error('Could not find a compatible GPU');
    }
    device = await adapter.requestDevice();
```

packages/wigsill/src/wgslRuntime.ts
```diff
@@ -55,4 +55,37 @@ class WGSLRuntime {
...
+    device = await adapter.requestDevice();
+    return new WGSLRuntime(device);
+  }
+
+  if (options instanceof GPUDevice) {
+    return new WGSLRuntime(options);
+  }
+
+  adapter = await navigator.gpu.requestAdapter(options.adapter);
+  device = await adapter!.requestDevice(options.device);
```
iwoplaza: Same down here.

```suggestion
  if (!adapter) {
    throw new Error('Could not find a compatible GPU');
  }
  device = await adapter.requestDevice(options.device);
```

PR #42, opened 2024-07-15
Add fluid simulation example

reczkok:


apps/wigsill-examples/examples/fluid-simulation.js
```diff
@@ -0,0 +1,713 @@
...
+        break;
+      case 'drain':
+        drawType = 3 << 24;
+        break;
+      default:
+        drawType = 100;
+        break;
+    }
+  },
+);
```
iwoplaza: This might be more efficient in the long run, but I would be tempted to store `brushType` instead of `drawType`, put it in `Options.brushType` like the other options, and compute `drawType` on each `mousemove`.

```suggestion

const BrushTypes = ['wall', 'source', 'drain', 'water'];

function encodeBrushType(brushType) {
  switch (brushType) {
    case 'wall':
      return 1 << 24;
    case 'source':
      return 2 << 24;
    case 'drain':
      return 3 << 24;
    default:
      return 100;
  }
}

addParameter(
  'brushType',
  { initial: 'water', options: BrushTypes },
  (value) => {
    Options.brushType = value;
  },
);

```

apps/wigsill-examples/examples/fluid-simulation.js
```diff
@@ -0,0 +1,713 @@
...
+      }
+    }
+
+    if (isErasing) {
+      for (const cell of allAffectedCells) {
+        drawCanvasData[cell.y * Options.size + cell.x] = 4 << 24;
+      }
+    } else {
+      for (const cell of allAffectedCells) {
+        drawCanvasData[cell.y * Options.size + cell.x] = drawType;
```
iwoplaza:
```suggestion
        drawCanvasData[cell.y * Options.size + cell.x] = encodeBrushType(Options.brushType);
```

apps/wigsill-examples/examples/fluid-simulation.js
reczkok: Looks much cleaner - hated that floating `let`

PR #46, opened 2024-07-15
Typescript support in example code editor

### LINKED ISSUE #33: TypeScript support for examples.
### Transforming the source code before executing in `exampleRunner.ts`

TypeScript can be compiled into JavaScript via the `typescript` library. According to ChatGPT, you can do:

```ts
import * as ts from 'typescript';

const tsCode = `
    import { something } from "some-module";
    const x: number = 10;
`;

const result = ts.transpileModule(tsCode, {
  compilerOptions: { module: ts.ModuleKind.ESNext }
});

console.log(result.outputText);
```

### Changing preferred language in Monaco

The editor language is currently set to `javascript`, should work if we plug in `typescript` instead in `CodeEditor.tsx`.

### Creating a custom `.d.ts` file to declare the fake `@wigsill/example-toolkit` module.

mhawryluk:
closes #33

### DIFF
--- a/apps/wigsill-examples/src/CodeEditor.tsx
+++ b/apps/wigsill-examples/src/CodeEditor.tsx
@@ -1,10 +1,10 @@
 import Editor, { Monaco } from '@monaco-editor/react';
 import webgpuTypes from '@webgpu/types/dist/index.d.ts?raw';
-import wigsill from './types/wigsill-types.d.ts?raw';
-import toolkitTypes from './types/example-toolkit.d.ts?raw';
 import typedBinary from 'typed-binary/dist/index.d.ts?raw';
 import useEvent from './common/useEvent';
 import { tsCompilerOptions } from './embeddedTypeScript';
+import toolkitTypes from './types/example-toolkit.d.ts?raw';
+import wigsill from './types/wigsill-types.d.ts?raw';
 
 function handleEditorWillMount(monaco: Monaco) {
   const tsDefaults = monaco?.languages.typescript.typescriptDefaults;
--- a/apps/wigsill-examples/src/example/exampleRunner.ts
+++ b/apps/wigsill-examples/src/example/exampleRunner.ts
@@ -3,10 +3,10 @@ import type TemplateGenerator from '@babel/template';
 import type { TraverseOptions } from '@babel/traverse';
 import { GUI } from 'dat.gui';
 import { filter, isNonNull, map, pipe } from 'remeda';
-import { ExampleState } from './exampleState';
-import { LayoutInstance } from './layout';
 import { transpileModule } from 'typescript';
 import { tsCompilerOptions } from '../embeddedTypeScript';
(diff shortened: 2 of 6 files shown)


pnpm-lock.yaml
mhawryluk, line 1: not sure what happened here. I can revert this change and install tsx once again if it's necessary

apps/wigsill-examples/src/CodeEditor.tsx
```diff
@@ -10,11 +15,40 @@ type Props = {
 export function CodeEditor(props: Props) {
   const { code, onCodeChange } = props;
 
+  function handleEditorWillMount(monaco: Monaco) {
```
iwoplaza: Since this function does not use anything inside the `CodeEditor` component, I would move it outside into the global scope to avoid recreating it on every render.

apps/wigsill-examples/src/CodeEditor.tsx
```diff
@@ -10,11 +15,40 @@ type Props = {
...
+      strict: true,
+      esModuleInterop: true,
+      module: languages.typescript.ModuleKind.ESNext,
+      noUncheckedIndexedAccess: true,
+      noEmit: true,
+      skipLibCheck: true,
+      exactOptionalPropertyTypes: true,
+      baseUrl: '.',
+      lib: ['dom', 'es2021'],
+      typeRoots: ['./node_modules/@webgpu/types', './node_modules/@types'],
```
iwoplaza: Since we inject the `typeRoots` with `addExtraLib`, this property seems to be redundant.

apps/wigsill-examples/src/example/exampleRunner.ts
```diff
@@ -55,6 +58,12 @@ const staticToDynamicImports = {
   } satisfies TraverseOptions,
 };
 
+function tsToJs(code: string): string {
+  return ts.transpileModule(code, {
+    compilerOptions: { module: ts.ModuleKind.ESNext },
```
iwoplaza: Would be nice to compile with the same compilerOptions that are provided to the monaco editor. Perhaps we could extract those options from `CodeEditor.tsx` and put them in like a `src/embeddedTypeScript.ts` file?

apps/wigsill-examples/src/example/layout.ts
```diff
@@ -1,34 +1,13 @@
...
-  type: ElementType,
-  options: ElementOptions,
-) => Promise<HTMLElement>;
+import {
+  AddElement,
+  ElementDef,
+  ElementOptions,
+  ElementType,
+  LayoutDef,
+} from '@wigsill/example-toolkit';
```
iwoplaza: Tiny nitpick, we could emphasize that we only care about the types from that module.

```suggestion
import type {
  AddElement,
  ElementDef,
  ElementOptions,
  ElementType,
  LayoutDef,
} from '@wigsill/example-toolkit';
```

PR #55, opened 2024-07-17
Granular bindings

### LINKED ISSUE #15: Instead of providing bindables only on the top level, allow for more granular bindings.
### Utility module code:

```ts
const colorSlot = wgsl.slot<WGSLSegment>();

const render = wgsl.fn`() -> {
  let color = ${colorSlot};
  // ...
}`;
```

### User code:

```ts
const mainCode = wgsl.fn`() -> {
  ${render}();
}`.with(colorSlot, 'vec3f(0., 1., 0.)');

```

or:

```ts
const renderWithColor = render.with(colorSlot, 'vec3f(0., 1., 0.)');

const mainCode = wgsl.fn`() -> {
  ${renderWithColor}();
}`;

```

or:

```ts
const renderWithRed = render.with(colorSlot, 'vec3f(1., 0., 0.)');
const renderWithGreen = render.with(colorSlot, 'vec3f(0., 1., 0.)');

const mainCode = wgsl.fn`() -> {
  ${renderWithRed}();
  ${renderWithGreen}();
}`;

```

# Dependency graph and resolution

In the following example, we use the variable `itemsRendered` in the `render` function. The question is, do we create
a unique variable per scope (in this case per unique value of `colorSlot`), or is the use of `itemsRendered` global across the program by default? With the latter approach, I think we can have greater flexibility.

### Utility module code:

```ts
const colorSlot = wgsl.slot<WGSLSegment>();

const itemsRendered = wgsl.var(u32, 0);

const render = wgsl.fn`() -> {
  ${itemsRendered} += 1;

  let color = ${colorSlot};
  // ...
}`;
```

If we wanted to have a unique variable per `colorSlot` value, we could do something like this:

```ts
const colorSlot = wgsl.slot<WGSLSegment>();

const itemsRendered = wgsl.var(u32, 0);
const colorRenderedVars = (() => {
  const colorToVarMap = new Map<string, WGSLVariable>();

  return (color: string) => {
    if (colorToVarMap.has(color)) {
      return colorToVarMap.get(color);
    }
    
    const variable = wgsl.var(u32, 0);
    colorToVarMap.set(color, variable);
    return variable;
  };
})();

const render = wgsl.fn`() -> {
  ${itemsRendered} += 1;
  ${(get) => colorRenderedVars(get(colorSlot))} += 1;

  let color = ${colorSlot};
  // ...
}`;
```

Or if we wanted to create a general utility out of it:

```ts
const colorSlot = wgsl.slot<WGSLSegment>();

type Potential<T> = T | Slot<T>;

const itemsRendered = wgsl.var(u32, 0);
const colorRenderedVars = wgsl.family((color: string) => wgsl.var(u32, 0));

const render = wgsl.fn`() -> {
  ${itemsRendered} += 1;
  // If we pass a slot into the function, the utility can return
  // another function in to compute the value.
  ${colorRenderedVars(colorSlot)} += 1;

  let color = ${colorSlot};
  // ...
}`;
```

iwoplaza:
Resolves #15 and #56

### DIFF
--- a/packages/wigsill/src/resolutionCtx.ts
+++ b/packages/wigsill/src/resolutionCtx.ts
@@ -5,7 +5,7 @@ import {
   type ResolutionCtx,
   type SlotValuePair,
   type Wgsl,
-  type WgslBufferBindable,
+  type WgslBindable,
   type WgslResolvable,
   type WgslSlot,
   isResolvable,
@@ -35,7 +35,7 @@ class SharedResolutionState {
   >();
 
   private _nextFreeBindingIdx = 0;
-  private readonly _usedBindables = new Set<WgslBufferBindable>();
+  private readonly _usedBindables = new Set<WgslBindable>();
   private readonly _declarations: string[] = [];
 
   constructor(
@@ -43,7 +43,7 @@ class SharedResolutionState {
     private readonly _bindingGroup: number,
   ) {}
 
-  get usedBindables(): Iterable<WgslBufferBindable> {
+  get usedBindables(): Iterable<WgslBindable> {
     return this._usedBindables;
   }
 
@@ -87,7 +87,7 @@ class SharedResolutionState {
     return result;
   }
 
-  reserveBindingEntry(_bindable: WgslBufferBindable) {
+  reserveBindingEntry(_bindable: WgslBindable) {
     this._usedBindables.add(_bindable);
 
     return { group: this._bindingGroup, idx: this._nextFreeBindingIdx++ };
@@ -115,7 +115,7 @@ export class ResolutionCtxImpl implements ResolutionCtx {
     throw new Error('Call ctx.resolve(item) instead of item.resolve(ctx)');
   }
 
-  addBinding(_bindable: WgslBufferBindable, _identifier: WgslIdentifier): void {
+  addBinding(_bindable: WgslBindable, _identifier: WgslIdentifier): void {
     throw new Error('Call ctx.resolve(item) instead of item.resolve(ctx)');
   }
 
@@ -160,7 +160,7 @@ class ScopedResolutionCtx implements ResolutionCtx {
     this._shared.addDeclaration(this.resolve(declaration));
   }
 
-  addBinding(bindable: WgslBufferBindable, identifier: WgslIdentifier): void {
+  addBinding(bindable: WgslBindable, identifier: WgslIdentifier): void {
     const { group, idx } = this._shared.reserveBindingEntry(bindable);
 
     this.addDeclaration(
--- a/packages/wigsill/src/wgslConstant.ts
+++ b/packages/wigsill/src/wgslConstant.ts
@@ -23,17 +23,21 @@ export function constant(expr: Wgsl): WgslConst {
 // --------------
 
 class WgslConstImpl implements WgslConst {
-  public label?: string | undefined;
+  private _label: string | undefined;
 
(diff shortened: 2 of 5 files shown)


packages/wigsill/tests/slot.test.ts
```diff
@@ -74,4 +97,154 @@ describe('wgsl.slot', () => {
...
+    const wrapperFn = wgsl.fn('wrapper')`() {
+      return ${getColorWithGreen}();
+    }`;
+
+    const actual = wgsl`
+      fn main() {
+        ${getColorWithRed}();
+        ${wrapperFn}();
+      }
+    `;
```
reczkok: To actually check if it prefers closer bindings, shouldn't we pass `getColorWithRed` into `wrapperFn` and then call `.with(colorSlot, Green)` on it?

iwoplaza: It sure should, good catch! Will change it to be correct.

packages/wigsill/src/resolutionCtx.ts
```diff
@@ -0,0 +1,206 @@
...
+}
+
+export class ResolutionCtxImpl implements ResolutionCtx {
+  private readonly _shared: SharedResolutionState;
+
+  usedSlots = new Set<WgslSlot<unknown>>();
+
+  /**
+   * @throws {MemoryArenaConflict}
+   */
```
reczkok:
```suggestion
```
Arenas are gone🎉

packages/wigsill/src/resolutionCtx.ts
```diff
@@ -0,0 +1,204 @@
...
+import type { WgslIdentifier } from './wgslIdentifier';
+
+export type ResolutionCtxImplOptions = {
+  readonly names: NameRegistry;
+  readonly bindingGroup?: number;
+};
+
+type SlotToValueMap = Map<WgslSlot<unknown>, unknown>;
+
+const usageToVarTemplateMap: Record<BufferUsage, string> = {
```
mhawryluk, line 23: either import it from `programBuilder.ts` or remove it there. not sure which option I prefer, so I leave it up to you

packages/wigsill/src/resolutionCtx.ts
```diff
@@ -0,0 +1,204 @@
...
+  private _nextFreeBindingIdx = 0;
+  private readonly _usedBindables = new Set<WgslBufferBindable>();
+  private readonly _declarations: string[] = [];
+
+  constructor(
+    public readonly names: NameRegistry,
+    private readonly _bindingGroup: number,
+  ) {}
+
+  get usedBindables(): Iterable<WgslBufferBindable> {
```
mhawryluk: I believe `WgslBufferBindable` was meant to be renamed to `WgslBindable`

packages/wigsill/src/wgslConstant.ts
```diff
@@ -23,20 +23,20 @@ export function constant(expr: Wgsl): WgslConst {
 // --------------
 
 class WgslConstImpl implements WgslConst {
-  public debugLabel?: string | undefined;
-  public identifier = new WgslIdentifier();
+  public label?: string | undefined;
```
mhawryluk: why public here? we have label as either private _label, readonly or public in different places, is there some kind of logic behind that?

iwoplaza: No particular reason, changing it now!

PR #62, opened 2024-07-18
Rename alias to name and fix typing in fluid sim

### LINKED ISSUE #49: Fix typing in the fluid simulation example

### LINKED ISSUE #61: Rename all uses of `.alias` to `.$name`
## Why?

After internal discussion with @reczkok and @mhawryluk, we think it would be best to distinguish between methods that are part of the builder, and other methods.

### Builder methods
To create a terse and unified way of adding optional properties onto WGSL items, additional methods can be called on that item in chain. For example:

```ts
const timeBuffer = wgsl.buffer(f32)
  .$name('time')
  .$allowUniform();
```

These methods mutate the original object, so should be limited to use only during the item's declaration.

There are however other methods on these items, and we want to distinguish them from the builder methods, for example:

```ts
const timeBuffer = wgsl.buffer(f32)
  .$name('time')
  .$allowUniform();

// returns a resolvable item to be embedded in the program, not mutating the original item.
const timeUniform = timeBuffer.asUniform();
```

The `$builder` naming scheme allows users of the library to distinguish between these two behaviors.

reczkok:
Resolves #61. Also should resolve #49? I think adding the overloads to the toolkit is enough?

### DIFF
--- a/apps/wigsill-examples/src/types/example-toolkit.d.ts
+++ b/apps/wigsill-examples/src/types/example-toolkit.d.ts
@@ -34,7 +34,7 @@ declare module '@wigsill/example-toolkit' {
 
   export const addElement: AddElement;
 
-  function addParameter(
+  export function addParameter(
     label: string,
     options: {
       initial: number;
@@ -44,7 +44,7 @@ declare module '@wigsill/example-toolkit' {
     },
     onChange: (newValue: number) => void,
   ): void;
-  function addParameter<TOption extends string | number>(
+  export function addParameter<TOption extends string | number>(
     label: string,
     options: {
       initial: TOption;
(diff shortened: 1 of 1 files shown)


apps/wigsill-examples/src/types/example-toolkit.d.ts
```diff
@@ -44,6 +44,29 @@ declare module '@wigsill/example-toolkit' {
...
     onChange: (newValue: number) => void,
   ): void;
+  export function addParameter(
+    label: string,
+    options: {
+      initial: string;
+      options: string[];
+    },
+    onChange: (newValue: string) => void,
+  ): void;
```
iwoplaza: We could try making the addParameter function infer the option type. Because of this, values passed into onChange are typed properly and we can depend on them.

```suggestion
  export function addParameter<TOption extends string>(
    label: string,
    options: {
      initial: TOption;
      options: TOption[];
    },
    onChange: (newValue: TOption) => void,
  ): void;
```

reczkok: Great idea - implemented!

PR #77, opened 2024-07-22
Added missing params in RenderPipelineExecutor for passEncoder.draw()

reczkok:


packages/wigsill/src/wigsillRuntime.ts
```diff
@@ -299,7 +302,12 @@ class RenderPipelineExecutor extends PipelineExecutor<GPURenderPipeline> {
       this.program.bindGroup,
     );
 
-    passEncoder.draw(vertexCount);
+    passEncoder.draw(
+      vertexCount,
+      options.instanceCount,
```
mhawryluk: add these 3 new parameters to object deconstructing and use them without `options.` same as `vertexCount`

PR #84, opened 2024-07-24
Create voxel rendering example based on raytracing

### LINKED ISSUE #65: Add a 3d voxel renderer example
- Camera rotating automatically around subject, we can zoom in and out.
- Subject is a static model made out of voxels, multi-color (one color per voxel)
- Lambert lighting model (only diffuse and ambient)

mhawryluk:
the current implementation suffers quite low performance. the straight-forward raytracing is to be replaced with another, more efficient, approach in the following PRs

+ fix creating buffers for structs concerning byteAlignment (kudos to @reczkok for finding the fix)
+ separate shader modules for vertex and fragment in `makeRenderPipeline`
+ make example page refresh on example source change
+ allow passing vertex buffers to `makeRenderPipeline`

closes #65

### DIFF
--- a/apps/wigsill-examples/src/examples/voxel-rendering.ts
+++ b/apps/wigsill-examples/src/examples/voxel-rendering.ts
@@ -55,7 +55,7 @@ const cubeSize = [X * VOXEL_SIZE, Y * VOXEL_SIZE, Z * VOXEL_SIZE];
 
 const voxelMatrixBuffer = wgsl
   .buffer(arrayOf(arrayOf(arrayOf(voxelStruct, Z), Y), X))
-  .$name('vexel_array')
+  .$name('voxel_array')
   .$allowReadonlyStorage();
 const voxelMatrixData = voxelMatrixBuffer.asReadonlyStorage();
 
@@ -68,6 +68,7 @@ const cameraPositionData = cameraPositionBuffer.asReadonlyStorage();
 const cameraAxesBuffer = wgsl
   .buffer(
     struct({
+      right: vec3f,
       up: vec3f,
       forward: vec3f,
     }),
@@ -82,28 +83,12 @@ const canvasDimsBuffer = wgsl
   .$allowReadonlyStorage();
 const canvasDimsData = canvasDimsBuffer.asReadonlyStorage();
 
-const normalizeVectorFn = wgsl.fn('normalize')` (vector: ${vec3f}) -> ${vec3f} {
-  let length = sqrt(pow(vector.x, 2) + pow(vector.y, 2) + pow(vector.z, 2));
-  return vector / length;
-}`;
-
-const crossProductFn = wgsl.fn(
-  'cross_product',
-)` (vectorA: ${vec3f}, vectorB: ${vec3f}) -> ${vec3f} {
-  return vec3f(
-    vectorA.y * vectorB.z - vectorA.z * vectorB.y,
-		vectorA.z * vectorB.x - vectorA.x * vectorB.z,
-		vectorA.x * vectorB.y - vectorA.y * vectorB.x  
-  );
-}
-`;
-
 const getBoxIntersectionFn = wgsl.fn('box_intersection')`(
   boundMin: vec3f,
   boundMax: vec3f,
   ray: ${rayStruct}
 ) -> ${intersectionStruct} {
-    var output = ${intersectionStruct}();
+    var output: ${intersectionStruct};
 
     var tMin: f32; 
     var tMax: f32;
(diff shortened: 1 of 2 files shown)


apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+const X = 10;
+const Y = 11;
+const Z = 12;
+
+const VOXEL_SIZE = 15;
+const cubeSize = [X * VOXEL_SIZE, Y * VOXEL_SIZE, Z * VOXEL_SIZE];
+
+const voxelMatrixBuffer = wgsl
+  .buffer(arrayOf(arrayOf(arrayOf(voxelStruct, Z), Y), X))
+  .$name('vexel_array')
```
iwoplaza: Small typo
```suggestion
  .$name('voxel_array')
```

apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+const canvasDimsBuffer = wgsl
+  .buffer(struct({ width: u32, height: u32 }))
+  .$name('canvas_dims')
+  .$allowReadonlyStorage();
+const canvasDimsData = canvasDimsBuffer.asReadonlyStorage();
+
+const normalizeVectorFn = wgsl.fn('normalize')` (vector: ${vec3f}) -> ${vec3f} {
+  let length = sqrt(pow(vector.x, 2) + pow(vector.y, 2) + pow(vector.z, 2));
+  return vector / length;
+}`;
```
iwoplaza: There exists a built-in `normalize` function in WGSL.

```suggestion
```

apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+const crossProductFn = wgsl.fn(
+  'cross_product',
+)` (vectorA: ${vec3f}, vectorB: ${vec3f}) -> ${vec3f} {
+  return vec3f(
+    vectorA.y * vectorB.z - vectorA.z * vectorB.y,
+		vectorA.z * vectorB.x - vectorA.x * vectorB.z,
+		vectorA.x * vectorB.y - vectorA.y * vectorB.x  
+  );
+}
+`;
```
iwoplaza: There exists a built-in `cross` function in WGSL.

```suggestion
```

apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+
+const getBoxIntersectionFn = wgsl.fn('box_intersection')`(
+  boundMin: vec3f,
+  boundMax: vec3f,
+  ray: ${rayStruct}
+) -> ${intersectionStruct} {
+    var output = ${intersectionStruct}();
```
iwoplaza:
```suggestion
    var output: ${intersectionStruct};
```

apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+  },
+
+  fragment: {
+    args: ['@builtin(position) pos: vec4f'],
+    code: wgsl.code`
+      let minDim = f32(min(${canvasDimsData}.width, ${canvasDimsData}.height));
+
+      var ray: ${rayStruct};
+      ray.origin = ${cameraPositionData};
+      ray.direction += -${crossProductFn}(${cameraAxesData}.forward, ${cameraAxesData}.up) * (pos.x - f32(${canvasDimsData}.width)/2)/minDim;
```
iwoplaza: To increase efficiency, I would compute the cross product on the CPU (in this case we would not have to compute the cross product, just compute the resulting vector). In this case, the vector is computed by each fragment.

```suggestion
      ray.direction += ${cameraAxesData}.right * (pos.x - f32(${canvasDimsData}.width)/2)/minDim;
```

apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+      ray.direction += ${cameraAxesData}.up * (pos.y - f32(${canvasDimsData}.height)/2)/minDim;
+      ray.direction += ${cameraAxesData}.forward;
+      ray.direction = ${normalizeVectorFn}(ray.direction);
```
iwoplaza:
```suggestion
      ray.direction = normalize(ray.direction);
```

apps/wigsill-examples/src/examples/voxel-rendering.ts
```diff
@@ -0,0 +1,358 @@
...
+}
+
+const boxCenter = cubeSize.map((value) => value / 2);
+
+let frame = 0;
+let timeDelta = 0.05;
+let cameraDist = 350;
+
+onFrame(() => {
+  frame += timeDelta;
```
iwoplaza: To make it framerate independent, I would suggest renaming `timeDelta` to something like `radiansPerSecond` and multiplying it by the provided `deltaTime` parameter.

```suggestion
onFrame((deltaTime) => {
  frame += radiansPerSecond * deltaTime;
```

packages/wigsill/src/wigsillRuntime.ts
```diff
@@ -28,7 +28,10 @@ class WigsillRuntime {
     if (!buffer) {
       buffer = this.device.createBuffer({
         usage: allocatable.flags,
-        size: allocatable.dataType.size,
+        size:
+          Math.ceil(
+            allocatable.dataType.size / allocatable.dataType.byteAlignment,
+          ) * allocatable.dataType.byteAlignment,
```
iwoplaza: `roundUp` does the same faster, but works only if the second parameter is a power of two, which in this case, it has to be.

```suggestion
        size:
          roundUp(allocatable.dataType.size, allocatable.dataType.byteAlignment),
```

PR #86, opened 2024-07-24
Added inline resolution.

### LINKED ISSUE #16: Create a primitive for processing WGSL items inline.
Lets consider an example of a repeat macro. Currently, we can only code-gen based on literal values (strings, numbers, objects, etc.) but not on WGSLItems, since their values are known at resolve-time.

That why we need a primitive for describing a custom WGSLItem inline, that can act on the resolved values.

```ts
function repeat(snippet: WGSLSegment, iterations: number | Slot<number>) {
  return resolved(
    iterations,
    (iter) => {
      // Here we can act on the value of `iterations` at resolve time, and create more code.
      return wgsl`${Array.from({ length: iter }, () => snippet)}`
    }
  );
}
```

```ts
const fractalDepth = wgsl.slot(16); // Slot<number>

const main = wgsl.fn`() -> {
  ${repeat(wgsl`${crawlFractal}();`, fractalDepth)}
}`;
```

iwoplaza:
Closes #16 #88 #89

### DIFF
--- a/packages/wigsill/src/wgslFunction.ts
+++ b/packages/wigsill/src/wgslFunction.ts
@@ -1,5 +1,6 @@
 import type {
   Eventual,
+  InlineResolve,
   ResolutionCtx,
   SlotValuePair,
   Wgsl,
@@ -22,7 +23,10 @@ export interface WgslFn extends WgslResolvable {
 export type BoundWgslFn = Omit<WgslFn, '$name'>;
 
 export function fn(label?: string) {
-  return (strings: TemplateStringsArray, ...params: Wgsl[]): WgslFn => {
+  return (
+    strings: TemplateStringsArray,
+    ...params: (Wgsl | InlineResolve)[]
+  ): WgslFn => {
     const func = new WgslFnImpl(code(strings, ...params));
     if (label) {
       func.$name(label);
--- a/packages/wigsill/tests/inlineResolve.test.ts
+++ b/packages/wigsill/tests/inlineResolve.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from 'vitest';
+import { wgsl } from 'wigsill';
+import { parseWGSL } from './utils/parseWGSL';
+
+describe('inline resolve', () => {
+  it('injects returned constant in code', () => {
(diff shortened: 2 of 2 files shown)


packages/wigsill/src/macro/repeat.ts
```diff
@@ -1,13 +1,45 @@
...
+    const countValue = get(count);
+    const snippetValue = get(snippet);
+
+    if (typeof countValue !== 'number') {
+      const index = new WgslIdentifier().$name('i');
+
+      if (typeof snippetValue === 'function') {
+        return code`
+for (var ${index} = 0; ${index} < ${countValue}; ${index} += 1) {
+  ${snippetValue(index as unknown as number)}
```
mhawryluk: please add a comment explaining what's happening here

packages/wigsill/src/macro/repeat.ts
```diff
@@ -1,13 +1,45 @@
...
+  return code`${(get) => {
+    const countValue = get(count);
+    const snippetValue = get(snippet);
+
+    if (typeof countValue !== 'number') {
+      const index = new WgslIdentifier().$name('i');
+
+      if (typeof snippetValue === 'function') {
+        return code`
+for (var ${index} = 0; ${index} < ${countValue}; ${index} += 1) {
```
mhawryluk: if we're planning on creating a prettifier tool outputting nicely formatted wgsl from parsed AST in the near future, then this piece of code should be indented to not stick out as much

packages/wigsill/tests/macro.test.ts
```diff
@@ -1,14 +1,61 @@
...
 describe('repeat', () => {
-  it('repeats a string', () => {
-    const result = repeat(3, 'a');
-    expect(result).toEqual(wgsl`${['a', 'a', 'a']}`);
+  it('repeats a function call N times', () => {
+    const actual = parseWGSL(wgsl`
+      { ${repeat(3, 'a();')} }
+    `);
+
+    const expected = parseWGSL(wgsl`
```
mhawryluk: the `wgsl` string interpolation function for expected values is unnecessary

PR #100, opened 2024-07-25
Added plums, and a way to use them to synchronize buffers.

iwoplaza:


packages/wigsill/src/plumStore.test.ts
```diff
@@ -0,0 +1,389 @@
...
+      },
+    });
+
+    store.set(fooPlum, 5);
+
+    expect(doubledListener).toBeCalledTimes(1);
+    expect(tripledListener).toBeCalledTimes(1);
+    expect(store.get(doubledPlum)).toEqual(10);
+    expect(store.get(tripledPlum)).toEqual(15);
+  });
```
reczkok: It does not actually check it - instead it's the same as 'should handle multiple dependents'

apps/wigsill-docs/src/content/docs/guides/state-management.mdx
```diff
@@ -0,0 +1,129 @@
...
+Determining **when** to synchronize the state depends on the use case, but it usually falls into one of two strategies:
+
+import ListItem from '../../../components/ListItem.astro';
+
+<ol>
+  <ListItem idx="A">
+    <details>
+    <summary>Before each shader execution</summary>
+    
+    Optimal when states updates more frequently than shaders are ran.
```
mhawryluk:
```suggestion
    Optimal when state updates more frequently than shaders are run.
```

apps/wigsill-docs/src/content/docs/guides/state-management.mdx
```diff
@@ -0,0 +1,129 @@
...
+
+    loop();
+    ```
+    </details>
+  </ListItem>
+  <ListItem idx="B">
+    <details>
+    <summary>After each state change</summary>
+
+    Optimal when states updates are rarer than shader executions.
```
mhawryluk:
```suggestion
    Optimal when state updates are rarer than shader executions.
```

packages/wigsill/src/plumStore.ts
```diff
@@ -0,0 +1,235 @@
...
+    if (state.active) {
+      // Unsubscribing from old dependencies
+      for (const unsub of state.active.unsubs) {
+        unsub();
+      }
+    }
+
+    const newDependencies = new Map<WgslPlum, number>();
+
+    const getter = (<T>(dep: WgslPlum<T>) => {
```
mhawryluk, line 84: same as in previous method, can be moved to a private method to remove duplication

packages/wigsill/src/plumStore.ts
```diff
@@ -0,0 +1,235 @@
...
+    state.version++;
+
+    if (state.active) {
+      // copying, because listeners may change after we notify our dependents.
+      const listeners = [...state.active.listeners];
+
+      for (const listener of listeners) {
+        listener();
+      }
+    }
```
mhawryluk: also can be moved to a separate method to avoid duplication

packages/wigsill/src/wgslPlum.ts
```diff
@@ -0,0 +1,215 @@
...
+    this._label = label;
+    return this;
+  }
+
+  get label(): string | undefined {
+    return this._label;
+  }
+
+  resolve(ctx: ResolutionCtx): string {
+    throw new Error('Method not implemented.');
```
mhawryluk: why declare plum as WgslResolvable when it in fact is not. Is it because in the future we want plums to be resolvable?

iwoplaza: Yup, that's why. I guess we can remove it for now until we come around to implementing it.

PR #102, opened 2024-07-26
Add button element support in the example toolkit

### LINKED ISSUE #93: Create a generic button element into the example toolkit
Useful for examples like `Increment`, `Matrix Multiplication`, etc.

```ts
// HTMLButtonElement
const button = addElement('button', { label: 'Click me!' });
```

mhawryluk:
Closes #93

### DIFF
--- a/apps/wigsill-examples/src/types/example-toolkit.d.ts
+++ b/apps/wigsill-examples/src/types/example-toolkit.d.ts
@@ -39,20 +39,11 @@ declare module '@wigsill/example-toolkit' {
 
   export type ElementType = keyof ElementOptions;
 
-  type Merge<A, B> = {
-    [K in keyof A]: K extends keyof B ? B[K] : A[K];
-  } & B extends infer O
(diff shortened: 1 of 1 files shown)


apps/wigsill-examples/src/types/example-toolkit.d.ts
```diff
@@ -1,50 +1,67 @@
...
+
+  type ElementDefs = {
+    [K in ElementType]: Merge<
+      ElementOptions[K],
+      {
+        type: K;
+        key: string;
+      }
+    >;
+  };
```
iwoplaza: This seems to work well, is there a reason for using the more complex merge?

```suggestion
  type ElementDefs = {
    [K in ElementType]: ElementOptions[K] & {
      type: K;
      key: string;
    };
  };
```

mhawryluk: I tried that and thought it didn't work so I went for the more complex way, but now I've realized I just made a mistake hah

PR #106, opened 2024-07-26
Added autogenerated API docs.

iwoplaza:
image

### DIFF
--- a/packages/wigsill/src/web/webWigsillRuntime.ts
+++ b/packages/wigsill/src/web/webWigsillRuntime.ts
@@ -377,17 +377,11 @@ export async function createRuntime(
     throw new Error('WebGPU is not supported by this browser.');
   }
 
-  const adapterOptions =
-    options && 'adapter' in options ? options.adapter : undefined;
-
-  const deviceOptions =
-    options && 'device' in options ? options.device : undefined;
-
-  const adapter = await navigator.gpu.requestAdapter(adapterOptions);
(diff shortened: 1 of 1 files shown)


packages/wigsill/src/errors.ts
```diff
@@ -1,5 +1,8 @@
 import type { WgslSlot } from './types';
 
+/**
+ * @category Errors
```
mhawryluk, line 4: I don't think this does anything at the moment, does it? to have these grouped in the navigation, I believe we need to enable this behavior in typedoc config, via "navigation.includeCategories"

iwoplaza: I tried that configuration, and tried fiddling with other config options, but to no avail ;-;. I will look into the implementation later on how to fix this, in a separate PR.

packages/wigsill/src/web/webWigsillRuntime.ts
```diff
@@ -335,38 +333,61 @@ class ComputePipelineExecutor extends PipelineExecutor<GPUComputePipeline> {
...
-  if (!options) {
-    adapter = await navigator.gpu.requestAdapter();
-    if (!adapter) {
-      throw new Error('Could not find a compatible GPU');
-    }
-    device = await adapter.requestDevice();
-    return new WebWigsillRuntime(device);
-  }
+  const adapterOptions =
+    options && 'adapter' in options ? options.adapter : undefined;
```
mhawryluk: `options?.adapter` and below `options?.device` and we don't need these stored in constants

PR #109, opened 2024-07-29
WgslResolvable toString overhaul

### LINKED ISSUE #87: Make `toString()` methods of WgslResolvable implementations throw errors during the resolution process.
This is to dissuade developers from injecting items into template literals that are not tagged functions.

Behavior before the change:
```ts
const valueSlot = wgsl.slot(123).$name('value');

console.log(ctx.resolve(
  wgsl`const something = ${valueSlot};` // > const something = 123;
));

console.log(ctx.resolve(
  `const something = ${valueSlot};` // > const something = slot:value;
));

console.log(valueSlot); // > slot:value

```

Expected after the change:

```ts
const valueSlot = wgsl.slot(123).$name('value');

console.log(ctx.resolve(
  wgsl`const something = ${valueSlot};` // > const something = 123;
));

console.log(ctx.resolve(
  `const something = ${valueSlot};` // ❌ error thrown, e.g. "Use wgsl`...` when interpolating wgsl code."
));

// \/ this still works
console.log(valueSlot); // > slot:value

```

Since some errors are created during resolution, we need a different way to access this inherent name & purpose. Maybe:
`readonly typeInfo: string` on the WgslResolvable interface?

mhawryluk:
Closes #87


packages/wigsill/src/wgslResolvableBase.ts
```diff
@@ -0,0 +1,19 @@
+import { ResolvableToStringError } from './errors';
+
+export abstract class WgslResolvableBase {
+  abstract typeInfo: string;
```
iwoplaza: To allow classes that extend this base class to compute their typeInfo instead of having it constant (for example to include additional context), lets make it readonly.

```suggestion
  abstract readonly typeInfo: string;
```

packages/wigsill/src/wgslResolvableBase.ts
```diff
@@ -0,0 +1,19 @@
...
+  public label: string | undefined;
+
+  $name(label?: string | undefined) {
+    this.label = label;
+    return this;
+  }
```
iwoplaza: Lets make label settable only through `$name`.
```suggestion
  private _label: string | undefined;
  
  get label() {
    return this._label;
  }

  $name(label?: string | undefined) {
    this._label = label;
    return this;
  }
```

packages/wigsill/src/wgslSlot.ts
```diff
@@ -23,15 +24,15 @@ export function slot<T>(defaultValue?: T): WgslSlot<T> {
 // Implementation
 // --------------
 
-class WgslSlotImpl<T> implements WgslResolvable, WgslSlot<T> {
+class WgslSlotImpl<T>
+  extends WgslResolvableBase
+  implements WgslResolvable, WgslSlot<T>
+{
+  typeInfo = 'slot';
```
iwoplaza:
```suggestion
  readonly typeInfo = 'slot';
```

PR #117, opened 2024-07-30
Display example code execution errors with a new Snackbar component

### LINKED ISSUE #105: Code editor can't handle infinite loops
When there's an infinite loop in the code editor, the whole `wigsill-examples` site freezes forevermore. Even though it seems like a user skill issue, it is actually easy to do involuntarily when starting to write a for loop. enter `for ()` and it's game over. We should limit allowed code execution time to avoid the issue.

mhawryluk:
closes #105 & #111

### DIFF
--- a/apps/wigsill-examples/src/example/exampleRunner.ts
+++ b/apps/wigsill-examples/src/example/exampleRunner.ts
@@ -58,11 +58,11 @@ const staticToDynamicImports = {
   } satisfies TraverseOptions,
 };
 
-/**
- * from https://github.com/facebook/react/blob/d906de7f602df810c38aa622c83023228b047db6/scripts/babel/transform-prevent-infinite-loops.js
- * **/
 const MAX_ITERATIONS = 2000;
 
+/**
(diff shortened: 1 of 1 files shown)


apps/wigsill-examples/src/example/exampleRunner.ts
```diff
@@ -58,6 +58,49 @@ const staticToDynamicImports = {
   } satisfies TraverseOptions,
 };
 
+/**
+ * from https://github.com/facebook/react/blob/d906de7f602df810c38aa622c83023228b047db6/scripts/babel/transform-prevent-infinite-loops.js
+ * **/
```
iwoplaza: Small nitpick, I would move this comment above the `preventInfiniteLoops` function, since I think linking the source to the implementation rather than used constants is more readable.

```suggestion
/**
 * from https://github.com/facebook/react/blob/d906de7f602df810c38aa622c83023228b047db6/scripts/babel/transform-prevent-infinite-loops.js
 */
```

PR #131, opened 2024-08-02
Fixed bug in fluid sim and made the canvas not have a fixed size

### LINKED ISSUE #130: Fluid simulation drawing not working as intended.
On my machine the brush is off by half from the desired coordinates. @mhawryluk please check if you can reproduce the bug. ⚙︎

reczkok:
Closes #130 if the bug is reproducible.

### DIFF
--- a/apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics.ts
+++ b/apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics.ts
@@ -536,8 +536,11 @@ canvas.onmousemove = (event) => {
   }
 
   const cellSize = canvas.width / options.size;
-  const x = Math.floor((event.offsetX * 2) / cellSize);
-  const y = options.size - Math.floor((event.offsetY * 2) / cellSize) - 1;
+  const x = Math.floor((event.offsetX * window.devicePixelRatio) / cellSize);
+  const y =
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics.ts
```diff
@@ -536,8 +536,8 @@ canvas.onmousemove = (event) => {
   }
 
   const cellSize = canvas.width / options.size;
-  const x = Math.floor(event.offsetX / cellSize);
-  const y = options.size - Math.floor(event.offsetY / cellSize) - 1;
+  const x = Math.floor((event.offsetX * 2) / cellSize);
```
mhawryluk:
```suggestion
  const x = Math.floor((event.offsetX * window.devicePixelRatio) / cellSize);
```

I think the bug appeared after changes in `Canvas.tsx` where we moved scaling canvas by `window.devicePixelRatio`. hardcoding `2` here fixes it only for the 100% zoom level, but there's still a noticable offset on different scales. replacing `2` with `window.devicePixelRatio` seems to fix the issue. same with y

reczkok: Great - thanks!

PR #132, opened 2024-08-05
Blur example

reczkok:
An example using storage textures, sampled textures, samplers, plums and slots. Also cute plums picture.

### DIFF
--- a/packages/typegpu/src/plumStore.test.ts
+++ b/packages/typegpu/src/plumStore.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it, vi } from 'vitest';
-import { PlumStore } from '../src/plumStore';
-import { type Getter, plum, plumFromEvent } from '../src/wgslPlum';
+import { PlumStore } from './plumStore';
+import { type Getter, plum, plumFromEvent } from './wgslPlum';
 
 function makeSubject<T>(initial: T) {
   let value = initial;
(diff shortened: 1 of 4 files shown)


apps/typegpu-docs/src/content/examples/image-processing/blur.ts
```diff
@@ -0,0 +1,296 @@
...
+  flip: number,
+) {
+  return runtime.makeComputePipeline({
+    workgroupSize: [32],
+    code: wgsl`
+      ${mainComputeFun
+        .with(inTextureSlot, inTexture.asSampled(inParams))
+        .with(outTextureSlot, outTexture.asStorage(outParams))
+        .with(flipSlot, flip)}
+        (${builtin.workgroupId}, ${builtin.localInvocationId});
```
mhawryluk: why do we move compute code into a function and pass builtins here instead of using them in place of `wid` and `lid`? was it just for readability to not have to repeat ${builtin.workgroupId} and ${builtin.localInvocationId} multiple times or was there some other reason?

reczkok: This was caused by the fact that functions could not use builtins and code could not use bindings. I needed to extract logic to a function to apply bindings but couldn't use builtins (maybe in the future we can allow it with some parameter automation). This is now fixed as code can be bound 😁 [24e46f4](https://github.com/software-mansion/TypeGPU/pull/132/commits/24e46f42889e52792d3fa039542e3ddfa8615818) and [2cc166f](https://github.com/software-mansion/TypeGPU/pull/132/commits/2cc166f51518b1ddc4b92865d0ef0788292ce9c3)

PR #133, opened 2024-08-05
Allow for parsing textures and buffers (and other global vars)

reczkok:
While trying to implement textures into our parser I discovered that we actually already support them as well as nearly all declarations. The way we handled global variable declarations seems wrong so I updated it to better match [the specification](https://www.w3.org/TR/WGSL/#var-and-value-decl-grammar).

### DIFF
--- a/packages/typegpu-wgsl-parser/src/grammar.ne
+++ b/packages/typegpu-wgsl-parser/src/grammar.ne
@@ -286,7 +286,6 @@ global_variable_decl ->
 optionally_typed_ident ->
   ident (":" type_specifier):? {% ([ident, typespec]) => ({ ident: ident.value, typespec: typespec ? typespec[1] : null }) %}
 
-# TODO: Add support for 'override' global values
 @{% export type ValueDecl = { type: 'value_decl', ident: string, typespec: TypeSpecifier | null, expr: Expression }; %}
 value_decl ->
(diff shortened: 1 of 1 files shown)


packages/typegpu-wgsl-parser/src/grammar.ne
```diff
@@ -279,9 +279,14 @@ let_decl ->
...
   "var" template_list:? optionally_typed_ident ("=" expression):? {% ([ , template_list, typed_ident, opt_expr]) => ({ type: 'variable_decl', template_list: template_list, ...typed_ident, expr: opt_expr ? opt_expr[1] : null }) %}
 
+@{% export type GlobalVariableDecl = {type: 'global_variable_decl', attributes: Attribute[] | null, variable_decl: VariableDecl}; %}
+global_variable_decl ->
+  attribute:* variable_decl {% ([attrs, variable_decl]) => ({ type: 'global_variable_decl', attributes: attrs, variable_decl }) %}
+
 optionally_typed_ident ->
   ident (":" type_specifier):? {% ([ident, typespec]) => ({ ident: ident.value, typespec: typespec ? typespec[1] : null }) %}
 
+# TODO: Add support for 'override' global values
```
reczkok:
```suggestion
```

PR #135, opened 2024-08-05
Add a way to copy TypeGPU buffers

reczkok:
## Update: The masking idea is abandoned for now 
I'm thinking about adding a way to 'mask' when copying. In the following example:
```ts
const s1 = struct({
  aa: vec3u,
  bb: arrayOf(f32, 2),
});

const s2 = struct({
  a: u32,
  b: i32,
  c: s2,
});

const b1 = wgsl
  .buffer(s2, {
    a: 1,
    b: -2,
    c: {
      aa: [3, 4, 5],
      bb: [6.1, 7.1],
    },
  })

const b2 = wgsl
  .buffer(s2, {
    a: 11,
    b: -11,
    c: {
      aa: [11, 11, 11],
      bb: [11, 11],
    },
  })
```
Let's say we wanted to copy only a part of b2 into b1 - having a mask parameter could work like this:
```ts
runtime.copyBuffer(b2, b1, {
    a: 1,
    b: 0,
    c: {
      aa: [1, 0, 0],
      bb: [1, 1],
    },
  });
```
The result would be:
```ts
{
    a: 11,
    b: -2,
    c: {
      aa: [11, 4, 5],
      bb: [11, 11],
    }
}
```
If no mask is passed the entire buffer is copied. Is this approach worth further development?

### DIFF
--- a/apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics.ts
+++ b/apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics.ts
@@ -379,9 +379,9 @@ function resetGameData() {
       instanceCount: options.size ** 2,
     });
 
-    runtime.writeBuffer(currentStateBuffer, nextStateBuffer);
-
     runtime.flush();
+
+    runtime.writeBuffer(currentStateBuffer, nextStateBuffer);
   };
 
   applyDrawCanvas = () => {


packages/typegpu/src/createRuntime.ts
```diff
@@ -173,6 +173,66 @@ class TypeGpuRuntimeImpl {
     this.device.queue.writeBuffer(gpuBuffer, 0, hostBuffer, 0, size);
   }
 
+  copyBuffer<TData extends AnyWgslData>(
+    source: WgslAllocatable<TData>,
+    destination: WgslAllocatable<TData>,
+    mask?: Parsed<TData>,
```
mhawryluk: Are you sure that the mask can have the same data type as buffers? What about bool values? will it still work?

packages/typegpu/src/createRuntime.ts
```diff
@@ -173,6 +173,66 @@ class TypeGpuRuntimeImpl {
...
+          sourceBuffer,
+          offset,
+          destinationBuffer,
+          offset,
+          size,
+        );
+      }
+      this.device.queue.submit([commandEncoder.finish()]);
+    } else {
+      const commandEncoder = this.device.createCommandEncoder();
```
mhawryluk: you can simplify this code by pushing { 0, size } to the `toCopy` array in the else block and extract the commandEncoder logic after the if/else

reczkok: Very true, thanks

packages/typegpu/src/createRuntime.ts
reczkok: At the moment it works with all combination of structs, arrays and 32bit values. Bools and f16 will not work as expected in this primitive approach. I'm pretty sure I can get the proper offsets using typed-binary but for now I just wanted to know if it's worth to work on this.

mhawryluk: I see. so I think it's worth it, but if it turns out too difficult, then we can leave it for another time. I leave the decision up to you

PR #139, opened 2024-08-08
Tests for the runtime

### LINKED ISSUE #150: Mock "GPUDevice" to test buffer allocation.

reczkok:
Tests for the TypeGPU runtime.

closes #150

### DIFF
--- a/packages/typegpu/src/createRuntime.ts
+++ b/packages/typegpu/src/createRuntime.ts
@@ -539,12 +539,7 @@ export type CreateRuntimeOptions = {
 export async function createRuntime(
   options?: CreateRuntimeOptions | GPUDevice,
 ): Promise<TypeGpuRuntime> {
-  if (
-    options &&
-    typeof options === 'object' &&
-    'queue' in options &&
-    'createBuffer' in options
-  ) {
+  if (doesResembleDevice(options)) {
     return new TypeGpuRuntimeImpl(options);
   }
 
@@ -561,6 +556,11 @@ export async function createRuntime(
   return new TypeGpuRuntimeImpl(await adapter.requestDevice(options?.device));
 }
 
-export const exportedForTesting = {
-  TypeGpuRuntimeImpl,
-};
+function doesResembleDevice(value: unknown): value is GPUDevice {
+  return (
+    !!value &&
+    typeof value === 'object' &&
+    'createBuffer' in value &&
+    'queue' in value
+  );
+}
--- a/packages/typegpu/tests/runtime.test.ts
+++ b/packages/typegpu/tests/runtime.test.ts
@@ -1,10 +1,9 @@
 import { wgsl } from 'typegpu';
 import { describe, expect, it, vi } from 'vitest';
 import { afterEach } from 'vitest';
-import { createRuntime, exportedForTesting } from '../src/createRuntime';
+import { createRuntime } from '../src/createRuntime';
 import { struct, u32, vec3i, vec4u } from '../src/data';
 import { plum } from '../src/wgslPlum';
-const { TypeGpuRuntimeImpl } = exportedForTesting;
 
 global.GPUBufferUsage = {
   MAP_READ: 1,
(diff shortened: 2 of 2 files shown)


packages/typegpu/tests/buffer.test.ts
```diff
@@ -0,0 +1,238 @@
...
+  queue: {
+    copyExternalImageToTexture: vi.fn(),
+    onSubmittedWorkDone: vi.fn(),
+    submit: vi.fn(),
+    writeBuffer: vi.fn(),
+    writeTexture: vi.fn(),
+  },
+}));
+
+vi.stubGlobal('GPUDevice', mockDevice);
```
mhawryluk: `GPUDevice` set to `mockDevice` won't allow the mocked device to be detected as instanceOf it. The only way I got it to work though is to set GPUDevice to Object. it is definitely not ideal, cause we block ourselves from passing anything other than device to `createRuntime`, but we don't do so in any of these tests. I think it's the best way to go here, because then we don't need the `vi.mock` below, and we don't need to export `TypeGPURuntimeImpl` at all. we just need to remember to pass `mockDevice()` to createRuntime in each of these tests. but if you prefer the current approach, then it's okay with me. or maybe you can think of a way to make the returned mock object be a subclass of a specific class that we can then set as GPUDevice

packages/typegpu/tests/buffer.test.ts
```diff
@@ -0,0 +1,238 @@
...
+
+vi.mock('../src/createRuntime', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../src/createRuntime')>();
+  return {
+    ...actual,
+    // @ts-ignore
```
mhawryluk: don't think that's needed

packages/typegpu/src/createRuntime.ts
```diff
@@ -539,7 +539,12 @@ export type CreateRuntimeOptions = {
 export async function createRuntime(
   options?: CreateRuntimeOptions | GPUDevice,
 ): Promise<TypeGpuRuntime> {
-  if (options instanceof GPUDevice) {
+  if (
+    options &&
+    typeof options === 'object' &&
+    'queue' in options &&
+    'createBuffer' in options
+  ) {
```
iwoplaza: Could we move this check into a function like:
```ts
function doesResembleDevice(value: unknown): value is GPUDevice {
  return options &&
    typeof options === 'object' &&
    'createBuffer' in options
}
```

```suggestion
  if (doesResembleDevice(options)) {
```

packages/typegpu/tests/runtime.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+    const mockBuffer = runtime.bufferFor(bufferData);
+    expect(mockBuffer).toBeDefined();
+    expect(mockBuffer.getMappedRange).not.toBeCalled();
+
+    expect(testPipeline).toBeDefined();
+    expect(runtime.device.createBuffer).toBeCalledWith({
+      mappedAtCreation: false,
+      size: 4,
+      usage: 76,
```
iwoplaza: Could we use the combination of flags here for better readability?

packages/typegpu/tests/runtime.test.ts
```diff
@@ -0,0 +1,226 @@
+import { wgsl } from 'typegpu';
+import { describe, expect, it, vi } from 'vitest';
+import { afterEach } from 'vitest';
+import { createRuntime, exportedForTesting } from '../src/createRuntime';
+import { struct, u32, vec3i, vec4u } from '../src/data';
+import { plum } from '../src/wgslPlum';
+const { TypeGpuRuntimeImpl } = exportedForTesting;
```
iwoplaza: Is this still used?

packages/typegpu/src/createRuntime.ts
```diff
@@ -555,3 +560,7 @@ export async function createRuntime(
 
   return new TypeGpuRuntimeImpl(await adapter.requestDevice(options?.device));
 }
+
+export const exportedForTesting = {
```
mhawryluk: ✂️

PR #156, opened 2024-08-13
Example controls and page overhaul

### LINKED ISSUE #145: Implement new styles for Widgets like Buttons, Switch, etc.

### LINKED ISSUE #146: Implement new styles for the Example Controls.

### LINKED ISSUE #155: Update logo on Examples page to match the landing page.

mhawryluk:
known bugs that I couldn't figure out yet:
- [ ]  no underline under the link with gradient and background clip
- [x] when editing image blur example sometimes the example controls double
- [x] dropdown component styling is off -> I don't think the trigger can be styled, maybe via global css
- [ ] not sure about the design of buttons when they are in the example controls panel, nowhere to be found in figma
- [ ] flickering of canvas with certain window sizes, when size nor aspectRatio is specified. probably due to overflow: auto

closes #145
closes #146 
closes #155


apps/typegpu-docs/public/assets/discord-icon.svg
iwoplaza, line 1: Since we do not care for these files to be available via a stable URL, and Astro would not optimize them anyway, I think we can put them in `typegpu-docs/src/assets`, and import them.

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -0,0 +1,171 @@
...
+    <>
+      <div className="text-sm text-pretty text-ellipsis">{label}</div>
+
+      <Slider
+        min={min ?? 0}
+        max={max ?? 1}
+        step={step ?? 0.1}
+        initial={initial}
+        onChange={onChange}
+      />
```
iwoplaza: I would change the `Slider` to accept the current value of the slider instead of the initial value. That way, we can use this component in other places, where a controlled input might be more desired. In this case I would move the state into the <SliderRow /> component, and do the same for other reusable design components (except for the Toggle, I see it is already like I described).

apps/typegpu-docs/src/components/ExampleLayout.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { menuShownAtom } from '../utils/examples/menuShownAtom';
+import ExampleList from './ExampleList';
+import ExamplePage from './ExamplePage';
+
+export function ExampleLayout() {
+  const menuShown = useAtomValue(menuShownAtom);
+
+  return (
+    <div className="flex h-screen p-4 gap-4 bg-grayscale-20">
+      {menuShown ? (
```
iwoplaza: Could we extract the menu into a separate component for readability? It can still be in the same file.

apps/typegpu-docs/src/components/ExampleLink.tsx
```diff
@@ -29,13 +33,15 @@ export function ExampleLink(props: Props) {
...
         active
-          ? 'border-gray-500 text-black'
-          : 'border-transparent text-gray-600',
+          ? 'bg-clip-text bg-gradient-to-r from-gradient-purple-dark to-gradient-blue-dark text-transparent'
+          : '',
       )}
     >
       {children}
+
+      {active ? <img src="/assets/selected-dot.svg" alt="" /> : null}
```
iwoplaza: Import from asset file.

apps/typegpu-docs/src/utils/examples/exampleRunner.ts
```diff
@@ -167,66 +132,117 @@ export async function executeExample(
...
+        onChange,
+      },
+    ]);
+
+    // Eager run to initialize the values.
+    onChange(initial);
+  }
+
+  function addButtonParameter(label: string, onClick: () => void) {
+    setExampleControlParams((controls) => [
```
iwoplaza: I think the duplicate parameters might be caused by two example codes running at the same time. This can be patched by checking whether or not we were disposed of when creating new parameters.

This if statement should be added to every `add*Parameter` function.

```suggestion
    if (disposed) {
      return;
    }
    
    setExampleControlParams((controls) => [
```

apps/typegpu-docs/public/typegpu-logo-light.svg
iwoplaza, line 1: Because the size of the svg changed, the logo now appears tiny in the README (on light mode).

image

apps/typegpu-docs/public/typegpu-logo-light.svg
```diff
@@ -1,25 +1,62 @@
...
+<svg width="158" height="40" viewBox="0 0 158 40" fill="none" xmlns="http://www.w3.org/2000/svg">
```
iwoplaza: We can make the logo slightly bigger.
```suggestion
<svg width="474" height="120" viewBox="0 0 158 40" fill="none" xmlns="http://www.w3.org/2000/svg">
```

PR #182, opened 2024-08-20
Added a FAQ to the documentation.

iwoplaza:


apps/typegpu-docs/src/content/docs/faq.mdx
```diff
@@ -0,0 +1,46 @@
...
+
+The API differs from what can be seen in WebGPU due to a difference in behavior between the two.
+Resource management in TypeGPU is, for the most part, automated. When defining a GPU buffer, we
+can do so using the `wgsl.buffer` API like so:
+
+```ts
+const fooBuffer = wgsl.buffer(f32).$name('foo');
+```
+
+This however does not allocate any memory on the GPU, as opposed to the `device.createBuffer` WebGPU
```
mhawryluk, line 21: maybe we could specify here that it does not allocate memory right away, but it is allocated eventually when used in a shader, cause I think it may suggest otherwise

apps/typegpu-docs/src/content/docs/faq.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: FAQ
+description: Frequently Asked Questions
+---
+
+<details>
+<summary>🙋 Why does the API differ from WebGPU?</summary>
+
+The API differs from what can be seen in WebGPU due to a difference in behavior between the two.
```
mhawryluk, line 9: I would add that the API also differs, cause we want to eliminate much of the boilerplate to make writing GPU programs easier

apps/typegpu-docs/src/content/docs/faq.mdx
```diff
@@ -0,0 +1,51 @@
...
+<summary>🙋 Which browsers support TypeGPU?</summary>
+
+TypeGPU can be used on any browser that supports WebGPU. The current coverage can be [checked here](https://caniuse.com/webgpu).
+Additionally, thanks to [react-native-wgpu](https://github.com/wcandillon/react-native-webgpu), TypeGPU also works on React Native!
+
+</details>
+
+
+<details>
+<summary>🙋 Can I use it on React Native?</summary>
```
reczkok:
```suggestion
<summary>🙋 Can I use it in React Native?</summary>
```
I think this is better? Or:
```suggestion
<summary>🙋 Can I use it in my React Native project?</summary>
```

apps/typegpu-docs/src/content/docs/faq.mdx
```diff
@@ -0,0 +1,51 @@
...
+each other with the common language of our primitives, even though they are built with different use
+cases in mind (Visualization, AI inference, etc.).
+
+</details>
+
+<details>
+<summary>🙋 Which browsers support TypeGPU?</summary>
+
+TypeGPU can be used on any browser that supports WebGPU. The current coverage can be [checked here](https://caniuse.com/webgpu).
+Additionally, thanks to [react-native-wgpu](https://github.com/wcandillon/react-native-webgpu), TypeGPU also works on React Native!
```
reczkok:
```suggestion
Additionally, thanks to [react-native-wgpu](https://github.com/wcandillon/react-native-webgpu), TypeGPU also works in React Native!
```

apps/typegpu-docs/src/content/docs/faq.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+</details>
+
+
+<details>
+<summary>🙋 Can I use it on React Native?</summary>
+
+Yes! Thanks to [react-native-wgpu](https://github.com/wcandillon/react-native-webgpu), TypeGPU also works on React Native.
```
reczkok:
```suggestion
Yes! Thanks to [react-native-wgpu](https://github.com/wcandillon/react-native-webgpu), TypeGPU also works in React Native.
```

Issue #197, opened 2024-08-23
Auto align buffers if Uniform is allowed.


iwoplaza:
There is often a situation when we want to have a uniform consisting of non 16byte aligned variables (3 separate u32 for example). When calling `.allowUniform()` it would be nice if it auto-aligned the data.

PR #200, opened 2024-08-23
Custom wgsl data alignment and size

mhawryluk:
references #197

### DIFF
--- a/packages/typegpu/src/data/size.ts
+++ b/packages/typegpu/src/data/size.ts
@@ -64,8 +64,7 @@ export class WgslDataCustomSizedImpl<
     value: MaxValue | ParseUnwrapped<TData>,
     measurer: IMeasurer = new Measurer(),
   ): IMeasurer {
-    measurer.add(this.size - this.data.size);
-    return this.data.measure(value, measurer);
+    return measurer.add(this.size);
   }
 
   resolve(ctx: ResolutionCtx): string {


packages/typegpu/src/data/align.ts
```diff
@@ -0,0 +1,60 @@
...
+{
+  public readonly size: number;
+
+  constructor(
+    private data: AnyWgslData,
+    public readonly byteAlignment: number,
+  ) {
+    super();
+
+    this.size = this.measure(MaxValue).size;
```
reczkok: Maybe we could avoid measuring the data again if we just use the data size? I don't think there is a situation when it's different. 
```suggestion
    this.size = this.data.size;
```

mhawryluk: yup, already fixed that

packages/typegpu/src/data/size.ts
```diff
@@ -0,0 +1,74 @@
...
+    return this.data.read(input) as ParseUnwrapped<TData>;
+  }
+
+  measure(
+    value: MaxValue | ParseUnwrapped<TData>,
+    measurer: IMeasurer = new Measurer(),
+  ): IMeasurer {
+    measurer.add(this.size - this.data.size);
+    return this.data.measure(value, measurer);
+  }
```
iwoplaza: Size is known, since WGSL data types cannot have varying size based on the value they hold.

```suggestion
  measure(
    value: MaxValue | ParseUnwrapped<TData>,
    measurer: IMeasurer = new Measurer(),
  ): IMeasurer {
    return this.size;
  }
```

mhawryluk: measure has IMeasurer return type. I don't think I can return a number

iwoplaza: Oh right, my bad. But you can return this:

`return measurer.add(this.size)`

PR #206, opened 2024-08-27
TS to WGSL compiler proof-of-concept

iwoplaza:


packages/tgsl-tools/README.md
```diff
@@ -0,0 +1,92 @@
+# @typegpu/tgsl-tools
+
+A set of utilities for working with TGSL (TypeGPU Shader Language), a subset of JavaScript that can be ran on the GPU through [**TypeGPU**](https://docs.swmansion.com/typegpu).
```
mhawryluk:
```suggestion
A set of utilities for working with TGSL (TypeGPU Shader Language), a subset of JavaScript that can be run on the GPU through [**TypeGPU**](https://docs.swmansion.com/typegpu).
```

packages/typegpu/src/tgpuFn.ts
```diff
@@ -0,0 +1,193 @@
...
+ */
+interface TgpuFnShell<
+  Args extends AnyTgpuDataTuple,
+  Return extends AnyTgpuData | undefined,
+> {
+  readonly argTypes: Args;
+  readonly returnType: Return | undefined;
+
+  /**
+   * Creates an type-safe implementation of this signature
```
mhawryluk:
```suggestion
   * Creates a type-safe implementation of this signature
```

packages/typegpu/src/tgpuFn.ts
```diff
@@ -0,0 +1,193 @@
...
+
+export function fn<
+  Args extends AnyTgpuDataTuple,
+  Return extends AnyTgpuData | undefined = undefined,
+>(first?: Args | Return, second?: Return): TgpuFnShell<Args, Return> {
+  if (Array.isArray(first)) {
+    return new TgpuFnShellImpl((first ?? []) as Args, second as Return);
+  }
+
+  // If only one argument, at is is not an array, it is the return type.
```
mhawryluk:
```suggestion
  // If only one argument and it is not an array, it is the return type.
```

packages/typegpu/tests/builtin.test.ts
```diff
@@ -9,8 +9,8 @@ describe('builtin', () => {
     });
 
     const code = wgsl`
-      let x = ${builtin.position};
-      let y = ${builtin.frontFacing};
+      let x = ${builtin.position as unknown as symbol};
+      let y = ${builtin.frontFacing as unknown as symbol};
```
mhawryluk: pretty sure it works without it

packages/tgsl-tools/src/parsers.ts
```diff
@@ -0,0 +1,210 @@
...
+    if (node.async) {
+      throw new Error('tgpu.fn cannot be async.');
+    }
+
+    if (node.params.some((p) => p.type !== 'Identifier')) {
+      throw new Error('tgpu.fn implementations require concrete parameters');
+    }
+
+    const params = node.params as acorn.Identifier[];
+    const header = wgsl`(${params.map((p, idx) => wgsl`${p.name}: ${ctx.argTypes[idx] ?? ''}${idx < params.length - 1 ? ', ' : ''}`)}) ${() => (ctx.returnType ? wgsl`-> ${ctx.returnType}` : '')}`;
```
reczkok: Maybe a little formatting help here would help readability (this should not break the produced code **(?)**)
```suggestion
    const header = wgsl`
      (
        ${params.map(
          (p, idx) =>
            wgsl`
            ${p.name}: ${ctx.argTypes[idx] ?? ''}${idx < params.length - 1 ? ', ' : ''}
          `,
        )}
      ) ${() => (ctx.returnType ? wgsl`-> ${ctx.returnType}` : '')}`;
```

packages/typegpu/src/wgslSlot.ts
```diff
@@ -53,4 +55,11 @@ class TgpuSlotImpl<T> implements TgpuResolvable, TgpuSlot<T> {
   toString(): string {
     return `slot:${this.label ?? '<unnamed>'}`;
   }
+
+  get value(): ValueOf<T> {
+    if (!inGPUMode()) {
+      throw new Error(`Cannot access wgsl.slot's value directly in JS.`);
```
mhawryluk: why not?

iwoplaza: Because slot values are available at resolve time, and can differ between nodes in the resolution tree. We currently have no mechanism for running TGSL functions in JS with specific slot values in their context, but if this changes, then we can remove this error throwing.

mhawryluk: okie, makes sense ^^

packages/tgsl-tools/src/parsers.ts
mhawryluk: you can map `params` and pass it to our `valueList` util

apps/typegpu-docs/src/content/examples/simple/gradient-tiles-tgsl.ts
```diff
@@ -0,0 +1,123 @@
...
+  min: 1,
+  max: 16,
+  step: 1,
+});
+
+const spanPlum = wgsl.plum((get) => ({ x: get(xSpanPlum), y: get(ySpanPlum) }));
+const spanBuffer = wgsl
+  .buffer(struct({ x: u32, y: u32 }), spanPlum)
+  .$usage(tgpu.Uniform)
+  .$name('span');
```
reczkok: Use the new API

reczkok (commented):
Looks good so far 👍 Left some comments to consider

reczkok (approved):
Great job! 👏

PR #214, opened 2024-08-28
Make BufferUsage unnamable + create WgslNamable interface

### LINKED ISSUE #144: Make BufferUsages unnamable.
There is no sense in naming buffer usages, since they are not created with the usual `wgsl.*` API.

What we can do now:
```ts
const fooBuffer = wgsl.buffer(...);
const fooUniform = fooBuffer.asUniform().$name('foo');
// asUniform() returns the same value as last time, so we are renaming
// the usage to a different name accidentally.
const anotherFooUniform = fooBuffer.asUniform().$name('another_foo');
```

All usages should inherit the debugLabel from their owner buffer.

```ts
const fooBuffer = wgsl.buffer(...).$name('foo');
const fooUniform = fooBuffer.asUniform();
const anotherFooUniform = fooBuffer.asUniform();
```

mhawryluk:
Closes #144

### DIFF
--- a/packages/typegpu/src/wgslBufferUsage.ts
+++ b/packages/typegpu/src/wgslBufferUsage.ts
@@ -47,7 +47,7 @@ class WgslBufferUsageImpl<TData extends AnyWgslData, TUsage extends BufferUsage>
   }
 
   resolve(ctx: ResolutionCtx): string {
-    const identifier = new WgslIdentifier();
+    const identifier = new WgslIdentifier().$name(this.label);
 
     ctx.addBinding(this, identifier);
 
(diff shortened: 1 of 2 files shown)


packages/typegpu/src/wgslBufferUsage.ts
```diff
@@ -35,26 +33,19 @@ export function bufferUsage<
...
     return this.buffer;
   }
 
-  $name(label: string | undefined) {
-    this._label = label;
-    return this;
-  }
-
   resolve(ctx: ResolutionCtx): string {
     const identifier = new WgslIdentifier();
```
iwoplaza:
```suggestion
    const identifier = new WgslIdentifier().$name(this.label);
```

PR #218, opened 2024-08-29
Rework vector data to resemble WGSL

iwoplaza:


packages/typegpu/src/builtinTypes.ts
```diff
@@ -0,0 +1,42 @@
...
+  [builtin.fragDepth]: f32,
+  [builtin.sampleIndex]: u32,
+  [builtin.sampleMask]: u32,
+  [builtin.fragment]: vec4f,
+  [builtin.localInvocationId]: vec3u,
+  [builtin.localInvocationIndex]: u32,
+  [builtin.globalInvocationId]: vec3u,
+  [builtin.workgroupId]: vec3u,
+  [builtin.numWorkgroups]: vec3u,
+};
```
reczkok: I know it's not that relevant right now but this should use the new arrays.
```suggestion
import {
  type F32,
  type WgslArray,
  type U32,
  type Vec3u,
  type Vec4f,
  arrayOf,
  f32,
  u32,
  vec3u,
  vec4f,
} from './data';
import { builtin } from './wgslBuiltin';

export type BuiltInPossibleTypes = U32 | F32 | Vec3u | Vec4f | WgslArray<U32>;

export const builtinToType: Record<symbol, BuiltInPossibleTypes> = {
  [builtin.vertexIndex]: u32,
  [builtin.instanceIndex]: u32,
  [builtin.position as unknown as symbol]: vec4f,
  [builtin.clipDistances]: arrayOf(u32, 8),
  [builtin.frontFacing]: f32,
  [builtin.fragDepth]: f32,
  [builtin.sampleIndex]: u32,
  [builtin.sampleMask]: u32,
  [builtin.fragment]: vec4f,
  [builtin.localInvocationId]: vec3u,
  [builtin.localInvocationIndex]: u32,
  [builtin.globalInvocationId]: vec3u,
  [builtin.workgroupId]: vec3u,
  [builtin.numWorkgroups]: vec3u,
};
```

packages/typegpu/src/data/vector.ts
```diff
@@ -0,0 +1,618 @@
...
+    this.make = make;
+    this.makeFromScalar = makeFromScalar;
+  }
+
+  _call(...args: number[]): T {
+    const values = args; // TODO: Allow users to pass in vectors that fill part of the values.
+
+    if (values.length <= 1) {
+      return this.makeFromScalar(values[0]);
+    }
```
reczkok: I know this isn't an error as `makeFromScalar()` accepts undefined but to me it was a little confusing. Maybe break it down into the `length === 0` case?

apps/typegpu-docs/src/content/examples/rendering/box-raytracing.ts
```diff
@@ -13,17 +13,24 @@ import {
...
-import { asReadonly, asUniform, builtin, createRuntime, wgsl } from 'typegpu';
 import { arrayOf, bool, f32, struct, u32, vec3f, vec4f } from 'typegpu/data';
+import {
+  asReadonly,
+  asUniform,
+  builtin,
+  createRuntime,
+  std,
+  wgsl,
+} from 'typegpu/experimental';
```
mhawryluk, line 24: Zrzut ekranu 2024-09-2 o 17 26 15

`'typegpu/experimental': ['typegpu/dist/experimental/index.d.ts'],` missing in `setCompilerOptions` in `CodeEditor.tsx`

packages/typegpu/src/builtinTypes.ts
```diff
@@ -0,0 +1,42 @@
...
+  | U32
+  | F32
+  | Vec3u
+  | Vec4f
+  | WgslData<TB.Unwrap<U32>[]>;
+
+export const builtinToType: Record<symbol, BuiltInPossibleTypes> = {
+  [builtin.vertexIndex]: u32,
+  [builtin.instanceIndex]: u32,
+  [builtin.position as unknown as symbol]: vec4f,
```
mhawryluk: why? 🧐 isn't `position` already a symbol?

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -0,0 +1,164 @@
...
+      );
+    },
+    vec3u: (a: vec3u, b: vec3u) => {
+      return vec3u(
+        a.y * b.z - a.z * b.y,
+        a.z * b.x - a.x * b.z,
+        a.x * b.y - a.y * b.x,
+      );
+    },
+  } as Record<VecKind, <T extends vecBase>(a: T, b: T) => T>,
```
reczkok: This should only accept vec3

packages/typegpu/src/std.ts
```diff
@@ -0,0 +1,19 @@
...
+export const std = {
+  add: <T extends vecBase>(lhs: T, rhs: T): T =>
+    VectorOps.add[lhs.kind](lhs, rhs),
+  sub: <T extends vecBase>(lhs: T, rhs: T): T =>
+    VectorOps.sub[lhs.kind](lhs, rhs),
+  mul: <T extends vecBase>(s: number, v: T): T => VectorOps.mul[v.kind](s, v),
+  dot: <T extends vecBase>(lhs: T, rhs: T): number =>
+    VectorOps.dot[lhs.kind](lhs, rhs),
+  normalize: <T extends vecBase>(v: T): T => VectorOps.normalize[v.kind](v),
+  cross: <T extends vecBase>(a: T, b: T): T => VectorOps.cross[a.kind](a, b),
```
reczkok: Should only accept vec3. Right now it tells the user it's okey if he uses another length and throws
```
TypeError: v.cross[e.kind] is not a function
```

packages/typegpu/src/data/vector.ts
```diff
@@ -0,0 +1,618 @@
...
+  get yy(): T2 {
+    return this.make2(this.y, this.y);
+  }
+}
+
+abstract class vec2Impl<T2> extends Swizzle2Impl<T2> implements vec2 {
+  readonly length = 2;
+
+  constructor(
+    private _x: number,
```
mhawryluk: I don't see a point in making these private fields, then defining getters and setters that don't do anything different than if x and y were public. even when we add support for r,g,b in the future, we can define those as getters/setter using public xyz

iwoplaza: Definitely, was an oversight on my part. Thanks!

PR #229, opened 2024-08-30
Add unmanaged buffers

### LINKED ISSUE #225: Create unmanaged buffers.
When assigning a device to a `tgpu.buffer`, it is an "unmanaged" buffer, meaning it is up to the user to manage that memory (not managed or owned by the runtime).

There should exist a new property on these buffers, `buffer`, which is a get property. It should throw an error if the buffer has not been passed a `device` (is managed). If the buffer is unmanaged, it should either return the buffer that was passed to it when calling `tgpu.buffer`, or create and cache one that fits the description.

There should exist a `dispose()` method that cleans up any cached (or passed in) buffer.

reczkok:
closes #225

### DIFF
--- a/packages/typegpu/src/wgslBuffer.ts
+++ b/packages/typegpu/src/wgslBuffer.ts
@@ -229,13 +229,15 @@ function asUsage<
   TUsage extends BufferUsage,
   TType extends AllowVertex | AllowUniform | AllowReadonly | AllowMutable,
 >(usage: TUsage, _: TType) {
-  return <TData extends AnyWgslData>(buffer: WgslBuffer<TData> & TType) => {
+  return <TData extends AnyWgslData>(
+    buffer: WgslBuffer<TData> & TType,
+  ): WgslBufferUsage<TData, TUsage> => {
     if (buffer._usages[usage] === null) {
       throw new Error(
         `Cannot pass ${buffer} to as${capitalizeFirstLetter(usage)} function, as the buffer does not allow ${usage} usage. To allow it, use $allow${capitalizeFirstLetter(usage)} WgslBuffer method.`,
       );
     }
-    return buffer._usages[usage];
+    return buffer._usages[usage] as WgslBufferUsage<TData, TUsage>;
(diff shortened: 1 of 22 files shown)


packages/typegpu/src/wgslBuffer.ts
```diff
@@ -76,16 +115,57 @@ class WgslBufferImpl<
...
-    public readonly initial?: Parsed<TData> | WgslPlum<Parsed<TData>>,
-  ) {}
+    public readonly initialOrBuffer?:
+      | Parsed<TData>
+      | WgslPlum<Parsed<TData>>
+      | GPUBuffer
+      | undefined,
+  ) {
+    if (isGPUBuffer(initialOrBuffer)) {
+      this._buffer = initialOrBuffer as GPUBuffer;
```
mhawryluk: unnecessary `as`

reczkok:
```suggestion
      this._buffer = initialOrBuffer;
```

packages/typegpu/src/wgslBuffer.ts
```diff
@@ -13,137 +14,185 @@ import type { WgslPlum } from './wgslPlum';
...
-  TData extends AnyWgslData,
-  TAllows extends BufferUsage = never,
-> = {
-  uniform: WgslBufferUsage<TData, TAllows | 'uniform'> | null;
-  mutable: WgslBufferUsage<TData, TAllows | 'mutable'> | null;
-  readonly: WgslBufferUsage<TData, TAllows | 'readonly'> | null;
-  vertex: WgslBufferUsage<TData, TAllows | 'vertex'> | null;
+export interface Unmanaged {
+  get device(): GPUDevice;
+  get buffer(): GPUBuffer;
```
iwoplaza: Tiny nitpick, from the user's perspective its just a readonly property.

```suggestion
  readonly device: GPUDevice;
  readonly buffer: GPUBuffer;
```

Issue #234, opened 2024-08-30
Rename types to match the new convention.


iwoplaza:
```ts
WgslBuffer -> TgpuBuffer
WgslArray -> TgpuArray
WgslStruct -> TgpuStruct
WgslDataCustomAligned -> TgpuAligned // (more terse babyyyy ✨)
WgslDataCustomSized -> TgpuSized // (more terse babyyyy ✨)
// ...
```

PR #237, opened 2024-09-02
Rename code from Wgsl to Tgpu

mhawryluk:
references #234

### DIFF
--- a/apps/typegpu-docs/src/content/examples/simulation/fluid-double-buffering.ts
+++ b/apps/typegpu-docs/src/content/examples/simulation/fluid-double-buffering.ts
@@ -15,16 +15,6 @@ import {
 // --
 
 import {
-  type TgpuBufferUsage,
-  type Wgsl,
-  asMutable,
-  asReadonly,
-  asUniform,
-  builtin,
-  createRuntime,
-  wgsl,
-} from 'typegpu/experimental';
-import {
   type Parsed,
   arrayOf,
   f32,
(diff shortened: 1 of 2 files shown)


packages/typegpu/src/data/size.ts
```diff
@@ -8,31 +8,29 @@ import {
...
 
-export interface WgslDataCustomSized<
-  TSize extends number,
-  TData extends AnyWgslData,
-> extends WgslData<Unwrap<TData>> {}
+export interface TgpuSized<TSize extends number, TData extends AnyTgpuData>
+  extends TgpuData<Unwrap<TData>> {}
 
-export class WgslDataCustomSizedImpl<
+export class TgpuDataCustomSizedImpl<
```
iwoplaza:
```suggestion
export class TgpuSizedImpl<
```

packages/typegpu/src/types.ts
```diff
@@ -32,24 +32,24 @@ export interface ResolutionCtx {
...
+export function isResolvable(value: unknown): value is TgpuResolvable {
   return (
     !!value &&
     (typeof value === 'object' || typeof value === 'function') &&
     'resolve' in value
   );
 }
 
-export function isWgsl(value: unknown): value is Wgsl {
+export function isTgpu(value: unknown): value is Wgsl {
```
iwoplaza:
```suggestion
export function isWgsl(value: unknown): value is Wgsl {
```

PR #263, opened 2024-09-05
Add a way to destroy Tgpu buffers

reczkok:


packages/typegpu/src/wgslBuffer.ts
```diff
@@ -135,12 +141,16 @@ class TgpuBufferImpl<TData extends AnyTgpuData> implements TgpuBuffer<TData> {
   get device() {
     if (!this._device) {
       throw new Error(
-        'This buffer is managed by TypeGPU and cannot be used directly',
+        'This buffer has not been assigned a device. Use $device to assign a device',
```
iwoplaza: A small suggestion, we can make it more explicit.

```suggestion
        'This buffer has not been assigned a device. Use .$device(device) to assign a device',
```

PR #267, opened 2024-09-05
Implemented the `.$usage()` API for buffers.

iwoplaza:


packages/typegpu/tests/asUsage.test.ts
```diff
@@ -23,36 +22,48 @@ global.GPUBufferUsage = {
 
 describe('asUsage', () => {
   it('allows creating bufferUsages only for buffers allowing them', () => {
-    asReadonly(wgsl.buffer(u32, 2).$allowReadonly());
-    asReadonly(wgsl.buffer(u32, 2).$allowReadonly().$allowUniform());
-    asReadonly(wgsl.buffer(u32, 2).$allowVertex('vertex').$allowReadonly());
+    asReadonly(tgpu.createBuffer(u32, 2).$usage(tgpu.Storage));
```
reczkok, line 25: Should the tests be wrapped in `expect().not.toThrow()` or does it not matter?

packages/typegpu/src/wgslBuffer.ts
```diff
@@ -72,6 +70,24 @@ export function buffer<TData extends AnyTgpuData>(
   return new TgpuBufferImpl(typeSchema, initialOrBuffer);
 }
 
+export function usableAsUniform<T extends TgpuBuffer<AnyTgpuData>>(
```
mhawryluk: I would name those `isUsableAsUniform` etc. to match convention of `isWgsl` and such

packages/typegpu/tests/asUsage.test.ts
```diff
@@ -23,36 +22,48 @@ global.GPUBufferUsage = {
...
+    asVertex(
+      tgpu.createBuffer(u32, 2).$usage(tgpu.Storage, tgpu.Vertex),
+      'instance',
+    );
     // @ts-expect-error
-    expect(() => asVertex(wgsl.buffer(u32, 2))).toThrow();
+    expect(() => asVertex(tgpu.createBuffer(u32, 2))).toThrow();
     // @ts-expect-error
-    expect(() => asVertex(wgsl.buffer(u32, 2).$allowMutable())).toThrow();
+    expect(() => asVertex(tgpu.createBuffer(u32, 2).$allowMutable())).toThrow();
```
mhawryluk:
```suggestion
    expect(() => asVertex(tgpu.createBuffer(u32, 2).$usage(tgpu.Storage))).toThrow();
```

packages/typegpu/tests/asUsage.test.ts
mhawryluk: imo as long as we don't specify what error exactly we expect, it's just a stylistic choice, cause it's going to work the same either way. using `expect().not.toThrow()` makes all assertions look consistent, but introduces more code that isn't necessary

PR #268, opened 2024-09-06
Swizzle rework using Proxy

reczkok:


packages/typegpu/src/data/vector.ts
```diff
@@ -444,9 +802,9 @@ export const vec3f = makeVecSchema({
   byteAlignment: 16,
   length: 3,
   label: 'vec3f',
-  make: (x, y, z) => new vec3fImpl(x, y, z),
-  makeFromScalar: (x) => new vec3fImpl(x, x, x),
-}) as Vec3f;
+  make: (x, y, z) => new Proxy(new vec3fImpl(x, y, z), vecProxyHandler),
+  makeFromScalar: (x) => new Proxy(new vec3fImpl(x, x, x), vecProxyHandler),
+}) as unknown as Vec3f;
```
iwoplaza: Are these `as unknown` necessary after the newest changes?
```suggestion
}) as Vec3f;
```

iwoplaza (approved):
Great work! 💚
(One small nitpick)

packages/typegpu/src/data/vector.ts
reczkok: I think they are? 
```
Type 'TgpuData<vecBase> & { expressionCode: string; } & ((...args: number[]) => vecBase)' 
is not comparable to type 'TgpuData<vec4u>'.
  Types of property '__unwrapped' are incompatible.
  Type 'vecBase' is missing the following properties from type 'vec4u': x, y, z, w, and 299 more.
```

iwoplaza: Hmm, I experimented a bit and it should work if we do this:

```ts
  make: (x: number, y: number) =>
    new Proxy(new vec2fImpl(x, y), vecProxyHandler) as vec2f,
  makeFromScalar: (x) =>
    new Proxy(new vec2fImpl(x, x), vecProxyHandler) as vec2f,
}) as Vec2f;
```

iwoplaza (approved):
Awesome!
(  ◦°^°◦) 💚

PR #269, opened 2024-09-06
Separated the dev scripts into a separate package, so it can be used throughout the monorepo.

iwoplaza:


packages/tgpu-dev-cli/log.mjs
```diff
@@ -0,0 +1 @@
+export const Frog = '(  ◦°^°◦)';
```
reczkok, line 1: 🐸

PR #294, opened 2024-09-10
Changed image-blur refresh logic

### LINKED ISSUE #280: Fix image blur refreshing
At the moment the blur example has it's logic inside OnFrame() which is unnecessary and resource intensive. The blur pipeline should be performed only when parameters change.

reczkok:
Closes #280


iwoplaza (approved):
Awesome sauce 🍅 ❤️

PR #301, opened 2024-09-11
Fix build on main after tgsl poc merge

mhawryluk:
we can bring back gradient tiles (tgsl) example after implementing callable primitive type constructors, fixing builtins usage in tgsl code and implementing render function on runtime.


iwoplaza (approved):
All good changes, thanks! 💚 🐸

PR #309, opened 2024-09-11
Fix write size big

### LINKED ISSUE #307: Wrong size inference when writing to buffer
```ts
import tgpu from 'typegpu';
import { vec3f } from 'typegpu/data';

const adapter = await navigator.gpu.requestAdapter();
const device = await adapter?.requestDevice();

if (!device ) throw new Error();

const b = tgpu.createBuffer(vec3f).$device(device);

tgpu.write(b, vec3f(2,2,2))
```
Results in:
```
Write range (bufferOffset: 0, size: 16) does not fit in [Buffer (unlabeled)] size (12).
 - While calling [Queue].WriteBuffer([Buffer (unlabeled)], (0 bytes), data, (16 bytes))
```

reczkok:
# Description in #307 
closes #307


iwoplaza:
I am wondering if we could simplify this fix by always allocating a buffer rounded up to the alignment of the data-type. This would remove the need to check if the type we are writing to is a primitive, but would allocate more memory than necessary in some cases.

reczkok:
The only case when this would make a difference is when we have a buffer of just `vec3`. I don't think this would be a big deal but this would also mean our `vec3` buffer are not the same size than WebGPU's. I think this could lead to some migration bugs.

iwoplaza:
What if we always create buffers that are aligned, but write the minimum amount of bytes (unaligned). This should cover the case of any existing WebGPU buffer, and our own allocations. Would this raise any WebGPU errors?

PR #310, opened 2024-09-11
Fix write size 

### LINKED ISSUE #307: Wrong size inference when writing to buffer

reczkok:
closes #307

### DIFF
--- a/packages/typegpu/src/createRuntime.ts
+++ b/packages/typegpu/src/createRuntime.ts
@@ -246,17 +246,32 @@ class TgpuRuntimeImpl implements TgpuRuntime {
   ) {
     const gpuBuffer = this.bufferFor(allocatable);
 
-    const size = allocatable.dataType.size;
+    const size = roundUp(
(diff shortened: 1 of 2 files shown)


packages/typegpu/src/createRuntime.ts
iwoplaza, line 1: Does the host buffer need to be of the aligned size? If not, we could assign `allocatable.dataType.size` to `const size`.

PR #316, opened 2024-09-12
Add tags to examples and hide those marked as `experimental`

reczkok:


apps/typegpu-docs/src/components/ExampleList.tsx
iwoplaza, line 1: I wanted to bring back categories someday, especially when the amount of examples is going to increase, so maybe we could filter the examples as they are being rendered?

```ts
        {exampleCategories.map((category) =>
          (examplesByCategory[category.key] ?? []).map((example) => {
            if (example.metadata.tags?.some((tag) => excludeTags.includes(tag))) {
              return null;
            }
            
            return (
              <ExampleLink key={example.key} exampleKey={example.key}>
                {example.metadata.title}
              </ExampleLink>
            );
          }),
        )}
```

iwoplaza (approved):
Apart from the single nitpick, awesome work! 💚 🐸

iwoplaza (approved):
(  ◦°^°◦) 🌺

PR #317, opened 2024-09-12
Control button definitions via comments + phase 1 gradient tiles variant

mhawryluk:
so far it works (i think), but isn't pretty

### DIFF
--- a/apps/typegpu-docs/src/components/ExampleView.tsx
+++ b/apps/typegpu-docs/src/components/ExampleView.tsx
@@ -115,8 +115,7 @@ export function ExampleView({ example, isPlayground = false }: Props) {
 
       <div className="grid gap-4 grid-cols-[1fr_18.75rem] h-full">
         <div
-          className={cs('grid gap-4', codeEditorShowing ? 'grid-rows-2' : '')}
-        >
+          className={cs('grid gap-4', codeEditorShowing ? 'grid-rows-2' : '')}>
           <div
             style={{
               scrollbarGutter: 'stable both-edges',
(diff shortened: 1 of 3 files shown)


apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -126,6 +126,8 @@ export function ExampleView({ example, isPlayground = false }: Props) {
               codeEditorShowing ? 'max-h-[calc(50vh-3rem)]' : '',
             )}
           >
+            {def.elements.length === 0 ? <Canvas aspectRatio={1} /> : null}
```
iwoplaza, line 130: Could you add a comment here to signal that this is a temporary solution and why? For the future generations (us most likely) 🧒

```suggestion
            {/* Note: This is a temporary measure to allow for simple examples that do not require the @typegpu/example-toolkit package. */}
            {def.elements.length === 0 ? <Canvas aspectRatio={1} /> : null}
```

apps/typegpu-docs/src/content/examples/simple/gradient-tiles-typedbuffers.ts
```diff
@@ -0,0 +1,171 @@
...
+export function spanYMinus() {
+  spanY = Math.max(1, spanY - 1);
+  draw(spanX, spanY);
+}
+
+// @button "span y ++"
+export function spanYPlus() {
+  spanY = Math.min(spanY + 1, 20);
+  draw(spanX, spanY);
+}
```
iwoplaza: Imo doc comments are a little more suitable, as they have the whole @prop convention.

```suggestion
/** @button "↔️ -" */
export function spanXMinus() {
  spanX = Math.max(1, spanX - 1);
  draw(spanX, spanY);
}

/** @button "↔️ +" */
export function spanXPlus() {
  spanX = Math.min(spanX + 1, 20);
  draw(spanX, spanY);
}

/** @button "↕️ -" */
export function spanYMinus() {
  spanY = Math.max(1, spanY - 1);
  draw(spanX, spanY);
}

/** @button "↕️ +" */
export function spanYPlus() {
  spanY = Math.min(spanY + 1, 20);
  draw(spanX, spanY);
}

```

iwoplaza (approved):
Awesome 🤩

iwoplaza (approved):
🐸 💚

PR #331, opened 2024-09-16
Change redirection rules

reczkok:


mhawryluk (changes requested):
1. create filtered out map of examples once globally
2. in RedirectToFlagship component check if FLAGSHIP is in the filtered examples
3A. if yes set flagship
3B. if not set random from filtered examples

iwoplaza (approved):
Let's goooo 🚀

mhawryluk (approved):
Awesome! 😎

PR #348, opened 2024-09-17
Fix vertex shader types and tests

### LINKED ISSUE #345: Fix vertex buffer type checking
Make vertex buffers properly check if they are one on the supported types. Currently it only works for vectors and primitives - it should also work for vectors.

reczkok:
closes #345

### DIFF
--- a/packages/typegpu/tests/runtime.test.ts
+++ b/packages/typegpu/tests/runtime.test.ts
@@ -1,8 +1,13 @@
 import { describe, expect, it, vi } from 'vitest';
 import { afterEach } from 'vitest';
 import { createRuntime } from '../src/createRuntime';
-import { struct, u32, vec3i, vec4u } from '../src/data';
-import tgpu, { asReadonly, asUniform, wgsl } from '../src/experimental';
+import { arrayOf, struct, u32, vec2f, vec3f, vec3i, vec4u } from '../src/data';
+import tgpu, {
+  asReadonly,
+  asUniform,
+  asVertex,
+  wgsl,
+} from '../src/experimental';
 import { plum } from '../src/tgpuPlum';
 
 global.GPUBufferUsage = {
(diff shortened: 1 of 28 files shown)


mhawryluk (commented):
looks good to me! let's add some tests to make sure it works from now on

reczkok:
# TLDR
In the lates commit I also reworked the `unmanagedIO` and `runtime` tests. The previous implementation was less than optimal and made it difficult to write complex tests. Now they should be a little easier to work with 😁

# Details
Previously when creating mocks I used the following method
```ts
const mockCommandEncoder = vi.fn(() => ({
  beginComputePass: vi.fn(() => mockComputePassEncoder()),
  beginRenderPass: vi.fn(),
  ...
}));

const mockDevice = vi.fn(() => ({
  createBindGroup: vi.fn(() => 'mockBindGroup'),
  createBindGroupLayout: vi.fn(() => 'mockBindGroupLayout'),
  createBuffer: vi.fn(() => mockBuffer()),
  createCommandEncoder: vi.fn(() => mockCommandEncoder()),
  ...
});
```
Every mock was essentially a mock creator (i.e. each call to `mockDevice()` (or any `mockSth`) would create a new Mock object. This meant extracting info about calls from nested Mocks was very inconvenient as there were no handles to them.
Now all mocks are objects so we have all the necessary handles. Using `beforeEach()` and `restoreAllMocks()` now has sense and properly resets mocks. All this should mean tests that are easier to understand and expand🫡

mhawryluk (approved):
Great!!

packages/typegpu/tests/runtime.test.ts
```diff
@@ -257,4 +270,163 @@ describe('TgpuRuntime', () => {
...
+    });
+  });
+
+  it('should throw an error when trying to create an invalid vertex buffer', async () => {
+    const bufferData = tgpu
+      .createBuffer(
+        struct({
+          i: vec2f,
+          should: vec3f,
+          throw: u32,
```
iwoplaza, line 410: 💨🏈

iwoplaza (approved):
Awesome! 🔧 👷

PR #353, opened 2024-09-18
impr: Added intermediate AST to allow Vite plugins to parse JS as a build step.

iwoplaza:
### Things working:
- `@typegpu/tgsl-tools` now tranpiles a sub-set of JS into SMoL instead of WGSL directly, meaning it does not need to know the runtime context.
- `@typegpu/jit` uses tgsl-tools to transpile JS into SMoL at runtime.
- `typegpu` is now responsible for generating WGSL from SMoL. This makes it very easy for the future build plugin, as all it has to do is inject SMoL in JSON form so that it can be used at runtime.

### Things left to do:
- Calls to functions have to accept `Resource`s in order to adjust their functionality based on the type of arguments.
- The function call epxression can deduce the type by accessing `fn.shell.returnType`.
- Types of expressions are always unknown, this can be changed. Every operation has to correctly infer the type based on its arguments, and pass it along.
- The `.value` property of some WgslResolvables has to be a proxy that handles arbitrary member access, and returns another proxied resolvable that resolves the whatever the parent resolves to + `.${key}` of the member access operation.


packages/tgsl-tools/src/parsers.ts
```diff
@@ -1,172 +1,220 @@
...
   },
 
-  BinaryExpression(ctx, node) {
-    // TODO: Verify if all binary operators map 1-to-1 (they probably do not)
-    return wgsl`${generate(ctx, node.left)} ${node.operator} ${generate(ctx, node.right)}`;
+  BinaryExpression(node) {
+    const wgslOp = BINARY_OP_MAP[node.operator];
+    const left = transpile(node.left);
+    const right = transpile(node.right);
+    return { x2: [left, wgslOp, right] } as smol.AnyNode;
```
mhawryluk: why do we name this property `x2`? i may be missing something, but i don't think it's clear

packages/tgsl-tools/src/parsers.ts
```diff
@@ -1,172 +1,220 @@
...
 
-  LogicalExpression(ctx, node) {
-    const left = generate(ctx, node.left);
-    const right = generate(ctx, node.right);
-    return wgsl`${left} ${node.operator} ${right}`;
+  LogicalExpression(node) {
+    const wgslOp = LOGICAL_OP_MAP[node.operator];
+    const left = transpile(node.left);
+    const right = transpile(node.right);
+    return { x2: [left, wgslOp, right] } as smol.AnyNode;
```
mhawryluk: why not `as smol.LogicalExpression`? same for `BinaryExpression` and `AssignmentExpression`

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -123,8 +126,25 @@ type SlotBindingLayer = {
...
 };
 
+type FunctionScopeLayer = {
+  type: 'functionScope';
+  args: Resource[];
+  externalMap: Record<string, Wgsl>;
+  returnType: AnyTgpuData | undefined;
+};
+
+type BlockScopeLayer = {
```
mhawryluk, line 136: block scope layer is never pushed to the stack

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -173,13 +206,88 @@ class ItemStateStack {
...
+    }
+
+    let str = '';
+    let i = this.identLevel;
+    while (i > 8) {
+      str += INDENT[8];
+      i >>= 3;
+    }
+
+    return str;
```
mhawryluk: shouldn't that be `str + INDENT[i]`?

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -191,6 +299,7 @@ export class ResolutionCtxImpl implements ResolutionCtx {
   >();
 
   private readonly _shared: SharedResolutionState;
+  private readonly _identController = new IndentController();
```
mhawryluk:
```suggestion
  private readonly _indentController = new IndentController();
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -214,6 +323,67 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
+
+    this._itemStateStack.pushFunctionScope(
+      args,
+      fn.shell.returnType,
+      externalMap,
+    );
+    const str = generateFunction(this, body);
+    this._itemStateStack.pop();
+
+    // TODO: Actually generate WGSL from SMoL
```
mhawryluk:
```suggestion
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -214,6 +323,67 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
+    return {
+      head:
+        fn.shell.returnType !== undefined
+          ? `(${args.map((arg) => `${arg.value}: ${this.resolve(arg.dataType)}`)}) -> ${this.resolve(fn.shell.returnType)}`
```
mhawryluk: please extract this common argument map logic to a constant, cause it's the same in the next line

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -214,6 +323,67 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
+    // TODO: Provide data type information
+    return (
+      this._itemStateStack.getResourceById(id) ?? {
+        value: id,
+        dataType: UnknownData,
+      }
+    );
+  }
+
+  transpileFn(
```
mhawryluk: I suggest naming this function something different to distinguish the `transpile` functions as the ones that change JS to SMoL. this function changes tgpu to WGSL. maybe `resolveFn`?

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -0,0 +1,233 @@
...
+      // TODO: Pass in resources, not just values.
+      const result = (idValue as unknown as (...args: unknown[]) => unknown)(
+        ...argValues,
+      ) as Wgsl;
+      // TODO: Make function calls return resources instead of just values.
+      return { value: result, dataType: UnknownData };
+    }
+
+    return {
+      value: `${String(idValue)}(${argValues.join(', ')})`,
```
mhawryluk, line 150: use `valueList` if possible

packages/tgsl-tools/src/parsers.ts
iwoplaza: It was supposed to encompass any binary expression (x - expression, 2 - binary). I do agree that it is not very clear, but the format will not be read by people anyways, so brevity is the focus. Do you have other suggestions? Maybe `bi` or just `b`?

packages/tgsl-tools/src/parsers.ts
iwoplaza: Fixing it rn! 🙌

packages/tgsl-tools/src/parsers.ts
mhawryluk: no, I couldn't think of anything better tbh and I see that we want to prioritize the format being concise. so I'm okay with `x2`. thanks for the explanation

packages/typegpu/src/resolutionCtx.ts
iwoplaza: I added a comment in the wgsl generation code to implement this, thanks 💜

packages/typegpu/src/resolutionCtx.ts
iwoplaza: It should, good catch! 💜

packages/typegpu/src/smol/wgslGenerator.ts
iwoplaza: `valueList` produces a resolvable from a list of resolvables. `argValues` is a list of strings, so it is more compact and faster to join directly.

mhawryluk (approved):
Great job!! 😎

PR #356, opened 2024-09-19
Docs for how a migration from plain WGSL into typed WGSL could look like.

iwoplaza:


apps/typegpu-docs/src/content/docs/guides/typed-wgsl.mdx
```diff
@@ -0,0 +1,659 @@
...
+
+export const Flip = d.struct({
+  value: d.u32,
+});
+
+export const layout0 = tgpu.bindGroupLayout({
+  samp: { type: 'sampler' },
+  params: { type: 'uniform', data: Params },
+});
+const { samp, params } = layout0.bound;
```
mhawryluk, line 118: maybe add a comment to signify what these are (buffer usage "slots"?)

mhawryluk (approved):
the API looks very good so far imo! 🎉

apps/typegpu-docs/src/content/docs/guides/typed-wgsl.mdx
iwoplaza, line 118:
```suggestion
// Both `samp` and `params` can be used in code, as if they were already bound to
// some buffer, but they act more like slots.
const { samp, params } = layout0.bound;
```

PR #368, opened 2024-09-19
Make structures impose byte alignment, instead of the primitives doing it themselves

### LINKED ISSUE #365: Make structures and arrays impose byte alignment, instead of the primitives doing it themselves.
This is because when `loose` data types are introduced, we want to allow them to pack `vec3f` together tightly.

mhawryluk:
closes #365


iwoplaza:
Could you add tests to check if structs align `vec3f` properties? 👀

iwoplaza (approved):
Awesome job! 🌟💛

PR #370, opened 2024-09-20
The Root API

iwoplaza:
### Why?
- To ease people into using a runtime-like API (the runtime is being rebranded to root).
- To minimize having to pass `device` everywhere at the unmanaged phase.
- To simplify the read and write APIs

### Still left to do in a separate PR
- Rename `TgpuRuntime` to `TgpuRoot` (same with implementation)
- Rename `createRuntime.ts` to `init.ts`
- Rename `runtime.test.ts` to `root.test.ts`


packages/typegpu/src/legacyBufferApi.ts
```diff
@@ -0,0 +1,33 @@
+import type { Parsed } from 'typed-binary';
+import { type TgpuBuffer, createBufferImpl } from './tgpuBuffer';
+import type { TgpuPlum } from './tgpuPlumTypes';
+import type { AnyTgpuData } from './types';
+
+/**
+ * @deprecated Use `root.createBuffer` instead.
```
mhawryluk: will it be clear what `root` is in this context? can we specify that it's a result of the init function?

packages/typegpu/src/types.ts
```diff
@@ -114,7 +115,7 @@ export interface TgpuBindable<
   TData extends AnyTgpuData = AnyTgpuData,
   TUsage extends BufferUsage = BufferUsage,
 > extends TgpuResolvable {
-  readonly allocatable: TgpuAllocatable<TData>;
+  readonly allocatable: TgpuBuffer<TData>;
```
mhawryluk, line 97: is `TgpuAllocatable` interface still useful? can't remember what other types of allocatables we were planning to introduce

mhawryluk (approved):
Amazing job! I think we're heading in the right direction 🎉

PR #376, opened 2024-09-20
Fix mobile preview positioning

mhawryluk:
Tiny fix.

Before:
Zrzut ekranu 2024-09-20 o 17 28 26

After:
Zrzut ekranu 2024-09-20 o 17 28 35


iwoplaza (approved):
🌮 💛

PR #378, opened 2024-09-23
Typed bind groups

iwoplaza:
Left to do in another PR:
- Runtime arrays


apps/typegpu-docs/src/content/examples/simulation/boids-typed-bindgroups.ts
```diff
@@ -0,0 +1,450 @@
...
+  )[0].view = context.getCurrentTexture().createView();
+
+  const commandEncoder = root.device.createCommandEncoder();
+  const computePass = commandEncoder.beginComputePass();
+  computePass.setPipeline(computePipeline);
+  computePass.setBindGroup(
+    0,
+    even
+      ? root.unwrap(computeBindGroups[0])
+      : root.unwrap(computeBindGroups[1]),
```
mhawryluk: `root.unwrap(computeBindGroups[even ? 0 : 1])`

same with render pass encoder below

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -0,0 +1,449 @@
...
+export function bindGroupLayout<
+  Entries extends Record<string, TgpuLayoutEntry | null>,
+>(entries: Entries): TgpuBindGroupLayout<Entries> {
+  return createBindGroupLayout(entries);
+}
+
+export function isBindGroupLayout<T extends TgpuBindGroupLayout>(
+  value: T | unknown,
+): value is T {
+  return (value as T).resourceType === 'bind-group-layout';
```
mhawryluk: maybe `value && (value as T).resourceType === 'bind-group-layout'`?
so that it doesn't crash for null and undefined
same below

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -0,0 +1,449 @@
...
+
+class TgpuBindGroupLayoutImpl<
+  Entries extends Record<string, TgpuLayoutEntry | null>,
+> implements TgpuBindGroupLayout<Entries>
+{
+  private _label: string | undefined;
+
+  public readonly resourceType = 'bind-group-layout' as const;
+
+  // TODO: Fill bound values.
```
mhawryluk, line 228: in another PR, si?

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -0,0 +1,449 @@
...
+
+          if (
+            'texture' in entry ||
+            'storageTexture' in entry ||
+            'sampler' in entry ||
+            'externalTexture' in entry
+          ) {
+            return {
+              binding: idx,
+              resource: value as GPUSampler,
```
mhawryluk: `as GPUSampler` to make TS not complain, but the type is inferred correctly regardless?

and so far we only support raw WebGPU samplers and textures here, not our tgpu objects, right?

packages/typegpu/src/tgpuBufferUsage.ts
```diff
@@ -23,38 +23,55 @@ import type {
...
 }
 
+export function isBufferUsage<
+  T extends
+    | TgpuBufferUniform<AnyTgpuData>
+    | TgpuBufferReadonly<AnyTgpuData>
+    | TgpuBufferMutable<AnyTgpuData>
+    | TgpuBufferVertex<AnyTgpuData>,
+>(value: T | unknown): value is T {
+  return (value as T).resourceType === 'buffer-usage';
```
mhawryluk: same concern for null and undefined

packages/typegpu/src/types.ts
```diff
@@ -14,6 +14,25 @@ export type Resource = {
...
+  T,
+  {
+    [Key in keyof T]: T[Key] extends Prop ? never : Key;
+  }[keyof T]
+>;
+
+/**
+ * Utility type that takes an object type and makes the hover overlay more readable.
+ */
+export type Prettify<T> = {
```
mhawryluk: what difference does it make? and where do we plan on using it? right now it's unused

packages/typegpu/tests/runtime.test.ts
```diff
@@ -8,25 +8,7 @@ import tgpu, {
...
-  INDIRECT: 256,
-  QUERY_RESOLVE: 512,
-};
-
-global.GPUShaderStage = {
-  VERTEX: 1,
-  FRAGMENT: 2,
-  COMPUTE: 4,
-};
+import './utils/webgpuGlobals';
```
mhawryluk, line 11: great idea, let's use it in other test files too

packages/typegpu/src/tgpuBindGroupLayout.ts
iwoplaza: Si 🌮

packages/typegpu/src/tgpuBindGroupLayout.ts
iwoplaza: Yes, we only support raw samplers and textures for now, until we are sure our tgpu equivalents also match 1-to-1 with WebGPU.

packages/typegpu/src/types.ts
iwoplaza: It was being used before, but no longer is. I will remove it for now.

mhawryluk (approved):
Great job! 😎

packages/typegpu/src/index.ts
```diff
@@ -11,8 +12,9 @@ export const tgpu = {
   Storage,
   Vertex,
 
-  createBuffer,
+  bindGroupLayout,
```
reczkok, line 16: Is the whitespace intentionally there for separation?

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -0,0 +1,452 @@
...
+export class MissingBindingError extends Error {
+  constructor(groupLabel: string | undefined, key: string) {
+    super(
+      `Bind group '${groupLabel ?? '<unnamed>'}' is missing a required binding '${key}'`,
+    );
+
+    // Set the prototype explicitly.
+    Object.setPrototypeOf(this, MissingBindingError.prototype);
+  }
+}
```
reczkok, line 201: Maybe move this to `errors`?

reczkok (approved):
Great work 🛠️💜

packages/typegpu/src/index.ts
iwoplaza: Yess, I wanted to separate the actually doing something methods, the resource-creating methods and the legacy methods.

packages/typegpu/src/tgpuBindGroupLayout.ts
iwoplaza: It was in errors at first, but I thought it would make more sense to keep this local to here, as it strictly relates to bind group layouts.

PR #386, opened 2024-09-24
Make builtins a hybrid type (symbol & data)

### LINKED ISSUE #371: Make builtins a hybrid type (symbol & data)
This should allow them to be put in struct definitions, returned from vertex/fragment functions, and used in TGSL code.

### Example of this refactor for `builtin.position`

```ts
export type BuiltinPosition = TgpuBuiltin & Vec4f;

export const builtin = {
  // ...
  position: new TgpuBuiltin('position') as BuiltinPosition,
  // ...
} as const;
```

```ts
import tgpu, { builtin } from 'typegpu';

builtin.position;
//      ^? BuiltinPosition

const VertexOutput = d.struct({
  Position: builtin.position,
  UV: d.vec2f,
}); // => TgpuStruct<{ Position: BuiltinPosition, UV: Vec2f }>

const foo = tgpu.procedure(() => {
  const x = builtin.position.x;
  //                         ^? number
});
```

mhawryluk:
closes #371


packages/typegpu/src/builtin.ts
```diff
@@ -1,45 +1,75 @@
...
+export type BuiltinFrontFacing = symbol & Bool;
+export type BuiltinFragDepth = symbol & F32;
+export type BuiltinSampleIndex = symbol & U32;
+export type BuiltinSampleMask = symbol & vec4f;
+export type BuiltinFragment = symbol & vec4f;
+export type BuiltinLocalInvocationId = symbol & vec3u;
+export type BuiltinLocalInvocationIndex = symbol & U32;
+export type BuiltinGlobalInvocationId = symbol & vec3u;
+export type BuiltinWorkgroupId = symbol & vec3u;
+export type BuiltinNumWorkgroups = symbol & vec3u;
```
iwoplaza: The builtins should be the value type, not the schema type, so they can be used as if they were just normal values in from TypeScript's point of view.

```suggestion
export type BuiltinVertexIndex = symbol & number;
export type BuiltinInstanceIndex = symbol & number;
export type BuiltinPosition = symbol & vec4f;
export type BuiltinClipDistances = symbol & number[];
export type BuiltinFrontFacing = symbol & boolean;
export type BuiltinFragDepth = symbol & number;
export type BuiltinSampleIndex = symbol & number;
export type BuiltinSampleMask = symbol & vec4f;
export type BuiltinFragment = symbol & vec4f;
export type BuiltinLocalInvocationId = symbol & vec3u;
export type BuiltinLocalInvocationIndex = symbol & number;
export type BuiltinGlobalInvocationId = symbol & vec3u;
export type BuiltinWorkgroupId = symbol & vec3u;
export type BuiltinNumWorkgroups = symbol & vec3u;
```

packages/typegpu/tests/builtin.test.ts
```diff
@@ -16,4 +22,30 @@ describe('builtin', () => {
...
+      d: builtin.localInvocationIndex,
+    };
+
+    type X = typeof x;
+    type Omitted = OmitSymbols<X>;
+
+    expectTypeOf(x as Omitted).toEqualTypeOf({
+      a: u32,
+      c: f32,
+    });
```
iwoplaza: The generic parameter can be used to test a specific type, not the type of a value.
```suggestion
    expectTypeOf<Omitted>().toEqualTypeOf({
      a: u32,
      c: f32,
    });
```

iwoplaza (changes requested):

mhawryluk:
apparently `symbol & <some primitive type>` is `never`. i'm afraid that's a big issue for us here

Zrzut ekranu 2024-09-25 o 10 14 42

iwoplaza:
> apparently `symbol & <some primitive type>` is `never`. i'm afraid that's a big issue for us here
> 
> Zrzut ekranu 2024-09-25 o 10 14 42

It is an issue, hmm. We could go about it a few different ways. The current use of Vertex and Fragment output structs makes it so we are not using builtins as keys, so they do not really have to be symbols anymore. We can make a Builtin class that has a parameter we can check at runtime, and then do `Builtin & number`, which should work fine.

iwoplaza (approved):
Beautiful 🥹

PR #387, opened 2024-09-24
Passing raw WGSL to tgpu.fn

### LINKED ISSUE #379: Allow passing raw WGSL string to `.implement` function.
For clarity, this could be a different class from `TgpuFnImpl`, a separate implementation just for raw WGSL functions.

During resolution:
- Extract the name of the function, replace with identifier during resolution
- Find and replace external identifiers with their resolved names.

mhawryluk:
closes #379

### DIFF
--- a/packages/typegpu/src/tgpuFn.ts
+++ b/packages/typegpu/src/tgpuFn.ts
@@ -4,7 +4,13 @@ import { type TgpuNamable, isNamable } from './namable';
 import { valueList } from './resolutionUtils';
 import { code } from './tgpuCode';
 import { identifier } from './tgpuIdentifier';
-import type { AnyTgpuData, ResolutionCtx, TgpuResolvable, Wgsl } from './types';
+import {
+  type AnyTgpuData,
+  type ResolutionCtx,
+  type TgpuResolvable,
+  type Wgsl,
+  isWgsl,
+} from './types';
 
 // ----------
 // Public API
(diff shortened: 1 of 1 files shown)


packages/typegpu-wgsl-parser/src/grammar.ne
```diff
@@ -149,15 +149,22 @@ template_arg_comma_list ->
...
 @{%
+export type FunctionArgument = { type: 'func_argument', ident: string, typespec: TypeSpecifier };
 export type FunctionDecl = { type: 'function_decl', header: FunctionHeader, body: CompoundStatement, attrs: Attribute[] };
-export type FunctionHeader = { type: 'function_header', identifier: string };
+export type FunctionHeader = { type: 'function_header', ident: string, returntype: ReturnType | null, args: FunctionArgument[] | null };
+export type ReturnType = { type: 'return_type', typespec: TypeSpecifier };
 
 %}
+return_type ->
+  "-" ">" type_specifier {% ([,, typespec]) => ({ type: 'return_type', typespec }) %}
```
iwoplaza: I think this would accept whitespace between `-` and `>`, which is not permitted in WGSL. Lets add a new token `->` and place it before the minus sign, so that it gets recognized first.

```suggestion
  "->" type_specifier {% ([, typespec]) => ({ type: 'return_type', typespec }) %}
```

iwoplaza (approved):
Amazing work! 🙌 💜 
Left one nitpick about the grammar.

PR #397, opened 2024-09-27
Added utility functions to check schema types.

iwoplaza:


packages/typegpu/src/data/struct.ts
```diff
@@ -121,13 +127,16 @@ class TgpuStructImpl<TProps extends Record<string, AnyTgpuData>>
...
+  if (isAlignedSchema(field as unknown)) {
     return `@align(${field.byteAlignment}) `;
   }
-  if (field instanceof TgpuSizedImpl) {
+
+  if (isSizedSchema(field as unknown)) {
     return `@size(${field.size}) `;
   }
+
+  return undefined;
```
mhawryluk, line 139: this line is unnecessary, was it added to make the undefined return more explicit?

packages/typegpu/src/data/struct.ts
```diff
@@ -121,13 +127,16 @@ class TgpuStructImpl<TProps extends Record<string, AnyTgpuData>>
   }
 }
 
-function getAttribute(field: AnyTgpuData): string | undefined {
-  if (field instanceof TgpuAlignedImpl) {
+function getAttribute<T extends AnyTgpuData>(field: T): string | undefined {
```
mhawryluk, line 130: as noticed by @reczkok, this function doesn't work properly when both properties are present. we should fix that in another PR

packages/typegpu/src/data/struct.ts
```diff
@@ -121,13 +127,16 @@ class TgpuStructImpl<TProps extends Record<string, AnyTgpuData>>
...
+  if (isAlignedSchema(field as unknown)) {
     return `@align(${field.byteAlignment}) `;
   }
-  if (field instanceof TgpuSizedImpl) {
+
+  if (isSizedSchema(field as unknown)) {
```
mhawryluk, line 135: why are these `as unknown` necessary?

mhawryluk (approved):
Great! 😊

packages/typegpu/src/data/struct.ts
iwoplaza: Yess 🙌

packages/typegpu/src/data/struct.ts
iwoplaza: For some reason, TypeScript assumes after each of these `if` statements, that the `field` property can be nothing else that that condition, so it becomes `never`. I will debug it and fix it in another PR when I figure out what is wrong.

iwoplaza:
@reczkok Any thoughts before I merge? 👀

reczkok (approved):
Looks great💜

PR #405, opened 2024-09-30
BindGroupLayout cli generator

### LINKED ISSUE #366: Generate tgpu.bindGroupLayout from wgsl code in `@typegpu/tgpu-cli`

mhawryluk:
closes #366

### DIFF
--- a/packages/tgpu-cli/gen.mjs
+++ b/packages/tgpu-cli/gen.mjs
@@ -189,10 +189,10 @@ function generateGroupLayout(group) {
     .map((_, index) => group[index])
     .map((variable) =>
       variable
-        ? `${variable.name}: ${generateVariable(variable)}`
-        : `_${emptyCount++} : null`,
+        ? `${variable.name}: ${generateVariable(variable)},`
+        : `_${emptyCount++}: null,`,
     )
-    .join(',\n  ');
+    .join('\n  ');
 }
 
 /**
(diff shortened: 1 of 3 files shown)


packages/tgpu-cli/gen.mjs
```diff
@@ -52,7 +56,7 @@ ${structs.map((struct) => generateStruct(struct, toTs)).join('\n\n')}`
  * @param { boolean } toTs
  */
 function generateStruct(struct, toTs) {
-  return `const ${struct.name} =${hasVarLengthMember(struct) ? ` (${LENGTH_VAR}${toTs ? ': number' : ''}) =>` : ''} d.struct({
+  return `export const ${struct.name} =${hasVarLengthMember(struct) ? ` (${LENGTH_VAR}${toTs ? ': number' : ''}) =>` : ''} d.struct({
```
iwoplaza: Tiny nitpick, imo this placement of whitespace make this a bit more readable.
```suggestion
  return `export const ${struct.name} = ${hasVarLengthMember(struct) ? `(${LENGTH_VAR}${toTs ? ': number' : ''}) => ` : ''}d.struct({
```

packages/tgpu-cli/gen.mjs
```diff
@@ -136,4 +141,162 @@ function replaceWithAlias(type) {
...
+function generateGroupLayout(group) {
+  let emptyCount = 0;
+
+  return Array(group.length)
+    .fill(null)
+    .map((_, index) => group[index])
+    .map((variable) =>
+      variable
+        ? `${variable.name}: ${generateVariable(variable)},`
+        : `_${emptyCount++}: null,`,
```
iwoplaza: What do you think about embedding a comment into the generated code, like `// skipping binding N`?
Just so that the purpose of this field is apparent.
Also, we could simplify the empty indexing by breaking an requirement I made initially XD. We can just number them based on the binding index instead.

```suggestion
        : `_${index}: null, // skipping binding ${index}`,
```

packages/tgpu-cli/tests/bindGroupLayouts.test.ts
```diff
@@ -0,0 +1,213 @@
...
+    const wgsl = `
+struct TriangleData {
+    position : vec4f,
+    velocity : vec2f,
+};
+
+@binding(0) @group(0) var<storage> trianglePos: array<TriangleData, 1000>;
+@binding(1) @group(0) var<storage, read> colorPalette: vec3<f32>;
+@binding(2) @group(0) var<storage, read_write> threshold: u32;
+@binding(3) @group(0) var<storage, write> threshold2: i32;
```
iwoplaza: Not valid WGSL, can remove this line.

iwoplaza (approved):
Great job! 🙌
Left a few small nitpicks

packages/tgpu-cli/gen.mjs
mhawryluk: sure, i'll add a comment. and using index means we remove the only `let` in the file, which I'm fond of

PR #409, opened 2024-10-02
Rollup plugin

iwoplaza:
### Changes
- `@typegpu/tgsl-tools` no longer act on raw JS code as a string, but rather process an estree-compliant JavaScript AST. This allows the rollup plugin to plugin the ast it generated to transpile it to SMoL. The JIT is now responsible for generating an AST from the raw JS code with `acorn`.
- Anonymous functions (`function() {}`) are now supported as parameters to `.implement` and `.procedure`

### Usage

```js
// @ts-check
import { defineConfig } from 'vite';
import typegpu from 'rollup-plugin-typegpu';

export default defineConfig({
  plugins: [
    typegpu(),
  ],
});
```

### Functionality
- Targets files that import `typegpu` or any submodule `typegpu/*`.
- Gathers how the `tgpu` object is imported, and how it is aliased.
- Augments every tgpu function creation with calls to `$__ast` and `$uses`.

image

### Further work

- #417
- #416


packages/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,176 @@
...
+          // The default export of both 'typegpu' and 'typegpu/experimental' is the `tgpu` object.
+          ctx.tgpuAliases.add(spec.local.name);
+        } else if (
+          spec.type === 'ImportSpecifier' &&
+          spec.imported.type === 'Identifier' &&
+          spec.imported.name === 'tgpu'
+        ) {
+          // Aliasing 'tgpu' while importing, e.g. import { tgpu as t } from 'typegpu';
+          ctx.tgpuAliases.add(spec.local.name);
+        } else if (spec.type === 'ImportNamespaceSpecifier') {
```
reczkok: Merge the two identical branches

packages/tgsl-tools/src/parsers.ts
```diff
@@ -259,62 +259,74 @@ export type TranspilationResult = {
...
-  ) {
-    const mainExpression = programBody.expression;
-
-    if (mainExpression.async) {
-      throw new Error('tgpu.fn cannot be async');
-    }
+  let functionNode:
+    | acorn.ArrowFunctionExpression
+    | acorn.FunctionExpression
+    | acorn.FunctionDeclaration
```
mhawryluk, line 269: not sure what's the difference between these expressions vs declarations. couldn't find acorn docs, could you please link it if you have it?

packages/rollup-plugin/src/index.ts
```diff
@@ -0,0 +1,176 @@
...
+/**
+ * Checks if `node` is an alias for the 'tgpu' object, traditionally
+ * available via `import tgpu from 'typegpu'`.
+ */
+function isTgpu(ctx: Context, node: AnyNode): boolean {
+  let path = '';
+
+  let tail = node;
+  while (true) {
+    if (node.type === 'MemberExpression') {
```
mhawryluk: `node` is not modified inside this condition. if this condition is true and the one below it isn't, there will be an infinite loop

packages/tgsl-tools/src/parsers.ts
iwoplaza: FunctionDeclaration nodes are statements, so when we just declare a function, instead of e.g., passing that declaration immediately as an argument. FunctionExpressions are the opposite.

Example of FunctionDeclaration:

```js
function foo() {}
```

Example of FunctionExpression:
```js
bar(function foo() {})
//  ^ here
```

reczkok (approved):
Great job 👏🆙

mhawryluk (approved):
LGTM 🚀

PR #410, opened 2024-10-02
Allowed passing a prebuild AST to TGSL functions.

iwoplaza:


packages/typegpu/src/tgpuFn.ts
```diff
@@ -204,27 +210,32 @@ function createFn<
...
 
+    $__ast(argNames: string[], body: Block): TgpuTgslFnBase<Args, Return> {
+      prebuiltAst = { argNames, body };
+      return this;
+    },
+
     $name(newLabel: string): TgpuTgslFnBase<Args, Return> {
-      // @ts-ignore
-      this.label = newLabel;
+      // Can only set the name once, other calls are ignored.
```
mhawryluk: won't this be quite confusing for the users? case: they import our cli-generated function which we will name automatically, then they run $name with a new name, but it will not apply

mhawryluk (approved):
Awesome 😄

PR #412, opened 2024-10-02
Loose structs

### LINKED ISSUE #364: Create `looseStruct` schema.
Depends on #351

reczkok:
closes #364

### DIFF
--- a/packages/typegpu/src/data/array.ts
+++ b/packages/typegpu/src/data/array.ts
@@ -137,7 +137,7 @@ class TgpuLooseArrayImpl<TElement extends AnyTgpuData>
   }
 
   measure(
-    value: MaxValue | Parsed<Unwrap<TElement>>[],
+    _: MaxValue | Parsed<Unwrap<TElement>>[],
     measurer: IMeasurer = new Measurer(),
   ): IMeasurer {
     return measurer.add(this.stride * this.elementCount);
--- a/packages/typegpu/src/data/array.ts
+++ b/packages/typegpu/src/data/array.ts
@@ -10,6 +10,7 @@ import {
 import { roundUp } from '../mathUtils';
 import type {
   AnyTgpuData,
+  AnyTgpuLooseData,
   ResolutionCtx,
   TgpuData,
   TgpuLooseData,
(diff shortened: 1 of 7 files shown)


packages/typegpu/src/data/array.ts
```diff
@@ -162,7 +165,7 @@ class TgpuLooseArrayImpl<TElement extends AnyTgpuData>
   }
 
   measure(
-    value: MaxValue | Parsed<Unwrap<TElement>>[],
+    _: MaxValue | Parsed<Unwrap<TElement>>[],
     measurer: IMeasurer = new Measurer(),
   ): IMeasurer {
     return measurer.add(this.stride * this.elementCount);
```
iwoplaza: Tiny nit, we can simplify this, and line 153.
```suggestion
    return measurer.add(this.size);
```

packages/typegpu/src/data/vertexAttribute.ts
```diff
@@ -0,0 +1,458 @@
...
+  uint16x2: vec2u,
+  uint16x4: vec4u,
+  sint16x2: vec2i,
+  sint16x4: vec4i,
+  unorm16x2: vec2f,
+  unorm16x4: vec4f,
+  snorm16x2: vec2f,
+  snorm16x4: vec4f,
+  float16x2: vec2f,
+  float16x4: vec4f,
```
iwoplaza, line 277: According to the docs, these (and only these) formats are represented as `vec2<f16>` in WGSl (in short `vec2h`). We do not have vec*h vectors implemented yet, perhaps it's time do finally add them in.

Also, not sure if I analyzed it right, but in the current setup, `float16x2` and `float16x4` will both be encoded as unorm16x2 and unorm16x4 respectively, since their underlying data is both `f32[Math.log(2)]`:

```ts
  const getters = {
    u32: [view.getUint8, view.getUint16, view.getUint32],
    f32: [readf8, readf16 /* HERE */, view.getFloat32],
    i32: [view.getInt8, view.getInt16, view.getInt32],
  };
```

reczkok: Yes, that's correct. I treated them the same as normalized. Until we have the implementation of `f16` maybe we should leave it as a TODO?

iwoplaza: We can, but we definitively should change this behavior before 0.3.0

iwoplaza: #422

iwoplaza (approved):
Glad you made it back alive 🫂

packages/typegpu/src/data/vertexAttribute.ts
```diff
@@ -0,0 +1,458 @@
...
+  };
+
+  const setters = {
+    u32: [view.setUint8, view.setUint16, view.setUint32],
+    f32: [writef8, writef16, view.setFloat32],
+    i32: [view.setInt8, view.setInt16, view.setInt32],
+  };
+
+  const setter = setters[primitive][Math.log2(size)];
+  if (setter) setter.call(view, 0, value);
```
mhawryluk: use block with curly braces to match our convention (also in a few other places in this file)

packages/typegpu/src/data/vertexAttribute.ts
```diff
@@ -0,0 +1,458 @@
...
+} as const;
+
+function writeSizedVector(
+  output: ISerialOutput,
+  value: vec2u | vec3u | vec4u | vec2f | vec3f | vec4f | vec2i | vec3i | vec4i,
+  elementSize: 1 | 2 | 4,
+  isSigned: boolean,
+): void {
+  const primitive = vectorKindToPrimitive[value.kind];
+  if (!primitive) throw new Error('Invalid vector kind');
```
mhawryluk:
```suggestion
```

mhawryluk (approved):
great job!

PR #415, opened 2024-10-03
Make tgpu-cli generator accept glob pattern

### LINKED ISSUE #394: tgpu-cli: Add an ability to generate many files with a glob pattern.
https://github.com/micromatch/picomatch

mhawryluk:
closes #394 

How it works:
- you can pass in glob pattern as input, generator will be run for each matched file
- example: `npx tgpu-cli gen -i "**/*.wgsl"`
- quotes around pattern are important, so that shell doesn't match a file before passing the argument
- when output is provided, there can only be a single file matched, then the output name is used
- if output is not provided, the pattern has to end with `.wgsl`, the names of the output files will have `.wgsl` replaced with `.ts`

### DIFF
--- a/packages/tgpu-cli/tgpu-cli.mjs
+++ b/packages/tgpu-cli/tgpu-cli.mjs
@@ -2,6 +2,7 @@
 import { readFileSync } from 'node:fs';
 import { exit } from 'node:process';
 import arg from 'arg';
+import { glob } from 'glob';
 import color from './colors.mjs';
 import generate from './gen.mjs';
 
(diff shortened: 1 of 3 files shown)


packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -23,17 +24,54 @@ const COMMANDS = {
...
+
+      if (!output && !input.endsWith('.wgsl')) {
+        console.error(
+          `${color.Red}Error: No output name provided and provided input doesn't end with .wgsl ${color.Reset}`,
+        );
+        exit(1);
+      }
+
+      if (output && files.length > 1) {
+        if (!input.endsWith('.wgsl')) {
```
iwoplaza: What is this check for?

iwoplaza (approved):
Awesome! 🌲 💚

packages/tgpu-cli/tgpu-cli.mjs
mhawryluk: oh, must be a copy-paste error. i'll remove it here

PR #421, opened 2024-10-03
Rutime sized arrays

### LINKED ISSUE #406: Allow for runtime-sized arrays.
Accept `0` as a valid array length, and interpret it as a 'runtime sized' array.
- When resolving, resolve it to `array<${data}>`.
- When trying to get the `size` property of the array, throw an error.
- When trying to measure with `MaxValue`, make the measurer unbounded.

reczkok:
Closes #406

### DIFF
--- a/packages/typegpu/src/data/array.ts
+++ b/packages/typegpu/src/data/array.ts
@@ -1,7 +1,7 @@
 import type * as TB from 'typed-binary';
 import {
   type IMeasurer,
-  type MaxValue,
+  MaxValue,
   Measurer,
   type Parsed,
   Schema,
@@ -62,13 +62,14 @@ class TgpuArrayImpl<TElement extends AnyTgpuData>
   extends Schema<Unwrap<TElement>[]>
   implements TgpuArray<TElement>
 {
+  private _size: number;
   readonly elementType: TElement;
   readonly elementCount: number;
   readonly byteAlignment: number;
-  readonly size: number;
   readonly stride: number;
   readonly isLoose = false;
   readonly isCustomAligned = false;
+  readonly isRuntimeSized: boolean;
 
   constructor(elementType: TElement, count: number) {
     super();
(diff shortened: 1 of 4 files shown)


packages/typegpu/src/data/array.ts
```diff
@@ -101,17 +116,22 @@ class TgpuArrayImpl<TElement extends AnyTgpuData>
   }
 
   measure(
-    _: MaxValue | Parsed<Unwrap<TElement>>[],
+    value: MaxValue | Parsed<Unwrap<TElement>>[],
     measurer: IMeasurer = new Measurer(),
   ): IMeasurer {
+    if (this.isRuntimeSized && value === MaxValue) {
+      return measurer.unbounded;
+    }
```
iwoplaza: Tiny nit, this will return result in a size of 0 if a normal array is passed as `value`. I'd return `measurer.unbounded` every time.
```suggestion
    if (this.isRuntimeSized) {
      return measurer.unbounded;
    }
```

packages/typegpu/src/data/struct.ts
```diff
@@ -118,16 +136,28 @@ class TgpuStructImpl<TProps extends Record<string, AnyTgpuData>>
...
+    let structMeasurer = measurer;
+    alignIO(structMeasurer, this.byteAlignment);
 
     const maxing = value === MaxValue;
     for (const [key, property] of Object.entries(this._properties)) {
-      alignIO(measurer, property.byteAlignment);
-      property.measure(maxing ? MaxValue : value[key], measurer);
+      if (maxing && measurer.isUnbounded) {
+        throw new Error('Only the last property of a struct can be unbounded');
+      }
```
iwoplaza: Not sure, but using `measurer` instead of `structMeasurer` here could have been a mistake.
Also, why does this check only apply when **maxing** 💪 ?

```suggestion
      if (structMeasurer.isUnbounded) {
        throw new Error('Only the last property of a struct can be unbounded');
      }
```

packages/typegpu/src/data/struct.ts
```diff
@@ -118,16 +136,28 @@ class TgpuStructImpl<TProps extends Record<string, AnyTgpuData>>
...
+
+      alignIO(structMeasurer, property.byteAlignment);
+      structMeasurer = property.measure(
+        maxing ? MaxValue : value[key],
+        structMeasurer,
+      );
+
+      if (structMeasurer.isUnbounded && !isArraySchema(property)) {
+        throw new Error('Cannot nest unbouded struct within another struct');
```
iwoplaza:
```suggestion
        throw new Error('Cannot nest unbounded struct within another struct');
```

packages/typegpu/src/data/struct.ts
reczkok: An oversight after merge with main. Thanks!

reczkok: Now that I see this - I wonder why the tests passed? 😕
Edit:
The test was not very good (at least at what it was supposed to do) 😅 
Added an additional test for this case

iwoplaza (approved):
Awesome job! 🫚🧡

PR #423, opened 2024-10-03
Reworked how attributes work.

iwoplaza:


packages/typegpu/src/data/array.ts
```diff
@@ -121,50 +118,53 @@ class TgpuLooseArrayImpl<TElement extends AnyTgpuData>
...
+    output.seekTo(beginning + this.size);
   }
 
   read(input: TB.ISerialInput): Parsed<Unwrap<TElement>>[] {
+    alignIO(input, this.byteAlignment);
     const elements: Parsed<Unwrap<TElement>>[] = [];
     for (let i = 0; i < this.elementCount; i++) {
-      if (this.elementType.isCustomAligned) {
-        alignIO(input, this.elementType.byteAlignment);
+      const customAlignment = getCustomAlignment(this.elementType);
```
mhawryluk: custom alignment is the same for every element in the array, so there's no need to invoke this function for each of them. I think we can just use the value stored in `this.byteAlignment`

packages/typegpu/src/data/attributes.ts
```diff
@@ -0,0 +1,335 @@
...
+}
+
+export interface Size<T extends number> {
+  type: 'size';
+  size: T;
+}
+
+export type AnyAttribute = Align<number> | Size<number>;
+
+export type AnyAttributeTuple = [...AnyAttribute[]];
```
mhawryluk: doesn't that evaluate to just `AnyAttribute[]`? what's the difference?

mhawryluk (approved):
Love it! 🤩

packages/typegpu/src/data/attributes.ts
```diff
@@ -0,0 +1,335 @@
...
+    if (this.size <= 0) {
+      throw new Error(
+        `Custom data size must be a positive number. Got: ${this.size}.`,
+      );
+    }
+  }
+
+  resolveReferences(ctx: IRefResolver): void {
+    throw new RecursiveDataTypeError();
+  }
```
reczkok:
```suggestion
  resolveReferences(_: IRefResolver): void {
    throw new RecursiveDataTypeError();
  }
```

packages/typegpu/src/data/array.ts
```diff
@@ -100,6 +96,7 @@ class TgpuArrayImpl<TElement extends AnyTgpuData>
       alignIO(input, this.byteAlignment);
       elements.push(this.elementType.read(input) as Parsed<Unwrap<TElement>>);
     }
+    alignIO(input, this.byteAlignment);
```
iwoplaza, line 99: Aligning here as well, to make it match with how the array is measured.

packages/typegpu/src/data/attributes.ts
mhawryluk: I meant `type AnyAttributeTuple = AnyAttribute[];`
it looks to yield the same results, but I may be missing something

mhawryluk: or AnyAttribute[] directly instead of [...AnyAttribute[]]

iwoplaza: Good point, seems to work fine!

reczkok (approved):
Great idea, amazing implementation 🙇‍♂️🪿

PR #424, opened 2024-10-04
CommonJS support in tgpu-cli and other improvements

### LINKED ISSUE #407: Add an option to generate CommonJS with tgpu-cli

mhawryluk:
closes #407

### DIFF
--- a/packages/tgpu-cli/gen.mjs
+++ b/packages/tgpu-cli/gen.mjs
@@ -7,56 +7,73 @@ const cwd = new URL(`file:${process.cwd()}/`);
 
 const LENGTH_VAR = 'arrayLength';
 
+let exportsList = [];
+
 /**
  * @param { string } input
  * @param { string } output
+ * @param { boolean } toTs
+ * @param { boolean } toCjs
  */
-async function main(input, output) {
+async function main(input, output, toTs, toCjs) {
   const inputPath = new URL(input, cwd);
   const outputPath = new URL(output, cwd);
-  const toTs = output.endsWith('.ts');
   const inputContents = await fs.readFile(inputPath, 'utf8');
 
-  await fs.writeFile(outputPath, generate(inputContents, toTs));
+  await fs.writeFile(outputPath, generate(inputContents, toTs, toCjs));
 }
 
 /**
  * @param { string } wgsl
  * @param { boolean } toTs
+ * @param { boolean } toCjs
  */
-export function generate(wgsl, toTs = true) {
+export function generate(wgsl, toTs = true, toCjs = false) {
   const reflect = new WgslReflect(wgsl);
+
+  exportsList = [];
 
   return `/* generated via tgpu-cli by TypeGPU */
 
+${
+  toCjs
+    ? `\
+const tgpu = require('typegpu').default;
+const d = require('typegpu/data');`
+    : `\
 import tgpu from 'typegpu';
-import * as d from 'typegpu/data';
+import * as d from 'typegpu/data';`
+}
 
-${generateStructs(reflect.structs, toTs)}
+${generateStructs(reflect.structs, toTs, toCjs)}
 
-${generateAliases(reflect.aliases)}
+${generateAliases(reflect.aliases, toCjs)}
 
-${generateBindGroupLayouts(reflect.getBindGroups())}
+${generateBindGroupLayouts(reflect.getBindGroups(), toCjs)}
+
+${generateExports(toCjs)}
 `.trim();
 }
 
 /**
  * @param { import('wgsl_reflect').StructInfo[] } structs
  * @param { boolean } toTs
+ * @param { boolean } toCjs
  */
-function generateStructs(structs, toTs) {
+function generateStructs(structs, toTs, toCjs) {
   return structs.length > 0
     ? `/* structs */
-${structs.map((struct) => generateStruct(struct, toTs)).join('\n\n')}`
+${structs.map((struct) => generateStruct(struct, toTs, toCjs)).join('\n\n')}`
     : '';
 }
 
 /**
  * @param { import('wgsl_reflect').StructInfo } struct
  * @param { boolean } toTs
+ * @param { boolean } toCjs
  */
-function generateStruct(struct, toTs) {
-  return `export const ${struct.name} = ${
+function generateStruct(struct, toTs, toCjs) {
+  return `${declareConst(struct.name, toCjs)} = ${
     hasVarLengthMember(struct)
       ? `(${LENGTH_VAR}${toTs ? ': number' : ''}) => `
       : ''
@@ -75,12 +92,16 @@ function hasVarLengthMember(struct) {
 
 /**
  * @param { import('wgsl_reflect').AliasInfo[] } aliases
+ * @param { boolean } toCjs
  */
-function generateAliases(aliases) {
+function generateAliases(aliases, toCjs) {
   return aliases.length > 0
     ? `/* aliases */
 ${aliases
-  .map((alias) => `export const ${alias.name} = ${generateType(alias.type)};`)
+  .map(
+    (alias) =>
+      `${declareConst(alias.name, toCjs)} = ${generateType(alias.type)};`,
+  )
   .join('\n')}`
     : '';
 }
@@ -147,14 +168,15 @@ function replaceWithAlias(type) {
 
 /**
  * @param { import('wgsl_reflect').VariableInfo[][] } bindGroups
+ * @param { boolean } toCjs
  */
-function generateBindGroupLayouts(bindGroups) {
+function generateBindGroupLayouts(bindGroups, toCjs) {
   return bindGroups.length > 0
     ? `/* bindGroupLayouts */
 ${bindGroups
   .flatMap(
     (group, index) => `\
-export const layout${index} = tgpu.bindGroupLayout({
+${declareConst(`layout${index}`, toCjs)} = tgpu.bindGroupLayout({
   ${generateGroupLayout(group)}
 }).$forceIndex(${index});`,
   )
@@ -301,6 +323,32 @@ function generateExternalTextureVariable(variable) {
   return `{
     externalTexture: {},
   }`;
+}
+
+/**
+ * @param { string } ident
+ * @param { boolean } toCjs
+ */
+function declareConst(ident, toCjs) {
+  if (toCjs) {
+    exportsList.push(ident);
+    return `const ${ident}`;
+  }
+  return `export const ${ident}`;
+}
+
+/**
+ * @param { boolean } toCjs
+ */
+function generateExports(toCjs) {
+  if (!toCjs) {
+    exportsList = [];
+    return '';
+  }
+
+  const exports = `module.exports = {${exportsList.join(', ')}};`;
+  exportsList = [];
+  return exports;
 }
 
 export default main;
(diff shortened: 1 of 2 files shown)


packages/tgpu-cli/gen.mjs
```diff
@@ -7,56 +7,73 @@ const cwd = new URL(`file:${process.cwd()}/`);
...
 const LENGTH_VAR = 'arrayLength';
 
+let exportsList = [];
+
 /**
  * @param { string } input
  * @param { string } output
+ * @param { boolean } toTs
+ * @param { boolean } toCjs
  */
```
reczkok: Maybe it would be better to replace this with an enum?
```suggestion
/**
 * @enum {string}
 */
const Target = {
  TS: 'ts',
  CJS: 'cjs',
};

/**
 * @param { string } input
 * @param { string } output
 * @param { Target } target
 */
```

mhawryluk: don't think so, these flags are not mutually exclusive

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -11,18 +14,24 @@ const args = arg({
...
   '--input': String,
   '--output': String,
+  '--commonjs': Boolean,
+  '--js': Boolean,
 
   '-v': '--version',
   '-h': '--help',
   '-i': '--input',
   '-o': '--output',
+  '-c': '--commonjs',
```
iwoplaza: Imo, we should not give a shortcut for `--commonjs`, hard to interpret `c` as `commonjs` when somebody first sees it.

```suggestion
```

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -11,18 +14,24 @@ const args = arg({
   '--help': Boolean,
   '--input': String,
   '--output': String,
+  '--commonjs': Boolean,
+  '--js': Boolean,
```
iwoplaza: Could we provide a way to configure the output extension, and infer whether to generate TypeScript or JavaScript from that? Just added a task for it #427 

```suggestion
  '--ext': String,
```

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -21,32 +27,52 @@ const args = arg({
...
       const watch = args['--watch'] ?? false;
 
-      if (!input) {
+      if (!input || !output) {
         console.error(
-          `${color.Red}Error: Missing required argument: ${color.Yellow}--input${color.Reset}`,
+          `${color.Red}Error: Missing some of the required arguments: ${color.Yellow}--input, --output${color.Reset}`,
         );
         exit(1);
       }
```
reczkok: We could be more specific with the error.

reczkok, line 61: We could be more specific with the error.

packages/tgpu-cli/gen.mjs
```diff
@@ -1,64 +1,107 @@
...
+    .access(output)
+    .then(() => true)
+    .catch(() => false);
+
+  if (fileExists && !overwriteExisting) {
+    console.warn(
+      `File ${output} already exists. Use --overwrite option to replace existing files.`,
+    );
+    return;
+  }
```
reczkok: This interacts with watch mode in a non-intuitive way. I think maybe watch mode should implicitly have the `--overwrite` flag?

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -21,32 +27,52 @@ const args = arg({
...
-  --watch, -w   Watch for changes in the input file(s) and regenerate the output file(s)`),
+  --input, -i\t The input file or glob pattern.
+  --output, -o\t The output name or pattern for generated file(s).
+  --watch, -w\t Watch for changes in the input file(s) and regenerate the output file(s).
+  --commonjs\t Generate a CommonJS style file.
+`),
+
     execute: async () => {
       const input = args['--input'] ?? args._[1];
       const output = args['--output'];
```
reczkok: Now that output is not optional maybe we could also make it a positional argument for _convenience_.

packages/tgpu-cli/gen.mjs
mhawryluk: I'm afraid it's quite dangerous to just treat watch mode as having the overwrite flag. While overwriting is expected when editing source file, it isn't after we just booted the script. Imo there are 2 options: add an immediate warning when someone runs the script with `watch`, but with no `overwrite`; ooor in that case only check if we're overwriting the first time we generate a file, not when the file changes. what do you think?

packages/tgpu-cli/tgpu-cli.mjs
mhawryluk: wouldn't that be confusing which argument should come first?

packages/tgpu-cli/gen.mjs
reczkok: I think i like the second approach better. It is how I would assume `watch` works by default.

iwoplaza: Same here, second approach makes the most sense to me.

packages/tgpu-cli/tgpu-cli.mjs
reczkok: I don't think so but i'm not 100% on that. Maybe if someone is not sure he can use `--help`. I will take a look how other cli's handle that and get back to you.

packages/tgpu-cli/gen.mjs
iwoplaza: I would still lean towards having both modes be explicit (either keep existing files, or overwrite existing files), and by default, exit with an error if a matching file already exists. In that way, running `watch` a second time will result in a proper error, not just a set of warnings for existing files. Then the user can choose whether they want to `--overwrite` or `--keep`.

packages/tgpu-cli/tgpu-cli.mjs
reczkok: It looks like what I proposed isn't common practice. I still think we should provide a way to shorten the command (like defaulting to `.wgsl -> .ts` like before). The command to produce a single file (which would be a common use case imo) is a bit wordy.

mhawryluk: now that I think of it I also believe people may expect that typing multiple file names would result in all of them being inputs, not that the second one is the output

packages/tgpu-cli/gen.mjs
```diff
@@ -1,64 +1,107 @@
...
+
 /**
  * @param { string } input
  * @param { string } output
+ * @param { boolean } toTs
+ * @param { boolean } toCommonJs
+ * @param { boolean } overwriteExisting
  */
-async function main(input, output) {
+async function main(input, output, toTs, toCommonJs, overwriteExisting) {
```
iwoplaza: Tiny nitpick, not fond of this many primitive parameters being passed in as positional parameters. Maybe we could merge the Context and Options into one, and pass that around instead? (this would also get rid of the class in favor of a plain object).

```ts
/**
 * @typedef {object} Options
 * @prop {string} input
 * @prop {string} output
 * @prop {boolean} toTs
 * @prop {'commonjs'|'esmodule'} moduleSyntax
 * @prop {'keep'|'overwrite'} [existingFileStrategy]
 */
```

```suggestion
/**
 * @param {Options} options
 */
async function main(options) {
```

packages/tgpu-cli/gen.mjs
```diff
@@ -1,64 +1,107 @@
...
   const inputPath = new URL(input, cwd);
   const outputPath = new URL(output, cwd);
-  const toTs = output.endsWith('.ts');
   const inputContents = await fs.readFile(inputPath, 'utf8');
 
-  await fs.writeFile(outputPath, generate(inputContents, toTs));
+  const generated = generate(inputContents, new Context({ toTs, toCommonJs }));
```
iwoplaza:
```suggestion
  const generated = generate(inputContents, options);
```

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -21,32 +27,53 @@ const args = arg({
...
 
 Options:
-  --input, -i   The input file or glob pattern.
-  --output, -o  The output file. If not provided, the input file will be used with a .ts extension. Cannot be used with glob patterns.
-  --watch, -w   Watch for changes in the input file(s) and regenerate the output file(s)`),
+  --input, -i\t The input file or glob pattern.
+  --output, -o\t The output name or pattern for generated file(s).
+  --watch, -w\t Watch for changes in the input file(s) and regenerate the output file(s).
+  --commonjs\t Generate a CommonJS style file.
+  --overwrite\t Force overwriting existing files.
```
iwoplaza:
```suggestion
  --overwrite\t Overwrite existing files.
  --keep\t Keep existing files.
```

iwoplaza (changes requested):

packages/tgpu-cli/gen.mjs
```diff
@@ -1,100 +1,153 @@
...
 
 /**
- * @param { string } input
- * @param { string } output
+ * @typedef {object} Options
+ * @prop {string} inputPath
+ * @prop {string} outputPath
+ * @prop {boolean} toTs
+ * @prop {'commonjs' | 'esmodule'} moduleSyntax
+ * @prop {'keep'|'overwrite'} [existingFileStrategy]
```
reczkok:
```suggestion
 * @prop {'keep' | 'overwrite'} [existingFileStrategy]
```

reczkok (approved):
Great work! 🛠️ Really good idea with the output path tests, it was a pain to do manually💜

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -56,26 +103,44 @@ Options:
...
 
-      if (output && files.length > 1) {
+      const fileNames = files.map((file) => path.parse(file).name);
+      const duplicates = fileNames.filter(
+        (name, index) => fileNames.indexOf(name) !== index,
+      );
+      if (duplicates.length > 0 && !/\*\*\/.*\*.*/.test(output)) {
         console.error(
-          `${color.Red}Error: More than one file found, while a single output name was provided ${color.Reset}`,
+          `${color.Red}Error: Duplicates found with name(s): [${duplicates.join(', ')}], while a single directory output pattern was provided. Make sure your pattern contains "**/*" to keep the original directory structure. ${color.Reset}`,
```
iwoplaza, line 119: 🤩 💛

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -56,26 +103,44 @@ Options:
...
+      /**
+       * @param {{ exitOnError: boolean, files: string[], checkExisting: boolean }} options
+       */
+      const processFiles = async ({ exitOnError, files, checkExisting }) => {
         const results = await Promise.allSettled(
           files.map(async (file) => {
-            const out = output ?? file.replace('.wgsl', '.ts');
-            console.log(`Generating ${file} >>> ${out}`);
-            return generate(file, out).catch((error) => {
+            const outputhPath = outputPathCompiler(file);
```
iwoplaza: Tiny nit
```suggestion
            const outputPath = outputPathCompiler(file);
```

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -56,26 +103,44 @@ Options:
...
+
+            console.log(`Generating ${file} >>> ${outputhPath}`);
+            return generate({
+              inputPath: file,
+              outputPath: outputhPath,
```
iwoplaza:
```suggestion
              outputPath,
```

iwoplaza (approved):
Amazing work!! 🙌 🐇

PR #437, opened 2024-10-07
Cleaning `out` directories only on production builds.

iwoplaza:


mhawryluk (approved):
Seems to work great! Good job 😎

PR #441, opened 2024-10-07
Typed entry functions

iwoplaza:
Left to do in other PRs:
- Allow for usage of builtins and ioStructs in function args.
- Implement the new buffer vertex usage.
- Add tests to see if functions resolve properly.
- Implement the new pipeline API and use the entry functions.


packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -0,0 +1,122 @@
...
+    },
+
+    $name(newLabel: string): This {
+      core.label = newLabel;
+      return this;
+    },
+
+    resolve(ctx: ResolutionCtx): string {
+      return core.resolve(ctx, '@fragment ');
+    },
```
reczkok, line 120: Is it necessary to pass the attribute? Couldn't entry points resolve to their respective attributes?

reczkok: Ohh, I think I see it now. We use `addDeclaration` so the attributes would be in the wrong place.

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -0,0 +1,172 @@
...
+  $__ast(argNames: string[], body: Block): this;
+}
+
+export type TgpuFn<
+  Args extends AnyTgpuData[],
+  Return extends AnyTgpuData | undefined = undefined,
+> = TgpuFnBase<Args, Return> &
+  ((...args: UnwrapArgs<Args>) => UnwrapReturn<Return>);
+
+export function fn<Args extends [...AnyTgpuData[]] | []>(
```
mhawryluk: curious what's the reasoning behind `[...AnyTgpuData[]]` vs `AnyTgpuData[]`

packages/typegpu/src/errors.ts
```diff
@@ -86,9 +86,9 @@ export class NotStorageError extends Error {
 }
 
 export class MissingLinksError extends Error {
-  constructor(fnLabel: string | undefined, externalNames: string[]) {
+  constructor(externalNames: string[]) {
     super(
-      `The function '${fnLabel ?? '<unnamed>'}' is missing links to the following external values: ${externalNames}.`,
+      `A function is missing links to the following external values: ${externalNames}.`,
```
mhawryluk: I would keep passing the label to make it clear which function is missing externals in $uses

mhawryluk (approved):
Great 👍

reczkok (approved):
Looks good to me ⚙️🚀

packages/typegpu/src/core/function/tgpuFn.ts
iwoplaza: I could've sworn it did not work without it, haha. Seems like `AnyTgpuData[] | []` still behaves correctly, but just `AnyTgpuData[]` starts to break down.

iwoplaza: Without it, the type of function arguments is inferred to be an array of a union, not a tuple, so for example `(F32|U32)[]` instead of `[F32, U32]`

PR #442, opened 2024-10-08
fix: i32, f32 and u32 should not be assignable to each other.

iwoplaza:


packages/typegpu/src/data/numeric.ts
```diff
@@ -4,15 +4,16 @@ import { inGPUMode } from '../gpuMode';
...
   return {
     // Type-token, not available at runtime
     __unwrapped: undefined as unknown as number,
-    isLoose: false as const,
 
+    isLoose: false as const,
+    kind: code,
     size: 4,
     byteAlignment: 4,
     expressionCode: code,
```
reczkok, line 19: Is it necessary to have both `kind` and `expressionCode`?

reczkok (approved):
Great fix 🛠️

packages/typegpu/src/data/numeric.ts
iwoplaza: I wanted to eventually move away from having `expressionCode` as a public member, since code should be retrieved through `resolve`ing. I do not want people to depend on this property being available, and `kind` would not have to be a 1-to-1 WGSL representation, but instead just a unique string that identifies that schema.

mhawryluk (approved):
LGTM 👍

Issue #443, opened 2024-10-09
Add `d.location()` attribute for explicit locations.


iwoplaza:
### Explicit locations in vertex attributes
Source WGSL:
```wgsl
@vertex
fn vertex_main(@location(0) position: vec3f, @location(5) uv: vec2f) -> VertexOut {
  // ...
}
```
Equivalent JS:
```ts
const vertex_main = tgpu
  .vertexFn([d.vec3f, d.location(5, d.vec2f)], VertexOut)
  .implement(/* wgsl */`(@location(0) position: vec3f, @location(5) uv: vec2f) -> VertexOut {
    // ...
  }`);
```

### (Maybe?) Explicit locations in bind group layouts

```ts
const layout0 = tgpu.bindGroupLayout({
  foo: { uniform: d.location(1, d.vec3f) },
  bar: { uniform: d.location(3, d.vec3f) },
});
```

PR #444, opened 2024-10-09
feat: Add a location attribute.

iwoplaza:


packages/typegpu/src/data/attributes.ts
```diff
@@ -105,7 +106,7 @@ export type Decorate<
       >
     : never;
 
-function attribute<
+export function attribute<
```
mhawryluk, line 109: why? this function doesn't seem to be useful in other files

mhawryluk (approved):
Looks good! I'm not sure how this attribute is going to be useful in any real scenarios though. We don't really need to set location explicitly in ioStruct and maybe we can avoid it in vertex attributes too 🤔

packages/typegpu/src/data/attributes.ts
```diff
@@ -186,11 +210,41 @@ export function getCustomAlignment(
...
-    return data.customAlignment;
+    return data.alignAttrib;
   }
   return undefined;
 }
 
+export function getCustomLocation(
+  data: AnyTgpuData | AnyTgpuLooseData,
+): number | undefined {
+  return data ? (data as unknown as BaseDecorated).locationAttrib : undefined;
```
reczkok: According to types the second branch of ternary will never run. Shouldn't we check for decorations to avoid casting?
```suggestion
  return isDecorated(data) || isLooseDecorated(data)
    ? data.locationAttrib
    : undefined;
```

iwoplaza:
@mhawryluk
> Looks good! I'm not sure how this attribute is going to be useful in any real scenarios though. We don't really need to set location explicitly in ioStruct and maybe we can avoid it in vertex attributes too 🤔

I tried to summarize in the related issue #443. It is mostly for the WGSL->JS workflow, to be able to represent almost any WGSL as equivalent TypeGPU definitions.

packages/typegpu/src/data/attributes.ts
iwoplaza: My bad. It is not useful in this PR, but it does get used by builtins in "the future" 🚀

packages/typegpu/src/data/attributes.ts
iwoplaza: I'll remove the ternary, but the casting is mostly to reduce the runtime complexity, as accessing `locationAttrib` on an object that does not have it should result in `undefined` anyway, but is much less JS code than checking if the passed in value is a decorated object.

reczkok: Hmmm, okey. Then we should probably do the same for `getCustomAlignment` as it behaves nearly identically.

reczkok (approved):

PR #447, opened 2024-10-09
Add docs for numerics.

### LINKED ISSUE #403: Document `bool`, `u32`, `i32` and `f32`

reczkok:
closes #403

### DIFF
--- a/packages/typegpu/src/data/numeric.ts
+++ b/packages/typegpu/src/data/numeric.ts
@@ -56,7 +56,7 @@ const primitiveNumeric = <TKind extends string>(
 };
 
 /**
- * Boolean schema representing a single WGSL boolean value.
+ * Boolean schema representing a single WGSL bool value.
  * Cannot be used inside buffers as it is not host-shareable.
  */
 export type Bool = TgpuData<boolean>;
--- a/packages/typegpu/src/data/numeric.ts
+++ b/packages/typegpu/src/data/numeric.ts
@@ -93,8 +93,9 @@ const u32Cast = (v: number | boolean) => {
   return Math.max(0, Math.min(0xffffffff, Math.floor(v)));
 };
 /**
- * Creates a new unsigned 32-bit integer schema representing a single WGSL u32 value.
- * Can also be used to cast a value to a u32 in accordance with WGSL casting rules.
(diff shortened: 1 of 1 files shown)


packages/typegpu/src/data/numeric.ts
```diff
@@ -120,4 +158,11 @@ const f32Cast = (v: number | boolean) => {
   }
   return v;
 };
+/**
+ * Creates a new 32-bit float schema representing a single WGSL f32 value.
+ * Can also be used to cast a value to an f32.
```
iwoplaza: Since it by itself is a schema, rather than a function that creates a schema, I would make that clear in this doc string.

```suggestion
 * A schema that represents a 32-bit float value. (equivalent to `f32` in WGSL)
 * 
 * Can also be called to cast a value to an f32.
```

packages/typegpu/src/data/numeric.ts
```diff
@@ -106,8 +130,22 @@ const i32Cast = (v: number | boolean) => {
   const value = v < 0 ? Math.ceil(v) : Math.floor(v);
   return Math.max(-0x80000000, Math.min(0x7fffffff, value));
 };
+/**
+ * Creates a new signed 32-bit integer schema representing a single WGSL i32 value.
+ * Can also be used to cast a value to an i32 in accordance with WGSL casting rules.
```
iwoplaza:
```suggestion
 * A schema that represents a signed 32-bit integer value. (equivalent to `i32` in WGSL)
 *
 * Can also be called to cast a value to an i32 in accordance with WGSL casting rules.
```

iwoplaza (changes requested):

packages/typegpu/src/data/numeric.ts
```diff
@@ -55,13 +55,23 @@ const primitiveNumeric = <TKind extends string>(
   };
 };
 
+/**
+ * Boolean schema representing a single WGSL bool value.
+ * Cannot be used inside buffers as it is not host-shareable.
+ */
 export type Bool = TgpuData<boolean>;
+/**
+ * Creates a new boolean schema representing a single WGSL boolean value.
```
iwoplaza: The same tweaks here as the other docs, I overlooked it 👀

iwoplaza (approved):
🙌🙌🙌

PR #448, opened 2024-10-09
docs: Updated palette and font to better match the branding

iwoplaza:
**Note:** A lot of the headings and containers had to get an additional `m-0` or `box-border` class, because now, Tailwind is responsible for styling both the docs and the landing page/examples, and there is a mismatch between default starlight styles and what Tailwind has by default.

Updates to the Starlight palette to better match the TypeGPU logo and branding.

Light mode:
|Before|After|
|-|-|
|||

Dark mode:
|Before|After|
|-|-|
|||


apps/typegpu-docs/astro.config.mjs
```diff
@@ -103,12 +112,14 @@ export default defineConfig({
...
         {
-          label: '🙋 FAQ',
+          label: 'FAQ',
           slug: 'faq',
         },
       ]),
     }),
-    tailwind(),
+    tailwind({
+      applyBaseStyles: false,
```
mhawryluk, line 121: that's because instead we import tailwind.css file? what's the difference?

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -167,7 +167,7 @@ export function ControlPanel() {
     <div
       className={cs(
         isGPUSupported ? '' : 'hidden md:flex',
-        'flex flex-col gap-4 p-6 bg-grayscale-0 rounded-xl max-h-[50%] md:max-h-full',
+        'box-border flex flex-col gap-4 p-6 bg-grayscale-0 rounded-xl max-h-[50%] md:max-h-full',
```
mhawryluk, line 170: could we instead of writing `box-border` and `m-0` everywhere add a global css reset?

something like:
```css
*, *::after, *::before {
  box-sizing: border-box;
  margin: 0;
}
```
or using tailwind \at apply

apps/typegpu-docs/src/components/design/Button.tsx
```diff
@@ -2,19 +2,18 @@ import cs from 'classnames';
...
 
 export const Button = forwardRef<HTMLButtonElement, Props>((props, ref) => {
-  const { label, onClick, accent, children } = props;
+  const { onClick, accent, children } = props;
 
   return (
     <button
       className={cs(
-        'focus:ring-2 focus:ring-gradient-blue rounded-[6.25rem] text-sm px-5 py-2.5',
+        'inline-flex justify-center items-center font-sans box-border focus:ring-2 focus:ring-gradient-blue rounded-[6.25rem] text-sm px-5 py-2.5',
```
mhawryluk: instead of using `font-sans` here I would add a global css reset to make buttons and inputs inherit font properties. now the `Select` component has a wrong font family

mhawryluk (commented):
Docs look much better now!

apps/typegpu-docs/astro.config.mjs
iwoplaza: Yes. That's what the Starlight docs about Tailwind integration now say. I assume its because Starlight applies its own reset styles.

apps/typegpu-docs/src/components/ControlPanel.tsx
iwoplaza: We could, but I am wondering if then we would have to deal with components working differently between pages (since we cannot apply those resets to the Starlight docs page, due to those components relying on these defaults).

reczkok (approved):
Great job 🐙

apps/typegpu-docs/src/components/ControlPanel.tsx
mhawryluk: I don't think we reuse any components between the docs and the landing/examples page, do we?

iwoplaza: Not yet 👀

PR #452, opened 2024-10-10
Document Atomic types and objects via comments

### LINKED ISSUE #401: Document `Atomic`, `d.atomic()` and `isAtomicSchema`

mhawryluk:
closes #401

### DIFF
--- a/packages/typegpu/src/data/atomic.ts
+++ b/packages/typegpu/src/data/atomic.ts
@@ -17,7 +17,7 @@ import type { I32, U32 } from './numeric';
 // ----------
 
 /**
- * Marks a concrete integer scalar type schema (u32 or i32) as WGSL atomic.
+ * Marks a concrete integer scalar type schema (u32 or i32) as a WGSL atomic.
  *
  * @example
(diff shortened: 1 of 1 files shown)


packages/typegpu/src/data/atomic.ts
```diff
@@ -16,15 +16,34 @@ import type { I32, U32 } from './numeric';
 // Public API
 // ----------
 
+/**
+ * Marks a concrete integer scalar type schema (u32 or i32) as the WGSL atomic.
```
reczkok:
```suggestion
 * Marks a concrete integer scalar type schema (u32 or i32) as a WGSL atomic.
```

reczkok (approved):
LGTM 🦭 I don't know how specific we want to be with the explanations? Do we assume people should know what an atomic does?

mhawryluk:
@reczkok I was also wondering that. I assume we don't want to duplicate the WGSL documentation though

iwoplaza (approved):
📖📖📖

PR #456, opened 2024-10-10
chore: Split standard and experimental root API.

iwoplaza:
This is in preparation for exporting the root with limited functionality.


packages/typegpu/src/createRoot.ts
```diff
@@ -456,7 +456,7 @@ export type CreateRootOptions = {
  */
 export async function createRoot(
   options?: CreateRootOptions,
-): Promise<TgpuRoot> {
+): Promise<ExperimentalTgpuRoot> {
```
mhawryluk, line 459: `createRoot` aka `init` as a public API should return promise of `TgpuRoot`. I assume that's going to be changed in a following PR, oui?

mhawryluk (approved):
🎉

PR #458, opened 2024-10-10
Link to blog on GPUUnsupportedPanel

mhawryluk:
Link to blog instead of the FAQ.

\+ some visual adjustments and fixes

Before | After

<div class="flex">
Zrzut ekranu 2024-10-10 o 17 50 23
Zrzut ekranu 2024-10-10 o 17 49 42
</div>


iwoplaza (approved):
Awesome, definitely an improvement! 👏

PR #475, opened 2024-10-17
Implement new buffer $usage API

### LINKED ISSUE #470: Implement new `$usage` API
Implement a new API, deprecating the current one as described in discussion #465.

mhawryluk:
closes #470

### DIFF
--- a/apps/typegpu-docs/src/content/docs/guides/tgpu-buffer-api.mdx
+++ b/apps/typegpu-docs/src/content/docs/guides/tgpu-buffer-api.mdx
@@ -195,8 +195,8 @@ When creating a buffer using the TypeGPU API, you can use the `.$usage()` builde
 
 ```ts
 const buffer = tgpu.createBuffer(u32)
-  .$usage(tgpu.Uniform)
-  .$usage(tgpu.Storage)
+  .$usage('uniform')
+  .$usage('storage')
   .$device(device);
 ```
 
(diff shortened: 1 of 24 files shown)


packages/typegpu/src/index.ts
```diff
@@ -2,15 +2,17 @@
...
 
-import { Storage, Uniform, Vertex } from './core/buffer/buffer';
 import { createBuffer } from './legacyBufferApi';
 import { bindGroupLayout } from './tgpuBindGroupLayout';
 import { read, write } from './tgpuBufferUtils';
 
 export const tgpu = {
-  Uniform,
-  Storage,
-  Vertex,
```
iwoplaza, line 13: If we're not exporting the types anymore as part of `tgpu`, we need to export them another way.

We can just add that to the last export line:
```ts
export type { TgpuBuffer, Uniform, Storage, Vertex } from './core/buffer/buffer';
```

This is so that, e.g., devs can properly type function arguments that accept buffers that have a specific usage.

apps/typegpu-docs/src/content/docs/guides/tgpu-buffer-api.mdx
iwoplaza, line 1: I would move these public doc changes to a separate PR, which we can merge once 0.2 is released, otherwise the docs will be out of sync with what is public.

mhawryluk: well then maybe I can just wait with merging this entire PR until just before releasing 0.2? cause in addition to the docs, the live examples would also use the new unreleased API. and I don't think any other task for 0.2 depends on the changes made here anyways

reczkok (approved):
Great work 🛵🌤️

iwoplaza (approved):
Merge Party! 🎊 🎉

PR #477, opened 2024-10-17
Fix overriding files in tgpu-cli with watch mode and --keep option

### LINKED ISSUE #471: Fix overriding files in `tgpu-cli` with watch mode and --keep option + better duplicate error handling
* with --keep option overwrite only newly created files,
* with --overwrite overwrite all

at the beginning of the program get a list of existing files, if no strategy provided end the program with a descriptive error. with --keep mode watch only new files, exclude existing from the watch list.

mhawryluk:
closes #471

### DIFF
--- a/packages/tgpu-cli/tgpu-cli.mjs
+++ b/packages/tgpu-cli/tgpu-cli.mjs
@@ -124,33 +124,49 @@ Options:
 
       const outputPathCompiler = createOutputPathCompiler(input, output);
 
-      const existingFiles = await Promise.all(
-        allMatchedFiles.map(outputPathCompiler).map(
-          async (outputFile) =>
-            await access(outputFile)
-              .then(() => true)
-              .catch(() => false),
-        ),
-      ).then((existsResults) =>
-        existsResults.flatMap((exists, index) =>
-          exists ? [allMatchedFiles[index]] : [],
-        ),
-      );
+      const existingFilesIO =
+        existingFileStrategy === 'overwrite'
+          ? []
+          : await Promise.all(
+              allMatchedFiles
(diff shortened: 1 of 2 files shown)


packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -123,12 +124,53 @@ Options:
...
+          ? []
+          : await Promise.all(
+              allMatchedFiles
+                .map((input) => ({ input, output: outputPathCompiler(input) }))
+                .map(
+                  async ({ input, output }) =>
+                    await access(output)
+                      .then(() => ({ input, output, outputExists: true }))
+                      .catch(() => ({ input, output, outputExists: false })),
+                ),
```
iwoplaza: Tiny nit, this can be simplified a bit by returning the promise directly.

We could also return either an object, or null of we want to filter it out later.
```suggestion
                .map(
                  ({ input, output }) =>
                    access(output)
                      .then(() => ({ input, output }))
                      .catch(() => null),
                ),
```

packages/tgpu-cli/tgpu-cli.mjs
```diff
@@ -123,12 +124,53 @@ Options:
...
+            ).then((existsResultsIO) =>
+              existsResultsIO.flatMap(({ input, output, outputExists }) =>
+                outputExists ? [{ input, output }] : [],
+              ),
+            );
```
iwoplaza: Then we can just filter out the `null`s.
```suggestion
            ).then((existsResultsIO) =>
              existsResultsIO.filter((file): Exclude<typeof file, null> => !!file),
            );
```

iwoplaza (approved):
Great job! 🙌 💚
Left just a few nitpicks.

PR #491, opened 2024-10-24
Fix overflow in the roadmap section of the landing page

mhawryluk:
before:


iwoplaza:
How does it look after this is applied?

mhawryluk:
after:

iwoplaza (approved):
Ok cool 😎👌
Thanks for finding it and fixing it so quickly!

PR #497, opened 2024-10-28
Fix unstable increment example

mhawryluk:
should fix docs not building on main


iwoplaza (approved):
Thanks! Missed that one by mistake 🙏

PR #506, opened 2024-10-29
Deprecate the legacy buffer creation API

### LINKED ISSUE #455: (Only right before publishing) Deprecate the legacy buffer creation API.

mhawryluk:
closes #455


apps/typegpu-docs/src/content/examples/simple/gradient-tiles-typedbuffers.ts
```diff
@@ -16,7 +16,9 @@ const adapter = await navigator.gpu.requestAdapter();
 if (!adapter) {
   throw new Error('Could not find a compatible GPU.');
 }
-const device = await adapter.requestDevice();
+
+const root = await tgpu.init();
+const device = root.device;
```
iwoplaza: We can remove all code between lines 11 and 18 as well 🚀

apps/typegpu-docs/src/content/examples/simulation/boids-typed-bindgroups.ts
```diff
@@ -2,12 +2,11 @@
 {
   "title": "Boids (next)",
```
iwoplaza: Small nitpick, since we're making this one the default.

```suggestion
  "title": "Boids",
```

Also, I would do the following changes:
- Rename `boids.ts` to `boids-experimental.ts`
- Rename `boids-typed-bindgroups.ts` to `boids.ts`
- Delete `boids-typedbuffers.ts`

packages/typegpu/src/legacyBufferApi.ts
```diff
@@ -23,6 +25,9 @@ export function createBuffer<TData extends AnyTgpuData>(
   gpuBuffer: GPUBuffer,
 ): TgpuBuffer<TData>;
 
+/**
+ *  @deprecated Use the `root.createBuffer` API instead, accessible through `await tgpu.init()`
```
iwoplaza:
```suggestion
 * @deprecated Use the `root.createBuffer` API instead, accessible through `await tgpu.init()`
```

iwoplaza (approved):
Awesome! 🌲🌲🌲

PR #520, opened 2024-11-05
Fix deployment after removing addElement toolkit api

mhawryluk:
references #490 

\+ fixed one warning while at it


reczkok (approved):
🛠️

PR #529, opened 2024-11-05
FIx losing scroll position when switching between html/ts tabs in Live Examples

### LINKED ISSUE #526: Code editor loses scroll position when switching tabs in Live Examples

mhawryluk:
closes #526


iwoplaza (approved):
Awezzzzome!

PR #530, opened 2024-11-05
Add `d.location` attribute support in tgpu-gen

### LINKED ISSUE #518: Add `d.location` attribute to vertex output/fragment input when generating with tgpu-gen.

mhawryluk:
closes #518 

adds d.location attribute to annotated function argument and return types, as well as to struct members


iwoplaza (approved):
Awesome solution 💜

PR #533, opened 2024-11-05
Add `$idx` builder method to bindGroupLayouts

### LINKED ISSUE #523: Add `$idx` builder method to bindGroupLayouts.
This should (in the future) explicitly mark a bind-group layout to appear with a specific group index in the resulting WGSL code string.

mhawryluk:
closes #523 

* doesn't do anything yet
* isn't generated by tgpu-gen (also yet)

### DIFF
--- a/packages/typegpu/src/tgpuBindGroupLayout.ts
+++ b/packages/typegpu/src/tgpuBindGroupLayout.ts
@@ -136,6 +136,11 @@ export interface TgpuBindGroupLayoutExperimental<
     TgpuLayoutEntry | null
   >,
 > extends TgpuBindGroupLayout<Entries> {
+  /**
+   * Marks a bind group layout to appear with a specific group index in the resulting WGSL code
+   */
+  $idx(index?: number): this;
+
   readonly bound: {
     [K in keyof Entries]: BindLayoutEntry<Entries[K]>;
   };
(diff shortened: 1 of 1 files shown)


packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -136,6 +136,11 @@ export interface TgpuBindGroupLayoutExperimental<
     TgpuLayoutEntry | null
   >,
 > extends TgpuBindGroupLayout<Entries> {
+  /**
+   * Marks a bind group layout to appear with a specific group index in the resulting WGSL code
+   */
```
iwoplaza, line 141: Tiny nit, we might be able to rephrase this to also note on the automatic indexing if the method is not called.

```suggestion
  /**
   * Associates this bind group layout with an explicit numeric index. When a call to this
   * method is omitted, a unique numeric index is assigned to it automatically.
   */
```

iwoplaza: Tiny nit, we might be able to rephrase this to also note on the automatic indexing if the method is not called.

```suggestion
  /**
   * Associates this bind group layout with an explicit numeric index. When a call to this
   * method is omitted, a unique numeric index is assigned to it automatically.
   *
   * Used when generating WGSL code: `@group(${index}) @binding(...) ...;`
   */
```

iwoplaza (approved):
🥑🥑🥑

PR #536, opened 2024-11-07
Fix MNIST example type issue

reczkok:
`NodeListOf` type does not have an iterator in the live examples environment, this PR changes bars to `Array` type so it can be iterated over inside for..of (according to the types)

### DIFF
--- a/apps/typegpu-docs/src/content/examples/algorithms/mnist-inference/index.ts
+++ b/apps/typegpu-docs/src/content/examples/algorithms/mnist-inference/index.ts
@@ -272,7 +272,7 @@ const bars = document.querySelectorAll('.bar') as NodeListOf<HTMLDivElement>;
 const resetAll = () => {
   canvasData.fill(0);
   resetCanvas();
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/content/examples/algorithms/mnist-inference/index.ts
```diff
@@ -267,7 +267,9 @@ const network = createNetwork([
 
 const context = canvas.getContext('2d') as CanvasRenderingContext2D;
 
-const bars = document.querySelectorAll('.bar') as NodeListOf<HTMLDivElement>;
+const bars = Array.from(
+  document.querySelectorAll('.bar') as NodeListOf<HTMLDivElement>,
+);
```
iwoplaza:
```suggestion
const bars = Array.from(
  document.querySelectorAll('.bar'),
) as HTMLDivElement[];
```

iwoplaza (approved):
YEEEESSSS

PR #538, opened 2024-11-08
StackBlitz integration

mhawryluk:


apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -182,6 +186,12 @@ export function ControlPanel() {
           />
         </label>
 
+        {currentExample && currentExample in examples ? (
+          <Button onClick={() => openInStackBlitz(examples[currentExample])}>
+            Open in StackBlitz
```
iwoplaza: Small suggestion, we can suggest a bit more by changing the words, since not everyone might be familiar with StackBlitz.
```suggestion
            Edit on StackBlitz
```

PR #541, opened 2024-11-08
Better MNIST network and image processing

reczkok:
What changed?
- Much bigger 1.5M parameter network
- Normalized and augmented training data
- Image centering before inference
- Interpolation between points when drawing for smoother results


apps/typegpu-docs/src/content/examples/algorithms/mnist-inference/index.ts
```diff
@@ -309,47 +296,94 @@ const observer = new ResizeObserver(() => {
...
 });
 
-let lastPos = { x: 0, y: 0 };
+function centerImage(data: number[]) {
+  const x =
+    data.reduce((acc, value, i) => acc + value * (i % SIZE), 0) /
+    data.reduce((acc, value) => acc + value, 0);
+  const y =
+    data.reduce((acc, value, i) => acc + value * Math.floor(i / SIZE), 0) /
+    data.reduce((acc, value) => acc + value, 0);
```
iwoplaza:
```suggestion
  const mass = data.reduce((acc, value) => acc + value, 0);
  const x =
    data.reduce((acc, value, i) => acc + value * (i % SIZE), 0) /
    mass;
  const y =
    data.reduce((acc, value, i) => acc + value * Math.floor(i / SIZE), 0) /
    mass;
```

iwoplaza (approved):
👏👏👏

PR #543, opened 2024-11-12
Example controls definition via exported object

### LINKED ISSUE #540: Rework example controls from exported objects + Slider control

mhawryluk:
closes #540


apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -30,7 +30,7 @@ function ToggleRow({
 
       <label className="grid items-center justify-end h-10 cursor-pointer">
         <Toggle
-          checked={value}
+          checked={value ?? false}
```
iwoplaza: Tiny nitpick, we can assign the default value at the very beginning when destructuring props.

```ts
{
  label,
  initial = false,
  onChange,
}
```

```suggestion
          checked={value}
```

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -66,7 +66,7 @@ function SliderRow({
         min={min ?? 0}
         max={max ?? 1}
```
iwoplaza: We can also do the same here, giving initial values on destructuring.

```suggestion
        min={min}
        max={max}
```

Then give the initial value a default when using `useState`

```ts
const [value, setValue] = useState(initial ?? min);
```

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -94,7 +94,7 @@ function SelectRow({
       <div className="text-sm">{label}</div>
 
       <Select
-        value={value}
+        value={value ?? options[0]}
```
iwoplaza:
```ts
const [value, setValue] = useState(initial ?? options[0]);
```

```suggestion
        value={value}
```

iwoplaza (approved):
Awesome! 🕺 🪩

PR #567, opened 2024-11-15
A rework of builtins as decorated data-types.

iwoplaza:


packages/typegpu/src/programBuilder.ts
iwoplaza, line 1: We are deprecating the program builder and the old pipelines, so I stripped the ability to use builtins in these APIs for now.

mhawryluk:
seems that there's a circular import:

```"src/data/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
packages/typegpu dev:watch: Either change the import in "src/builtin.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
packages/typegpu dev:watch: Export "Decorated" of module "src/data/attributes.ts" was reexported through module "src/data/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
packages/typegpu dev:watch: Either change the import in "src/builtin.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.```

mhawryluk (approved):
Awesome 🚀

PR #571, opened 2024-11-19
Removed legacy pipelines & refactored the resolution context.

iwoplaza:


packages/typegpu/src/core/texture/texture.ts
```diff
@@ -537,7 +520,19 @@ class TgpuBindableSampledTextureImpl
...
     const id = ctx.names.makeUnique(this.label);
-    ctx.addRenderResource(this, id);
+    const { group, binding } = ctx.allocateFixedEntry(this);
+
+    let type: string;
+    if ((this._texture.props.sampleCount ?? 1) > 1) {
+      type = 'texture_multisampled_2d';
+    } else {
+      type = `texture_${dimensionToCodeMap[this.dimension]}`;
+    }
```
mhawryluk: replace with const and ternany operator

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -528,28 +569,10 @@ class TgpuBindGroupImpl<
...
-            }
-
-            return {
-              binding: idx,
-              resource,
-            };
-          }
-
-          if ('sampler' in entry) {
+          if ('externalTexture' in entry || 'sampler' in entry) {
```
mhawryluk, line 572: why don't we separate it to return GPUExternalTexture for externalTexture and GPUSampler for sampler?

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -366,50 +307,29 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
-
-    this.addDeclaration(
-      `@group(${group}) @binding(${idx}) var<${usageToVarTemplateMap[bindable.usage]}> ${identifier}: ${this.resolve((bindable.allocatable as TgpuBuffer<AnyTgpuData>).dataType)};`,
-    );
+    this._declarations.push(declaration);
   }
 
-  addRenderResource(resource: TgpuRenderResource, identifier: string): void {
-    const { group, idx } = this._shared.reserveRenderResourceEntry(resource);
+  allocateLayoutEntry(layout: TgpuBindGroupLayout): string {
```
mhawryluk, line 313: maybe we could write some tests for this group index resolution process

mhawryluk (approved):
awesome! 🔪

packages/typegpu/src/tgpuBindGroupLayout.ts
iwoplaza: Its the same JS behavior in both branches. We will probably split them later down the road when we refactor typed samplers.

PR #574, opened 2024-11-20
 Fix resolution when no catchall bind group is needed

mhawryluk:


packages/typegpu/src/resolutionCtx.ts
```diff
@@ -462,38 +462,43 @@ export function resolve(
 
   const automaticIds = naturalsExcept(takenIndices);
 
+  const createCatchallGroup = () => {
+    const catchallIdx = automaticIds.next().value;
+    const catchallLayout = bindGroupLayout(Object.fromEntries(layoutEntries));
+    bindGroupLayouts[catchallIdx] = catchallLayout;
+    code = code.replaceAll(CATCHALL_BIND_GROUP_IDX_MARKER, String(catchallIdx));
+    return [
```
iwoplaza, line 476: Tiny nitpick ⛏️
```suggestion

    return [
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -462,38 +462,43 @@ export function resolve(
...
+          ),
+        ),
+      ),
+    ] as [number, TgpuBindGroup];
+  };
+
+  const layoutEntries = ctx.fixedBindings.map(
+    (binding, idx) =>
+      [String(idx), binding.layoutEntry] as [string, TgpuLayoutEntry],
+  );
```
iwoplaza: Could we move this definition before `createCatchallGroup`? It might be easier to reason about the dependencies between values, as `createCatchallGroup` uses `layoutEntries` from the outer scope.

iwoplaza (approved):
Awesome, thanks! 🐸 💚

PR #581, opened 2024-11-26
CodeSandbox CI config

### LINKED ISSUE #570: Nightly experimental build on CI for every commit on main.
Every commit on main should produce an experimental build as an artifact, ready to install via pnpm.

mhawryluk:
closes #570


packages/typegpu/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "typegpu",
-  "private": true,
+  "private": false,
```
iwoplaza: Did this prohibit the package from being built?

package.json
```diff
@@ -11,7 +11,8 @@
     "coverage": "vitest --coverage",
     "test": "DEV=true EXPERIMENTAL=true pnpm -r dev:build && pnpm run '/^test:.*/'",
     "test:types": "pnpm run -r test:types",
-    "test:spec": "vitest run"
+    "test:spec": "vitest run",
+    "nightly-build": "EXPERIMENTAL=true pnpm -r dev:build && cd packages/typegpu && EXPERIMENTAL=true pnpm prepare-package"
```
iwoplaza: Suggestion, this can also be done by filtering the package name
```suggestion
    "nightly-build": "EXPERIMENTAL=true pnpm -r dev:build && EXPERIMENTAL=true pnpm --filter typegpu prepare-package"
```

packages/typegpu/package.json
mhawryluk: not from being built, but from being published yes. maybe changing this field in dist only would work. does it bother us that it's public though?

mhawryluk: wait, we do set it to false in dist. I think it's therefore not necessary, but imma test it rn

mhawryluk: it works 🎉

PR #585, opened 2024-11-27
Slot bindings on new pipelines.

iwoplaza:
### Changes
- `withPrimitive()` on render pipelines,
- `with()` on root to allow for slot bindings,
- Migrated fluid double buffering to new pipelines.
- A new `Infer<T>` type for "parsing", and a "__repr" type token to replace "__unwrapped"
- Making slot.value **double infer**.

### Additional Context
Changing "Parsed" to "Infer": https://github.com/software-mansion/TypeGPU/discussions/580
Making slot.value **double infer**: https://github.com/software-mansion/TypeGPU/discussions/586

### DIFF
--- a/packages/typegpu/src/core/function/tgpuComputeFn.ts
+++ b/packages/typegpu/src/core/function/tgpuComputeFn.ts
@@ -1,4 +1,5 @@
 import type { Block } from 'tinyest';
+import type { AnyBuiltin } from '../../builtin';
 import type { TgpuNamable } from '../../namable';
 import type { ResolutionCtx, TgpuResolvable } from '../../types';
 import { createFnCore } from './fnCore';
@@ -14,6 +15,7 @@ import type { Implementation } from './fnTypes';
 export interface TgpuComputeFnShell {
   readonly argTypes: [];
   readonly returnType: undefined;
+  readonly workgroupSize: [number, number, number];
 
   /**
    * Creates a type-safe implementation of this signature
@@ -36,6 +38,10 @@ export interface TgpuComputeFn extends TgpuResolvable, TgpuNamable {
   $__ast(argNames: string[], body: Block): this;
 }
 
+export interface ComputeFnOptions {
+  workgroupSize: number[];
+}
+
 /**
  * Creates a shell of a typed entry function for the compute shader stage. Any function
  * that implements this shell can perform general-purpose computation.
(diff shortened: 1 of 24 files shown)


packages/typegpu/src/data/vertexFormatData.ts
```diff
@@ -42,6 +42,9 @@ class TgpuVertexFormatDataImpl<T extends VertexFormat>
   extends Schema<FormatToWGSLType<T>>
   implements TgpuVertexFormatData<T>
 {
+  /** Type-token, not available at runtime */
+  public readonly __repr!: FormatToWGSLType<T>;
```
mhawryluk: do we need `underlyingType` then?

packages/typegpu/src/core/function/tgpuComputeFn.ts
```diff
@@ -36,17 +38,31 @@ export interface TgpuComputeFn extends TgpuResolvable, TgpuNamable {
...
  * that implements this shell can perform general-purpose computation.
  *
  * @param workgroupSize
  *   Size of blocks that the thread grid will be divided into (up to 3 dimensions).
  */
-export function computeFn(workgroupSize: number[]): TgpuComputeFnShell {
+export function computeFn<Args extends AnyBuiltin[] | []>(
+  argTypes: Args,
+  options: ComputeFnOptions,
+): TgpuComputeFnShell {
```
reczkok: I don't think we need to extend `[]`. Do we need this template? Is `argTypes` only here to create a type hint?

packages/typegpu/src/core/root/init.ts
```diff
@@ -59,60 +60,88 @@ import type {
...
   SetPlumAction,
+  WithBinding,
   WithCompute,
   WithFragment,
   WithVertex,
 } from './rootTypes';
 
+class WithBindingImpl implements WithBinding {
+  constructor(
+    private readonly _getRoot: () => ExperimentalTgpuRoot,
```
mhawryluk, line 71: why is this a function, not just root?

packages/typegpu/src/core/function/tgpuComputeFn.ts
```diff
@@ -36,17 +38,31 @@ export interface TgpuComputeFn extends TgpuResolvable, TgpuNamable {
...
 /**
  * Creates a shell of a typed entry function for the compute shader stage. Any function
  * that implements this shell can perform general-purpose computation.
  *
  * @param workgroupSize
  *   Size of blocks that the thread grid will be divided into (up to 3 dimensions).
  */
-export function computeFn(workgroupSize: number[]): TgpuComputeFnShell {
+export function computeFn<Args extends AnyBuiltin[] | []>(
+  argTypes: Args,
```
mhawryluk: this argument seems quite useless ngl

iwoplaza: It does not "do" anything rn, but its meant to be used when generating WGSL from TGSL. It will find its use someday 🫡.

packages/typegpu/src/core/root/init.ts
iwoplaza: Because the root implementation extends `WithBinding`, and passes itself to the super call. It cannot pass `this` directly before actually calling super, so this is a workaround.

packages/typegpu/src/data/vertexFormatData.ts
iwoplaza: Good point, I'll rework it to use `__repr` instead 👍

packages/typegpu/src/core/function/tgpuComputeFn.ts
iwoplaza: The `| []` is forcing the template to be a tuple instead of an array of a union. Though yeah, I think the template can be removed altogether as you suggest 💜.

mhawryluk (approved):
Awesome! 🚀

PR #590, opened 2024-11-28
docs: A draft of the slots API docs.

iwoplaza:
This is far from a complete docs page, but will allow us to iterate on the examples and think of good use-case examples to showcase its power.


mhawryluk (approved):
good start 🛫

PR #591, opened 2024-11-28
docs: Reference docs in DEV mode.

iwoplaza:
This will allow us to keep continuously improve how the API Reference documentation is structured, and ultimately make it publicly available along with the 0.3 release.


packages/typegpu/src/experimental/index.ts
```diff
@@ -29,8 +35,12 @@ export const tgpu = {
...
   bindGroupLayout,
 
-  init,
-  initFromDevice,
+  init: init as (
+    options?: InitOptions | undefined,
+  ) => Promise<ExperimentalTgpuRoot>,
+  initFromDevice: initFromDevice as (
+    options: InitFromDeviceOptions,
+  ) => ExperimentalTgpuRoot,
```
mhawryluk, line 43: what's the difference?

mhawryluk (approved):
cool, I think that's a good idea 💡

packages/typegpu/src/experimental/index.ts
iwoplaza: Forgot to change the default `init` and `initFromDevice` 🤦. It should make more sense now. We were exporting `init` and `initFromDevice` as returning `ExperimentalTgpuRoot` from the non-experimental entry point.

PR #594, opened 2024-12-02
DDD

iwoplaza:
### New terms:
- AnyWgslData - any data-type that can appear in WGSL code.
- AnyLooseData - data that does not adhere to WGSL's memory layout, for example, vertex buffer schemas.
- AnyData - AnyWgslData | AnyLooseData
- "exotic"/"native" - determines whether resources originate from TypeGPU (native), or from an external source like a WESL lib (exotic).
- Wgsl* - interfaces that are shared by both exotic and native resources.
- Tgpu* - interfaces of resources originating from TypeGPU.

### Why `~repr` and not `__repr`?

I am taking a note out of https://github.com/standard-schema/standard-schema#why-did-you-choose-to-prefix-the-standard-property-with- and choosing to prefix the type token with `~` so that it appears as a last suggestion in autocomplete.

`__repr`:
Screenshot 2024-12-04 at 11 25 21

`~repr`:

Screenshot 2024-12-04 at 11 34 21


packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -30,10 +35,10 @@ type LiteralToUsageType<T extends 'uniform' | 'storage' | 'vertex'> =
         ? Vertex
         : never;
 
-export interface TgpuBuffer<TData extends AnyTgpuData> extends TgpuNamable {
+export interface TgpuBuffer<TData extends BaseWgslData> extends TgpuNamable {
```
mhawryluk: when should we use `BaseWgslData` vs `AnyData`?

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -55,31 +56,31 @@ interface TgpuFnBase<
...
 > = TgpuFnBase<Args, Return> &
-  ((...args: UnwrapArgs<Args>) => UnwrapReturn<Return>);
+  ((...args: InferArgs<Args>) => InferReturn<Return>);
 
-export function fn<Args extends AnyTgpuData[] | []>(
+export function fn<Args extends AnyWgslData[] | []>(
   argTypes: Args,
   returnType?: undefined,
-): TgpuFnShell<Args, undefined>;
+): TgpuFnShell<ExoticArray<Args>, undefined>;
```
mhawryluk, line 67: not sure I fully grasp why we need the Exotic types here. could you please explain?

packages/typegpu/src/core/texture/textureFormats.ts
```diff
@@ -1,113 +1,105 @@
...
+  rgba16uint: u32 as U32,
+  rgba16sint: i32 as I32,
+  rgba16float: f32 as F32,
+  rgba32uint: u32 as U32,
+  rgba32sint: i32 as I32,
+  rgba32float: f32 as F32,
+  stencil8: f32 as F32, // NOTE: Honestly as Honestly have no idea if this is right
+  depth16unorm: f32 as F32,
+  depth24plus: f32 as F32, // NOTE: Honestly as Honestly have no idea if this is right
+  'depth24plus-stencil8': f32 as F32, // NOTE: Honestly as Honestly have no idea if this is right
```
mhawryluk:
```suggestion
  stencil8: f32 as F32, // NOTE: Honestly have no idea if this is right
  depth16unorm: f32 as F32,
  depth24plus: f32 as F32, // NOTE: Honestly have no idea if this is right
  'depth24plus-stencil8': f32 as F32, // NOTE: Honestly have no idea if this is right
```
hahah

packages/typegpu/src/data/alignmentOf.ts
```diff
@@ -0,0 +1,119 @@
...
+      .map((prop) => alignmentOf(prop))
+      .reduce((a, b) => (a > b ? a : b));
+  }
+
+  if (isWgslArray(data)) {
+    return alignmentOf(data.elementType);
+  }
+
+  if (isLooseStruct(data)) {
+    // A loose struct is alignment to its first property.
```
mhawryluk:
```suggestion
    // A loose struct is aligned to its first property.
```

packages/typegpu/src/data/alignmentOf.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  throw new Error(
+    `Cannot determine alignment of data: ${JSON.stringify(data)}`,
+  );
+}
+
+function computeCustomAlignment(data: BaseWgslData): number {
+  if (isLooseStruct(data)) {
+    // A loose struct is alignment to its first property.
```
mhawryluk:
```suggestion
    // A loose struct is aligned to its first property.
```

packages/typegpu/src/data/attributes.ts
```diff
@@ -174,14 +131,12 @@ export function attribute<
...
-  return attribute(data, { type: 'align', alignment }) as Decorate<
-    TData,
-    Align<TAlign>
-  >;
+export function align<TAlign extends number, TData extends AnyData>(
+  alignment: TAlign,
+  data: TData,
+): Decorate<Exotic<TData>, Align<TAlign>> {
+  // biome-ignore lint/suspicious/noExplicitAny: <tired of lying to types>
+  return attribute(data, { type: '@align', value: alignment }) as any;
```
mhawryluk, line 139: we should fix this. this `any` here is unacceptable, as it is a public API, not an internal function. using our attribute functions would remove type information for the data. same for the following functions

packages/typegpu/src/data/matrix.ts
```diff
@@ -123,11 +92,6 @@ abstract class mat2x2Impl<TColumn extends vec2f> implements mat2x2<TColumn> {
 
   abstract makeColumn(e0: number, e1: number): TColumn;
 
-  *elements() {
-    yield* this.columns[0];
-    yield* this.columns[1];
-  }
```
mhawryluk, line 129: whyy 😢

packages/typegpu/src/data/sizeOf.ts
```diff
@@ -0,0 +1,154 @@
...
+    return sizeOfStruct(data);
+  }
+
+  if (isLooseStruct(data)) {
+    return sizeOfLooseStruct(data);
+  }
+
+  if (isWgslArray(data)) {
+    if (data.length === 0) {
+      return Number.NaN;
```
mhawryluk, line 110: is the usage of NaN necessary? can't we use 0 instead, like we used to?

packages/typegpu/src/data/sizeOf.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  return size;
+}
+
+/**
+ * Returns the size (in bytes) of data represented by the `schema`.
+ */
+export function PUBLIC_sizeOf(schema: AnyData): number {
+  return sizeOf(schema);
+}
```
mhawryluk, line 154: why not just export `sizeOf` as public?

mhawryluk (changes requested):

packages/typegpu/src/data/attributes.ts
iwoplaza: The `: Decorate<Exotic<TData>, Align<TAlign>>` should narrow the type again, so that the public API is type-safe.

image

So the `as any` is only to omit the additional `as Decorate<Exotic<TData>, Align<TAlign>`

packages/typegpu/src/core/buffer/buffer.ts
iwoplaza: In this case, I was getting type errors and was trying to track down the source. Should be `AnyData` here.

packages/typegpu/src/core/function/tgpuFn.ts
iwoplaza: Of course. In this case, its mostly for cosmetic and consistency reasons. Without it, the shell and implementation types would look like the following:

```ts
const fooShell = tgpu.fn([f32, f32], vec3f);
// ^? TgpuFnShell<[F32 & F32Cast, F32 & F32Cast], Vec3f & Vec3fConstructor>
const foo = fooShell.does(...);
// ^? TgpuFn<[F32 & F32Cast, F32 & F32Cast], Vec3f & Vec3fConstructor>
```

But with exotic types, we have:
```ts
const fooShell = tgpu.fn([f32, f32], vec3f);
// ^? TgpuFnShell<[F32, F32], Vec3f>
const foo = fooShell.does(...);
// ^? TgpuFn<[F32, F32], Vec3f>
```

This means that, on the type level, a native function's inner types are no different than a function imported from a library.

packages/typegpu/src/data/matrix.ts
iwoplaza: I moved functionality to operate ON data, instead of being part of the data. This allows for native matrices and exotic matrices to match on the type level.

So something like the following:
```ts
const array = [...viewMatrix];
```
Will have to be written like this:
```ts
const array = mat3x3fToArray(viewMatrix);
```
We could compromise and make one function that turns any matrix to an array:
```ts
const array = matToArray(viewMatrix);
```

packages/typegpu/src/data/sizeOf.ts
iwoplaza: I think `NaN` was there always, unless I am misremembering. `Unbounded` in typed binary meant that the size is `NaN`. We can of course change this, and return `0` instead, but I think `NaN` better represents the undefined behavior of retrieving the size of a runtime-sized schema.

packages/typegpu/src/core/buffer/buffer.ts
iwoplaza: In general, `BaseWgslData` is safer to use on the side of internal types, as I found using `AnyData` can quickly lead to Infinite Instantiation errors ☠️.

packages/typegpu/src/data/sizeOf.ts
iwoplaza: The `sizeOf` defined above is a lot more permissive on the type level, which is helpful to avoid excessive type casting, but the public one is restricted to ingest only known types.

packages/typegpu/src/data/attributes.ts
mhawryluk: oh, that's right. thought I tested it and it showed as `any`, but I just needed to rerun the build haha

iwoplaza:
I'm still leaning towards the short type names alternative:
```ts
import { vec4f, mat4x4f } from 'typegpu/data';

mat4x4f; // => Mat4x4f
vec4f; //> Vec4f

const hello = mat4x4f(); // => m4x4f
const hello = mat4x4f(); // => v4f

const sup = vec4f(); // => v4f
```

mhawryluk (approved):
Great work 🤯🚀

reczkok (approved):
Amazing work! 🙌🚀

PR #600, opened 2024-12-04
Exposing textures as bound from a bind group layout.

iwoplaza:


mhawryluk (approved):
LGTM! I think we should make use of this in some of our texture examples (blur/ct/ck) 🚀

reczkok (approved):
Samplers rise up! 🎨🖌️

PR #610, opened 2024-12-09
Use plugin in examples

reczkok:
Makes the example use our plugin which means we don't have to write `.$uses()` anymore. Also makes the editor readonly and removes the infinite loop prevention.


apps/typegpu-docs/astro.config.mjs
```diff
@@ -18,12 +19,39 @@ const stripFalsy = (items) =>
...
+ *   transform(code: string, id: string): { code: string } | undefined;
+ * }}
+ */
+function toRawPlugin() {
+  return {
+    name: 'to-raw',
+    enforce: 'post',
+
+    transform(code, id) {
+      if (id.includes('tgpu=true')) {
```
iwoplaza: TIny nit, probably does not matter, but lets be safe just in case.
```suggestion
      if (id.endsWith('?tgpu=true')) {
```

iwoplaza (approved):
Great job!! 🌟💛

PR #615, opened 2024-12-10
feat: Separate API for writing from copying.

iwoplaza:


packages/typegpu/tests/runtime.test.ts
```diff
@@ -1,183 +1,106 @@
...
-    dataBuffer.write({
-      a: 3,
-      b: { a: 4, b: 5, c: vec3i(6, 7, 8) },
-      c: vec4u(9, 10, 11, 12),
+      expect(() => root.unwrap(buffer)).toThrowError();
     });
 
-    const mockBuffer = root.unwrap(dataBuffer);
-    expect(mockBuffer).toBeDefined();
+    it('should return an the same buffer that was passed into .createBuffer', () => {
```
mhawryluk:
```suggestion
    it('should return the same buffer that was passed into .createBuffer', () => {
```

packages/typegpu/tests/unmanagedIO.test.ts
```diff
@@ -1,245 +0,0 @@
...
-      .$device(device);
-    const data = await tgpu.read(buffer);
-
-    expect(device.createBuffer).not.toHaveBeenCalled();
-    expect(data).toBeDefined();
-    expect(mockBuffer.getMappedRange).toHaveBeenCalled();
-    expect(mockBuffer.unmap).toHaveBeenCalled();
-  });
-
-  it('should read from a buffer', async () => {
```
mhawryluk, line 194: we don't have any tests for `read` (and `copyFrom`) no more :( 
maybe we could modify these instead of deleting them?

reczkok (approved):
Great work! We should update the buffer docs to mention `copyFrom`

mhawryluk:
I suggest we rename `runtime.test.ts` to `root.test.ts`

mhawryluk (approved):
Amazing 🚀

PR #617, opened 2024-12-11
Automatically add output struct to $uses for raw-WGSL implemented functions

### LINKED ISSUE #576: Automatically add output struct to `$uses` for raw-WGSL implemented functions.

mhawryluk:
closes #576


packages/typegpu/src/core/function/ioOutputStruct.ts
```diff
@@ -0,0 +1,65 @@
...
+    const outputName = implementation
+      .match(/->(?<output>.*?){/s)
+      ?.groups?.output?.trim();
+
+    if (outputName) {
+      const externals: Record<string, unknown> = {};
+      externals[outputName] = Output;
+      core.applyExternals(externals);
+    }
+  }
```
reczkok: What about the situation when we have for example:
```wgsl
@vertex 
fn main(@builtin(vertex_index) vertexIndex: u32) -> @builtin(position) vec4f {
   // ....
}
```

mhawryluk: there will be one unnecessary string replace operation, but I think we can live with that

reczkok: I think we could just check for whitespace after trimming? This should be the only scenario in which whitespace appears here.

mhawryluk: without checking for whitespace: it adds to uses { \`@builtin(position) vec4f\`: Output}, then tries to replace `@builtin(position) vec4f` with `vertex_fn_Output_1` in resolved code, but it doesn't work, I assume because of the parentheses, as they aren't escaped in the regex

packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -68,7 +73,8 @@ export function fragmentFn<
     returnType: outputType as ExoticIO<Output>,
 
     does(implementation): TgpuFragmentFn<ExoticIO<Varying>, ExoticIO<Output>> {
-      return createFragmentFn(this, implementation as Implementation);
+      // biome-ignore lint/suspicious/noExplicitAny: <no need>
+      return createFragmentFn(this, implementation as Implementation) as any;
     },
```
reczkok: We could do this to avoid casting as any
```ts
does(implementation) {
  return createFragmentFn(
    this,
    implementation as Implementation,
  ) as TgpuFragmentFn<ExoticIO<Varying>, ExoticIO<Output>>;
},
 ```

File: packages/typegpu/src/core/function/tgpuFragmentFn.ts [hunk 3]
```diff
@@ -84,9 +90,11 @@ function createFragmentFn(
   type This = TgpuFragmentFn;
 
   const core = createFnCore(shell, implementation);
+  const Output = createOutputStruct(core, implementation, shell.returnType);
```
reczkok: Is there any reason to have `Output` instead of `output`?
```suggestion
  const output = createOutputStruct(core, implementation, shell.returnType);
```

packages/typegpu/src/core/function/ioOutputStruct.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  if (typeof implementation === 'string') {
+    const outputName = implementation
+      .match(/->(?<output>.*?){/s)
+      ?.groups?.output?.trim();
+
+    if (outputName && !/\s/g.test(outputName)) {
+      const externals: Record<string, unknown> = {};
+      externals[outputName] = Output;
+      core.applyExternals(externals);
```
reczkok: Nit / personal preference - feel free to ignore
```suggestion
      core.applyExternals({
        outputName: Output,
      });
```

reczkok (approved):
Works great 🚀 Left some nitpicks to consider

packages/typegpu/src/core/function/tgpuFragmentFn.ts
iwoplaza: I did it initially because it's a schema, and we write schemas in PascalCase, but we can debate on what we should do here. An alternate argument is that other resources have schemas on them as well, but they are written in camelCase, like `dataType` for buffers or `returnType` for functions.

iwoplaza: So maybe `outputType` instead?

packages/typegpu/tests/rawFn.test.ts
```diff
@@ -150,4 +150,54 @@ describe('tgpu.fn with raw string WGSL implementation', () => {
...
+
+    expect(
+      tgpu.resolve({ input: [fragmentFunction], names: 'strict' }),
+    ).toContain(`\
+struct fragment_Output {
+  @location(0) a: vec4f,
+  @location(1) b: vec4f,
+}`);
+  });
 });
```
reczkok, line 223: Test case I used locally
```suggestion

  it('does not add redundant struct definition', () => {
    const vertexFunction = tgpu
      .vertexFn({ vertexIndex: builtin.vertexIndex }, builtin.position)
      .does(/* wgsl */ `(@builtin(vertex_index) vertexIndex: u32) -> @builtin(position) vec4f {
    let pos = array<vec2f, 6>(
      vec2<f32>( 1,  1),
      vec2<f32>( 1, -1),
      vec2<f32>(-1, -1),
      vec2<f32>( 1,  1),
      vec2<f32>(-1, -1),
      vec2<f32>(-1,  1)
    );

    return vec4f(pos[vertexIndex], 0, 1);
  }`)
      .$name('vertex_fn');

    expect(
      tgpu.resolve({ input: [vertexFunction], names: 'strict' }),
    ).not.toContain('struct');
  });
});
```

packages/typegpu/src/core/function/tgpuFragmentFn.ts
reczkok: I like it 👍

packages/typegpu/src/core/function/ioOutputStruct.ts
mhawryluk: I don't think that's the same

mhawryluk: works with `[outputName]:`

reczkok: That makes sense, my bad

packages/typegpu/src/core/function/ioOutputType.ts
```diff
@@ -0,0 +1,63 @@
...
+import { type Decorate, attribute, location } from '../../data/attributes';
+import { getCustomLocation, isData } from '../../data/dataTypes';
+import type { BaseWgslData, Location, WgslStruct } from '../../data/wgslTypes';
+import type { FnCore } from './fnCore';
+import type { IOData, IOLayout, Implementation } from './fnTypes';
+
+export type IOLayoutToOutputSchema<T extends IOLayout> = T extends BaseWgslData
+  ? Decorate<T, Location<0>>
+  : T extends Record<string, BaseWgslData>
+    ? WgslStruct<T>
```
iwoplaza: Wrap every entry of the struct (that isn't a builtin) with a Decorate that adds a `Location<number>`.

packages/typegpu/tests/rawFn.test.ts
```diff
@@ -150,4 +150,82 @@ describe('tgpu.fn with raw string WGSL implementation', () => {
...
+  @location(0) a: vec4f,
+  @location(1) b: vec4f,
+}`);
+    expect(resolved).toContain('-> fragment_Output {');
+    expect(resolved).not.toContain(' Output');
+  });
+
+  it("does not add redundant struct definition when there's no struct output", () => {
+    const vertexFunction = tgpu
+      .vertexFn({ vertexIndex: builtin.vertexIndex }, { out: builtin.position })
```
iwoplaza: Lets make this a fragment entry function instead.

iwoplaza (approved):
Awesome! This will slim up the code so much 💜

iwoplaza (approved):
Yeeehaw! 🐄 🤠 🌵

PR #620, opened 2024-12-11
fix: Allow fragment functions to use a subset of the vertex output on the type level.

iwoplaza:
### Error when fragment input contains something that the vertex output does not have.
Screenshot 2024-12-11 at 22 20 03

### Error when fragment input contains a key from the vertex output, but the type does not match.
Screenshot 2024-12-11 at 22 35 31


packages/typegpu/src/core/function/tgpuVertexFn.ts
```diff
@@ -59,27 +69,27 @@ export interface TgpuVertexFn<
...
     returnType: outputType as ExoticIO<Output>,
 
-    does(
-      implementation,
-    ): TgpuVertexFn<ExoticIO<VertexAttribs>, ExoticIO<Output>> {
-      // biome-ignore lint/suspicious/noExplicitAny: <no need>
+    does(implementation) {
+      // biome-ignore lint/suspicious/noExplicitAny: <no thanks>
       return createVertexFn(this, implementation as Implementation) as any;
     },
```
reczkok, line 92: Any Police 🚨 (I only include this comment to be fair to @mhawryluk nitpicked earlier on the same thing - I actually think `any` might be more readable here)
```suggestion
    does(implementation) {
      return createVertexFn(
        this,
        implementation as Implementation,
      ) as TgpuVertexFn<
        OmitBuiltins<ExoticIO<VertexIn>>,
        OmitBuiltins<ExoticIO<Output>>
      >;
    },
```

packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -50,24 +59,26 @@ export interface TgpuFragmentFn<
...
+  outputType: FragmentOut,
+): TgpuFragmentFnShell<ExoticIO<FragmentIn>, ExoticIO<FragmentOut>> {
   return {
-    argTypes: [varyingTypes as ExoticIO<Varying>],
-    returnType: outputType as ExoticIO<Output>,
+    argTypes: [inputType as ExoticIO<FragmentIn>],
+    returnType: outputType as ExoticIO<FragmentOut>,
 
-    does(implementation): TgpuFragmentFn<ExoticIO<Varying>, ExoticIO<Output>> {
+    does(implementation) {
```
mhawryluk, line 81: please either add method return type annotation or `as any`, sometimes I get type errors here (but sometimes not lol)

mhawryluk (approved):
Great job! 🚀

reczkok (approved):
Great job 💜

PR #626, opened 2024-12-11
Improve labeling gpu objects

### LINKED ISSUE #582: Investigate missing labels on buffers
The label given to buffer during creation (via root) seems to vanish when actually created (on device). This could be caused by usage not properly populating descriptors. This could also be a bug in WebGPU Inspector (which led mi to find this issue).

mhawryluk:
closes #582


reczkok (approved):
Great work 🕵️‍♂️🔎

iwoplaza (approved):
Awesome! 🌟

PR #634, opened 2024-12-12
Bring back elementCount property on TgpuArray for now

mhawryluk:


iwoplaza (approved):
Awesome! Let's merge it 💜

PR #636, opened 2024-12-12
`root.createBindGroup` API replacing `layout.populate`

### LINKED ISSUE #631: Replace `layout.populate` with new `root.createBindGroup` API

mhawryluk:
closes #631


iwoplaza (approved):
Awesome! 🎉 🌮

reczkok (approved):
Great! Are we replacing the populate API or is this just an additional way? I don't see it hidden here.

iwoplaza:
> Great! Are we replacing the populate API or is this just an additional way? I don't see it hidden here.

We're gonna be deprecating it, but only on the 0.3 branch for now so that it still looks fine in the examples.

PR #637, opened 2024-12-12
chore: Made some APIs public for 0.3

iwoplaza:


packages/typegpu/src/index.ts
```diff
@@ -20,33 +19,65 @@ export const tgpu = {
...
+} from './core/sampler/sampler';
+export {
+  isSampledTextureView,
+  isStorageTextureView,
+  isTexture,
+} from './core/texture/texture';
+export {
+  isUsableAsRender,
+  isUsableAsSampled,
+} from './core/texture/usageExtension';
```
reczkok, line 48: Should we also expose `isBuffer`?

reczkok (approved):
Exciting! Bye bye legacy buffers [*]

packages/typegpu/src/index.ts
iwoplaza: That could be helpful as a public API, I exposed it! 🌄

PR #641, opened 2024-12-13
Deprecate `populate` method on bindGroupLayout

mhawryluk:
references #631


packages/typegpu/src/experimental/index.ts
iwoplaza, line 1: 🔫 Delete this

iwoplaza (approved):
Awesome! 🔔

PR #644, opened 2024-12-16
Allow only vertex usage for buffers containing loose data

### LINKED ISSUE #596: Allow only vertex usage for buffers with a loose schema.

reczkok:
closes #596

### DIFF
--- a/packages/typegpu/src/core/buffer/buffer.ts
+++ b/packages/typegpu/src/core/buffer/buffer.ts
@@ -1,11 +1,9 @@
 import { BufferReader, BufferWriter } from 'typed-binary';
+import { isWgslData } from '../../data';
 import { readData, writeData } from '../../data/dataIO';
-import {
-  type AnyData,
-  isLooseArray,
-  isLooseStruct,
-} from '../../data/dataTypes';
+import type { AnyData } from '../../data/dataTypes';
 import { sizeOf } from '../../data/sizeOf';
+import type { WgslTypeLiteral } from '../../data/wgslTypes';
 import type { Storage } from '../../extension';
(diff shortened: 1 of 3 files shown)


packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -58,6 +62,12 @@ export function INTERNAL_createBuffer<TData extends AnyData>(
   typeSchema: TData,
   initialOrBuffer?: Infer<TData> | GPUBuffer,
 ): TgpuBuffer<TData> {
+  if (isLooseArray(typeSchema) || isLooseStruct(typeSchema)) {
```
iwoplaza: Tiny nit, can be made more general in case we add more loose schemas.
```suggestion
  if (!isWgslData(typeSchema)) {
```

packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -83,6 +93,12 @@ export function isUsableAsVertex<T extends TgpuBuffer<AnyData>>(
 // Implementation
 // --------------
 
+type RestrictVertexUsages<TData extends AnyData> = TData extends
+  | { readonly type: 'loose-array' }
+  | { readonly type: 'loose-struct' }
+  ? 'vertex'[]
+  : ('uniform' | 'storage' | 'vertex')[];
```
iwoplaza: Tiny nit, can be made a bit more general, in case we want to add more loose schemas.

```suggestion
type RestrictVertexUsages<TData extends AnyData> = TData extends
  | { readonly type: WgslTypeLiteral }
  ? ('uniform' | 'storage' | 'vertex')[]
  : 'vertex'[];
```

PR #647, opened 2024-12-16
chore: Move slots into experimental `tgpu`.

iwoplaza:


packages/typegpu/src/core/resolve/tgpuResolve.ts
```diff
@@ -44,6 +44,12 @@ export function resolve(options: TgpuResolveOptions): string {
     },
   };
 
+  Object.defineProperty(resolutionObj, 'toString', {
+    value() {
+      return '<root>';
+    },
+  });
```
iwoplaza: Makes it look a lot better in the error logs.

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -150,6 +145,10 @@ function createFn<
     get: () => core.label,
   });
 
+  Object.defineProperty(fn, 'toString', {
+    value: () => `fn:${core.label ?? '<unnamed>'}`,
```
mhawryluk, line 149: why `value` and not `get`?

packages/typegpu/src/core/resolve/tgpuResolve.ts
```diff
@@ -44,6 +44,12 @@ export function resolve(options: TgpuResolveOptions): string {
     },
   };
 
+  Object.defineProperty(resolutionObj, 'toString', {
+    value() {
+      return '<root>';
+    },
```
mhawryluk: arrow function to make it one line

packages/typegpu/src/core/slot/slot.ts
```diff
@@ -0,0 +1,42 @@
...
+// Implementation
+// --------------
+
+class TgpuSlotImpl<T> implements TgpuSlot<T> {
+  readonly resourceType = 'slot';
+  public label?: string | undefined;
+
+  constructor(public defaultValue: T | undefined = undefined) {}
+
+  public $name(label: string) {
```
mhawryluk:
```suggestion
  $name(label: string) {
```

packages/typegpu/src/experimental/index.ts
```diff
@@ -33,6 +34,7 @@ export const tgpu = {
   computeFn,
   vertexLayout,
   bindGroupLayout,
+  slot,
```
mhawryluk, line 37: shouldn't we export other functions and types (`isSlot`, `TgpuSlot`, ...) too?

mhawryluk (approved):
great!! 🚀

packages/typegpu/src/core/function/tgpuFn.ts
iwoplaza: Because `get` makes this property a `computed` property, meaning `.toString` would be the string, instead of a function that returns a string.

mhawryluk: oh, got it

PR #656, opened 2024-12-17
Move standard functions to typegpu/std

### LINKED ISSUE #654: Move standard functions to `typegpu/std`

reczkok:
closes #654


packages/typegpu/src/std/index.ts
```diff
@@ -0,0 +1,5 @@
+import * as functions from './numeric';
+
+export const std = {
+  ...functions,
+};
```
iwoplaza: Lets just export these functions flat, outside of an object, so:

```ts
import { add, mul, ... } from 'typegpu/std';
```

or

```ts
import * as std from 'typegpu/std';
```

iwoplaza (commented):
We need to make `typegpu/std` a new entry point as well, so the `typegpu/package.json` needs a touchup, as well as the `typegpu/tsup.config.ts` file.
There are a couple places like this, best to look everywhere `typegpu/experimental` is being referenced in code.

packages/typegpu/src/std/index.ts
```diff
@@ -0,0 +1 @@
+export * from './numeric';
```
iwoplaza: I know its a lot, but lets list out every exported function, just so we can better control the public API.

iwoplaza (approved):
🦠🦠🦠

PR #671, opened 2024-12-18
Update fluid with atomics example

### LINKED ISSUE #657: Update fluid with atomics example

reczkok:
closes #657


mhawryluk:
```
packages/typegpu dev:watch: Export "AnyWgslData" of module "src/data/wgslTypes.ts" was reexported through module "src/data/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
packages/typegpu dev:watch: Either change the import in "src/core/slot/accessor.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
packages/typegpu dev:watch: Export "AnyWgslData" of module "src/data/wgslTypes.ts" was reexported through module "src/data/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
packages/typegpu dev:watch: Either change the import in "src/core/slot/slotTypes.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
packages/typegpu dev:watch: Export "AnyData" of module "src/data/dataTypes.ts" was reexported through module "src/data/index.ts" while both modules are dependencies of each other and will end up in different chunks by current Rollup settings. This scenario is not well supported at the moment as it will produce a circular dependency between chunks and will likely lead to broken execution order.
packages/typegpu dev:watch: Either change the import in "src/tgpuBindGroupLayout.ts" to point directly to the exporting module or reconfigure "output.manualChunks" to ensure these modules end up in the same chunk.
```

⚠️

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -54,157 +48,207 @@ function encodeBrushType(brushType: (typeof BrushTypes)[number]) {
...
-  let cell = ${getCellNext}(x, y);
-  let waterLevel = cell & ${maxWaterLevel};
+}`)
+  .$uses({ sizeData: sizeUniform });
+
+const getCell = tgpu
+  .fn([d.u32, d.u32], d.u32)
+  .does(/* wgsl */ `(x: u32, y: u32) -> u32 {
+  return currentStateData[getIndex(x, y)];
+}`)
```
mhawryluk:
```suggestion
const getCell = tgpu
  .fn([d.u32, d.u32], d.u32)
  .does(/* wgsl */ `(x: u32, y: u32) -> u32 {
    return currentStateData[getIndex(x, y)];
  }`)
```

same for all the other functions above and below

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -54,157 +48,207 @@ function encodeBrushType(brushType: (typeof BrushTypes)[number]) {
...
+  .does(/* wgsl */ `(x: u32, y: u32, value: u32) {
+  atomicStore(&nextStateData[getIndex(x, y)], value);
+}`)
+  .$uses({ nextStateData: nextStateStorage, getIndex });
+
+const addToCell = tgpu
+  .fn([d.u32, d.u32, d.u32])
+  .does(/* wgsl */ `(x: u32, y: u32, value: u32) {
+  let cell = getCellNext(x, y);
+  let waterLevel = cell & ${MAX_WATER_LEVEL};
```
mhawryluk: we can put `MAX_WATER_LEVEL` in `$uses` instead of using string interpolation and I'd go with that instead to make it more unified, but I don't insist. same everywhere else

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -54,157 +48,207 @@ function encodeBrushType(brushType: (typeof BrushTypes)[number]) {
...
 
-const getStableStateBelow = wgsl.fn`(upper: u32, lower: u32) -> u32 {
+const getStableStateBelow = tgpu.fn([d.u32, d.u32], d.u32).does(/* wgsl */ `(upper: u32, lower: u32) -> u32 {
   let totalMass = upper + lower;
-  if (totalMass <= ${maxWaterLevelUnpressurized}) {
+  if (totalMass <= ${MAX_WATER_LEVEL_UNPRESSURIZED}) {
     return totalMass;
-  } else if (totalMass >= ${maxWaterLevelUnpressurized}*2 && upper > lower) {
-    return totalMass/2 + ${maxCompress};
+  } else if (totalMass >= ${MAX_WATER_LEVEL_UNPRESSURIZED}*2 && upper > lower) {
```
mhawryluk:
```suggestion
  } else if (totalMass >= ${MAX_WATER_LEVEL_UNPRESSURIZED} * 2 && upper > lower) {
```

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -313,60 +381,53 @@ let renderChanges: () => void;
...
+  const compute = tgpu
+    .computeFn([d.builtin.globalInvocationId], {
+      workgroupSize: [options.workgroupSize, options.workgroupSize],
+    })
+    .does(/* wgsl */ `(@builtin(global_invocation_id) gid: vec3u) {
+    decideWaterLevel(gid.x, gid.y);
+  }`)
+    .$uses({ decideWaterLevel });
+
+  const cpp = root.withCompute(compute).createPipeline();
```
mhawryluk: `cpp` -> `computePipeline`

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -313,60 +381,53 @@ let renderChanges: () => void;
...
+  const vp = root
```
mhawryluk: `vp` -> `renderPipeline`

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -293,15 +351,25 @@ const fragWGSL = wgsl`
...
-  var r = f32((u32(cell) >> 16) & 0xFF)/255.;
-  var g = f32((u32(cell) >> 8) & 0xFF)/255.;
-  var b = f32(u32(cell) & 0xFF)/255.;
-  if (r > 0.) { g = 1.;}
-  if (g > 0.) { b = 1.;}
-  if (b > 0. && b < 0.5) { b = 0.5;}
+  let normalized = min(cell / f32(0xFF), 1.);
+
+  if (normalized == 0.) {
+    return vec4f(0.937, 0.937, 0.976, 1.);
```
mhawryluk:
```suggestion
    return vec4f();
```

mhawryluk (changes requested):
- remember to merge the main branch, because experimental is no more
- brush size value is effectively shifted by 1 (setting 0 draws 1 square)
- it is very difficult to draw anything with timestep set to low values (don't know if we can do anything about this, maybe rearranging the order of operations could help somehow?)

packages/typegpu/src/data/attributes.ts
```diff
@@ -337,6 +337,9 @@ class DecoratedImpl<TInner extends BaseWgslData, TAttribs extends unknown[]>
   implements Decorated<TInner, TAttribs>
 {
   public readonly type = 'decorated';
+  public readonly '~memIdent'!: TAttribs extends Location<number>[]
+    ? MemIdentity<TInner> | Decorated<MemIdentity<TInner>, TAttribs>
+    : Decorated<TInner, TAttribs>;
```
iwoplaza: I am not 100% sure, but wouldn't the MemIdentity of a Decorated type be this instead?:
```suggestion
  public readonly '~memIdent'!: TAttribs extends Location<number>[]
    ? MemIdentity<TInner> | Decorated<MemIdentity<TInner>, TAttribs>
    : Decorated<MemIdentity<TInner>, TAttribs>;
```

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -416,29 +494,25 @@ function resetGameData() {
...
 
-  const cellSize = canvas.width / options.size;
-  const x = Math.floor((event.offsetX * window.devicePixelRatio) / cellSize);
-  const y =
-    options.size -
-    Math.floor((event.offsetY * window.devicePixelRatio) / cellSize) -
-    1;
-  const allAffectedCells = [];
+const handleDrawing = (x: number, y: number) => {
+  const allAffectedCells = new Set<{ x: number; y: number }>();
```
iwoplaza: Turn back into array as the Set does not guarantee uniqueness in this case 🙏

iwoplaza (approved):
A W E S O M E

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
reczkok: It actually needed neither 🫡

PR #674, opened 2024-12-18
Make plugin not override anything passed in $uses

### LINKED ISSUE #670: Allow overriding function externals via $uses when using plugin
Reproduction:
```ts
  const x = 1;

  const main = tgpu
    .fn([], d.f32)
    .does(() => {
      return x;
    })
    .$uses({
      x: 2,
    })
    .$name('main');

  console.log(tgpu.resolve({ input: main, names: 'strict' }));
```

Current result:

```
fn main() -> f32{
  return 1;
}
```

Expected:

```
fn main() -> f32{
  return 2;
}
```

reczkok:
closes #670


mhawryluk (approved):
Thanks! 🎅

PR #687, opened 2024-12-19
Fix misleading error message when texture validation fails due to missing usage.

reczkok:
When missing the proper usage in texture we are getting a misleading error message. Fixed by swapping the order of intersection type.

Before:
Screenshot 2024-12-19 at 15 10 23

After:
Screenshot 2024-12-19 at 15 10 02


mhawryluk (approved):

iwoplaza (approved):
🫡🫡🫡

PR #695, opened 2024-12-20
Create `tgpu.accessor` API

### LINKED ISSUE #652: Add a resource locator/accessor API
## Purpose
- Inversion of control, being able to provide resources to libraries/generic modules.

## Common use-cases
- Giving shaders a context of the viewport size.
  - Its limiting to assume that its going to be the only data in a buffer, or a specific property on a structure. Resource locators allow the end developer to decide where this information should be allocated.
- (...)

## Capabilities
- Accessors can be provided with a buffer usage that matches the accessor's data-type.
- Accessors can be provided with a function that has a return type matching the accessor's data-type.

## API Sketch

```ts
//
// Utilities
//

const getViewportSize = tgpu.accessor(vec2f); // => TgpuAccessor<Vec2f>

const pixelToUV = tgpu
  .fn([vec2f], vec2f)
  .does((pixelCoord) => {
    const viewportSize = getViewportSize(); // => v2f
    return div(pixelCoord, viewportSize);
  });

//
// App code
//

const viewportSizeBuffer = root
  .createBuffer(vec2f)
  .$usage('uniform');

const main = tgpu
  .fragmentFn({ coord: builtin.position }, vec4f)
  .does(({ coord }) => {
    const uv = pixelToUV(coord);
    // ...
  })
  .with(getViewportSize, viewportSizeBuffer.asUniform());
```

mhawryluk:
closes #652

### DIFF
--- a/packages/typegpu/src/core/function/tgpuFn.ts
+++ b/packages/typegpu/src/core/function/tgpuFn.ts
@@ -1,9 +1,16 @@
+import type { Infer } from '../../data';
 import type { Exotic, ExoticArray } from '../../data/exotic';
 import type { AnyWgslData } from '../../data/wgslTypes';
 import { inGPUMode } from '../../gpuMode';
 import type { TgpuNamable } from '../../namable';
 import type { ResolutionCtx, TgpuResolvable, Wgsl } from '../../types';
-import type { Eventual, TgpuSlot } from '../slot/slotTypes';
+import type { TgpuBufferUsage } from '../buffer/bufferUsage';
+import {
+  type Eventual,
+  type TgpuAccessor,
+  type TgpuSlot,
+  isAccessor,
+} from '../slot/slotTypes';
 import { createFnCore } from './fnCore';
 import type { Implementation, InferArgs, InferReturn } from './fnTypes';
 
@@ -46,6 +53,10 @@ interface TgpuFnBase<
 
   $uses(dependencyMap: Record<string, unknown>): this;
   with<T>(slot: TgpuSlot<T>, value: Eventual<T>): TgpuFn<Args, Return>;
+  with<T extends AnyWgslData>(
+    accessor: TgpuAccessor<T>,
+    value: TgpuFn<[], T> | TgpuBufferUsage<T> | Infer<T>,
+  ): TgpuFn<Args, Return>;
 }
 
 export type TgpuFn<
@@ -121,8 +132,15 @@ function createFn<
       return this;
     },
 
-    with(slot, value): TgpuFn<Args, Return> {
-      return createBoundFunction(fn, slot, value);
+    with<T extends AnyWgslData>(
+      slot: TgpuSlot<T> | TgpuAccessor<T>,
+      value: T | TgpuFn<[], T> | TgpuBufferUsage<T> | Infer<T>,
+    ): TgpuFn<Args, Return> {
+      return createBoundFunction(
+        fn,
+        isAccessor(slot) ? slot.slot : slot,
+        value,
+      );
     },
 
     resolve(ctx: ResolutionCtx): string {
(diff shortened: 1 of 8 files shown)


packages/typegpu/src/core/slot/slotTypes.ts
```diff
@@ -15,7 +19,6 @@ export interface TgpuSlot<T> extends TgpuNamable {
 
   readonly value: Infer<T>;
 }
-
```
iwoplaza: Why the line deletion?

packages/typegpu/tests/accessor.test.ts
```diff
@@ -0,0 +1,177 @@
...
+const RED_RESOLVED = 'vec3f(1, 0, 0)';
+
+const resolutionRootMock = {
+  toString() {
+    return '<root>';
+  },
+} as TgpuResolvable;
+
+describe('tgpu.accessor', () => {
+  it('resolves to invokation of provided function', () => {
```
iwoplaza:
```suggestion
  it('resolves to invocation of provided function', () => {
```

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -0,0 +1,80 @@
...
+      | Infer<T>
+      | undefined = undefined,
+  ) {
+    this.slot = slot(defaultValue);
+  }
+
+  $name(label: string) {
+    this.label = label;
+    return this;
+  }
```
iwoplaza: Tiny nit, since the slot is internal, we can name is the same as the accessor.
```suggestion
  $name(label: string) {
    this.label = label;
    this.slot.$name(label);
    return this;
  }
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -84,6 +95,15 @@ export function procedure(implementation: () => void) {
   return fn([]).does(implementation);
 }
 
+export function isTgpuFn<
```
iwoplaza, line 99: Could we give TgpuFn-s a `resourceType` and check against that instead?

packages/typegpu/src/core/slot/slotTypes.ts
mhawryluk: by mistake

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -0,0 +1,81 @@
...
+export class TgpuAccessorImpl<T extends AnyWgslData>
+  implements TgpuAccessor<T>
+{
+  readonly resourceType = 'accessor';
+  public label?: string | undefined;
+  public slot: TgpuSlot<TgpuFn<[], T> | TgpuBufferUsage<T> | Infer<T>>;
+
+  constructor(
+    public schema: T,
+    public defaultValue:
```
reczkok: I think those could be readonly (I know they are on the interface but just to be consistent)

reczkok (approved):
Great work! Love the thorough testing 🛠️

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  get value(): Infer<T> {
+    const ctx = getResolutionCtx();
+    if (!ctx) {
+      throw new Error(
+        `Cannot access tgpu.accessor's value outside of resolution.`,
+      );
+    }
+
+    return ctx.unwrap(this.slot) as unknown as Infer<T>;
```
iwoplaza: I am not sure if that would be the case, but I think this might cause using accessors in TGSL not possible.
```suggestion
    return this as unknown as Infer<T>;
```

iwoplaza (approved):
Just one nitpick, otherwise, awesome work! 🙌

PR #697, opened 2024-12-20
🎁: Add an alternate slot-value propagation mechanism, allowing for slot values to pass through a derived boundary

iwoplaza:
This fixes some edge cases where derived was interaction with slots.


packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -111,6 +115,60 @@ export interface ColorAttachment {
   storeOp: GPUStoreOp;
 }
 
+export interface DepthStencilAttachment {
+  /**
+   * A {@link GPUTextureView} describing the texture subresource that will be output to
```
mhawryluk:
```suggestion
   * A {@link GPUTextureView | (TgpuTexture & Render)} describing the texture subresource that will be output to
```

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -235,37 +304,70 @@ class TgpuRenderPipelineImpl implements TgpuRenderPipeline {
...
+      if (buffer) {
+        missingVertexLayouts.delete(vertexLayout);
+        pass.setVertexBuffer(idx, branch.unwrap(buffer));
       }
-      pass.setVertexBuffer(idx, branch.unwrap(buffer));
     });
 
+    for (const layout of missingBindGroups) {
+      throw new MissingBindGroupError(layout.label);
+    }
```
mhawryluk: first thrown error will exit out of this method, so this loop is pointless. at least should be changed to work just as `missingVertexLayouts` below

packages/typegpu/src/data/index.ts
```diff
@@ -136,6 +136,7 @@ export {
   BuiltinNumWorkgroups,
   AnyBuiltin,
 } from '../builtin';
+export type { Exotic as Base } from './exotic';
```
mhawryluk, line 139: will this aliased export also change how Exotic appears as part of other types when hovering over them?

mhawryluk (approved):
if you have an example code snippet that prompted this change, then it would be great to paste it in as a new test, so we're sure it works from now on! 🚀

packages/typegpu/src/data/index.ts
iwoplaza: Honestly not sure. I did not want to pack too much as part of this PR, we can rename all exotics to base in a separate one.

reczkok (approved):
Great work! I agree with @mhawryluk that having some tests would be nice 🕵️‍♂️

PR #700, opened 2024-12-21
fix: Export `TgpuRoot` type, and narrower `init` return type.

iwoplaza:


packages/typegpu/src/index.ts
```diff
@@ -17,7 +18,7 @@ export const tgpu = {
 
   bindGroupLayout,
 
-  init,
+  init: init as (options?: InitOptions) => Promise<TgpuRoot>,
```
reczkok, line 21: Why does it have to be cast? Isn't the type already that?

reczkok (approved):

packages/typegpu/src/index.ts
iwoplaza: Not on the 0.2 branch, it still returns ExperimentalTgpuRoot there

afternoon2:
works, thanks! 🎉

PR #707, opened 2024-12-22
docs: Removed (potentially) redundant `onSubmittedWorkDone`

iwoplaza:
The example seems to work without the call to `onSubmittedWorkDone`.


iwoplaza:
It seems that, because the example currently uses both the internal command encoder (in the root) and an external command encoder, it has to be called to get properly synchronized.

PR #718, opened 2025-01-07
Introduce host shareable type

reczkok:


reczkok:
It's not worth it. WebGPU will provide a descriptive error at shader compilation time. Trying to properly type when a buffer is host-shareable is doable but tricky.

PR #720, opened 2025-01-07
docs: Add info about `$addFlags` method to buffer docs

### LINKED ISSUE #639: Write docs about `addFlags`, as it still has its uses.
### Uses:
- Adding `MAP_READ` or `MAP_WRITE` to a buffer.

mhawryluk:
closes #639 

> [!CAUTION]
> currently it is not possible to add the `MAP_READ` or `MAP_WRITE` flags, because we include both `COPY_DST` and `COPY_SRC` by default every time. is that okay and we just add info to docs or should we fix this?

### DIFF
--- a/apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
@@ -122,7 +122,8 @@ buffer.$addFlags(GPUBufferUsage.QUERY_RESOLVE);
 ```
 
 :::note
-Every typed buffer created without providing existing GPU buffer has `COPY_DST` and `COPY_SRC` flags included by default.
+Every typed buffer created without providing existing GPU buffer has `COPY_DST` and `COPY_SRC` flags included by default. 
+However once the `MAP_READ` flag is provided, the only other flag set is `COPY_DST`. Similarly when setting `MAP_WRITE` it is paired with `COPY_SRC`.
 :::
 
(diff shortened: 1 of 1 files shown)


reczkok:
> currently it is not possible to add the `MAP_READ` or `MAP_WRITE` flags, because we include both `COPY_DST` and `COPY_SRC` by default every time. is that okay and we just add info to docs or should we fix this?

I think this is fine. There should rarely be any reason to manually create those buffers with our IO utilities (especially after we implement masking of some sort). If someone really wants to they can just create a buffer via device and wrap it.

apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
```diff
@@ -102,17 +102,32 @@ const buffer = root.createBuffer(d.u32)
...
+It is also possible to add any of the `GPUBufferUsage` flags to a typed buffer object, using the `.$addFlags` method.
+Though it shouldn't be necessary in most scenarios as majority of the flags are handled automatically by the library 
+or indirectly through the `.$usage` method.
+
+```ts
+buffer.$addFlags(GPUBufferUsage.QUERY_RESOLVE);
+```
+
+:::note
+Every typed buffer created without providing existing GPU buffer has `COPY_DST` and `COPY_SRC` flags included by default.
```
reczkok: We should mention the new `MAP_READ/WRITE` behavior

mhawryluk: done ✅

PR #732, opened 2025-01-07
feat: Handling unary operators & order of binary operators better.

iwoplaza:
The way of handling order of operations in this PR is a bit eager, causing every binary expression to be surrounded by parentheses, but it... works for now 🤷

### DIFF
--- a/packages/typegpu/src/errors.ts
+++ b/packages/typegpu/src/errors.ts
@@ -1,8 +1,6 @@
 import type { TgpuBuffer } from './core/buffer/buffer';
-import type { TgpuDerived, TgpuSlot } from './core/slot/slotTypes';
+import type { TgpuSlot } from './core/slot/slotTypes';
 import type { AnyData } from './data/dataTypes';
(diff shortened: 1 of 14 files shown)


packages/typegpu/src/errors.ts
iwoplaza, line 1: Embrace simplicity... 🧘‍♂️

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -65,7 +86,21 @@ function generateExpression(
...
+    };
+  }
+
+  if ('u' in expression) {
+    // Unary Expression
+
+    const [op, arg] = expression.u;
+    const argExpr = resolveRes(ctx, generateExpression(ctx, arg));
+    return {
+      value: `${op}${argExpr}`,
```
mhawryluk, line 103: don't we need parentheses here too? around ${argExpr}

mhawryluk (approved):
cool, thanks! 🚀

reczkok (approved):
Looks great 👍

PR #752, opened 2025-01-09
Auto-use all struct types in the shell of WGSL-implemented functions

### LINKED ISSUE #701: Resolve all arg types of typed functions.
**Reason:** So that we do not need to put struct types used in the shell in `$uses` as well (for WGSL-implemented functions).

mhawryluk:
closes #701


packages/typegpu/src/core/resolve/externals.ts
```diff
@@ -22,6 +22,42 @@ export function applyExternals(
...
+    Object.fromEntries(
+      argTypes.flatMap((argType, i) => {
+        const argTypeName = argTypeNames ? argTypeNames[i] : undefined;
+        return isWgslStruct(argType) && argTypeName !== undefined
+          ? [[argTypeName, argType]]
+          : [];
+      }),
+    ),
+  );
+}
```
iwoplaza: It's not guaranteed that it will work with builtins (or other attributes), let's try to add some test coverage for that.

iwoplaza (approved):
Awesome!!!!

packages/typegpu/src/core/resolve/externals.ts
mhawryluk: it indeed didn't work. it should now

iwoplaza (approved):
✂️❤️✂️

PR #761, opened 2025-01-12
Add @interpolate attribute

deluksic:
I tried making it as typesafe as I can, by preventing usage of `perspective` or `linear` with integer types. We could also not handle this to simplify the code a bit. Let me know what you prefer.

Example usage:
```typescript
const Data = d.struct({
  a: d.f32, // has implicit 'perspective, center' interpolation
  b: d.interpolate('linear, sample', d.f32),
});
```

The only unsafe thing I see potentially is that this attribute [must](https://www.w3.org/TR/WGSL/#interpolate-attr) be defined together with `@location`:
```
Must only be applied to a declaration that has a location attribute applied.
```

I didn't know how to enforce this.

I also considered using `['perspective', 'sample']` approach, but typescript would not narrow down after writing `['perspective', ...]`. Then I considered an object approach, but that just felt too clunky. For me, string approach feels just right.

### DIFF
--- a/packages/typegpu/src/data/attributes.ts
+++ b/packages/typegpu/src/data/attributes.ts
@@ -16,7 +16,13 @@ import {
   type BaseWgslData,
   type Builtin,
   type Decorated,
+  type FlatInterpolatableData,
+  type FlatInterpolationType,
+  type Interpolate,
+  type InterpolationType,
   type Location,
+  type PerspectiveOrLinearInterpolatableData,
+  type PerspectiveOrLinearInterpolationType,
   type Size,
   type WgslTypeLiteral,
   isAlignAttrib,
@@ -53,7 +59,8 @@ export type AnyAttribute =
   | Align<number>
   | Size<number>
   | Location<number>
-  | Builtin<BuiltinName>;
+  | Builtin<BuiltinName>
+  | Interpolate<InterpolationType>;
 
 export type ExtractAttributes<T> = T extends {
   readonly attribs: unknown[];
@@ -184,6 +191,68 @@ export function location<TLocation extends number, TData extends AnyData>(
 ): Decorate<Exotic<TData>, Location<TLocation>> {
   // biome-ignore lint/suspicious/noExplicitAny: <tired of lying to types>
   return attribute(data, { type: '@location', value: location }) as any;
+}
+
+/**
+ * Specifies how user-defined vertex shader output (fragment shader input)
+ * must be interpolated.
+ *
+ * Tip: Integer outputs cannot be interpolated.
+ *
+ * @example
+ * const Data = d.ioStruct({
+ *   a: d.f32, // has implicit 'perspective, center' interpolation
+ *   b: d.interpolate('linear, sample', d.f32),
+ * });
+ *
+ * @param location The explicit numeric location.
+ * @param data The data-type to wrap.
+ */
+export function interpolate<
+  TInterpolation extends PerspectiveOrLinearInterpolationType,
+  TData extends PerspectiveOrLinearInterpolatableData,
+>(
+  interpolationType: TInterpolation,
+  data: TData,
+): Decorate<Exotic<TData>, Interpolate<TInterpolation>>;
+
+/**
+ * Specifies how user-defined vertex shader output (fragment shader input)
+ * must be interpolated.
+ *
+ * Tip: Default sampling method of `flat` is `first`. Unless you specifically
+ * need deterministic behavior provided by `'flat, first'`, prefer explicit
+ * `'flat, either'` as it could be slightly faster in hardware.
+ *
+ * @example
+ * const Data = d.ioStruct({
+ *   a: d.f32, // has implicit 'perspective, center' interpolation
+ *   b: d.interpolate('flat, either', d.u32), // integer outputs cannot interpolate
+ * });
+ *
+ * @param location The explicit numeric location.
+ * @param data The data-type to wrap.
+ */
+export function interpolate<
+  TInterpolation extends FlatInterpolationType,
+  TData extends FlatInterpolatableData,
+>(
+  interpolationType: TInterpolation,
+  data: TData,
+): Decorate<Exotic<TData>, Interpolate<TInterpolation>>;
+
+export function interpolate<
+  TInterpolation extends InterpolationType,
+  TData extends AnyData,
+>(
+  interpolationType: TInterpolation,
+  data: TData,
+): Decorate<Exotic<TData>, Interpolate<TInterpolation>> {
+  return attribute(data, {
+    type: '@interpolate',
+    value: interpolationType,
+    // biome-ignore lint/suspicious/noExplicitAny: <tired of lying to types>
+  }) as any;
 }
 
 export function isBuiltin<
--- a/packages/typegpu/src/data/attributes.ts
+++ b/packages/typegpu/src/data/attributes.ts
@@ -176,11 +176,11 @@ export function size<TSize extends number, TData extends AnyData>(
  * Assigns an explicit numeric location to a struct member or a parameter that has this type.
  *
  * @example
- * const Data = d.struct({
+ * const VertexOutput = {
  *   a: d.u32, // has implicit location 0
  *   b: d.location(5, d.u32),
  *   c: d.u32, // has implicit location 6
- * });
+ * };
  *
  * @param location The explicit numeric location.
  * @param data The data-type to wrap.
--- a/packages/typegpu/src/data/attributes.ts
+++ b/packages/typegpu/src/data/attributes.ts
@@ -176,7 +176,7 @@ export function size<TSize extends number, TData extends AnyData>(
  * Assigns an explicit numeric location to a struct member or a parameter that has this type.
  *
  * @example
- * const Data = d.ioStruct({
+ * const Data = d.struct({
  *   a: d.u32, // has implicit location 0
  *   b: d.location(5, d.u32),
  *   c: d.u32, // has implicit location 6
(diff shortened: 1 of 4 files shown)


packages/typegpu/src/data/attributes.ts
```diff
@@ -186,6 +193,68 @@ export function location<TLocation extends number, TData extends AnyData>(
...
+ * @example
+ * const Data = d.ioStruct({
+ *   a: d.f32, // has implicit 'perspective, center' interpolation
+ *   b: d.interpolate('flat, either', d.u32), // integer outputs cannot interpolate
+ * });
+ *
+ * @param location The explicit numeric location.
+ * @param data The data-type to wrap.
+ */
+export function interpolate<
```
deluksic, line 236: it has two overloads with slightly different types and descriptions to enforce compatibility of interpolation and data type.

reczkok:
Hey there! Thanks for the contribution. At first I thought I would prefer a double template implementation like this
```ts
export type InterpolationType = 'perspective' | 'linear' | 'flat';
export type InterpolationSamplingType<T extends InterpolationType> =
  T extends 'flat' ? 'first' | 'either' : 'center' | 'centroid' | 'sample';

export interface Interpolate<
  TType extends InterpolationType,
  TSample extends InterpolationSamplingType<TType>,
> {
  readonly type: '@interpolate';
  readonly value: `${TType}, ${TSample}`;
}
```
But when I wanted to create an implementation that is faithful to the spec when it comes to defaults and also consistent with other attributes when it comes to argument order it led me to this monstrosity:
```ts
export function interpolate<
  TData extends PerspectiveOrLinearInterpolatableData,
>(data: TData): Decorate<Exotic<TData>, Interpolate<'perspective', 'center'>>;

export function interpolate<
  TInterpolation extends InterpolationType,
  TData extends TInterpolation extends 'flat'
    ? FlatInterpolatableData
    : PerspectiveOrLinearInterpolatableData,
>(
  interpolationType: TInterpolation,
  data: TData,
): Decorate<
  Exotic<TData>,
  Interpolate<
    TInterpolation,
    TInterpolation extends 'flat' ? 'first' : 'center'
  >
>;

export function interpolate<
  TInterpolation extends InterpolationType,
  TSampling extends InterpolationSamplingType<TInterpolation>,
  TData extends TInterpolation extends 'flat'
    ? FlatInterpolatableData
    : PerspectiveOrLinearInterpolatableData,
>(
  interpolationType: TInterpolation,
  samplingType: TSampling,
  data: TData,
): Decorate<Exotic<TData>, Interpolate<TInterpolation, TSampling>>;

export function interpolate<
  TInterpolation extends InterpolationType,
  TSampling extends InterpolationSamplingType<TInterpolation>,
  TData extends AnyData,
>(
  dataOrInterpolation: TData | TInterpolation,
  dataOrSampling?: TData | TSampling,
  data?: TData,
): Decorate<Exotic<TData>, Interpolate<TInterpolation, TSampling>> {
  if (typeof dataOrInterpolation === 'string') {
    if (typeof dataOrSampling === 'string') {
      return attribute(data as TData, {
        type: '@interpolate',
        value: `${dataOrInterpolation}, ${dataOrSampling}`,
        // biome-ignore lint/suspicious/noExplicitAny: <tired of lying to types>
      }) as any;
    }

    return attribute(dataOrSampling as TData, {
      type: '@interpolate',
      value: `${dataOrInterpolation}, center`,
      // biome-ignore lint/suspicious/noExplicitAny: <tired of lying to types>
    }) as any;
  }

  return attribute(dataOrInterpolation as TData, {
    type: '@interpolate',
    value: 'perspective, center',
    // biome-ignore lint/suspicious/noExplicitAny: <tired of lying to types>
  }) as any;
}
```
I must admit the simplicity of your solution absolutely grew on me 😂

reczkok:
Since the `@interpolate` attribute can only be defined on vertex outputs and fragment inputs I don't think we have to worry about enforcing the location attribute. Every member should implicitly get decorated anyway. Unless the order of attributes matters and I don't think it does?

deluksic:
Quickly glanced at the spec, I did not find an explicit mention of ordering. However this tells me that it does not matter:

```
Unless explicitly permitted in an attribute’s description, an attribute must not be specified more than once per object or type.
```

deluksic:
> I must admit the simplicity of your solution absolutely grew on me 😂 I'm sure the double template approach could be shortened but I'm not convinced I would even prefer it 🤔 What are your thoughts on this?

From a library user's point of view, I think having a single enumeration to fully specify an interpolation mode might be preferable as I would instead need to carry around two values and make sure they match when unpacking into args.

```typescript 
const interpMode = ["flat", "either"]
...
interpolate(...interpMode, u32)
// vs
const interpMode = 'flat, either'
...
interpolate(interpMode, u32)
```

iwoplaza:
Aahh, `ioStruct`, I haven't seen you in a long time buddy... 🚬👴

iwoplaza:
I'm personally leaning towards a single value as well, purely for how nice the auto-complete would be 💜
```ts
interpolate('perspective, center', u32)
```

packages/typegpu/src/data/attributes.ts
```diff
@@ -186,6 +193,68 @@ export function location<TLocation extends number, TData extends AnyData>(
...
+ * Specifies how user-defined vertex shader output (fragment shader input)
+ * must be interpolated.
+ *
+ * Tip: Integer outputs cannot be interpolated.
+ *
+ * @example
+ * const Data = d.ioStruct({
+ *   a: d.f32, // has implicit 'perspective, center' interpolation
+ *   b: d.interpolate('linear, sample', d.f32),
+ * });
```
iwoplaza: The draft guides with `ioStruct` use legacy experimental (or "legsperimental") APIs, `ioStruct` being one of them. We shifted our approach to allowing devs to pass in a record into `tgpu.vertexFn` directly.

```suggestion
 * const VertexOutput = {
 *   a: d.f32, // has implicit 'perspective, center' interpolation
 *   b: d.interpolate('linear, sample', d.f32),
 * };
```

packages/typegpu/src/data/attributes.ts
```diff
@@ -186,6 +193,68 @@ export function location<TLocation extends number, TData extends AnyData>(
...
+ * need deterministic behavior provided by `'flat, first'`, prefer explicit
+ * `'flat, either'` as it could be slightly faster in hardware.
+ *
+ * @example
+ * const Data = d.ioStruct({
+ *   a: d.f32, // has implicit 'perspective, center' interpolation
+ *   b: d.interpolate('flat, either', d.u32), // integer outputs cannot interpolate
+ * });
+ *
+ * @param location The explicit numeric location.
```
iwoplaza: Tiny nit, mismatched docs 📚
```suggestion
 * @param interpolationType How data should be interpolated.
```

packages/typegpu/src/data/attributes.ts
```diff
@@ -186,6 +193,68 @@ export function location<TLocation extends number, TData extends AnyData>(
...
+ *
+ * Tip: Integer outputs cannot be interpolated.
+ *
+ * @example
+ * const Data = d.ioStruct({
+ *   a: d.f32, // has implicit 'perspective, center' interpolation
+ *   b: d.interpolate('linear, sample', d.f32),
+ * });
+ *
+ * @param location The explicit numeric location.
```
iwoplaza: Tiny nit, mismatched docs 📚
```suggestion
 * @param interpolationType How data should be interpolated.
```

iwoplaza (changes requested):
Awesome! I really like the API! 🙌
The one thing that's missing is tests to verify that the generated WGSL is the one we expect. You can take a look at `location.test.ts` for inspiration, only the first unit test really.

deluksic:
Done, @iwoplaza please take a look again.

packages/typegpu/src/data/attributes.ts
```diff
@@ -92,6 +99,12 @@ export type IsBuiltin<T> = ExtractAttributes<T>[number] extends []
     ? true
     : false;
 
+export type HasCustomInterpolate<T> = ExtractAttributes<T>[number] extends []
+  ? false
+  : ExtractAttributes<T>[number] extends Interpolate<InterpolationType>
+    ? true
+    : false;
+
```
iwoplaza: I don't think we need this type utility as of this moment. `HasCustomLocation` was useful for determining whether to add a location attribute or not on the type level when used by pipelines. In this case, interpolation is always explicit, so we don't need this utility.

```suggestion
```

iwoplaza (changes requested):
Almost there!
There seems to be a missing import to `interpolate` in the test file.

packages/typegpu/src/data/attributes.ts
deluksic: will remove

packages/typegpu/src/data/attributes.ts
```diff
@@ -169,7 +176,7 @@ export function size<TSize extends number, TData extends AnyData>(
  * Assigns an explicit numeric location to a struct member or a parameter that has this type.
  *
  * @example
- * const Data = d.ioStruct({
+ * const Data = d.struct({
```
deluksic: is this actually correct? or should it be `VertexOutput = {...`?

iwoplaza: It's technically correct, as regular struct fields can also have location attributes, but yeah, a more up-to-date doc would be using `VertexOutput`. Would you mind updating it? 🙌

deluksic: done

iwoplaza (approved):
Awesome work!! 🚀 🌔

PR #771, opened 2025-01-13
Fix invalid bound type for samplers

### LINKED ISSUE #750: Invalid bound type for samplers.
AFAIK, samplers can be either "TgpuSampler", or "TgpuComparisonSampler" types, but when binding, the type is always just "TgpuSampler". We should look into this.

Image

mhawryluk:
closes #750

### DIFF
--- a/packages/typegpu/tests/bindGroupLayout.test.ts
+++ b/packages/typegpu/tests/bindGroupLayout.test.ts
@@ -319,7 +319,7 @@ describe('TgpuBindGroup', () => {
       });
     });
 
-    it('accepts filtering/non-filtering sampler when creating bind group, but not comparison ', ({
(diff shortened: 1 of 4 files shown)


packages/typegpu/tests/bindGroupLayout.test.ts
```diff
@@ -315,6 +318,84 @@ describe('TgpuBindGroup', () => {
         ],
       });
     });
+
+    it('accepts filtering/non-filtering sampler when creating bind group, but not comparison ', ({
```
iwoplaza: Tiny nit
```suggestion
    it('accepts filtering/non-filtering sampler when creating bind group, but not comparison', ({
```

PR #775, opened 2025-01-13
chore: Hide internal resolve API

iwoplaza:
## Why?
Since we have objects that are resolved by an external mechanism (resolveData for schemas) and objects that can resolve themselves (fns, vector and matrix instances, slots, ...), the `resolve` method seems more and more like an internal API.

## This PR
- Makes the `resolve` method hidden from the public API,
- Alters the type we accept as externals in `tgpu.resolve`
- (BONUS) Makes matrix instances resolvable

### DIFF
--- a/packages/typegpu/src/core/function/tgpuComputeFn.ts
+++ b/packages/typegpu/src/core/function/tgpuComputeFn.ts
@@ -81,7 +81,7 @@ function createComputeFn(
 
   const core = createFnCore(shell, implementation);
 
-  const result: This = {
+  return {
     shell,
 
     get label() {
@@ -108,7 +108,5 @@ function createComputeFn(
     toString() {
       return `computeFn:${this.label ?? '<unnamed>'}`;
     },
-  };
-
-  return result;
+  } as This;
 }
--- a/packages/typegpu/src/core/function/tgpuComputeFn.ts
+++ b/packages/typegpu/src/core/function/tgpuComputeFn.ts
@@ -1,6 +1,6 @@
 import type { AnyBuiltin } from '../../builtin';
 import type { TgpuNamable } from '../../namable';
-import type { ResolutionCtx, TgpuResolvable } from '../../types';
+import type { Labelled, ResolutionCtx, SelfResolvable } from '../../types';
 import { createFnCore } from './fnCore';
 import type { Implementation } from './fnTypes';
 
@@ -30,7 +30,7 @@ export interface TgpuComputeFnShell {
   does(implementation: string): TgpuComputeFn;
 }
 
-export interface TgpuComputeFn extends TgpuResolvable, TgpuNamable {
+export interface TgpuComputeFn extends TgpuNamable {
   readonly shell: TgpuComputeFnShell;
 
   $uses(dependencyMap: Record<string, unknown>): this;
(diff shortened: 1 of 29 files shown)


reczkok (approved):
Great work! The matrix resolution could use some tests 🕵️‍♂️

packages/typegpu/src/core/function/tgpuComputeFn.ts
```diff
@@ -77,11 +77,11 @@ function createComputeFn(
   workgroupSize: number[],
   implementation: Implementation<[], void>,
 ): TgpuComputeFn {
-  type This = TgpuComputeFn;
+  type This = TgpuComputeFn & Labelled & SelfResolvable;
 
   const core = createFnCore(shell, implementation);
 
-  return {
+  const result: This = {
```
mhawryluk: do we need this variable? wouldn't returning with `as This` work?

packages/typegpu/src/data/vector.ts
```diff
@@ -106,12 +106,12 @@ abstract class vec2Impl {
     this.y = value;
   }
 
-  resolve(): string {
+  '~resolve'(): string {
```
mhawryluk, line 110: we should mark vectors and matrices with the `SelfResolvable` interface

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -508,6 +471,17 @@ export interface v4u
   w: number;
 }
 
+export type AnyVecInstance =
```
mhawryluk, line 474: f16 vectors are missing

packages/typegpu/src/core/function/tgpuVertexFn.ts
```diff
@@ -105,7 +104,7 @@ function createVertexFn(
     );
   }
 
-  return {
+  const result: This = {
```
mhawryluk: return { (...) as This;

packages/typegpu/src/core/resolve/externals.ts
```diff
@@ -82,14 +81,9 @@ export function replaceExternalsInWgsl(
...
 ) {
   return Object.entries(externalMap).reduce((acc, [externalName, external]) => {
-    const resolvedExternal =
-      isResolvable(external) || isWgslData(external) || isSlot(external)
-        ? ctx.resolve(external as Wgsl)
-        : String(external);
-
     return acc.replaceAll(
       new RegExp(`(?<![\\w_.])${externalName}(?![\\w_])`, 'g'),
-      resolvedExternal,
```
mhawryluk: `() => { return acc...}` -> `() => acc...`

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -470,9 +464,16 @@ class ResolutionCtxImpl implements ResolutionCtx {
...
+      let result: string;
+      if (isWgslData(item)) {
+        result = resolveData(this, item);
+      } else if (isDerived(item) || isSlot(item)) {
+        result = this.resolve(item);
+      } else if (isSelfResolvable(item)) {
+        result = item['~resolve'](this);
+      } else {
+        result = this.resolveValue(item);
+      }
```
mhawryluk, line 476: i'd go with a `const` and multiple ternary operators, but it's just a personal preference

packages/typegpu/src/types.ts
```diff
@@ -84,26 +125,31 @@ export interface ResolutionCtx {
...
 }
 
-export function isResolvable(value: unknown): value is TgpuResolvable {
-  return (
-    !!value &&
-    (typeof value === 'object' || typeof value === 'function') &&
-    'resolve' in value
-  );
+export interface Labelled {
+  readonly label?: string | undefined;
```
mhawryluk, line 139: this interface is used exclusively for functions now, but a lot of things have a `label` property :/

mhawryluk (changes requested):

mhawryluk (approved):
Great!

PR #778, opened 2025-01-13
chore: Make `typegpu/std` public

iwoplaza:


reczkok (approved):
Why not

mhawryluk (changes requested):
let's update `CodeEditor.tsx` and `astro.config.mjs` too

mhawryluk (approved):
Amazing!

PR #779, opened 2025-01-14
refactor: Vector implementation without using Proxy

deluksic:
Instead of using `Proxy` to implement swizzling, we can use `new Function` to compile small getters with no runtime if / switch. This should allow js engines to optimize these calls better, but more testing is needed. The most important thing is that `vec[0]` and `vec.x` are simple index access and a getter respectively.

Example swizzle getter:
```
get xxyw() {
  return new this._Vec4(this[0], this[0], this[1], this[3]);
}
```

One important change is the ability to unpack vectors like:
```
const [x, y] = v2
```
I achieve this by extending `Array`. Typescript doesn't understand yet that this is possible, but I think we should be able to tell it.

My first tests show almost 2x speedup in `buffer.write` speed, 25ms vs 40ms for 50k `Circle(position, radius, styleIndex)` structs.

### DIFF
--- a/packages/typegpu/src/data/vector.ts
+++ b/packages/typegpu/src/data/vector.ts
@@ -1,5 +1,19 @@
 import { inGPUMode } from '../gpuMode';
-import type { SelfResolvable } from '../types';
+import {
+  Vec2fImpl,
+  Vec2hImpl,
+  Vec2iImpl,
+  Vec2uImpl,
+  Vec3fImpl,
+  Vec3hImpl,
+  Vec3iImpl,
+  Vec3uImpl,
+  Vec4fImpl,
+  Vec4hImpl,
+  Vec4iImpl,
+  Vec4uImpl,
+  type VecBase,
+} from './vectorImpl';
 import type {
   Vec2f,
   Vec2h,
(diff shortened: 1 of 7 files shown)


packages/typegpu/src/data/vector.ts
```diff
@@ -58,11 +73,7 @@ function makeVecSchema<TType extends string, TValue>(
       return `${VecSchema.type}(${values.join(', ')})` as unknown as TValue;
     }
 
-    if (values.length <= 1) {
-      return options.makeFromScalar(values[0] ?? 0);
-    }
-
-    if (values.length === options.length) {
+    if (values.length <= 1 || values.length === options.length) {
       return options.make(...values);
```
deluksic: Arguably, we could pass the values directly to the constructor of the vector and not check lengths. In case someone generates thousands of vectors, these checks can add up. We could simply rely on typescript to handle this for us and not pay the runtime cost (even if small).

I think having the option to initialize using `vec4(vec3(), w)` would be super cool, but I did not find a way to do this without paying the cost of checking args. If someone wants to do this in js, the simplest way would be to `vec4(...vec3(), w)`. This would work just fine, we just need to adjust the transpiler to ignore the dots.

iwoplaza:
Awesome initiative! 👏
We have to see if creating functions on-demand works well in Hermes (React Native). If not, we can always do the generation as a build step, and ship the generated code in the package.
I'll take a closer look in the coming days 👀

deluksic:
If it can be autogenerated on build for everyone, that would be even better! Although downsides are code size, and not being able to use the library from node without compile step. (Node has experimental type stripping support)

deluksic:
Oooh I see

packages/typegpu/tsup.config.ts
```diff
@@ -11,7 +11,7 @@ export default defineConfig({
   outDir: 'dist',
   format: ['cjs', 'esm'],
   tsconfig: './tsconfig.json',
-  target: 'es2017',
+  target: 'es2022',
```
deluksic: Switching to `es2022` seems to improve performance of vector creation. I'm not sure why yet, but given WebGPU did not exist in es2017, we might as well switch it.

iwoplaza:
Okay, I verified that the \`new Function(`...`)\` works fine on React Native 🎉. However, doing this at runtime still makes it impossible for things like Static Hermes or Porffor (JS compilers) to optimize them well AOT.

I would still attempt to offload this work AOT as a prepublish step, and ship the generated code. For now though, we can continue experimenting and pushing the performance with a JIT approach here, as it's easier to prototype with.

iwoplaza:
Alright, I think if we're able to support both the slow path (proxy) and the fast path (JIT functions), then we're going to have the best of both worlds. We can go back to generating these functions AOT if we find that that's the better approach.

@deluksic Would you be willing to update this PR with a check to see if `new Function` is supported? And if it's not, use the old Proxy approach? We can check it like so (credit to [Brian Breiholz](https://github.com/Ctrlmonster)):

```ts
let EVAL_ALLOWED_IN_ENV: boolean;
try {
  new Function(`return true`);
  EVAL_ALLOWED_IN_ENV = true;
} catch {
  EVAL_ALLOWED_IN_ENV = false;
}

// ...
```

deluksic:
I'm looking at this and it is not super easy to switch this at runtime. Are you sure we absolutely must support the proxy way? Do you know when `new Function` would not be supported where proxy and WebGPU are?

deluksic:
OK, I found this https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#unsafe_eval_expressions

I will try to find a way to switch at runtime, hopefully by doing this check only once. It kinda defeats the purpose if every vec creation or access needs to check for `new Function` support.

deluksic:
I lied, I implemented the AOT version instead. It was actually easier. For now I created a `scripts` folder with the script that generates the getters which can be manually copied over to the abstract vector definitions. In the future we can make it automatic, but honestly, this is one of those things you do once.

packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -0,0 +1,830 @@
+import type { SelfResolvable } from '../types';
+import type { VecKind } from './wgslTypes';
+
+export abstract class VecBase extends Array implements SelfResolvable {
+  abstract get kind(): VecKind;
```
deluksic, line 6: switched `kind` to a getter, so we don't add a string to each instance of the vector.

packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -0,0 +1,830 @@
...
+  get wwyz() { return new this._Vec4(this[3], this[3], this[1], this[2]); }
+  get wwyw() { return new this._Vec4(this[3], this[3], this[1], this[3]); }
+  get wwzx() { return new this._Vec4(this[3], this[3], this[2], this[0]); }
+  get wwzy() { return new this._Vec4(this[3], this[3], this[2], this[1]); }
+  get wwzz() { return new this._Vec4(this[3], this[3], this[2], this[2]); }
+  get wwzw() { return new this._Vec4(this[3], this[3], this[2], this[3]); }
+  get wwwx() { return new this._Vec4(this[3], this[3], this[3], this[0]); }
+  get wwwy() { return new this._Vec4(this[3], this[3], this[3], this[1]); }
+  get wwwz() { return new this._Vec4(this[3], this[3], this[3], this[2]); }
+  get wwww() { return new this._Vec4(this[3], this[3], this[3], this[3]); }
```
deluksic, line 367: I expect this to compress pretty well, but we should check.

packages/typegpu/tsup.config.ts
deluksic: Will remove from this PR as it is kinda unrelated

deluksic:
I have an idea to reduce the amount of code.

What if we add the 4-way swizzle directly on the `VecBase`? Sure, some of those getters would not be applicable for `Vec2` and `Vec3`, but also users will never interact with these directly, so typescript would warn anyways. This way, only one manual copy is needed and we remove ~150 LOC.

deluksic:
@iwoplaza This PR is ready, if you want to have a look.

reczkok:
When doing some manual profiling it looks very good but the benchmarks seem to disagree 🤔 I'm pretty sure they are lying but I can't figure out why.

Here are my results from manual profiling 25.2% -> 8.4% which is great! 🚀
Before:
Screenshot 2025-02-12 at 15 34 49
After:
Screenshot 2025-02-12 at 15 34 27

deluksic:
> @deluksic what is the performance like for you in this implementation? Did you use our benchmark or did you profile it on your own?

It cuts down the buffer write time dramatically:

image

image

reczkok:
Looks great! We could add `[Symbol.iterator](): IterableIterator<number>` to make the unpacking work in types (We used to have it on `NumberArrayView` but decided to drop it so it matches `MutableNumberArray` [wgpu-matrix](https://github.com/greggman/wgpu-matrix) for interop). Maybe we could have it on the Swizzle interfaces?

> If my calculation is correct, we are paying 3KB in gzipped minified code for this. Not nothing...

12% increase is not insignificant but I think it's worth it

deluksic:
> Looks great! We could add `[Symbol.iterator](): IterableIterator<number>` to make the unpacking work in types (We used to have it on `NumberArrayView` but decided to drop it so it matches `MutableNumberArray` [wgpu-matrix](https://github.com/greggman/wgpu-matrix) for interop). Maybe we could have it on the Swizzle interfaces?

Could this be a separate PR as it is kind of orthogonal to the removal of proxies?

packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -0,0 +1,683 @@
...
+  set x(value: number) {
+    this[0] = value;
+  }
+
+  set y(value: number) {
+    this[1] = value;
+  }
+}
+
+// biome-ignore format: swizzles should not expand
```
iwoplaza: Tiny nit, there are no swizzles in this class, what formatting is this comment ignoring?
(Also, the `Vec2` class does not have this comment)

iwoplaza (approved):
Awesome improvement, thanks for your contribution! 🙌 💜
(tiny nits about biome comments)

packages/typegpu/src/data/vectorImpl.ts
deluksic: done

PR #795, opened 2025-01-16
Return proxy from `.value` properties

### LINKED ISSUE #361: Return proxy from `.value` properties.
The .value property of some WgslResolvables has to be a proxy that handles arbitrary member access, and returns another proxied resolvable that resolves the whatever the parent resolves to + .${key} of the member access operation.

mhawryluk:
closes #361

### DIFF
--- a/packages/typegpu/src/core/buffer/bufferUsage.ts
+++ b/packages/typegpu/src/core/buffer/bufferUsage.ts
@@ -2,6 +2,7 @@ import type { AnyWgslData, BaseWgslData } from '../../data/wgslTypes';
 import { type Storage, isUsableAsStorage } from '../../extension';
 import { inGPUMode } from '../../gpuMode';
 import type { Infer } from '../../shared/repr';
+import { valueProxyHandler } from '../../shared/valueProxyHandler';
 import type { LayoutMembership } from '../../tgpuBindGroupLayout';
 import type {
   BindableBufferUsage,
@@ -103,10 +104,10 @@ class TgpuFixedBufferImpl<
     if (!inGPUMode()) {
       throw new Error(`Cannot access buffer's value directly in JS.`);
     }
-    return this as Infer<TData>;
+
+    return new Proxy(this, valueProxyHandler) as Infer<TData>;
   }
 }
-
 export class TgpuLaidOutBufferImpl<
   TData extends BaseWgslData,
   TUsage extends BindableBufferUsage,
@@ -146,7 +147,7 @@ export class TgpuLaidOutBufferImpl<
     if (!inGPUMode()) {
       throw new Error(`Cannot access buffer's value directly in JS.`);
     }
-    return this as Infer<TData>;
+    return new Proxy(this, valueProxyHandler) as Infer<TData>;
   }
 }
 
--- a/packages/typegpu/src/core/constant/tgpuConstant.ts
+++ b/packages/typegpu/src/core/constant/tgpuConstant.ts
@@ -2,6 +2,7 @@ import type { AnyWgslData } from '../../data/wgslTypes';
 import { inGPUMode } from '../../gpuMode';
 import type { TgpuNamable } from '../../namable';
 import type { Infer } from '../../shared/repr';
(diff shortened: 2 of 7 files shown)


packages/typegpu/src/core/buffer/bufferUsage.ts
```diff
@@ -103,10 +95,16 @@ class TgpuFixedBufferImpl<
     if (!inGPUMode()) {
       throw new Error(`Cannot access buffer's value directly in JS.`);
     }
-    return this as Infer<TData>;
+
+    return new Proxy(
+      {
+        '~resolve': ((ctx: ResolutionCtx) => ctx.resolve(this)).bind(this),
+        toString: (() => `.value:${this.label ?? '<unnamed>'}`).bind(this),
```
iwoplaza: Since we're using arrow functions, we don't need to use `.bind(this)`. It actually does not affect the arrow function.
```suggestion
        '~resolve': (ctx: ResolutionCtx) => ctx.resolve(this),
        toString: () => `.value:${this.label ?? '<unnamed>'}`,
```

iwoplaza: Since we're using arrow functions, we don't need to use `.bind(this)`. It actually does not affect the arrow function.
```suggestion
        '~resolve': (ctx: ResolutionCtx) => ctx.resolve(this),
        toString: () => `.value:${this.label ?? '<unnamed>'}`,
```

packages/typegpu/src/core/buffer/bufferUsage.ts
```diff
@@ -146,7 +144,14 @@ export class TgpuLaidOutBufferImpl<
...
```
iwoplaza: Same here
```suggestion
        '~resolve': (ctx: ResolutionCtx) => ctx.resolve(this),
        toString: () => `.value:${this.label ?? '<unnamed>'}`,
```

packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -65,6 +66,13 @@ class TgpuConstImpl<TDataType extends AnyWgslData>
     if (!inGPUMode()) {
       return this._value;
     }
-    return this as Infer<TDataType>;
+
+    return new Proxy(
+      {
+        '~resolve': ((ctx: ResolutionCtx) => ctx.resolve(this)).bind(this),
+        toString: (() => `.value:${this.label ?? '<unnamed>'}`).bind(this),
```
iwoplaza: 🫡
```suggestion
        '~resolve': (ctx: ResolutionCtx) => ctx.resolve(this),
        toString: () => `.value:${this.label ?? '<unnamed>'}`,
```

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -62,7 +68,13 @@ export class TgpuAccessorImpl<T extends AnyWgslData>
       );
     }
 
-    return this as unknown as Infer<T>;
+    return new Proxy(
+      {
+        '~resolve': ((ctx: ResolutionCtx) => ctx.resolve(this)).bind(this),
+        toString: (() => `.value:${this.label ?? '<unnamed>'}`).bind(this),
```
iwoplaza:
```suggestion
        '~resolve': (ctx: ResolutionCtx) => ctx.resolve(this),
        toString: () => `.value:${this.label ?? '<unnamed>'}`,
```

iwoplaza (changes requested):

packages/typegpu/tests/constant.test.ts
```diff
@@ -24,4 +24,47 @@ describe('tgpu.const', () => {
...
+    });
+
+    expect(parse(resolved)).toEqual(
+      parse(`
+        struct Boid {
+          pos: vec3f,
+          vel: vec3u,
+        }
+
+        const boid = Boid(vec3f(1, 2, 3), vec3u(4, 5, 6));
```
reczkok, line 61: Didn't know that was legal in wgsl const. It's perfectly fine as long as the struct has a fixed footprint. Fun fact 🕵️‍♂️

PR #798, opened 2025-01-17
TGSL structs

reczkok:


packages/typegpu/src/data/struct.ts
```diff
@@ -18,7 +18,7 @@ import type { AnyWgslData, BaseWgslData, WgslStruct } from './wgslTypes';
 export interface TgpuStruct<TProps extends Record<string, BaseWgslData>>
   extends WgslStruct<TProps>,
     TgpuNamable {
-  readonly '~exotic': WgslStruct<ExoticRecord<TProps>>;
+  readonly '~exotic': WgslStruct<TProps>;
```
mhawryluk, line 20: why this change? and if it is indeed supposed to be like this, then I think the assigned value in `struct`  function below (`struct['~exotic'] = {} as WgslStruct<ExoticRecord<TProps>>;`) should match exactly

mhawryluk (approved):
Awesome! 😎

packages/typegpu/src/data/struct.ts
reczkok: **DISCLAIMER** - I could be totally wrong about this

I think this was an error in the original implementation. When we create a struct:
```ts
export const struct = <TProps extends Record<string, AnyWgslData>>(
  props: TProps,
): TgpuStruct<Prettify<ExoticRecord<TProps>>>
```
we pass props as `ExoticRecord<TProps>` to `TgpuStruct`. This means that we shouldn't do that an additional time in the `TgpuStruct` itself. For `struct['~exotic']` on the other hand we get props before it is wrapped in `ExoticRecord` hence the difference.

packages/typegpu/src/data/struct.ts
```diff
@@ -34,36 +34,36 @@ export interface TgpuStruct<TProps extends Record<string, BaseWgslData>>
  */
 export const struct = <TProps extends Record<string, AnyWgslData>>(
   props: TProps,
-): TgpuStruct<Prettify<ExoticRecord<TProps>>> =>
-  new TgpuStructImpl(props as ExoticRecord<TProps>);
+): TgpuStruct<Prettify<ExoticRecord<TProps>>> => {
+  const struct = <T>(props: T) => {
+    return props;
+  };
```
mhawryluk:
```suggestion
  const struct = <T>(props: T) => props;
```

packages/typegpu/src/data/struct.ts
mhawryluk: okie, makes sense

packages/typegpu/src/data/struct.ts
```diff
@@ -34,36 +34,34 @@ export interface TgpuStruct<TProps extends Record<string, BaseWgslData>>
...
 export const struct = <TProps extends Record<string, AnyWgslData>>(
   props: TProps,
-): TgpuStruct<Prettify<ExoticRecord<TProps>>> =>
-  new TgpuStructImpl(props as ExoticRecord<TProps>);
+): TgpuStruct<Prettify<ExoticRecord<TProps>>> => {
+  const struct = <T>(props: T) => props;
+  Object.setPrototypeOf(struct, TgpuStructImpl);
+  struct.propTypes = props as ExoticRecord<TProps>;
+  struct['~repr'] = {} as InferRecord<TProps>;
+  struct['~exotic'] = {} as WgslStruct<ExoticRecord<TProps>>;
```
iwoplaza: AFAIK, these props don't even actually have to exist at runtime.
```suggestion
```

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -198,6 +212,42 @@ function generateExpression(
...
+          throw new Error(
+            `Missing property ${key} in object literal for struct ${callee}`,
+          );
+        }
+        return val;
+      });
+
+      return {
+        value: generateEntries(values),
+        dataType: UnknownData,
```
iwoplaza: This is actually a rare case where we know the data type 😂 
```suggestion
        dataType: callee,
```

iwoplaza (approved):
🫡🫡🫡

packages/typegpu/src/data/struct.ts
reczkok: I thought they make TS happy - turns out they are useless

PR #803, opened 2025-01-22
Implement partial IO

### LINKED ISSUE #800: Allow for partial I/O operations

reczkok:
closes #800 

- [x] Investigate if having one ArrayBuffer for all data is practical and faster
- [ ] Figure a way to cache calls to avoid creating instructions for the same accessed data (to speed up subsequent calls)
- [x] Support unstructs and disarrays


reczkok:
I'm currently testing two approaches to partial IO.
The one on this draft:
Screenshot 2025-01-28 at 10 30 44

And one with one big array buffer. Instead of instructions containing the buffer itself they contain views for it:
Screenshot 2025-01-28 at 10 29 35

reczkok:
I was also wondering if combining instructions in the current draft is worth it. Here is a version without it:
image
It doubles the speed of contagious writes with almost no overhead in this particular benchmark.

reczkok:
Closed in favor of #808

PR #804, opened 2025-01-23
Create typegpu babel plugin

### LINKED ISSUE #796: Create build plugin for React Native
Same as our rollup-plugin, but for metro

mhawryluk:
closes #796


packages/plugin/package.json
iwoplaza, line 1: Since we're planning to rewrite the plugin in `unplugin` (at least the parts that will run on rollup and rollup-adjacent bundlers), we should IMO go with the name `unplugin-typegpu`.

https://github.com/unplugin/unplugin-starter/tree/main

It will be less ambiguous what the plugin is for, and will be easier to find 🙌

iwoplaza: Also, could we call the directory `unplugin-typegpu` instead of matching the current convention and calling it `unplugin`? I feel like for packages living under the `@typegpu` namespace it would be fine, but this will make it more clear which package it represents.

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -14,6 +15,8 @@ type Context = {
   stack: Scope[];
 };
 
+type JsNode = babel.Node | acorn.AnyNode;
```
reczkok, line 18: I'm wondering if it wouldn't be easier to switch to `@types/estree` and use the babel plugin to convert the AST? What do you think?

reczkok: Maybe then we could have a parser agnostic implementation

packages/rollup-plugin/README.md
```diff
@@ -2,7 +2,7 @@
 
 # rollup-plugin-typegpu
 
-🚧 **Under Construction** 🚧 - [GitHub](https://github.com/software-mansion/TypeGPU/tree/main/packages/rollup-plugin)
+⚠️ **This package is deprecated, please use `unplugin-typegpu/rollup` instead.** ⚠️
```
iwoplaza: Hmm, maybe lets redirect to just the package name, should be easy to then find how to use it in rollup.

```suggestion
⚠️ **This package is deprecated, please use `unplugin-typegpu` instead.** ⚠️
```

packages/unplugin-typegpu/README.md
```diff
@@ -0,0 +1,50 @@
+<div align="center">
+
+# unplugin-typegpu
+
+🚧 **Under Construction** 🚧 - [GitHub](https://github.com/software-mansion/TypeGPU/tree/main/packages/unplugin-typegpu)
+
+</div>
+
+Build plugins for [TypeGPU](https://typegpu.com), transpiling JavaScript functions into WGSL at build-time.
```
iwoplaza: Could be misinformation 😂. Maybe we can rephrase it while not going into more detail than necessary.

```suggestion
Build plugins for [TypeGPU](https://typegpu.com) that enable:
- Seamless JavaScript -> WGSL transpilation
- [🚧 TODO] Improved debugging with automatic naming of resources
```

packages/tinyest-for-wgsl/src/parsers.ts
mhawryluk: sure, I can try and see how viable this is

packages/unplugin-typegpu/package.json
```diff
@@ -0,0 +1,77 @@
+{
+  "name": "unplugin-typegpu",
+  "version": "0.0.0",
+  "description": "Build plugins for TypeGPU, transpiling JavaScript functions into WGSL at build-time.",
```
iwoplaza:
```suggestion
  "description": "Build plugins for TypeGPU, enabling seamless JavaScript -> WGSL transpilation and improved debugging.",
```

iwoplaza (changes requested):

PR #805, opened 2025-01-25
🐎: Scaffolding for a benchmarking tool

iwoplaza:
image

## Motivation
As we're focusing more and more on performance, I thought a tool for comparing different versions of TypeGPU in the same benchmarks would be a nice tool to have.

Leveraging [esm.sh](https://esm.sh), we can download and run any version of TypeGPU that exists on npm. They also promise that installing packages from PRs should work fine as well, which we could use to:
- Allow a "PR" type of version, synced to the URL like the rest of the state.
- Create a GitHub bot that comments on new PRs with a link to the benchmark, with the id of that PR.

This would allow us to see the performance impact of a PR's changes without having to checkout that code locally 🚀

## This PR
- Creates a new route in the docs site: `docs.swmansion.com/TypeGPU/benchmark`
- Adds two benchmarks that measure the speed of writing to buffers (one with our APIs, the other by manually writing values to the buffer).

### DIFF
--- a/apps/typegpu-docs/src/pages/benchmark/atom-with-url.ts
+++ b/apps/typegpu-docs/src/pages/benchmark/atom-with-url.ts
@@ -1,0 +1,68 @@
+import { atom } from 'jotai';
+import { atomWithLocation } from 'jotai-location';
+
+const locationAtom = atomWithLocation();
+
+export const stringParam = {
+  encode: (val: string) => val,
+  decode: (val: string) => val,
+};
+
+export const boolParam = {
+  encode: (val: boolean) => (val ? '1' : '0'),
+  decode: (val: string) => val === '1',
+};
+
+export const numberParam = {
+  encode: (val: number) => String(val),
+  decode: (val: string) => Number.parseFloat(val),
+};
+
+export const objParam = {
+  encode: JSON.stringify,
+  decode: JSON.parse,
+};
+
+export const typeToParam = {
+  string: stringParam,
+  boolean: boolParam,
+  number: numberParam,
+  object: objParam,
+};
+
+export const atomWithUrl = <T>(
+  key: string,
+  defaultValue: T,
+  options?: { encode: (val: T) => string; decode: (val: string) => unknown },
+) => {
+  const optionsOrInferred =
+    options ??
+    typeToParam[typeof defaultValue as keyof typeof typeToParam] ??
+    objParam;
+
+  const { encode, decode } = optionsOrInferred;
+
+  return atom(
+    (get) => {
+      const location = get(locationAtom);
+      return location.searchParams?.has(key)
+        ? (decode(location.searchParams.get(key) ?? '') as T)
+        : defaultValue;
+    },
+    (get, set, newValue: T) => {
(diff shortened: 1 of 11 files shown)


apps/typegpu-docs/src/pages/benchmark/atom-with-url.ts
```diff
@@ -0,0 +1,68 @@
...
+        ? (decode(location.searchParams.get(key) ?? '') as T)
+        : defaultValue;
+    },
+    (get, set, newValue: T) => {
+      const prev = get(locationAtom);
+      const searchParams = new URLSearchParams(prev.searchParams);
+      searchParams.set(key, encode(newValue));
+
+      set(
+        // biome-ignore lint/suspicious/noExplicitAny: <really annoying>
```
reczkok: I think it's fine?
```suggestion
```

apps/typegpu-docs/src/pages/benchmark/benchmark-app.tsx
```diff
@@ -0,0 +1,223 @@
...
+}
+
+async function runBench(params: BenchParameterSet): Promise<BenchResults> {
+  const { tgpu } = await importTypeGPU(params.typegpu);
+  const d = await importTypeGPUData(params.typegpu);
+
+  const bench = new Bench({
+    name: stringifyLocator('typegpu', params.typegpu),
+    time: 1000,
+  });
```
reczkok, line 28: We should be mindful of the default 16 warmup iterations in `tinybench`. Since as far as I checked the warmup consist of the actual tasks, during the benchmarks all our values that are cached globally will probably already be cached. This is not bad but something to keep in mind.

apps/typegpu-docs/src/pages/benchmark/parameter-set.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  return name;
+}
+
+function getFreeKey(get: Getter): number {
+  return Math.max(...get(parameterSetsAtom).map((params) => params.key)) + 1;
+}
+
+export const parameterSetsAtom = atomWithUrl<BenchParameterSet[]>('p', [
```
reczkok, line 38: I like the idea but maybe we could have some small debounce on this? Having browse history cluttered by every change is a bit bothersome.

reczkok (approved):
Great work! 🧱👷

apps/typegpu-docs/src/pages/benchmark/benchmark-app.tsx
iwoplaza: That's actually nice! Good thing to keep in mind for tasks where we measure init-time, but in this case, we are measuring performance of things that are going to be in the hot-loop anyway, so good to warm it up before-hand.

apps/typegpu-docs/src/pages/benchmark/parameter-set.ts
iwoplaza: Good catch! I changed it so that it does not create history entries at all 🙌

PR #807, opened 2025-01-26
🥗: Utility modules starter

iwoplaza:


apps/typegpu-docs/src/utils/examples/sandboxModules.ts
```diff
@@ -0,0 +1,84 @@
+import { entries, fromEntries, map, pipe } from 'remeda';
+
+import dtsWebGPU from '@webgpu/types/dist/index.d.ts?raw';
+import dtsWgpuMatrix from 'wgpu-matrix/dist/3.x/wgpu-matrix.d.ts?raw';
+
+interface SandboxModuleDefinition<TModuleType> {
```
mhawryluk: doesn't need to be a generic type

apps/typegpu-docs/src/content/examples/image-processing/chroma-keying-next/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Chroma Keying (next)",
+  "category": "image-processing",
+  "tags": ["camera", "experimental"]
```
mhawryluk, line 4: why a new example, instead of just updating the existing chroma keying? also, I don't see anything experimental. do we plan on further updating this example? (or do we consider the util modules experimental?)

apps/typegpu-docs/src/content/examples/simulation/fluid-double-buffering/index.ts
```diff
@@ -431,7 +414,7 @@ const mainCompute = tgpu['~unstable']
     const y = d.i32(input.gid.y);
     const index = coordsToIndex(x, y);
 
-    setupRandomSeed(d.vec2f(d.f32(index), timeUniform.value));
+    BPETER.seed(d.vec2f(d.f32(index), timeUniform.value));
```
mhawryluk: since we don't assign `BPETER.sample` to the `randomGeneratorSlot` in this example, we shouldn't use `BPETER.seed`, but rather `DefaultGenerator.seed`. I would also consider changing the slot to contain both sample and seed functions, so that user functions can use the correct seed function, taking it from `randomGeneratorSlot`

mhawryluk (changes requested):

apps/typegpu-docs/src/content/examples/image-processing/chroma-keying-next/meta.json
iwoplaza: I was considering the utility module as experimental 👀

reczkok (approved):
👨‍🦰➡️🧺❓  
👀🔎🤔  
🤯😲🎉  
🙌🥚🐣✨  
🏃‍♂️💨📸🎥

mhawryluk (approved):
📣 Awesome!

PR #808, opened 2025-01-28
Implement more performant partial IO

reczkok:
Performance vs #803:
| #803    | this |
| -------- | ------- |
| old  | new    |


packages/typegpu/src/data/partialIO.ts
```diff
@@ -0,0 +1,140 @@
...
+    if (partialValue === undefined || partialValue === null) {
+      return;
+    }
+
+    if (isWgslStruct(node) || isUnstruct(node)) {
+      const propOffsets = offsetsForProps(node);
+
+      const sortedProps = Object.entries(propOffsets).sort(
+        (a, b) => a[1].offset - b[1].offset, // Sort by offset
+      );
```
reczkok: The `propOffsets` should always be in the correct order already. But I don't think it's guaranteed. It doesn't cost much to sort it here.

packages/typegpu/src/data/partialIO.ts
```diff
@@ -0,0 +1,138 @@
...
+      const elementSize = roundUp(
+        sizeOf(arrSchema.elementType),
+        alignmentOf(arrSchema.elementType),
+      );
+
+      if (!Array.isArray(partialValue)) {
+        throw new Error('Partial value for array must be an array');
+      }
+
+      partialValue.sort((a, b) => a.idx - b.idx);
```
reczkok: This is necessary since we don't want to sort all the write instructions later. If we don't sort here the instructions will still be valid but will not get properly merged. Sorting here is way less of an overhead than sorting all the instructions.

packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -228,6 +230,34 @@ class TgpuBufferImpl<TData extends AnyData> implements TgpuBuffer<TData> {
...
+
+    if (gpuBuffer.mapState === 'mapped') {
+      const mappedRange = gpuBuffer.getMappedRange();
+      const mappedView = new Uint8Array(mappedRange);
+
+      for (const instruction of instructions) {
+        mappedView.set(instruction.data, instruction.data.byteOffset);
+      }
+
+      return;
```
mhawryluk: I'd prefer an `else` clause instead of multiple return points in this case

packages/typegpu/src/data/offests.ts
```diff
@@ -0,0 +1,73 @@
+import { Measurer } from 'typed-binary';
```
mhawryluk: rename file to `offsets.ts`

packages/typegpu/src/data/offests.ts
```diff
@@ -0,0 +1,73 @@
...
+  );
+  if (cached) {
+    return cached as Record<keyof T, OffsetInfo>;
+  }
+
+  const measurer = new Measurer();
+  const offsets: Record<keyof T, OffsetInfo> = {} as Record<
+    keyof T,
+    OffsetInfo
+  >;
```
mhawryluk:
```suggestion
  const offsets = {} as Record<keyof T, OffsetInfo>;
```

packages/typegpu/src/data/partialIO.ts
```diff
@@ -0,0 +1,138 @@
...
+    if (isWgslStruct(node) || isUnstruct(node)) {
+      const propOffsets = offsetsForProps(node);
+
+      const sortedProps = Object.entries(propOffsets).sort(
+        (a, b) => a[1].offset - b[1].offset, // Sort by offset
+      );
+
+      for (const [key, propOffset] of sortedProps) {
+        const subSchema = node.propTypes[key];
+        if (!subSchema) continue;
```
mhawryluk:
```suggestion
        if (!subSchema) {
          continue;
        }
```

packages/typegpu/src/data/partialIO.ts
```diff
@@ -0,0 +1,138 @@
...
+        if (childValue !== undefined) {
+          gatherAndWrite(
+            subSchema,
+            childValue,
+            offset + propOffset.offset,
+            propOffset.padding ?? padding,
+          );
+        }
+      }
+      return;
```
mhawryluk: also here, I think `if, else if, else` would be more readable than early returns

packages/typegpu/src/data/partialIO.ts
```diff
@@ -0,0 +1,138 @@
...
+      );
+
+      if (!Array.isArray(partialValue)) {
+        throw new Error('Partial value for array must be an array');
+      }
+
+      partialValue.sort((a, b) => a.idx - b.idx);
+
+      for (const i of partialValue) {
+        const { idx, value } = i;
```
mhawryluk:
```suggestion
      for (const { idx, value } of partialValue) {
```

also, there's something wrong with types. idx and value are any :/

packages/typegpu/tests/partialIo.test.ts
```diff
@@ -0,0 +1,305 @@
...
+  expect(instruction.data.byteLength).toBe(length);
+
+  const dataArrays = Array.isArray(expectedData)
+    ? expectedData
+    : [expectedData];
+
+  let totalByteLength = 0;
+  for (const arr of dataArrays) {
+    totalByteLength += arr.byteLength;
+  }
```
mhawryluk: you can write this in one line using `.reduce`

packages/typegpu/src/data/partialIO.ts
```diff
@@ -0,0 +1,138 @@
...
+import { alignmentOf } from './alignmentOf';
+import { writeData } from './dataIO';
+import { isDisarray, isUnstruct } from './dataTypes';
+import { offsetsForProps } from './offests';
+import { sizeOf } from './sizeOf';
+import type * as wgsl from './wgslTypes';
+import { isWgslArray, isWgslStruct } from './wgslTypes';
+
+export interface WriteInstruction {
+  data: Uint8Array;
```
mhawryluk, line 13: why Uint8Array?

packages/typegpu/src/data/partialIO.ts
reczkok: Yeah i left that in cause it didn't bother me much. TypeScript just asserted that it is **an** array so the return type of an iterator is any. I could just cast it or assert with something like this:
```ts
function isPartialValueArray(
  value: InferPartial<wgsl.BaseWgslData>,
): value is InferPartial<wgsl.WgslArray<wgsl.BaseWgslData>> {
  return (
    Array.isArray(value) && value.every((v) => typeof v === 'object')
   );
}
```
The type of `value[x].value` will still be unknown but I don't think there is a solution here.

packages/typegpu/tests/partialIo.test.ts
reczkok: Yeah, it used to do more

packages/typegpu/src/data/partialIO.ts
reczkok: Most convenient way to hold byte info. It also allows to define a host buffer so less allocations than if we just held array buffers.

packages/typegpu/src/data/partialIO.ts
reczkok: I just cast it 🤷‍♂️

mhawryluk (approved):

packages/typegpu/src/data/offsets.ts
```diff
@@ -0,0 +1,71 @@
...
+import type { BaseData } from './wgslTypes';
+
+export interface OffsetInfo {
+  offset: number;
+  size: number;
+  padding?: number | undefined;
+}
+
+const cachedOffsets = new WeakMap<
+  WgslStruct | Unstruct,
```
iwoplaza: Try with `AnyWgslStruct`, maybe it will eliminate the need to cast on line 25. You can create a synonymous type for Unstruct: `AnyUnstruct`
```suggestion
  AnyWgslStruct | AnyUnstruct,
```

packages/typegpu/src/data/partialIO.ts
iwoplaza: [It is guaranteed!](https://tc39.es/ecma262/multipage/ordinary-and-exotic-objects-behaviours.html#sec-ordinaryownpropertykeys)

packages/typegpu/src/shared/utilityTypes.ts
```diff
@@ -31,6 +31,30 @@ export type Mutable<T> = {
...
+ */
+export type NTuple<T, N extends number> = N extends N
+  ? number extends N
+    ? T[]
+    : _NTupleOf<T, N, []>
+  : never;
+
+type _NTupleOf<T, N extends number, R extends unknown[]> = R['length'] extends N
+  ? R
+  : _NTupleOf<T, N, [T, ...R]>;
```
iwoplaza: Delit dis 🔫

reczkok: 🏳️

iwoplaza (approved):
LET"S GOOOO

PR #811, opened 2025-01-30
feat(🗑️): Try to stabilize derived.with behavior

iwoplaza:


aleksanderkatan:
Is this draft going to be continued or can we close it?

PR #813, opened 2025-02-03
Implement write instruction compilation

reczkok:


reczkok:
Looks promising 🧑‍🍳 Need to sort out the actual benchmarks as they don't seem to be accurate for the new method (this results are from a temporary test example)
Screenshot 2025-02-04 at 15 25 33

packages/typegpu/src/data/compiledIO.ts
```diff
@@ -0,0 +1,199 @@
...
+
+  if (rootIsArray) {
+    const index = path.shift();
+    return `value${index}${path.map((p) => `.${p}`).join('')}`;
+  }
+
+  return path.length === 0 ? 'value' : `value.${path.join('.')}`;
+}
+
+const compiledWriters = new Map<
```
iwoplaza:
```suggestion
const compiledWriters = new WeakMap<
```

reczkok:
Turns out a lot of our overhead is the creation of buffers per write that triggers the garbage collector
Screenshot 2025-02-06 at 12 48 51
After changing from buffer per write to buffer per TgpuBuffer we are almost at vanilla performacne
Screenshot 2025-02-06 at 12 53 21

deluksic:
Permanently storing a buffer is likely a good approach as a default, but some apps might want to keep the buffer gpu-only to save on memory when the buffer doesn't change often. This becomes especially interesting when you start allowing partial writes, where the update buffer could be tiny compared to the full thing.

reczkok:
At the moment it only creates the buffer when you fist write to a schema. Partial writes don't trigger this behaviour so as long as you only partial write the main buffer will never be created. Of course that is subject to change if we want to incorporate compiling to partial write which will be more difficult. There is also a major optimisation to be made for the partial write algorithm since it currently always creates an ArrayBuffer the size of the schema but it should be feasible to only make it as big as necessary without much overhead.

reczkok:
Added some benchmarks. Here is a sample result:
bench

packages/typegpu/src/data/compiledIO.ts
```diff
@@ -0,0 +1,149 @@
...
+import * as wgsl from './wgslTypes';
+
+export let EVAL_ALLOWED_IN_ENV: boolean;
+
+try {
+  new Function('return true');
+  EVAL_ALLOWED_IN_ENV = true;
+} catch {
+  EVAL_ALLOWED_IN_ENV = false;
+}
```
iwoplaza: To ensure that we're getting the computed value of `EVAL_ALLOWED_IN_ENV` in other files, let's try something like this.

```suggestion
export const EVAL_ALLOWED_IN_ENV: boolean = (() => {
  try {
    new Function('return true');
    return true;
  } catch {
    // Expected error
  }
  return false;
})();
```

iwoplaza: I can foresee a different behavior in CJS and ESM, which would be hard to catch.

packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -245,24 +251,44 @@ class TgpuBufferImpl<TData extends AnyData> implements TgpuBuffer<TData> {
     return this;
   }
 
+  compileWriter(): void {
+    if (EVAL_ALLOWED_IN_ENV) {
+      getCompiledWriterForSchema(this.dataType);
+    } else {
+      throw new Error('This enviorment does not allow eval');
```
iwoplaza:
```suggestion
      throw new Error('This environment does not allow eval');
```

iwoplaza:
The changes fail with the following test:

```ts
  it('should compile a writer for a mat4x4f', () => {
    const Schema = d.struct({
      transform: d.mat4x4f,
    });
    const writer = getCompiledWriterForSchema(Schema);

    const arr = new ArrayBuffer(sizeOf(Schema));
    const dataView = new DataView(arr);

    writer(dataView, 0, {
      transform: d.mat4x4f(...Array.from({ length: 16 }).map((_, i) => i)),
    });

    expect([...new Float32Array(arr)]).toEqual(
      Array.from({ length: 16 }).map((_, i) => i),
    );
  });
```

PR #824, opened 2025-02-06
fix: Make babel plugin target only calls to does with a function expression inside

iwoplaza:
This makes it work for cases where `.does()` was already transformed, e.g., for libraries using TypeGPU. This also works with `.does` method calls that don't originate from TypeGPU.


mhawryluk (approved):
Cool 🚀

PR #868, opened 2025-02-13
feat: Allow `std.mul()` to accept a combination of vectors and matrices

reczkok:


packages/typegpu/src/std/numeric.ts
```diff
@@ -29,13 +30,50 @@ export function sub<T extends vBase>(lhs: T, rhs: T): T {
...
     return VectorOps.mulSxV[v.kind](s, v);
   }
+  if (
+    typeof s === 'object' &&
+    typeof v === 'object' &&
+    'kind' in s &&
+    'kind' in v
+  ) {
+    const sIsVector = !s.kind.includes('mat');
+    const vIsVector = !v.kind.includes('mat');
```
iwoplaza: Tiny nit, covers all the cases and might be more performant.
```suggestion
    const sIsVector = !s.kind.startsWith('mat');
    const vIsVector = !v.kind.startsWith('mat');
```

iwoplaza (approved):
JUICED UP 🥤🍊

mhawryluk (approved):
Awesome! 🎉

PR #869, opened 2025-02-14
fix: Make colors in game of life example independent from resolution

mhawryluk:


iwoplaza (approved):
More consistency, yes pleassseee

PR #888, opened 2025-02-17
fix(typegpu): Throw NotStorageError instead of NotUniformError when unwrapping invalid bind group storage entry.

iwoplaza:


mhawryluk (approved):
🥸

PR #892, opened 2025-02-17
chore: Use typegpu plugin in StackBlitz integration

### LINKED ISSUE #890: Use rollup plugin in StackBlitz integration

mhawryluk:
closes #890 

(shall be changed to importing from 'unplugin-typegpu/vite') once that's a thing


apps/typegpu-docs/src/components/stackblitz/openInStackBlitz.ts
```diff
@@ -61,9 +61,18 @@ ${example.htmlCode}
     },
     "dependencies": {
       "@webgpu/types": "^0.1.44",
-      "typegpu": "^0.4.1"
+      "typegpu": "latest",
+      "unplugin-typegpu": "latest"
```
iwoplaza: I would still like to pin these, since they can be used as starters for people's custom TypeGPU projects.
We can however import our package's package.json files and inject the latest versions that way, to avoid having to update them manually.

```suggestion
      "typegpu": "^0.4.1",
      "unplugin-typegpu": "^0.1.0-alpha.2"
```

iwoplaza (approved):

PR #903, opened 2025-02-19
Add multisampling to function visualizer

reczkok:
I think this example will benefit much from it as smooth lines are prone to aliasing

Before:
Screenshot 2025-02-19 at 10 55 50
After:
Screenshot 2025-02-19 at 10 55 55


iwoplaza (approved):
Smooooth 🌊🌊🌊

aleksanderkatan (approved):
Great idea!

PR #931, opened 2025-02-20
feat: Atomics, Pointers and Resources scaffolding

### LINKED ISSUE #829: Handle atomics in TGSL
Atomic values are currently inferred as numbers even though they aren't in WGSL.
Example:
```ts
const getCell = tgpu['~unstable'].fn([d.u32, d.u32], d.u32).does((x, y) => {
  const value = nextStateStorage.value[getIndex(x, y)]; // <- TS: number(!) | WGSL: atomic<u32>
  const sum = value + u32(12) // <- TS: ✅ | WGSL: 🚫
  const actualValue = atomicLoad(value); // <- TS: number | WGSL: u32
  const correctSum = actualValue + u32(12) // All good 😎
});
```
Having them as number is misleading. We should infer them as a separate entity (most probably a specific pointer) that is accepted by standard functions like `atomicLoad`.

reczkok:
closes #829


packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -823,17 +844,39 @@ export interface WgslArray<TElement extends BaseData = BaseData> {
...
   /** Type-token, not available at runtime */
   readonly '~repr': Infer<TInner>;
 }
 
+export function unwrapPtr<T extends BaseData>(
+  ptr: Ptr<AddressSpace, T, Access>,
+): T {
+  return ptr.inner;
+}
+
```
mhawryluk:
```suggestion
```

chop

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -159,9 +215,45 @@ function generateExpression(
...
+      // Since wgsl doesn't support binary literals, we'll convert it to a decimal number
+      return {
+        value: `${Number.parseInt(value.slice(2), 2)}`,
+        dataType: d.abstractInt,
+      };
+    }
+
+    const floatRegex = /^-?(?:\d+\.\d*|\d*\.\d+)$/;
+    if (floatRegex.test(value)) {
+      return { value: `${value}`, dataType: d.abstractFloat };
```
mhawryluk:
```suggestion
      return { value, dataType: d.abstractFloat };
```

same everywhere below

mhawryluk (approved):
mind=blown 🚀

packages/typegpu/src/data/wgslTypes.ts
reczkok: It will be useful later but I can chop it for now

packages/typegpu/src/std/atomic.ts
```diff
@@ -0,0 +1,69 @@
...
+    return `atomicLoad(&${a})` as unknown as number;
+  }
+  throw new Error('Atomic operations are not supported outside of GPU mode.');
+}
+
+export function atomicStore<T extends AnyAtomic>(a: T, value: number): void {
+  if (inGPUMode()) {
+    // biome-ignore lint/correctness/noVoidTypeReturn: <string-void duality>
+    // biome-ignore lint/suspicious/noConfusingVoidType: <^>
+    return `atomicStore(&${a}, ${value})` as unknown as void;
```
iwoplaza: We can get away with less Biome wrestling 🤼‍♂️
```suggestion
    // biome-ignore lint/correctness/noVoidTypeReturn: <string-void duality>
    return `atomicStore(&${a}, ${value})` as unknown as undefined;
```

packages/typegpu/src/data/index.ts
```diff
@@ -2,12 +2,20 @@
...
 
-export { bool, f32, f16, i32, u32 } from './numeric';
+export {
+  bool,
+  f32,
+  f16,
+  i32,
+  u32,
+  abstractInt,
+  abstractFloat,
```
iwoplaza: Is there a need for `abstractInt` and `abstractFloat` to be part of the public API, or not? From what I understand, WGSL essentially hides them, and does not allow their use in user-defined shaders.

iwoplaza (approved):
Amazing addition, great job! 👏

packages/typegpu/src/std/atomic.ts
reczkok: I don't even need to turn it from void to undefined as it turns out

PR #939, opened 2025-02-24
dx: Fetch dependency versions automatically from codebase in openInStackBlitz function

### LINKED ISSUE #937: Fetch dependency versions automatically from codebase in `openInStackBlitz` function

mhawryluk:
closes #937


iwoplaza (commented):
Awesome, this reduces a lot of the friction we encounter while deploying 🙌 🚀 

I have one request, could we try out [`@std/yaml`](https://jsr.io/@std/yaml) from the JSR & Deno folk instead of `js-yaml`? It seems to have type definitions out of the box, and it's more likely to get patch updates down the line.

As for the "as { ... }" after loading the Yaml, could we use [arktype](https://arktype.io/docs/intro/setup) for validating whether or not the config file matches what we expect? This way, if we ever change the workspace config file in an unexpected way, we'll be notified properly.

apps/typegpu-docs/src/components/stackblitz/openInStackBlitz.ts
```diff
@@ -1,8 +1,21 @@
...
 
+const pnpmWorkspaceYaml = type({
+  catalog: { typescript: 'string', '@webgpu/types': 'string' },
+})(parse(pnpmWorkspace));
+
 export function openInStackBlitz(example: Example) {
+  if (pnpmWorkspaceYaml instanceof type.errors) {
+    throw new Error(pnpmWorkspaceYaml.message);
+  }
+
```
iwoplaza: Imo, we could move this error check outside of the function to be aware of failure earlier, not requiring the button click.

iwoplaza (approved):
Awesome 🤩

apps/typegpu-docs/src/components/stackblitz/openInStackBlitz.ts
mhawryluk: then the types inside the function are no longer narrowed

mhawryluk: nevermind

PR #941, opened 2025-02-24
feat: Check for invalid length in arrayOf and disarrayOf

### LINKED ISSUE #889: Check for NaN length in `arrayOf` and `disarrayOf`, and throw descriptive error.

mhawryluk:
closes #889


packages/typegpu/src/data/array.ts
```diff
@@ -47,6 +47,16 @@ class WgslArrayImpl<TElement extends BaseData> implements WgslArray<TElement> {
     if (Number.isNaN(sizeOf(elementType))) {
       throw new Error('Cannot nest runtime sized arrays.');
     }
+
+    if (
+      Number.isNaN(elementCount) ||
+      !Number.isInteger(elementCount) ||
```
aleksanderkatan: The [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) of `Number.isInteger` say that "If the target value is an integer, return true, otherwise return false. If the value is [NaN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NaN) or [Infinity](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Infinity), return false.", is the `isNaN` check necessary?

packages/typegpu/src/data/disarray.ts
```diff
@@ -46,5 +46,15 @@ class DisarrayImpl<TElement extends AnyData> implements Disarray<TElement> {
   constructor(
     public readonly elementType: TElement,
     public readonly elementCount: number,
-  ) {}
+  ) {
+    if (
+      Number.isNaN(elementCount) ||
```
aleksanderkatan: Same comment as above

aleksanderkatan (approved):
Good job!

packages/typegpu/src/data/array.ts
mhawryluk: great catch, thank you!!

packages/typegpu/src/data/array.ts
```diff
@@ -47,6 +47,16 @@ class WgslArrayImpl<TElement extends BaseData> implements WgslArray<TElement> {
...
+      elementCount < 0
+    ) {
+      throw new Error(
+        `Cannot create array with invalid element count: ${elementCount}.`,
```
iwoplaza:
```suggestion
        `Cannot create array schema with invalid element count: ${elementCount}.`,
```

iwoplaza (approved):
Just tiny nit, awesome change!! 👏

aleksanderkatan (approved):
👍 👍 👍

PR #943, opened 2025-02-24
Simple 3d cube rasterization

reczkok:


reczkok:
The controls are not working as intended at the moment but it was just for reference so I will leave it

reczkok:
Closed in favor of #953

PR #957, opened 2025-02-25
fix: Function visualizer msaa texture size

reczkok:


apps/typegpu-docs/src/content/examples/rendering/function-visualizer/index.ts
```diff
@@ -481,6 +481,26 @@ canvas.onwheel = (event) => {
...
+    format: presentationFormat,
+    usage: GPUTextureUsage.RENDER_ATTACHMENT,
+  });
+  msView = msTexture.createView();
+
+  renderPassDescriptor.colorAttachments[0].view = msView;
+  renderBackgroundPassDescriptor.colorAttachments[0].view = msView;
+});
+
+resizeObserver.observe(canvas);
```
mhawryluk, line 502: observer shall be disconnected on cleanup

apps/typegpu-docs/src/content/examples/rendering/function-visualizer/index.ts
```diff
@@ -481,6 +481,26 @@ canvas.onwheel = (event) => {
   );
 };
 
+// add resize observer that recreates the multisampled texture on canvas resize
```
mhawryluk:
```suggestion
// recreate the multisampled texture on canvas resize
```

apps/typegpu-docs/src/content/examples/rendering/function-visualizer/index.ts
```diff
@@ -481,6 +481,26 @@ canvas.onwheel = (event) => {
...
+  msTexture = device.createTexture({
+    size: [
+      canvas.clientWidth * window.devicePixelRatio,
+      canvas.clientHeight * window.devicePixelRatio,
+    ],
+    sampleCount: 4,
+    format: presentationFormat,
+    usage: GPUTextureUsage.RENDER_ATTACHMENT,
+  });
+  msView = msTexture.createView();
```
mhawryluk, line 496: couuld be a function returning texture and view to not repeat oneself and ensure consistency, although I do not insist

mhawryluk (approved):
Awesome!!

apps/typegpu-docs/src/content/examples/rendering/function-visualizer/index.ts
reczkok: I think it's okey here. There is no much risk of inconsistencies since there are no moving parts (the descriptor is not likely to change). Abstracting it away would be a net gain in line count and net loss in readability for people who just want to see how the shader code works (without the specific example implementation)

iwoplaza (approved):
Great change! 🍸

PR #968, opened 2025-02-28
feat: TGSL resources

### LINKED ISSUE #357: Calls to functions should accept `Resources` instead of plain identifiers.
Calls to functions have to accept `Resources` in order to adjust their functionality based on the type of arguments.

E.g.:

Wrapping `i32` and `u32` expressions with `f32` to minimize explicit casts in these obvious cases.
```js
const val_u = u32(123);
const val_i = i32(123);

vec3f(val_u) // vec3f({ value: '123', dataType: U32 }) -> `vec3f(f32(val_u))`
vec3f(val_i) // vec3f({ value: '123', dataType: I32 }) -> `vec3f(f32(val_i))`
// AbstractInt and AbstractFloat are left as they are
vec3f(123) // vec3f({ value: '123', dataType: AbstractInt }) -> `vec3f(123)`
vec3f(0.5) // vec3f({ value: '123', dataType: AbstractFloat }) -> `vec3f(0.5)`
```

### LINKED ISSUE #987: bug: Allow for numbers with scientific notation in TGSL.
Right now, trying to parse the following function throws an error:
```typescript
const f = tgpu['~unstable'].fn([], d.f32).does(() => 1e3);
```
The same happens for:
```typescript
const f = tgpu['~unstable'].fn([], d.f32).does(() => 1000);
```
This seems not to be the case for numeric literals that do not simplify to scientific notation.

### LINKED ISSUE #998: impr: Throw an appropriate error when declaring a variable without an initial value.
When running a function containing a variable without an initial value, e.g.:
```typescript
const f = tgpu['~unstable'].fn([], d.f32).does(() => {
  let a;
  a = 3;
  return a;
});
```
the error that is thrown is not descriptive:

<details>
  <summary>Error message</summary>
  
  ```javascript
- renderPipeline:<unnamed>
- vertexFn:<unnamed>
    at ResolutionError.appendToTrace (errors.ts:67:12)
    at ResolutionCtxImpl._getOrInstantiate (resolutionCtx.ts:549:19)
    at resolutionCtx.ts:567:52
    at provideCtx (gpuMode.ts:11:12)
    at ResolutionCtxImpl.resolve (resolutionCtx.ts:567:24)
    at resolve (resolutionCtx.ts:614:18)
    at RenderPipelineCore.unwrap (renderPipeline.ts:424:52)
    at TgpuRenderPipelineImpl.draw (renderPipeline.ts:305:28)
    at frame (eval at executeExample (exampleRunner.ts:207:17), <anonymous>:480:165)
    at eval (eval at executeExample (exampleRunner.ts:207:17), <anonymous>:495:1)
    at async executeExample (exampleRunner.ts:214:5)Caused by: Error: Expected block scope layer to be on top.
    at ItemStateStack.popBlockScope (resolutionCtx.ts:136:13)
    at ResolutionCtxImpl.popBlockScope (resolutionCtx.ts:346:26)
    at generateBlock (wgslGenerator.ts:103:9)
    at generateFunction (wgslGenerator.ts:473:10)
    at ResolutionCtxImpl.fnToWgsl (resolutionCtx.ts:369:17)
    at Object.resolve (fnCore.ts:99:36)
    at ~resolve (tgpuVertexFn.ts:164:21)
    at ResolutionCtxImpl._getOrInstantiate (resolutionCtx.ts:532:33)
    at ResolutionCtxImpl.resolve (resolutionCtx.ts:571:19)
    at renderPipeline.ts:428:19
    at ResolutionCtxImpl.withSlots (resolutionCtx.ts:430:14)
    at ~resolve (renderPipeline.ts:427:17)
    at ResolutionCtxImpl._getOrInstantiate (resolutionCtx.ts:532:33)
    at resolutionCtx.ts:567:52
    at provideCtx (gpuMode.ts:11:12)
    at ResolutionCtxImpl.resolve (resolutionCtx.ts:567:24)
    at resolve (resolutionCtx.ts:614:18)
    at RenderPipelineCore.unwrap (renderPipeline.ts:424:52)
    at TgpuRenderPipelineImpl.draw (renderPipeline.ts:305:28)
    at frame (eval at executeExample (exampleRunner.ts:207:17), <anonymous>:480:165)
    at eval (eval at executeExample (exampleRunner.ts:207:17), <anonymous>:495:1)
    at async executeExample (exampleRunner.ts:214:5)
  ```
  
</details>

reczkok:
# ~~Very WIP~~
### Now should be ready for review

closes #987 
closes #357 
closes #998


packages/typegpu/src/std/numeric.ts
```diff
@@ -12,304 +14,420 @@ import type {
...
+export const add = createDualImpl(
+  // CPU implementation
+  <T extends vBase>(lhs: T, rhs: T): T => {
+    return VectorOps.add[lhs.kind](lhs, rhs);
+  },
+  // GPU implementation
+  (lhs, rhs) => {
+    return { value: `(${lhs.value} + ${rhs.value})`, dataType: lhs.dataType };
+  },
+);
```
iwoplaza: I think we can get away with using this shorthand, especially for gpu implementations.

```suggestion
export const add = createDualImpl(
  // CPU implementation
  <T extends vBase>(lhs: T, rhs: T): T => VectorOps.add[lhs.kind](lhs, rhs),
  // GPU implementation
  (lhs, rhs) => ({ value: `(${lhs.value} + ${rhs.value})`, dataType: lhs.dataType }),
);
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -153,22 +159,34 @@ function createFn<
...
-      );
-    }
+      return implementation(...args);
+    },
+    (...args) => {
+      return {
+        value: new FnCall(fn, args.map((arg) => arg.value) as Wgsl[]),
+        dataType: shell.returnType ?? UnknownData,
+      };
+    },
```
```suggestion
    (...args) => ({
      value: new FnCall(fn, args.map((arg) => arg.value) as Wgsl[]),
      dataType: shell.returnType ?? UnknownData,
    }),
```

packages/typegpu/src/core/buffer/bufferUsage.ts
```diff
@@ -22,6 +22,7 @@ export interface TgpuBufferUsage<
   TUsage extends BindableBufferUsage = BindableBufferUsage,
 > {
   readonly resourceType: 'buffer-usage';
+  readonly dataType: TData;
```
iwoplaza: Could we hold this information on the `$internal` property? This way, we can narrow down the public API, and change internal APIs with more confidence.

mhawryluk:
error

reczkok:
> error

This was caused by being overly cautious with variable definition. Since it checked for state that is invalid in TS anyway I removed the check instead of implementing shadowing validation (checking if a variable is redeclared) on our side.

packages/typegpu/src/data/vector.ts
```diff
@@ -257,3 +213,66 @@ export const vec4i = makeVecSchema(Vec4iImpl) as Vec4i;
...
+  vec2i: vec2i,
+  vec2u: vec2u,
+  vec3f: vec3f,
+  vec3h: vec3h,
+  vec3i: vec3i,
+  vec3u: vec3u,
+  vec4f: vec4f,
+  vec4h: vec4h,
+  vec4i: vec4i,
+  vec4u: vec4u,
```
mhawryluk:
```suggestion
  vec2f,
  vec2h,
  vec2i,
  vec2u,
  vec3f,
  vec3h,
  vec3i,
  vec3u,
  vec4f,
  vec4h,
  vec4i,
  vec4u,
```

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -835,4 +835,128 @@ export const VectorOps = {
       c: T | number,
     ) => T
   >,
+
+  sin: {
```
mhawryluk, line 839: you can use the new `unaryXY` functions and such to simplify these function definitions below

packages/typegpu/src/smol/generationHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+      return indexableTypeToResult[
+        resource.type as keyof typeof indexableTypeToResult
+      ];
+    }
+  }
+
+  return undefined;
+}
+
+export function getTypeFormWgsl(resource: Wgsl): AnyWgslData | UnknownData {
```
mhawryluk: from

packages/typegpu/src/shared/generators.ts
```diff
@@ -14,3 +17,21 @@ export function* naturalsExcept(
...
+  jsImpl: T,
+  gpuImpl: (...args: MapValueToResource<Parameters<T>>) => Resource,
+): T {
+  return ((...args: Parameters<T>) => {
+    if (inGPUMode()) {
+      return gpuImpl(
+        ...(args as unknown as MapValueToResource<Parameters<T>>),
+      ) as unknown as Resource;
+    }
+    // biome-ignore lint/suspicious/noExplicitAny: <t's very convenient>
```
mhawryluk:
```suggestion
    // biome-ignore lint/suspicious/noExplicitAny: <it's very convenient>
```

packages/typegpu/src/shared/getTypeForPropAccess.ts
mhawryluk, line 1: ✂️

packages/typegpu/src/smol/generationHelpers.ts
```diff
@@ -0,0 +1,230 @@
...
+  type AnyWgslData,
+  type BaseData,
+  isDecorated,
+  isWgslArray,
+  isWgslData,
+} from '../data/wgslTypes';
+import { getResolutionCtx } from '../gpuMode';
+import { type Resource, UnknownData, type Wgsl } from '../types';
+
+const indexableTypes = [
```
mhawryluk: from how it's used I guess it's more `swizzleableTypes`? and 'struct' isn't necessary here. cause now indexableTypes and indexableTypeToResult don't match and I would assume they should from their names

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -31,24 +36,27 @@ const parenthesizedOps = [
   '||',
 ];
 
-function binaryOperatorToType<
+const logicalOps = ['&&', '||', '!', '==', '!=', '<', '<=', '>', '>='];
+
+type OpratorTypes =
```
mhawryluk: `Operator`

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -31,24 +36,27 @@ const parenthesizedOps = [
   '||',
 ];
 
-function binaryOperatorToType<
+const logicalOps = ['&&', '||', '!', '==', '!=', '<', '<=', '>', '>='];
```
mhawryluk: make it binaryLogicalOps and remove `'!'`

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -154,46 +158,52 @@ export function generateExpression(
...
     };
   }
 
   if ('a' in expression) {
     // Member Access
-
     const [targetId, property] = expression.a;
     const target = generateExpression(ctx, targetId);
-    const propertyStr = resolveRes(ctx, generateExpression(ctx, property));
+    const propertyStr = property;
```
mhawryluk: ✂️

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -154,46 +158,52 @@ export function generateExpression(
...
+        value,
+        dataType:
+          getTypeForPropAccess(target.value as d.AnyWgslData, propertyStr) ??
+          UnknownData,
       };
     }
 
     if (typeof target.value === 'object') {
+      const dataType =
+        !target.value === null && isWgsl(value)
```
mhawryluk: `!target.value === null` will never be true. I believe this was supposed to be `!(target.value === null)`? regardless, the program will fail on line 187 if target.value is null and never reach this line

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -202,101 +212,71 @@ export function generateExpression(
...
     // Assuming that `id` is callable
-    // TODO: Pass in resources, not just values.
     const result = (idValue as unknown as (...args: unknown[]) => unknown)(
-      ...argValues,
-    ) as Wgsl;
-    // TODO: Make function calls return resources instead of just values.
-    return { value: result, dataType: UnknownData };
+      ...resolvedResources,
+    ) as Resource;
+    return result;
```
mhawryluk: `result` constant isn't necessary

packages/typegpu/src/std/atomic.ts
```diff
@@ -1,92 +1,185 @@
...
+export const workgroupBarrier = createDualImpl(
+  // CPU implementation
+  () => {
+    console.warn('workgroupBarrier is a no-op outside of GPU mode.');
+  },
+  // GPU implementation
+  () => {
+    return { value: 'workgroupBarrier()', dataType: Void };
+  },
+);
```
mhawryluk:
```suggestion
export const workgroupBarrier = createDualImpl(
  // CPU implementation
  () => console.warn('workgroupBarrier is a no-op outside of GPU mode.'),
  // GPU implementation
  () => ({ value: 'workgroupBarrier()', dataType: Void }),
);
```

same everywhere beloww

packages/typegpu/src/std/atomic.ts
```diff
@@ -1,92 +1,185 @@
...
 
-export function storageBarrier(): void {
-  if (inGPUMode()) {
-    // biome-ignore lint/correctness/noVoidTypeReturn: <string-void duality>
-    return 'storageBarrier()' as unknown as undefined;
-  }
-  console.warn('storageBarrier is a no-op outside of GPU mode.');
-}
+export const storageBarrier = createDualImpl(
+  // CPU implementation
```
mhawryluk, line 15: `createDualImpl` could accept object { cpu: ..., gpu: ... }, we wouldn't have to write these comments. but it's okay

packages/typegpu/src/std/numeric.ts
```diff
@@ -12,304 +14,420 @@ import type {
...
-}
+export const add = createDualImpl(
+  // CPU implementation
+  <T extends vBase>(lhs: T, rhs: T): T => {
+    return VectorOps.add[lhs.kind](lhs, rhs);
+  },
+  // GPU implementation
+  (lhs, rhs) => {
+    return { value: `(${lhs.value} + ${rhs.value})`, dataType: lhs.dataType };
+  },
```
mhawryluk: I suggest one-lining the functions in this file too, but I don't insist

mhawryluk (changes requested):

packages/typegpu/src/std/atomic.ts
reczkok: I don't know... It could be a little bloated

packages/typegpu/src/std/atomic.ts
reczkok: Everywhere? I would stop at barriers

mhawryluk: sure, just when it's possible and straight-forward. so up till barriers

packages/typegpu/src/std/atomic.ts
mhawryluk: i mean, it's fewer lines than with the comments

```
export const storageBarrier = createDualImpl({
  cpu: () => console.warn('storageBarrier is a no-op outside of GPU mode.'),
  gpu: () => ({ value: 'storageBarrier()', dataType: Void }),
});
```

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -31,24 +36,27 @@ const parenthesizedOps = [
   '||',
 ];
 
-function binaryOperatorToType<
+const binaryLogicalOps = ['&&', '||', '==', '!=', '<', '<=', '>', '>='];
+
+type Oprator =
```
mhawryluk: `Operator` (there's a typo)

apps/typegpu-docs/public/assets/atom.svg
mhawryluk, line 1: I assume changed by mistake?

packages/typegpu/src/smol/wgslGenerator.ts
reczkok: I seem to be very good at those

mhawryluk (approved):
Awesome! 👨‍🍳

packages/typegpu/src/smol/generationHelpers.ts
```diff
@@ -0,0 +1,232 @@
...
+    };
+  }
+
+  // Floating point literals
+  if (/^-?(?:\d+\.\d*|\d*\.\d+)$/i.test(value)) {
+    return { value, dataType: abstractFloat };
+  }
+
+  // Floating point literals with scientific notation
+  if (/^-?\d+\.\d+e-?\d+$/i.test(value)) {
```
iwoplaza: I think your force push removed my changes.
```suggestion
  if (/^-?\d+(?:\.\d+)?e-?\d+$/i.test(value)) {
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -131,9 +135,17 @@ class ItemStateStack {
   }
 
   popBlockScope() {
-    const layer = this._stack.pop();
-    if (layer?.type !== 'blockScope') {
-      throw new Error('Expected block scope layer to be on top.');
+    // pop block scope (if top layer is not block scope, traverse until block scope is found)
```
iwoplaza: Does this mean that there are layers that are not being popped when they're supposed to?

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -477,7 +487,9 @@ export class ResolutionCtxImpl implements ResolutionCtx {
       }
 
       // If we got here, no item with the given slot-to-value combo exists in cache yet
+      pushMode(RuntimeMode.CPU);
       const result = derived['~compute']();
+      popMode(RuntimeMode.CPU);
```
iwoplaza:
```suggestion
      pushMode(RuntimeMode.CPU);
      let result;
      try {
        result = derived['~compute']();
      } finally {
        popMode(RuntimeMode.CPU);
      }
```

packages/typegpu/src/smol/generationHelpers.ts
```diff
@@ -0,0 +1,240 @@
...
+  }
+
+  if (isDerived(targetType) || isSlot(targetType)) {
+    const ctx = getResolutionCtx();
+    if (!ctx) {
+      throw new Error(
+        'Resolution context not found when unwrapping slot or derived',
+      );
+    }
+    return ctx.unwrap(targetType) as BaseData;
```
iwoplaza: Is this a full solution? I would have thought it would look something like this:
```suggestion
    return getTypeForPropAccess(ctx.unwrap(targetType) as Wgsl, propName);
```

Also, when compared to index access, why are we not using `targetType.value` here?

packages/typegpu/src/resolutionCtx.ts
reczkok: I think it used to be the case in the earlier implementation but that is no longer the case. Will remove

packages/typegpu/src/resolutionCtx.ts
reczkok: Is there a concrete reason for this or just consistency?

packages/typegpu/src/smol/generationHelpers.ts
reczkok: I actually think that inside index access 
```
  if (isDerived(resource) || isSlot(resource)) {
    return getTypeForIndexAccess(resource.value as Wgsl);
  }
```
could be omitted as there should not be a way for them to be the target of index access (that would work in TGSL).

reczkok: I will create some tests for those cases and come back to this.

reczkok: I created some tests to cover this. The main takeaways are:
- The index access code was wrong but unreachable
- We don't recurse here since the only way to get there is via `.value` which is meaningless on the unwrapped type
- But there was an error since the unwrapped result needs to go through `getTypeFromWgsl`

iwoplaza (approved):
MERGE IT 🔈

PR #989, opened 2025-03-05
docs: Multiple files support

lursz:
Introduced support for multi-file examples, thus eliminating the need for tremendously long files.

Examples were previously loaded dynamically by reading files into a string of code, using a custom solution to handle imports, and then transpiling the code into `js`. This pull request changes that. The implementation extracts the URL from the `vite` plugin's import and uses it to load the example. Dynamic imports aren't viable with `vite` because the path changes during the `build`, thus forcing us to use workaround. Directly using import.meta.glob is also not an option, as the data gets cached, which is why we concatenate `&update=${Date.now()}` to the query.

image


apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -66,29 +66,37 @@ function useExample(
...
-  // biome-ignore lint/correctness/useExhaustiveDependencies: reset embedded html on code change
+  const editorTabsList = [
+    ...(Object.keys(tsCodes).includes('index.ts')
+      ? [
+          'index.ts',
+          ...Object.keys(tsCodes).filter((name) => name !== 'index.ts'),
+        ]
+      : Object.keys(tsCodes)),
+    'index.html',
+  ];
```
reczkok: Maybe introduce something like `const tsFiles = Object.keys(tsCodes)`

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -131,18 +139,35 @@ export function ExampleView({ example }: Props) {
...
+                            'px-4 py-2',
+                            currentTSFile === fileName
+                              ? 'rounded-t-lg rounded-bl-none rounded-br-none bg-gradient-to-br from-gradient-purple to-gradient-blue text-white hover:from-gradient-purple-dark hover:to-gradient-blue-dark'
+                              : 'rounded-t-lg rounded-bl-none rounded-br-none bg-white border-tameplum-100 border-2 hover:bg-tameplum-20',
+                          )}
+                        >
+                          {fileName}
+                        </button>
+                      ))}
+                    </div>
```
reczkok: I would re-introduce a component for this

apps/typegpu-docs/src/utils/examples/exampleContent.ts
```diff
@@ -10,9 +10,9 @@ function pathToExampleKey<T>(record: Record<string, T>): Record<string, T> {
         [
           pipe(
             path,
-            (path) => path.replace(/^..\/..\/content\/examples\//, ''), // removing parent folder
-            (path) => path.replace(/\/[^\/]*$/, ''), // removing leaf file names (e.g. meta.json, index.ts)
-            (path) => path.replace(/\//, '--'), // replacing path separators with '--'
+            (p) => p.replace(/^..\/..\/content\/examples\//, ''),
+            (p) => p.replace(/\/[^\/]*$/, ''),
+            (p) => p.replace(/\//, '--'),
```
reczkok: The comments were nice

apps/typegpu-docs/src/utils/examples/exampleContent.ts
```diff
@@ -21,21 +21,80 @@ function pathToExampleKey<T>(record: Record<string, T>): Record<string, T> {
...
+): Record<string, Record<string, T>> {
+  const groups: Record<string, Record<string, T>> = {};
+
+  for (const [path, value] of Object.entries(record)) {
+    const groupKey = pipe(
+      path,
+      (p) => p.replace(/^..\/..\/content\/examples\//, ''),
+      (p) => p.replace(/\/[^\/]*$/, ''),
+      (p) => p.replace(/\//, '--'),
+    );
```
reczkok: This pipe seems to be used quite a lot in this file so maybe we could reuse it

apps/typegpu-docs/src/components/ExampleView.tsx
lursz: It does not seem to be necessary, it's just a single `button` wrapped in a `div`. It just looks bloated because of the biome formatting. We are also almost certainly never going to reuse this.

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -131,18 +137,35 @@ export function ExampleView({ example }: Props) {
...
+                    {currentTSFile === 'index.html' ? (
+                      <HtmlCodeEditor shown code={htmlCode} />
+                    ) : (
+                      <TsCodeEditor shown code={tsCodes[currentTSFile]} />
+                    )}
+                  </>
+                }
+
+                {<HtmlCodeEditor shown code={htmlCode} />}
```
mhawryluk:
```suggestion
```

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -131,18 +137,35 @@ export function ExampleView({ example }: Props) {
...
+                            currentTSFile === fileName
+                              ? 'rounded-t-lg rounded-bl-none rounded-br-none bg-gradient-to-br from-gradient-purple to-gradient-blue text-white hover:from-gradient-purple-dark hover:to-gradient-blue-dark'
+                              : 'rounded-t-lg rounded-bl-none rounded-br-none bg-white border-tameplum-100 border-2 hover:bg-tameplum-20',
+                          )}
+                        >
+                          {fileName}
+                        </button>
+                      ))}
+                    </div>
+                    {currentTSFile === 'index.html' ? (
```
mhawryluk: `currentTSFile` should be called `currentFile`, as it can be html too

apps/typegpu-docs/src/content/examples/simple/triangle/index.ts
```diff
@@ -1,10 +1,12 @@
 import tgpu from 'typegpu';
 import * as d from 'typegpu/data';
-
```
mhawryluk: keep the empty line please

apps/typegpu-docs/src/utils/examples/exampleRunner.ts
```diff
@@ -160,59 +36,52 @@ export async function executeExample(
...
+  ): string | undefined {
+    const filePath = String(importFn);
+    const match = filePath.match(/\(\)\s*=>\s*import\("([^"]+)"\)/);
+    if (match?.[1]) {
+      return match[1];
+    }
+  }
 
-    const transformedCode =
-      Babel.transform(jsCode, {
```
mhawryluk, line 195: you can remove Babel from this package's package.json

apps/typegpu-docs/src/utils/examples/exampleRunner.ts
```diff
@@ -160,59 +36,52 @@ export async function executeExample(
...
 
-    const jsCode = tsToJs(exampleCode);
+  function myExtractUrlFromViteImport(
+    importFn: () => void,
+  ): string | undefined {
+    const filePath = String(importFn);
+    const match = filePath.match(/\(\)\s*=>\s*import\("([^"]+)"\)/);
+    if (match?.[1]) {
+      return match[1];
+    }
```
mhawryluk:
```suggestion
    return match?.[1];
```

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -131,18 +137,35 @@ export function ExampleView({ example }: Props) {
...
-
-                <TsCodeEditor shown={currentEditorTab === 'ts'} code={tsCode} />
-
-                <HtmlCodeEditor
-                  shown={currentEditorTab === 'html'}
-                  code={htmlCode}
-                />
+              <div className="absolute inset-1">
+                {
+                  <>
```
mhawryluk: remove these wrappers, not needed

mhawryluk (commented):
editor sticks out of the page and causes overflow:

Zrzut ekranu 2025-03-6 o 11 32 44

on desktop view there's also an unnecessary scroll (maybe because of inset-1)

apps/typegpu-docs/src/utils/examples/exampleRunner.ts
```diff
@@ -160,59 +36,52 @@ export async function executeExample(
...
+  }
 
-    const jsCode = tsToJs(exampleCode);
+  function myExtractUrlFromViteImport(
+    importFn: () => void,
+  ): string | undefined {
+    const filePath = String(importFn);
+    const match = filePath.match(/\(\)\s*=>\s*import\("([^"]+)"\)/);
+    if (match?.[1]) {
+      return match[1];
```
iwoplaza:
```suggestion
      return new URL(match[1], window.location.origin);
```

apps/typegpu-docs/src/utils/examples/exampleRunner.ts
```diff
@@ -160,59 +36,52 @@ export async function executeExample(
...
+    }
+    if ('onSliderChange' in param) {
+      return param.onSliderChange(param.initial ?? param.min ?? 0);
+    }
+  }
 
-    const jsCode = tsToJs(exampleCode);
+  function myExtractUrlFromViteImport(
+    importFn: () => void,
+  ): string | undefined {
```
iwoplaza:
```suggestion
  ): URL | undefined {
```

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -22,7 +22,8 @@ type Props = {
 };
 
 function useExample(
-  exampleCode: string,
+  exampleCode: Record<string, string>,
+  tsSources: Record<string, () => void>,
```
iwoplaza: I find the names a bit misleading, and to be frank, they were misleading before this change as well, haha. Now that I look at this change, `exampleCode` doesn't seems to be used at all (apart from the dependency array, which it should be removed from).

```suggestion
  tsImports: Record<string, () => Promise<unknown>>,
```

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -66,29 +66,35 @@ function useExample(
   }, [exampleCode, htmlCode, setSnackbarText, setExampleControlParams]);
```
iwoplaza: `exampleCode` is not used in this useEffect, while `tsSources` is.

```suggestion
  }, [tsSources, htmlCode, setSnackbarText, setExampleControlParams]);
```

iwoplaza (changes requested):

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -63,32 +62,36 @@ function useExample(
...
   const codeEditorMobileShowing = useAtomValue(codeEditorShownMobileAtom);
-
   const exampleHtmlRef = useRef<HTMLDivElement>(null);
 
-  // biome-ignore lint/correctness/useExhaustiveDependencies: reset embedded html on code change
+  const codeFiles = Object.keys(tsCodes);
+  const editorTabsList = [
+    ...(codeFiles.includes('index.ts')
+      ? ['index.ts', ...codeFiles.filter((name) => name !== 'index.ts')]
+      : codeFiles),
```
reczkok: I think we already assume that `index.ts` exists so maybe we could simplify this?

mhawryluk (approved):
Great job! 🎉

iwoplaza:
Awesome change! 🙌

PR #1006, opened 2025-03-11
fix: Make babel plugin work with boolean and string literals

### LINKED ISSUE #1003: Babel plugin doesn't support BooleanLiteral

mhawryluk:
closes #1003

### DIFF
--- a/packages/tinyest-for-wgsl/package.json
+++ b/packages/tinyest-for-wgsl/package.json
@@ -1,7 +1,7 @@
 {
   "name": "tinyest-for-wgsl",
   "private": true,
-  "version": "0.1.0-alpha.4",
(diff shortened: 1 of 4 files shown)


iwoplaza:
Looks good! Just see why the amount of changes is so large. Might have to change the base to something else, then change it back to main

mhawryluk:
> Looks good! Just see why the amount of changes is so large. Might have to change the base to something else, then change it back to main

looks okay to me 🧐

Zrzut ekranu 2025-03-11 o 17 07 22

iwoplaza (approved):
🚀 🚀 🚀 
(looks great here, it looked weird on mobile)

PR #1008, opened 2025-03-12
dx: Run CI on any PR, not only to main

mhawryluk:
previously biome check didn't run for PRs to the release branch

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,8 @@ name: CI
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -2,9 +2,8 @@ name: CI
 
 on:
   push:
-    branches: ['main']
+    branches: [main]
```
reczkok: We could also include release here

reczkok (approved):

PR #1017, opened 2025-03-13
feat: Add cubemap reflection example

reczkok:


iwoplaza:
Could we move the cubemap to public/assets/cubemap-reflection, to match the other examples? Also, what's the source and license for that cubemap?

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.html
```diff
@@ -0,0 +1,41 @@
...
+    border-radius: 10px;
+    user-select: none;
+    pointer-events: none;
+    transition: opacity 0.5s;
+  }
+</style>
+<div id="help">
+  <h3 class="controls-header">Controls (click to dismiss)</h3>
+  <p class="controls"><b>Left Mouse Button:</b> Look around</p>
+  <p class="controls"><b>Scroll Wheel:</b> Zoom in/out</p>
```
aleksanderkatan, line 34: I know that in code this really is a zoom in/out, but from the user perspective, I think it would be more intuitive to interpret this as scaling the icosphere, what do you think?

mhawryluk (commented):
please add support for the new vector slider controls in `stackBlitzIndex.ts`

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,478 @@
...
+    d.vec3f(0, 1, 0),
+    d.mat4x4f(),
+  );
+  cameraBuffer.writePartial({ view: newView, position: newCameraPos });
+}
+
+canvas.addEventListener('wheel', (event: WheelEvent) => {
+  event.preventDefault();
+  const zoomSensitivity = 0.05;
+  orbitRadius = Math.max(1, orbitRadius + event.deltaY * zoomSensitivity);
```
aleksanderkatan:
```suggestion
  orbitRadius = std.clamp(orbitRadius + event.deltaY * zoomSensitivity, 3, 100);
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/dataTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+export const Camera = d.struct({
+  view: d.mat4x4f,
+  projection: d.mat4x4f,
+  position: d.vec4f,
+});
+
+export const Vertex = d.unstruct({
+  position: d.float16x4,
+  normal: d.float16x4,
+});
```
aleksanderkatan, line 12: `vec4h` has size 8 and aligmnent 8, I don't see what is the advantage of `unstruct` here, and the example seems to work after changing it to a struct of two `vec4h`s

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/dataTypes.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export const ComputeVertex = d.struct({
+  position: d.vec2u, // 2x16 float
+  normal: d.vec2u, // 2x16 float
+});
```
aleksanderkatan: What is the advantage of using `vec2u` and then unpacking it to `vec4f` instead of just using `vec4h`? Also, I think the comment is a little misleading, I would change it to e.g. `2x(2x16 float)`

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+import * as d from 'typegpu/data';
+import * as std from 'typegpu/std';
+
+export const unpackVec2u = tgpu['~unstable'].fn(
+  { packed: d.vec2u },
+  d.vec4f,
+)(({ packed }) => {
+  const xy = std.unpack2x16float(packed.x);
+  const zw = std.unpack2x16float(packed.y);
+  return d.vec4f(xy.x, xy.y, zw.x, zw.y);
```
aleksanderkatan:
```suggestion
  return d.vec4f(xy, zw);
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+});
+
+export const packVec2u = tgpu['~unstable'].fn(
+  { toPack: d.vec4f },
+  d.vec2u,
+)(({ toPack }) => {
+  const xy = std.pack2x16float(d.vec2f(toPack.x, toPack.y));
+  const zw = std.pack2x16float(d.vec2f(toPack.z, toPack.w));
```
aleksanderkatan:
```suggestion
  const xy = std.pack2x16float(toPack.xy);
  const zw = std.pack2x16float(toPack.zw);
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+  return d.vec2u(xy, zw);
+});
+
+export const normalizeSafely = tgpu['~unstable'].fn(
+  { v: d.vec4f },
+  d.vec4f,
+)(({ v }) => {
+  const length = std.length(d.vec3f(v.x, v.y, v.z));
```
aleksanderkatan:
```suggestion
  const length = std.length(v.xyz);
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/icosphere.ts
```diff
@@ -0,0 +1,324 @@
...
+/**
+ * Safely normalizes a vector to prevent numerical instability
+ */
+function normalizeSafely(v: d.v4f): d.v4f {
+  const length = Math.sqrt(v.x * v.x + v.y * v.y + v.z * v.z);
+  if (length < 1e-8) {
+    return d.vec4f(0, 0, 1, 1);
+  }
+  return d.vec4f(v.x / length, v.y / length, v.z / length, 1);
+}
```
aleksanderkatan: This function is the same as TGSL function `normalizeSafety` from `helpers.ts`, right? Can that function be used instead of this one, or is this an optimization?

aleksanderkatan (commented):
The example looks amazing!

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/dataTypes.ts
reczkok: It will work because in this case they are functionally identical. They will get translated to identical attributes. The unstruct used to have one more field which made a difference. I don't see any advantage in using one over the other (unless we want to get rid of the Compute casting and packing/unpacking but since this is an example I think it's useful to showcase more features). I'm also curious that it worked in the compute shader 🤔 Did you use the changed buffer inside the compute shader or did you leave the casting as is. I think it will not work if you use a buffer with f16 (or any vech) inside a compute shader without an explicit feature flag.

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/dataTypes.ts
reczkok: (Replied at the end of previous comment)

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/icosphere.ts
reczkok: I actually forgot about our dual functions 😅 It 100% can be

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/dataTypes.ts
aleksanderkatan: All I did was replace this with 
```typescript
export const Vertex = d.struct({
  position: d.vec4h,
  normal: d.vec4h,
});
```
There are type errors, but  there are no WGSL errors and the example seems to be working. I'm a little lost in the code, it isn't obvious for me that the `unstruct` here was related to the packing.
I thought that getting rid of the packing it is an advantage (shorter, simpler code), but if it's here to showcase the feature, could you add comments explaining this situation above the `Vertex` and `ComputeVertex` definition?

reczkok: It's not about the struct or unstruct - in the context of a vertex buffer in this particular case both of them will produce identical results (the vec4h is transformed into float16x4 internally and there is no padding with two vec4h). Packing is related to the fact that we cannot have f16 inside compute shader without the flag I mentioned. So we can have struct/unstruct here interchangeably - I just prefer unstruct and explicit vertex formats as it signals the intent clearly. If we removed the packing our data would be twice as big and we have a lot of vertices at higher subdivisions.

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.html
reczkok: I have no strong opinion tbh. I'm open to changing it but it does seem unnecessary (and maybe a little misleading even) since it actually moves the camera - which is not really zooming too now that I think about it 😅

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -90,6 +92,42 @@ function SliderRow({
...
+  onChange,
+}: {
+  label: string;
+  initial?: T;
+  min: T;
+  max: T;
+  step: T;
+  onChange: (value: T) => void;
+}) {
+  const [value, setValue] = useState<T>(initial || min);
```
mhawryluk:
```suggestion
  const [value, setValue] = useState<T>(initial ?? min);
```

apps/typegpu-docs/src/components/design/VectorSlider.tsx
```diff
@@ -0,0 +1,89 @@
...
+  max,
+  step,
+  value,
+  onChange,
+}: Props<T>) {
+  const components: Array<'x' | 'y' | 'z' | 'w'> = [];
+  if ('x' in min) components.push('x');
+  if ('y' in min) components.push('y');
+  if ('z' in min) components.push('z');
+  if ('w' in min) components.push('w');
```
mhawryluk:
```suggestion
  const components = (['x', 'y', 'z', 'w'] as const).slice(0, min.length);
```

apps/typegpu-docs/src/components/design/VectorSlider.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const getComponentValue = (vec: T, comp: 'x' | 'y' | 'z' | 'w'): number =>
```
mhawryluk: please create type alias for component type

apps/typegpu-docs/src/components/design/VectorSlider.tsx
```diff
@@ -0,0 +1,89 @@
...
+    comp in vec
+      ? ((vec as unknown as Record<'x' | 'y' | 'z' | 'w', number | undefined>)[
```
mhawryluk:
```suggestion
      ? ((vec as Record<'x' | 'y' | 'z' | 'w', number | undefined>)[
```

apps/typegpu-docs/src/components/design/VectorSlider.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+  const handleComponentChange = (
+    componentName: 'x' | 'y' | 'z' | 'w',
+    newValue: number,
+  ) => {
+    const newVec: Partial<d.v4f> = {};
+    if ('x' in value) newVec.x = value.x;
+    if ('y' in value) newVec.y = value.y;
+    if ('z' in value) newVec.z = value.z;
+    if ('w' in value) newVec.w = value.w;
```
mhawryluk: my instant idea was doing `const newVec = { ...value };`, but it doesn't really work the first time, because of how much we lie with types. imo this is hanging by a thread, but mostly because of our complicated vector implementation

apps/typegpu-docs/src/components/design/VectorSlider.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+    if (componentName in min) {
+      newVec[componentName] = newValue;
+    }
+
+    onChange(newVec as T);
```
mhawryluk: like there's no `kind` in newVec. this is another lie

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,510 @@
...
+      cubemapTexture = newCubemapTexture;
+    },
+  },
+  'ambient color': {
+    initial: materialProps.ambient,
+    min: d.vec3f(0, 0, 0),
+    max: d.vec3f(1, 1, 1),
+    step: d.vec3f(0.01, 0.01, 0.01),
+    onVectorSliderChange: (value: d.v3f) => {
+      console.log('ambient', value);
```
mhawryluk:
```suggestion
```

apps/typegpu-docs/src/components/design/VectorSlider.tsx
mhawryluk: maybe a new `value.copy()` api could help here

packages/typegpu/src/std/packing.ts
```diff
@@ -0,0 +1,104 @@
...
+      dataType: vec2f,
+    };
+  },
+);
+
+/**
+ * @privateRemarks
+ * https://gpuweb.github.io/gpuweb/wgsl/#pack2x16float-builtin
+ Converts two floating point values to half-precision floating point numbers, and then combines them into one u32 value.
+ Component e[i] of the input is converted to a IEEE-754 binary16 value, which is then placed in bits 16 × i through 16 × i + 15 of the result. See § 15.7.6 Floating Point Conversion.
```
mhawryluk:
```suggestion
 * Converts two floating point values to half-precision floating point numbers, and then combines them into one u32 value.
 * Component e[i] of the input is converted to a IEEE-754 binary16 value, which is then placed in bits 16 × i through 16 × i + 15 of the result. See § 15.7.6 Floating Point Conversion.
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,521 @@
...
+import {
+  Camera,
+  CubeVertex,
+  DirectionalLight,
+  Material,
+  Vertex,
+} from './dataTypes';
+import { IcosphereGenerator } from './icosphere';
+
+// Initialization
```
mhawryluk, line 15:
```suggestion
// Initialization

```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,521 @@
...
+  .$usage('uniform');
+
+const materialBuffer = root
+  .createBuffer(Material, {
+    ambient: materialProps.ambient,
+    diffuse: materialProps.diffuse,
+    specular: materialProps.specular,
+    shininess: materialProps.shininess,
+    reflectivity: materialProps.reflectivity,
+  })
```
mhawryluk:
```suggestion
  .createBuffer(Material, materialProps)
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,521 @@
...
+  },
+})((input) => {
+  const worldPos = input.position;
+  const pos = std.mul(camera.value.view, input.position);
+  return {
+    pos: std.mul(camera.value.projection, pos),
+    normal: input.normal,
+    worldPos: worldPos,
+  };
+});
```
mhawryluk:
```suggestion
})((input) => ({
  pos: std.mul(
    camera.value.projection,
    std.mul(camera.value.view, input.position),
  ),
  normal: input.normal,
  worldPos: input.position,
}));
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,521 @@
...
+    texCoord: d.vec3f,
+  },
+})((input) => {
+  const viewRotationMatrix = d.mat4x4f(
+    camera.value.view.columns[0],
+    camera.value.view.columns[1],
+    camera.value.view.columns[2],
+    d.vec4f(0, 0, 0, 1),
+  );
+  const pos = std.mul(viewRotationMatrix, input.position);
```
mhawryluk: unnecessary constant

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/index.ts
```diff
@@ -0,0 +1,521 @@
...
+
+const cubeFragmentFn = tgpu['~unstable'].fragmentFn({
+  in: {
+    texCoord: d.vec3f,
+  },
+  out: d.vec4f,
+})((input) => {
+  const dir = std.normalize(input.texCoord);
+  return std.textureSample(cubemapBinding, texSampler, dir);
+});
```
mhawryluk: also could be one-lined

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/icosphere.ts
```diff
@@ -0,0 +1,313 @@
...
+      const edge1 = d.vec4f(v2.x - v1.x, v2.y - v1.y, v2.z - v1.z, 0);
+      const edge2 = d.vec4f(v3.x - v1.x, v3.y - v1.y, v3.z - v1.z, 0);
+      const faceNormal = helpers.normalizeSafely({
+        v: d.vec4f(
+          edge1.y * edge2.z - edge1.z * edge2.y,
+          edge1.z * edge2.x - edge1.x * edge2.z,
+          edge1.x * edge2.y - edge1.y * edge2.x,
+          0,
+        ),
+      });
```
mhawryluk: use `std.cross`

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/icosphere.ts
```diff
@@ -0,0 +1,313 @@
...
+      smoothFlag: this.smoothBuffer,
+    });
+
+    const triangleCount = getVertexAmount(wantedSubdivisions - 1) / 3;
+    const xGroups = Math.min(triangleCount, 65535);
+    const yGroups = Math.ceil(triangleCount / 65535);
+
+    this.pipeline
+      .with(generatorLayout, bindGroup)
+      .dispatchWorkgroups(xGroups, yGroups, 1);
```
mhawryluk: workgroup size is set to [64, 1, 1], so aren't we invoking 64x too many threads?

apps/typegpu-docs/src/components/design/VectorSlider.tsx
```diff
@@ -0,0 +1,56 @@
...
+          <RadixSlider.Track className="flex-1 h-full">
+            <RadixSlider.Range className="absolute h-full bg-gradient-to-br from-gradient-purple to-gradient-blue" />
+          </RadixSlider.Track>
+          <div className="absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2 text-xs text-center">
+            {currentValue.toFixed(2)}
+          </div>
+        </RadixSlider.Root>
+      </div>
+    );
+  };
```
mhawryluk:
```suggestion
  const renderSlider = (index: number) => (
    <div key={index} className="flex items-center flex-1">
      <RadixSlider.Root
        value={[value[index]]}
        min={min[index]}
        max={max[index]}
        step={step[index]}
        onValueChange={(values) => handleComponentChange(index, values[0])}
        className="bg-grayscale-20 h-10 rounded-[0.25rem] relative flex overflow-hidden flex-1"
      >
        <RadixSlider.Track className="flex-1 h-full">
          <RadixSlider.Range className="absolute h-full bg-gradient-to-br from-gradient-purple to-gradient-blue" />
        </RadixSlider.Track>
        <div className="absolute left-1/2 top-1/2 -translate-x-1/2 -translate-y-1/2 text-xs text-center">
          {value[index].toFixed(2)}
        </div>
      </RadixSlider.Root>
    </div>
  );
```

mhawryluk (approved):
Looks awesome! 

There's an error when switching to another example:
Zrzut ekranu 2025-04-7 o 13 02 43

reczkok:
> There's an error when switching to another example: Zrzut ekranu 2025-04-7 o 13 02 43

Should be fixed 🛠️

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as std from 'typegpu/std';
+
+export const unpackVec2u = tgpu['~unstable'].fn(
+  { packed: d.vec2u },
+  d.vec4f,
+)(({ packed }) => {
+  const xy = std.unpack2x16float(packed.x);
+  const zw = std.unpack2x16float(packed.y);
+  return d.vec4f(xy, zw);
+});
```
iwoplaza: Since we're keeping both versions, we could keep these in the "old" API 🤔. And the same for pack(...).

```suggestion
export const unpackVec2u = tgpu['~unstable'].fn(
  [d.vec2u],
  d.vec4f,
)((packed) => {
  const xy = std.unpack2x16float(packed.x);
  const zw = std.unpack2x16float(packed.y);
  return d.vec4f(xy, zw);
});
```

apps/typegpu-docs/src/content/examples/rendering/cubemap-reflection/helpers.ts
```diff
@@ -0,0 +1,47 @@
...
+  const edge2 = std.sub(v3.xyz, v1.xyz);
+  return std.normalize(d.vec4f(std.cross(edge1, edge2), 0));
+});
+
+export const calculateMidpoint = tgpu['~unstable'].fn(
+  { v1: d.vec4f, v2: d.vec4f },
+  d.vec4f,
+)(({ v1, v2 }) => {
+  return d.vec4f(std.mul(0.5, std.add(v1.xyz, v2.xyz)), 1);
+});
```
iwoplaza: Could be array args as well, wdyt?

packages/typegpu/src/data/unstruct.ts
```diff
@@ -27,33 +26,32 @@ import type { BaseData } from './wgslTypes.js';
  */
 export function unstruct<TProps extends Record<string, BaseData>>(
   properties: TProps,
-): Unstruct<TProps> {
-  return new UnstructImpl(properties as TProps);
+): Unstruct<Prettify<TProps>> {
```
iwoplaza: Prettify here, but not on the Unstruct type. If we complete it there, we can remove it from here.

packages/typegpu/src/data/dataTypes.ts
```diff
@@ -46,13 +48,19 @@ export interface Disarray<TElement extends wgsl.BaseData = wgsl.BaseData> {
 export interface Unstruct<
   TProps extends Record<string, wgsl.BaseData> = Record<string, wgsl.BaseData>,
 > extends TgpuNamable {
+  (props: Prettify<InferRecord<TProps>>): Prettify<InferRecord<TProps>>;
   readonly label?: string | undefined;
   readonly type: 'unstruct';
   readonly propTypes: TProps;
   readonly [$repr]: Prettify<InferRecord<TProps>>;
+  readonly '~gpuRepr': InferGPURecord<TProps>;
+  readonly '~memIdent': Unstruct<MemIdentityRecord<TProps>>;
```
iwoplaza:
```suggestion
  readonly '~gpuRepr': Prettify<InferGPURecord<TProps>>;
  readonly '~memIdent': Unstruct<Prettify<MemIdentityRecord<TProps>>>;
```

iwoplaza (approved):
REFLECT 🪩🪩🪩

reczkok:
I removed `Prettify` from `~memIdent` in structs and unstructs which results in a little less readable `copyFrom` signature:
image
But if I leave it it totally wrecks our `bindGroupLayout` types.
The `copyFrom` looks a little nicer:
image
But I get the following error:
image
Which I have no clue how to approach at the moment

PR #1018, opened 2025-03-13
feat: Add `std.arrayLength`, `std.distance`, handle Update Expression in TGSL, rewrite Boids(next)

reczkok:


packages/typegpu/src/std/numeric.ts
```diff
@@ -394,3 +394,10 @@ export const reflect = createDualImpl(
     };
   },
 );
+
+export const distance = createDualImpl(
+  // CPU implementation
+  <T extends vBase>(a: T, b: T): number => length(sub(a, b)),
+  // GPU implementation
+  (a, b) => ({ value: `distance(${a.value}, ${b.value})`, dataType: f32 }),
+);
```
aleksanderkatan: The [wgsl distance function](https://www.w3.org/TR/WGSL/#distance-builtin) can also accept numbers, I guess we could as well (`std.length` already supports numbers, `std.sub` does not yet)

aleksanderkatan (approved):
LGTM!

packages/typegpu/src/std/numeric.ts
reczkok: Oh yeah, I misread the spec - will add

PR #1019, opened 2025-03-14
feat: Implement `eq`

### LINKED ISSUE #961: feat: Implement `eq` std function for comparing vectors and matrices

aleksanderkatan:
Closes #961

### DIFF
--- a/packages/typegpu/src/std/index.ts
+++ b/packages/typegpu/src/std/index.ts
@@ -1,6 +1,7 @@
 export { discard } from './discard.js';
 export {
   // ops
+  eq,
   add,
   sub,
   mul,
--- a/packages/typegpu/src/std/numeric.ts
+++ b/packages/typegpu/src/std/numeric.ts
@@ -31,6 +31,16 @@ function isNumeric(element: Resource) {
 
 type vBase = { kind: VecKind };
 
+export const eq = createDualImpl(
+  // CPU implementation
+  <T extends vBase | number>(lhs: T, rhs: T): boolean => lhs === rhs,
+  // GPU implementation
+  (lhs, rhs) => ({
+    value: `(${lhs.value} == ${rhs.value})`,
+    dataType: bool,
+  }),
+);
+
 export const add = createDualImpl(
   // CPU implementation
   <T extends vBase>(lhs: T, rhs: T): T => VectorOps.add[lhs.kind](lhs, rhs),
@@ -436,7 +446,7 @@ export const isCloseTo = createDualImpl(
   (e1, e2, precision = { value: 0.01, dataType: f32 }) => {
     if (isNumeric(e1) && isNumeric(e2)) {
       return {
-        value: `abs(f32(${e1.value})-f32(${e2.value})) <= ${precision.value}`,
+        value: `(abs(f32(${e1.value})-f32(${e2.value})) <= ${precision.value})`,
         dataType: bool,
       };
     }


aleksanderkatan:
Sorry, I should have split this into two PRs.

Changes: 
- add `vecNb` vectors corresponding to `vecN<bool>` vectors in WGSL,
- split vector type categorisation into float, integer and boolean vectors; replace the general type with appropriate one in numeric functions,
- add boolean functions corresponding to operators: `eq`, `neq`, `lessThan`, `lessThanOrEqual`, `greaterThan`, `greaterThanOrEqual`, `not`, `or` (corresponds to `|`, not `||`), `and` (corresponds to `&`, not `&&`),
- add boolean functions corresponding to WGSL std functions: `all`, `any`, `select`,
- add `allEq` function corresponding to (`all` ∘ `eq`),
- add a new example containing tests for boolean functions (probably to be deleted before merging).

You can access `vecN<bool>` elements by indexing, you can swizzle it, you can use it in structs, you can pass it in TGSL functions (directly or in a struct).
Attempting to use `vecN<bool>` in uniform or storage buffers throws at WGSL level, attempting to do this with vertex buffers throws at JS level (behavior similar to that of `bool`).

aleksanderkatan:
In file `dataIO.ts`, should `vecN<bool>` functions throw? If so, should `bool` function throw as well?

apps/typegpu-docs/src/content/examples/tests/tests/boolean.ts
```diff
@@ -0,0 +1,151 @@
+import tgpu from 'typegpu';
+import {
```
mhawryluk: change these imports to `* as d` and `* as std`

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+ * Equivalent to `all(eq(lhs, rhs))`.
+ * @example
+ * allEq(vec2f(0.0, 1.0), vec2f(0.0, 2.0)) // returns false
+ * allEq(vec3u(0, 1, 2), vec3u(0, 1, 2)) // returns true
+ */
+export const allEq = createDualImpl(
+  // CPU implementation
+  <T extends AnyVecInstance>(lhs: T, rhs: T) => {
+    return all(VectorOps.eq[lhs.kind](lhs, rhs));
+  },
```
mhawryluk:
```suggestion
  <T extends AnyVecInstance>(lhs: T, rhs: T) =>
    all(VectorOps.eq[lhs.kind](lhs, rhs)),
```

mhawryluk: same everywhere below

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -106,6 +109,95 @@ const binaryComponentWise4u = (op: BinaryOp) => (a: wgsl.v4u, b: wgsl.v4u) =>
...
+    vec4i: (e1: wgsl.v4i, e2: wgsl.v4i) =>
+      vec4b(e1.x === e2.x, e1.y === e2.y, e1.z === e2.z, e1.w === e2.w),
+    vec4u: (e1: wgsl.v4u, e2: wgsl.v4u) =>
+      vec4b(e1.x === e2.x, e1.y === e2.y, e1.z === e2.z, e1.w === e2.w),
+    'vec4<bool>': (e1: wgsl.v4b, e2: wgsl.v4b) =>
+      vec4b(e1.x === e2.x, e1.y === e2.y, e1.z === e2.z, e1.w === e2.w),
+  } as Record<
+    VecKind,
+    <T extends wgsl.AnyVecInstance>(e1: T, e2: T) => wgsl.AnyBooleanVecInstance
+  >,
```
mhawryluk:
```suggestion
  } as Record<
    VecKind,
    <T extends wgsl.AnyVecInstance>(
      e1: T,
      e2: T,
    ) => T extends wgsl.AnyVec2Instance
      ? wgsl.v2b
      : T extends wgsl.AnyVec3Instance
        ? wgsl.v3b
        : wgsl.v4b
  >,
```
what do you think about this? I'd prefer that, cause then we don't have to cast to `AnyToBooleanComponentWise` in `boolean.ts`. and I think the type is still quite readable. same for the other ones like neq and lessThan...

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+  // GPU implementation
+  (value) => ({
+    value: `!(${value.value})`,
+    dataType: value.dataType,
+  }),
+);
+
+/**
+ * Returns **component-wise** `lhs | rhs`.
```
mhawryluk:
```suggestion
 * Returns **component-wise** logical `or` result.
```

`lhs | rhs` is true for wgsl, but in JS it's something different. same for `and`

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+  (lhs, rhs) => ({
+    value: `(${lhs.value} | ${rhs.value})`,
+    dataType: lhs.dataType,
+  }),
+);
+
+/**
+ * Returns **component-wise** `lhs & rhs`.
+ * @example
+ * and(vec2b(false, true), vec2b(true, true)) // returns vec2b(true, false)
```
mhawryluk:
```suggestion
 * and(vec2b(false, true), vec2b(true, true)) // returns vec2b(false, true)
```

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+  (value) => ({
+    value: `!(${value.value})`,
+    dataType: value.dataType,
+  }),
+);
+
+/**
+ * Returns **component-wise** `lhs | rhs`.
+ * @example
+ * or(vec2b(false, true), vec2b(false, false)) // returns vec2b(true, false)
```
mhawryluk:
```suggestion
 * or(vec2b(false, true), vec2b(false, false)) // returns vec2b(false, true)
```

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+  (value) => ({
+    value: `any(${value.value})`,
+    dataType: bool,
+  }),
+);
+
+// other
+
+/**
+ * Checks whether the given elements differ by at most 0.01.
```
mhawryluk:
```suggestion
 * Checks whether the given elements differ by at most the `precision` value.
```

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+    if (typeof lhs !== 'number' && typeof rhs !== 'number') {
+      return VectorOps.isCloseToZero[lhs.kind](sub(lhs, rhs), precision);
+    }
+    return false;
+  },
+  // GPU implementation
+  (lhs, rhs, precision = { value: 0.01, dataType: f32 }) => {
+    if (isNumeric(lhs) && isNumeric(rhs)) {
+      return {
+        value: `(abs(f32(${lhs.value})-f32(${rhs.value})) <= ${precision.value})`,
```
mhawryluk:
```suggestion
        value: `(abs(f32(${lhs.value}) - f32(${rhs.value})) <= ${precision.value})`,
```

packages/typegpu/src/std/boolean.ts
```diff
@@ -0,0 +1,371 @@
...
+        dataType: bool,
+      };
+    }
+    if (!isNumeric(lhs) && !isNumeric(rhs)) {
+      return {
+        // https://www.w3.org/TR/WGSL/#vector-multi-component:~:text=Binary%20arithmetic%20expressions%20with%20mixed%20scalar%20and%20vector%20operands
+        // (a-a)+prec creates a vector of a.length elements, all equal to prec
+        value: `all(abs(${lhs.value}-${rhs.value}) <= (${lhs.value} - ${lhs.value})+${precision.value})`,
```
mhawryluk:
```suggestion
        value: `all(abs(${lhs.value} - ${rhs.value}) <= (${lhs.value} - ${lhs.value}) + ${precision.value})`,
```

packages/typegpu/tests/std/boolean/all.test.ts
```diff
@@ -0,0 +1,26 @@
+import { describe, expect, it } from 'vitest';
+import { vec2b, vec3b, vec4b } from '../../../src/data';
+import { all } from '../../../src/std/boolean';
+
+describe('all', () => {
+  it('calculates for 2 element vectors', () => {
+    expect(all(vec2b(false, false))).toBeFalsy();
```
mhawryluk: if `all` returned e.g. `undefined`, this assertion would pass. I'd prefer `.toBe(false)` instead of `.toBeFalsy()` and `.toBe(true)` instead of `.toBeTruthy()`. same everywhere

packages/typegpu/src/data/vectorOps.ts
aleksanderkatan: Yeah, I think that is way better. I still left the `SelectOverload`, but there is no function casting anymore.

mhawryluk (approved):
that's huge, good job! 🌯

apps/typegpu-docs/src/content/examples/tests/tests/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+// #region Example controls and cleanup
+
+export const controls = {
+  'Run tests': {
+    onButtonClick: async () => {
+      const result = await runTests();
+      const table = document.querySelector('.result') as HTMLDivElement;
+      table.innerText = result === 0 ? 'Tests failed.' : 'Tests succeeded!';
+    },
```
iwoplaza: Tint nit
```suggestion
    async onButtonClick() {
      const table = document.querySelector<HTMLDivElement>('.result');
      if (!table) {
        throw new Error('Nowhere to display the results');
      }
      table.innerText = await runTests() ? 'Tests succeeded!' : 'Tests failed.';
    },
```

apps/typegpu-docs/src/content/examples/tests/tests/index.ts
```diff
@@ -0,0 +1,48 @@
+import tgpu from 'typegpu';
+import * as d from 'typegpu/data';
+import { booleanTests } from './boolean';
+
+const root = await tgpu.init();
+
+const resultBuffer = root.createBuffer(d.i32, 0).$usage('storage');
+const result = resultBuffer.as('mutable');
```
iwoplaza: We can use the unstable `createMutable` API to create `result`, as it's only used as mutable storage.
```suggestion
const root = await tgpu.init();
const result = root['~unstable'].createMutable(d.i32, 0);
```

apps/typegpu-docs/src/content/examples/tests/tests/index.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const computeRunTests = tgpu['~unstable']
+  .computeFn({ in: { num: d.builtin.numWorkgroups }, workgroupSize: [1] })
```
iwoplaza: No input seems to be used in the function.
```suggestion
  .computeFn({ workgroupSize: [1] })
```

apps/typegpu-docs/src/content/examples/tests/tests/index.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  });
+
+const pipeline = root['~unstable']
+  .withCompute(computeRunTests)
+  .createPipeline();
+
+async function runTests() {
+  pipeline.dispatchWorkgroups(1);
+  return await resultBuffer.read();
```
iwoplaza: To adapt with the 'createMutable' changes.

```suggestion
  return await resultBuffer.buffer.read();
```

iwoplaza (changes requested):

packages/typegpu/src/data/alignmentOf.ts
iwoplaza, line 1: Technically, since boolean vectors are not host-shareable, they don't need alignment.

apps/typegpu-docs/src/content/examples/tests/tests/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Tests",
```
iwoplaza: Could we find a more descriptive name for this example?
Maybe something like: "Logical Expressions"?

packages/typegpu/src/data/sizeOf.ts
iwoplaza, line 1: Same as alignment

packages/typegpu/tests/struct.test.ts
```diff
@@ -226,6 +229,17 @@ describe('struct', () => {
...
 
+  it('supports boolean vectors', () => {
+    const TestStruct = struct({
+      a: vec2b,
+      b: vec3b,
+      c: vec4b,
+    });
+
+    expect(sizeOf(TestStruct)).toEqual(48);
+    expect(alignmentOf(TestStruct)).toEqual(16);
```
iwoplaza: No need, I don't think 👀
@reczkok @mhawryluk  Please verify 🙏

```suggestion
```

packages/typegpu/tests/vector.test.ts
```diff
@@ -127,6 +127,93 @@ describe('vec2i', () => {
...
 });
 
+describe('vec2<bool>', () => {
+  it('should span 8 bytes', () => {
+    expect(sizeOf(d.vec2b)).toEqual(8);
+  });
+
+  it('should align to 8 bytes', () => {
+    expect(d.alignmentOf(d.vec2b)).toEqual(8);
+  });
```
iwoplaza:
```suggestion
```

apps/typegpu-docs/src/content/examples/tests/tests/meta.json
aleksanderkatan: I thought that this could be an example for all TGSL parsing, is "TGSL parsing tests" alright?

packages/typegpu/tests/vector.test.ts
```diff
@@ -257,6 +344,101 @@ describe('vec3i', () => {
...
 
+describe('vec3<bool>', () => {
+  it('should span 12 bytes', () => {
+    expect(sizeOf(d.vec3b)).toEqual(12);
+  });
+
+  it('should align to 16 bytes', () => {
+    expect(d.alignmentOf(d.vec3b)).toEqual(16);
+  });
+
```
iwoplaza:
```suggestion
```

iwoplaza:
I'm wondering if we should do `std.greaterThanOrEqual` or `std.gte` 🤔. So: `std.eq`, `std.lt`, `std.gt`, `std.lte`, `std.gte`

iwoplaza:
@mhawryluk @reczkok ^?

mhawryluk:
good idea. I prefer the acronyms :D

reczkok (approved):
Great work! ⨈

iwoplaza (approved):
Uuuuuuge 🦾

PR #1024, opened 2025-03-17
dx: Don't run pkg-pr workflow on PRs from forks

### LINKED ISSUE #1009: Don't run pkg-pr workflow on PRs from forks
these fail because of the lack of permissions

mhawryluk:
closes #1009


iwoplaza (approved):
🫡

PR #1045, opened 2025-03-19
feat: Replace `does` method with shell invocation

### LINKED ISSUE #1031: Implement the new does-less tgpu.fn API

mhawryluk:
closes #1031 

1st PR related to the new function API, more to come

to do in PRs to follow:
* ~~"kernel" directive~~
* "kernel & js" directive
* ~~passing parameters to shell and call in objects~~
* ~~passing function body without header into wgsl implemented functions~~


apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -19,14 +19,13 @@ These functions can reference outside resources, like other user-defined or help
 ## Creating a function
 
 Functions are constructed by first defining their shells, which specify their inputs and outputs.
-Then the actual WGSL implementation is passed in through the `does` method.
+Then the actual WGSL implementation is passed in as an argument to shell invocation.
```
iwoplaza:
```suggestion
Then the actual WGSL implementation is passed in as an argument to a shell invocation.
```

reczkok (approved):
Looks good 🏗️ I would restore the JSdoc on exported types

PR #1062, opened 2025-03-24
fix: Function call member access type inference

### LINKED ISSUE #1059: fix: Allow member access on FnCall return type and fix member access inference.

reczkok:
closes #1059


packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -195,6 +195,20 @@ export function generateExpression(
     // biome-ignore lint/suspicious/noExplicitAny: <sorry TypeScript>
     const propValue = (target.value as any)[property];
 
+    if (isWgsl(target.dataType)) {
```
iwoplaza:
```suggestion
    if (target.dataType !== UnknownData) {
```

iwoplaza (changes requested):

packages/typegpu/src/data/matrix.ts
```diff
@@ -52,7 +53,7 @@ function createMatSchema<
     '~repr': undefined as unknown as ValueType,
     type: options.type,
     label: options.type,
-  };
+  } as unknown as AnyWgslData;
```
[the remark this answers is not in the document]
iwoplaza: These are library internals, so type casting is useful here.

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -273,9 +291,14 @@ export function generateExpression(
...
     // Assuming that `id` is callable
-    return (idValue as unknown as (...args: unknown[]) => unknown)(
+    const fnRes = (idValue as unknown as (...args: unknown[]) => unknown)(
       ...resolvedResources,
     ) as Resource;
+
+    return {
+      value: resolveRes(ctx, fnRes),
+      dataType: fnRes.dataType,
+    };
```
iwoplaza, line 327: Let's not resolve the fn call here, in case it's not a TGPU function we're calling. We can branch here to see if we're calling a TGPU fn, and if not, we throw an error.

The only JS functions we want to allow are our Dual Impl functions. Could we mark them with `[$internal]: {}` props to see if they originate from us?

iwoplaza (changes requested):

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -189,16 +190,32 @@ export function generateExpression(
     if (typeof target.value === 'string') {
       return {
         value: `${target.value}.${property}`,
-        dataType: getTypeForPropAccess(target.dataType as Wgsl, property),
+        dataType: isWgsl(target.dataType)
```
iwoplaza: Since `LooseData` don't fall under the umbrella of Wgsl, we don't get type inference for them here. We can use `isData` instead to both reduce the amount of unnecessary checks, and to increase the scope.
```suggestion
        dataType: isData(target.dataType)
```

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -189,16 +190,32 @@ export function generateExpression(
...
+          ? getTypeForPropAccess(target.dataType, property)
+          : UnknownData,
       };
     }
     // biome-ignore lint/suspicious/noExplicitAny: <sorry TypeScript>
     const propValue = (target.value as any)[property];
 
+    if (target.dataType.type !== 'unknown') {
+      if (target.dataType.type.startsWith('mat') && property === 'columns') {
```
iwoplaza: Could we create an `isMatSchema` function for this? Will be easier to refactor once we transition to symbols.

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -226,7 +243,9 @@ export function generateExpression(
 
     return {
       value: `${targetStr}[${propertyStr}]`,
-      dataType: getTypeForIndexAccess(targetExpr.dataType as d.AnyWgslData),
+      dataType: isWgsl(targetExpr.dataType)
```
iwoplaza:
```suggestion
      dataType: isData(targetExpr.dataType)
```

packages/typegpu/src/smol/wgslGenerator.ts
```diff
@@ -273,9 +292,18 @@ export function generateExpression(
...
 
     // Assuming that `id` is callable
-    return (idValue as unknown as (...args: unknown[]) => unknown)(
+    const fnRes = (idValue as unknown as (...args: unknown[]) => unknown)(
       ...resolvedResources,
     ) as Resource;
+
+    if (!isMarkedInternal(idValue)) {
+      throw new Error(`Function ${String(idValue)} is not marked as internal.`);
+    }
```
iwoplaza: Also, I feel like we should check this earlier, before calling the function.

```suggestion
    if (!isMarkedInternal(idValue)) {
      throw new Error(`Function ${String(idValue)} has not been created using TypeGPU APIs. Did you mean to wrap the function with tgpu.fn(args, return)(...) ?`);
    }
```

packages/typegpu/src/smol/wgslGenerator.ts
iwoplaza: Still yet to be addressed 👀

packages/typegpu/src/types.ts
```diff
@@ -234,3 +228,13 @@ export function hasInternalDataType(
...
 }
+
+export function isMarkedInternal(
+  value: unknown,
+): value is { [$internal]: Record<string, unknown> } {
+  return (
+    !!value &&
+    (typeof value === 'object' || typeof value === 'function') &&
+    $internal in (value as object)
+  );
```
iwoplaza: I believe this should be enough 👍
```suggestion
  return !!(value as { [$internal]: Record<string, unknown> })?.[$internal];
```

iwoplaza (changes requested):

packages/typegpu/src/smol/wgslGenerator.ts
reczkok: But how exactly do you want to handle this? We didn't want to resolve in case it's not a Tgpu function. Do we not want to resolve at all? What about FnCall

reczkok: Do we want to check for `FnCall` specifically or change the index/member access logic?

packages/typegpu/src/types.ts
reczkok: Not for TS
```ts
// biome-ignore lint/suspicious/noExplicitAny: <they told me to>
return !!(value as any)?.[$internal];
```
or
```ts
return !!(
  value as {
    [$internal]: Record<string, unknown>;
  }
)?.[$internal];
```

packages/typegpu/src/smol/wgslGenerator.ts
iwoplaza: Oh wait... right, we disabled non-tgpu functions, so objects returned from functions will always be resolvable. Let's keep it as it is for now, until we want to implement compile-time functions.

packages/typegpu/src/types.ts
iwoplaza: Right! Updated my suggestion 🙌. (should be a one-liner, biome does not fold it for me)

iwoplaza (approved):
Let's gooo 🎉

PR #1067, opened 2025-03-24
chore: Verify that the tag used for publishing matches the version number

iwoplaza:


packages/tgpu-dev-cli/verify-publish-tag.mjs
```diff
@@ -0,0 +1,33 @@
...
+      }
+
+      tagVerified = true;
+      break;
+    }
+  }
+
+  if (!tagVerified) {
+    throw new Error(
+      `Publishing under a mismatched tag "${chosenPublishTag}" for version ${packageJSON.version}.`,
```
mhawryluk, line 33: If I'm not mistaken if there's no tag provided (it's undefined) then this will always throw. and e.g. for `typegpu` package (currently not alpha) we want there to be no tag. so I think the logic for undefined needs to be changed, we cannot look for "undefined" in version number

packages/tgpu-dev-cli/prepack.mjs
```diff
@@ -131,6 +133,12 @@ async function main() {
   consola.start('Preparing the package for publishing');
   console.log('');
 
+  const args = arg({ '--skip-publish-tag-check': Boolean });
```
mhawryluk, line 136: this option should be utilized in the `nightly-build` script

mhawryluk (approved):

reczkok (approved):
Looks good 🛠️

PR #1079, opened 2025-03-25
feat: Added std.identity4x4 & std.translate4x4

lursz:
std.identity4x4 generates a 4 by 4 identity matrix.


reczkok:
I'm not quite sure I would expect identity to be an std function. My first thought would be something like:
```ts
const foo = tgpu.fn([]).does(() => {
  const x = d.mat4x4f.identity()
})
```
It would also be more familiar for `wgpu-matrix` users
```ts
const foo = m.mat4.identity();
```

PR #1089, opened 2025-03-27
impr: Make the compiledIO benchmarks fairer

reczkok:


reczkok:
Very stale and fails to fix much

PR #1090, opened 2025-03-27
chore: Discriminated union benchmark

iwoplaza:


apps/infra-benchmarks/src/discriminated-union.ts
```diff
@@ -1,22 +1,189 @@
...
+        if (obj[$internal].type === 'aaa') {
+          count += 2;
+        } else if (obj[$internal].type === 'bbb') {
+          count += 3;
+        } else if (obj[$internal].type === 'ccc') {
+          count += 4;
+        }
+      }
+
+      console.log(count);
```
mhawryluk: let's remove all of the console.logs, please

mhawryluk (approved):
🏃‍♂️‍➡️🏃‍♂️‍➡️🏃‍♂️‍➡️

apps/infra-benchmarks/src/discriminated-union.ts
iwoplaza: I wanted to keep the console logs to ensure that the count variable wouldn't be optimised away, since it would not be used anywhere.

mhawryluk: sure, but it affects the result in that we count in the IO operation, which is slow and unpredictable

mhawryluk: from what I see, with logging the results are 2x slower, but their respective placements seem to be unaffected. so I guess it doesn't matter either way

iwoplaza: I see your point 👍. I'll remove it 🫡

PR #1094, opened 2025-03-28
Autogenerate function header for wgsl-implemented functions

### LINKED ISSUE #1093: Autogenerate function header for wgsl-implemented functions

### LINKED ISSUE #1099: docs: Update functions doc with the new shell API

mhawryluk:
closes #1093 
closes #1099


packages/typegpu/tests/rawFn.test.ts
```diff
@@ -264,26 +264,28 @@ struct fragment_Output {
...
         [d.vec4f, Point],
         undefined,
       )(/* wgsl */ `(a: vec4f, b: Point) {
-    var newPoint: Point;
-    newPoint = b;
-  }`)
+        var newPoint: Point;
+        newPoint = b;
+      }`)
       .$name('newPointF');
```
iwoplaza: I think we should add `$uses()` here, since we're counting on the name of the struct to be `Point` exactly, and that only applies when in `strict` naming scheme.

iwoplaza (changes requested):

packages/typegpu/tests/rawFn.test.ts
mhawryluk: but then we wouldn't be checking if Point struct is added to externals automatically. I will remove the usage in the function block

mhawryluk: and I assume that's a comment to the test with record argTypes below? cause with the approach with array args we don't need the name to be exact

iwoplaza (approved):
Yuhuu! 🎉

PR #1097, opened 2025-03-29
feat: array.length in TGSL

iwoplaza:


reczkok (approved):
Very convenient❣️I just hope once we start supporting some JS specific syntax users won't get confused what they can and can't use. I could see that being frustrating 😐

iwoplaza:
> Very convenient❣️I just hope once we start supporting some JS specific syntax users won't get confused what they can and can't use. I could see that being frustrating 😐

I can see that. It's definitely a balance. Perhaps with good error messaging, we can make these frustrations less... frustrating! We have a lot of context around the validity of code, and what syntax is supported, so we can strictly validate that only a subset of properties are accessible on values of datatype "WgslArray" or "Disarray", and throw a descriptive error otherwise.

PR #1105, opened 2025-04-03
fix: Apply missing Prettify calls

### LINKED ISSUE #1026: fix: Identify missing Prettify calls

reczkok:
I think that is the only one? I thought there were more but could not find them. If someones encounters one we could reopen the ticket
closes #1026


iwoplaza (approved):
‼️‼️‼️

PR #1128, opened 2025-04-08
feat: Allow passing raw wgsl implementation to shell à la template literal

aleksanderkatan:


packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -127,14 +130,39 @@ export function fn<
...
+      typeof implementation !== 'function' &&
+      typeof implementation !== 'string'
+    ) {
+      return createFn(
+        shell,
+        templateLiteralIdentity(implementation, ...values),
+      );
+    }
+    return createFn(shell, implementation);
+  };
```
mhawryluk:
```suggestion
  ) =>
    createFn(
      shell,
      typeof implementation !== 'function' && typeof implementation !== 'string'
        ? templateLiteralIdentity(implementation, ...values)
        : implementation,
    );
```

packages/typegpu/tests/function.test.ts
```diff
@@ -141,6 +140,30 @@ describe('tgpu.fn', () => {
       TgpuFn<[d.F32, d.U32], d.Bool>
     >();
   });
+
+  it('parses template literal without arguments', () => {
```
mhawryluk: please add tests using the new function api too (object args, instead of array)

mhawryluk (changes requested):
please implement the same thing for `vertexFn`, `fragmentFn` and `computeFn`. and update some examples to use this api (I suggest box raytracing)

packages/typegpu/src/core/function/tgpuFn.ts
aleksanderkatan: I extracted the second argument parsing to a function since now it is used in more places

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -19,17 +19,27 @@ These functions can reference outside resources, like other user-defined or help
...
     return color;
   }`)
   .$name('getGradientColor');
+
+// or
+
+const getGradientColor = tgpu['~unstable'].fn({ ratio: d.f32 }, d.vec4f)`{
+    let color = mix(vec4f(0.769, 0.392, 1.0, 1), vec4f(0.114, 0.447, 0.941, 1), ratio);
+    return color;
+  }`.$name('getGradientColor');
```
mhawryluk:
```suggestion
  let color = mix(vec4f(0.769, 0.392, 1.0, 1), vec4f(0.114, 0.447, 0.941, 1), ratio);
  return color;
}`.$name('getGradientColor');
```

packages/typegpu/src/core/function/tgpuComputeFn.ts
```diff
@@ -110,11 +115,14 @@ export function computeFn<
...
-  const call = (implementation: Implementation) =>
+  const call = (
+    arg: Implementation | TemplateStringsArray,
+    ...values: unknown[]
+  ) =>
     createComputeFn(
       shell,
       options.workgroupSize,
-      implementation as Implementation,
+      stripTemplate(arg as Implementation | TemplateStringsArray, ...values),
```
mhawryluk:
```suggestion
      stripTemplate(arg, ...values),
```

and same for all the other functions. I believe arg parameter of the call function can be typed as `Implementation | TemplateStringsArray` in fragment and vertex too, but we should check

apps/typegpu-docs/src/content/examples/simulation/confetti/index.ts
```diff
@@ -90,28 +90,25 @@ const dataLayout = tgpu
...
+const mainVert = tgpu['~unstable'].vertexFn({
+  in: {
+    tilt: d.f32,
+    angle: d.f32,
+    color: d.vec4f,
+    center: d.vec2f,
+    index: d.builtin.vertexIndex,
+  },
+  out: VertexOutput,
+})`/* wgsl */ {
```
mhawryluk: the backtick needs to be after the wgsl comment for the extension to work. + please dedent the implementation by one level. same for mainCompute

packages/typegpu/tests/functionTagged.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      const vertexFn = tgpu['~unstable'].vertexFn({
+        in: {},
+        out: {},
+      })`{}`.$name('vertexFn');
+
+      const actual = parseResolved({ vertexFn });
+
+      const expected = parse(`
+        struct vertexFn_Input {}
+        struct vertexFn_Output {}
```
mhawryluk: not a valid WGSL. once we fix this in a second, these tests will fail. could we please change these entry functions to have non-empty input and output?

apps/typegpu-docs/src/content/examples/simulation/confetti/index.ts
aleksanderkatan: How did this even happen

packages/typegpu/src/core/function/tgpuComputeFn.ts
aleksanderkatan: I think we can replace those argument types with raw `Implementation | TemplateStringsArray`, however I think that our `Implementation` type needs tweaking. I failed to come up with something right now, so I created issue #1147

PR #1131, opened 2025-04-08
fix: Function resolve bug fixes

### LINKED ISSUE #658: Generate argument code for TGSL functions based on the shell, not on the AST.
Currently, resolving the following function will generate a WGSL function with no arguments:
```ts
tgpu
  .fn([d.arrayOf(d.f32, gridSize)])
  .does(() => {
    // do something
  })
  .$name('fill');
```
That's because the arguments are based on what's in the AST, and not what's in the shell.
Arguments omitted from the AST should be given unique but random names, like param_0, param_1, etc...

### LINKED ISSUE #1130: fix: Don't construct args and return type structs for empty objects

mhawryluk:
closes #1130 
closes #658

### DIFF
--- a/packages/typegpu/src/core/function/fnCore.ts
+++ b/packages/typegpu/src/core/function/fnCore.ts
@@ -168,9 +168,12 @@ export function createFnCore(
 
         const args: Resource[] = Array.isArray(shell.argTypes)
           ? ast.argNames.type === 'identifiers'
-            ? ast.argNames.names.map((name, idx) => ({
-                value: name,
-                dataType: (shell.argTypes as unknown[])[idx] as AnyWgslData,
+            ? shell.argTypes.map((arg, i) => ({
+                value:
+                  (ast.argNames.type === 'identifiers'
+                    ? ast.argNames.names[i]
+                    : undefined) ?? `param_${i}`,
+                dataType: arg as AnyWgslData,
               }))
             : []
           : Object.entries(shell.argTypes).map(([name, dataType]) => ({
--- a/packages/typegpu/src/core/function/tgpuFragmentFn.ts
+++ b/packages/typegpu/src/core/function/tgpuFragmentFn.ts
@@ -47,7 +47,7 @@ type TgpuFragmentFnShellHeader<
   FragmentIn extends FragmentInConstrained,
   FragmentOut extends FragmentOutConstrained,
 > = {
-  readonly argTypes: [AnyWgslStruct];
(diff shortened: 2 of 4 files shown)


packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -139,9 +139,14 @@ export function fragmentFn<
...
 }): TgpuFragmentFnShell<FragmentIn, FragmentOut> {
   const shell: TgpuFragmentFnShellHeader<FragmentIn, FragmentOut> = {
-    argTypes: [createStructFromIO(options.in ?? {})],
+    argTypes:
+      options.in && Object.keys(options.in).length !== 0
+        ? [createStructFromIO(options.in)]
+        : [],
     targets: options.out,
-    returnType: createOutputType(options.out) as FragmentOut,
+    returnType: (Object.keys(options.out).length !== 0
```
mhawryluk, line 157: not sure what to do when someone passes an empty object to `out` (we do so in multiple tests). should we throw an error instead?

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -175,9 +175,12 @@ export function createFnCore(
...
           ? ast.argNames.type === 'identifiers'
-            ? ast.argNames.names.map((name, idx) => ({
-                value: name,
-                dataType: (shell.argTypes as unknown[])[idx] as AnyWgslData,
+            ? shell.argTypes.map((arg, i) => ({
+                value:
+                  (ast.argNames.type === 'identifiers'
+                    ? ast.argNames.names[i]
+                    : undefined) ?? `arg_${i}`,
+                dataType: arg as AnyWgslData,
```
reczkok, line 183: I don't like the double `ast.argNames.type === 'identifiers'` check but I see TS has no intention of narrowing in the map body 🤔

packages/typegpu/src/core/function/tgpuFragmentFn.ts
reczkok: I think it's valid for fragment functions to have no output [in no-color-output mode]( https://www.w3.org/TR/webgpu/#no-color-output) so we shouldn't throw.

reczkok (approved):
Great changes 🕵️‍♂️

packages/typegpu/src/core/function/tgpuFragmentFn.ts
mhawryluk: ok, and vertex?

PR #1136, opened 2025-04-08
impr: Add automatic type conversions when possible

### LINKED ISSUE #970: feat: Automatically convert resource (aka snippet) types in expressions
When we encounter different but automatically convertible types we should convert according to [this ruleset](https://gpuweb.github.io/gpuweb/wgsl/#conversion-rank)

### LINKED ISSUE #990: impr: Properly handle TGSL pointers
We should properly apply the load rule where relevant based on expression type

reczkok:
closes #990 
closes #970


packages/typegpu/src/data/vector.ts
```diff
@@ -317,10 +338,24 @@ function makeVecSchema<TValue, S extends number | boolean>(
...
+        let argType = args[i]!.dataType;
+        while (isDecorated(argType)) {
+          argType = (argType as Decorated).inner as AnyWgslData;
+        }
+
+        argTypes[i] = isVec(argType) ? argType : vecTypeToPrimitive[type];
+      }
+
+      return argTypes;
+    },
```
iwoplaza: I think we can map the original args here, which will let us not fight with TS ⚔️ 
Also, things can be wrapped in Decorated only at one level, so we can safely replace the `while` with an `if`

```suggestion
    (...args) =>
      args.map((arg) => {
        let argType = arg.dataType;
        if (isDecorated(argType)) {
          argType = (argType as Decorated).inner as AnyWgslData;
        }
    
        return isVec(argType) ? argType : vecTypeToPrimitive[type];
      });
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -279,3 +292,430 @@ export function numericLiteralToSnippet(value: string): Snippet | undefined {
...
+  action: 'none',
+};
+
+function unwrapDecorated(type: AnyData): AnyData {
+  let currentType = type;
+  while (currentType.type === 'decorated') {
+    currentType = currentType.inner as AnyData;
+  }
+  return currentType;
+}
```
iwoplaza: Same here
```suggestion
function unwrapDecorated(data: AnyData): AnyData {
  if (data.type === 'decorated') {
    return data.inner as AnyData;
  }
  return data;
}
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -279,3 +292,430 @@ export function numericLiteralToSnippet(value: string): Snippet | undefined {
...
+  [vec2h.type]: f16,
+  [vec3h.type]: f16,
+  [vec4h.type]: f16,
+  [vec2i.type]: i32,
+  [vec3i.type]: i32,
+  [vec4i.type]: i32,
+  [vec2u.type]: u32,
+  [vec3u.type]: u32,
+  [vec4u.type]: u32,
+};
```
iwoplaza: Seems like a duplicate of `vecTypeToPrimitive` 👀

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -339,34 +389,62 @@ export function generateExpression(
...
+        }
+
+        return {
+          value: snippets,
+          dataType: UnknownData,
+        };
+      }
+    }
+
     return {
```
iwoplaza: Seems like we covered every case, so we can loudly fail here

iwoplaza (changes requested):

packages/typegpu/src/tgsl/wgslGenerator.ts
reczkok: It seems like can't. We allow raw objects as return values for TGSL entry-points.

iwoplaza: But those are implicitly wrapped in structs as well, right? If we don't provide it on the call-stack yet, we should, otherwise the order of properties in the returned object literal might not match with the order in the output struct.

iwoplaza (approved):
Apart from the unresolved discussion, amazing work!! 🚀

iwoplaza (approved):
yes yes yes! 🎉

PR #1149, opened 2025-04-11
docs: Add Gravity example & ops: Use git lfs for large files

aleksanderkatan:


iwoplaza:
Since this PR introduces changes to the library, it has to target `main` as the base branch 🙌

apps/typegpu-docs/src/content/examples/simulation/gravity/helpers.ts
```diff
@@ -0,0 +1,172 @@
...
+        modelMesh.attributes.NORMAL.value[3 * i + 1],
+        modelMesh.attributes.NORMAL.value[3 * i + 2],
+      ),
+      uv: d.vec2f(
+        modelMesh.attributes.TEXCOORD_0.value[2 * i],
+        1 - modelMesh.attributes.TEXCOORD_0.value[2 * i + 1],
+      ),
+    });
+  }
+  modelVertices.reverse();
```
iwoplaza: Uu, that's interesting! Why do we reverse the vertices? Is it because of culling? 👀

apps/typegpu-docs/src/content/examples/simulation/gravity/compute.ts
```diff
@@ -0,0 +1,149 @@
...
+          continue;
+        }
+
+        const dist = std.distance(current.position, other.position);
+        // are bodies disjoint?
+        if (dist >= radiusOf(current) + radiusOf(other)) {
+          continue;
+        }
+        // is the collision skipped?
+        if (current.collisionBehavior === 0 || other.collisionBehavior === 0) {
```
mhawryluk: i don't like that we have to remember which number is which collisionBehavior here. I would change `CollisionBehavior` into an enum:

```ts
export const enum CollisionBehavior {
  none = 0,
  bounce = 1,
  merge = 2,
}
```

and then refer to it here by `CollisionBehavior.none` instead of `0`

apps/typegpu-docs/src/content/examples/simulation/gravity/compute.ts
```diff
@@ -0,0 +1,149 @@
...
+            std.mul(m2 / (m1 + m2), other.velocity),
+          );
+          updatedCurrent.mass = m1 + m2;
+        }
+      }
+    }
+
+    collisionsLayout.$.outState[input.gid.x] = updatedCurrent;
+  })
+  .$name('Compute collisions shader');
```
mhawryluk:
```suggestion
  .$name('collisions');
```
the type of gpu object is visible in the error stack anyway. same everywhere

apps/typegpu-docs/src/content/examples/simulation/gravity/compute.ts
```diff
@@ -0,0 +1,149 @@
...
+        const otherId = d.u32(i);
+        const other = collisionsLayout.$.inState[otherId];
+        if (d.u32(i) === input.gid.x || other.destroyed === 1) {
+          continue;
+        }
+
+        const dist = std.distance(current.position, other.position);
+        // are bodies disjoint?
+        if (dist >= radiusOf(current) + radiusOf(other)) {
+          continue;
```
mhawryluk: I'd prefer else ifs instead of continue statements, but I don't insist

apps/typegpu-docs/src/content/examples/simulation/gravity/enums.ts
```diff
@@ -0,0 +1,29 @@
+export const collisionBehaviors = ['none', 'bounce', 'merge'] as const;
+export type CollisionBehavior = (typeof collisionBehaviors)[number];
+
+export type SkyBox = 'milky-way';
```
mhawryluk: since there's only one, we can simplify the example and remove this parameter

apps/typegpu-docs/src/content/examples/simulation/gravity/index.ts
```diff
@@ -0,0 +1,460 @@
...
+  renderVertexLayout,
+  skyBoxBindGroupLayout,
+  skyBoxVertexLayout,
+} from './schemas.ts';
+
+const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
+const canvas = document.querySelector('canvas') as HTMLCanvasElement;
+const context = canvas.getContext('webgpu') as GPUCanvasContext;
+const root = await tgpu.init();
+const device = root.device;
```
mhawryluk: remove this constant and replace its one (1) use

apps/typegpu-docs/src/content/examples/simulation/gravity/index.ts
```diff
@@ -0,0 +1,460 @@
...
+
+const { vertexBuffer: sphereVertexBuffer, vertexCount: sphereVertexCount } =
+  await loadModel(root, '/TypeGPU/assets/gravity/sphere.obj');
+
+const sampler = device.createSampler({
+  magFilter: 'linear',
+  minFilter: 'linear',
+});
+
+let cameraPosition = examplePresets['Solar System'].initialCameraPos;
```
mhawryluk: assign initial preset value to a constant, so it's possible to change it just in one place

apps/typegpu-docs/src/content/examples/simulation/gravity/index.ts
```diff
@@ -0,0 +1,460 @@
...
+          velocity: element.velocity ?? d.vec3f(),
+          mass: element.mass,
+          radiusMultiplier: element.radiusMultiplier ?? 1,
+          collisionBehavior: element.collisionBehavior
+            ? collisionBehaviors.indexOf(element.collisionBehavior)
+            : 0,
+          textureIndex: sphereTextureNames.indexOf(group.texture),
+          ambientLightFactor: element.ambientLightFactor ?? 0.6,
+        };
+      }),
```
mhawryluk: `() => { return {...}; }` -> `() => ({...})`

apps/typegpu-docs/src/content/examples/simulation/gravity/compute.ts
reczkok: I would avoid enums and just do an object

mhawryluk: could be an object, sure. though enums are faster

apps/typegpu-docs/src/content/examples/simulation/gravity/enums.ts
aleksanderkatan: I thought this example might be expanded with more presets in the future, but I guess this logic can be added back if this happens

apps/typegpu-docs/public/assets/gravity/skyboxes/milky-way/negx.jpg
deluksic, line 1: these jpegs look huge. Do you already use `git-lfs`? Having blobs in history can be detrimental to `git clone` performance down the line.

aleksanderkatan: We don't use it yet, we'll look into it

apps/typegpu-docs/src/content/examples/simulation/gravity/helpers.ts
aleksanderkatan: Honestly this is just a piece of code copied from 3D Fish. ~~Since we're using the depth stencil, I don't think either this or the cull mode matters (everything looks alright after removing those).~~ We can either use the reverse and cull mode `front`, or just use the cull mode `back` or `none`.

apps/typegpu-docs/src/content/examples/simulation/gravity/compute.ts
aleksanderkatan: I have refactored the logic a little, can you check if it's better now?

apps/typegpu-docs/src/content/examples/simulation/gravity/helpers.ts
iwoplaza: I'd use cull mode back, and remove the reverse.

apps/typegpu-docs/public/assets/gravity/skyboxes/milky-way/negx.jpg
iwoplaza: I applied Git LFS on this branch. I've heard it's limited to 1 GB, but maybe it will be sufficient for us for a while.

apps/typegpu-docs/src/content/examples/simulation/gravity/compute.ts
mhawryluk: great, I like it!

reczkok (approved):
Great work! 🪐

apps/typegpu-docs/src/content/examples/simulation/gravity/helpers.ts
```diff
@@ -0,0 +1,174 @@
...
+
+export const radiusOf = tgpu['~unstable']
+  .fn(
+    [CelestialBody],
+    d.f32,
+  )(
+    (body) =>
+      std.pow((body.mass * 0.75) / Math.PI, 0.333) * body.radiusMultiplier,
+  )
+  .$name('radiusOf');
```
iwoplaza: This should be fine with Deno, let's format it like this
```suggestion
export const radiusOf = tgpu['~unstable'].fn([CelestialBody], d.f32)((body) => {
  return std.pow((body.mass * 0.75) / Math.PI, 0.333) * body.radiusMultiplier;
});
```

apps/typegpu-docs/src/content/examples/simulation/gravity/render.ts
```diff
@@ -0,0 +1,125 @@
...
+      skyBoxLayout.$.camera.view.columns[2],
+      d.vec4f(0, 0, 0, 1),
+    );
+    return {
+      pos: std.mul(
+        skyBoxLayout.$.camera.projection,
+        std.mul(viewRotationMatrix, input.position),
+      ),
+      texCoord: input.position.xyz,
+    };
```
iwoplaza: Instead of creating a new matrix, we can "turn off" translation by setting the 4th parameter of the position vector to `0`.

```suggestion
    return {
      pos: std.mul(
        skyBoxLayout.$.camera.projection,
        std.mul(
          skyBoxLayout.$.camera.view,
          d.vec4f(input.position.xyz, 0),
        ),
      ),
      texCoord: input.position.xyz,
    };
```

apps/typegpu-docs/src/content/examples/simulation/gravity/render.ts
```diff
@@ -0,0 +1,125 @@
...
+    );
+    const cosTheta = std.dot(normal, lightDirection);
+    const diffuse = std.mul(
+      std.max(0, cosTheta),
+      std.mul(textureColor, lightColor),
+    );
+
+    const lightedColor = std.add(ambient, diffuse);
+
+    return d.vec4f(lightedColor.xyz, 1);
```
iwoplaza:
```suggestion
    const litColor = std.add(ambient, diffuse);

    return d.vec4f(litColor.xyz, 1);
```

packages/typegpu/src/std/index.ts
iwoplaza, line 1: Was this formatted with biome by accident?

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -316,7 +314,7 @@ export type InferLayoutEntry<T extends TgpuLayoutEntry | null> = T extends
   : T extends TgpuLayoutSampler ? TgpuSampler
   : T extends TgpuLayoutComparisonSampler ? TgpuComparisonSampler
   : T extends TgpuLayoutTexture ? TgpuSampledTexture<
-      Default<GetDimension<T['viewDimension']>, '2d'>,
+      Default<T['viewDimension'], '2d'>,
```
iwoplaza, line 317: What does this fix? 👀

aleksanderkatan: image

The `TDimension` should be `"cube"` because the dimensions of this texture are `[2048, 2048, 6]`. I'm not sure though why this particular change fixes this issue (it was advised to me by someone else from the team). Should I investigte it more/add tests for this?

packages/typegpu/src/std/index.ts
aleksanderkatan: It must have, I'll fix this. Edit: it was not formatting, it was organize imports on save

apps/typegpu-docs/src/content/examples/simulation/gravity/render.ts
aleksanderkatan: From what I see, you also need to add (0, 0, 0, w) to the intermediate result, like this: 
```typescript
return {
      pos: std.mul(
        skyBoxLayout.$.camera.projection,
        std.add(
          std.mul(
            skyBoxLayout.$.camera.view,
            d.vec4f(input.position.xyz, 0),
          ),
          d.vec4f(0, 0, 0, input.position.w),
        ),
      ),
      texCoord: input.position.xyz,
    };
```
or
```typescript
    return {
      pos: std.mul(
        skyBoxLayout.$.camera.projection,
        d.vec4f(
          std.mul(
            skyBoxLayout.$.camera.view,
            d.vec4f(std.div(input.position.xyz, input.position.w), 0),
          ).xyz,
          1,
        ),
      ),
      texCoord: input.position.xyz,
    };
```

Without this, the skybox doesn't seem to work.

packages/typegpu/src/tgpuBindGroupLayout.ts
reczkok: `GetDimension` translates a texture view dimension into texture dimension (since textures themselves can be only 1d, 2d or 3d. Since we are not binding textures but views this seems like an error. If we were translating `TgpuLayoutTexture` into a `TgpuTexture` and not it's view then we would need to use `GetDimension`.

iwoplaza: If it fixes a bug, let's keep it 💜

iwoplaza (approved):
Great job, awesome example!! 🪐

PR #1164, opened 2025-04-15
fix(docs): Cubemap Reflection example assets

iwoplaza:


apps/typegpu-docs/astro.config.mjs
```diff
@@ -23,7 +23,8 @@ export default defineConfig({
   site: 'https://docs.swmansion.com',
   base: 'TypeGPU',
   vite: {
-    plugins: [typegpu({})],
+    // Allowing query params, for invalidation
+    plugins: [typegpu({ include: [/\.m?[jt]sx?/] })],
```
iwoplaza, line 27: Fixes TGSL parsing during dev, this bug seems to have no effect in production

PR #1181, opened 2025-04-22
docs: Add fluid simulation based image distortion example

reczkok:


apps/infra-benchmarks/deno.json
```diff
@@ -1,5 +1,7 @@
 {
+  "exclude": ["."],
   "fmt": {
+    "exclude": ["!."],
```
mhawryluk, line 4: what are these for?

apps/typegpu-docs/src/content/examples/simulation/stable-fluid/meta.json
```diff
@@ -0,0 +1,4 @@
+{
+  "title": "Stable Fluids",
+  "category": "simulation"
+}
```
mhawryluk: is it not "experimental" on purpose? it uses unstable features

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -328,6 +328,12 @@ export type ExtractBindGroupInputFromLayout<
   [K in keyof T]: LayoutEntryToInput<T[K]>;
 };
 
+export type BindGroupLayoutToEntries<
+  T extends Record<string, TgpuLayoutEntry | null>,
+> = {
+  [K in keyof T]: LayoutEntryToInput<T[K]>;
+};
+
```
mhawryluk: ?

packages/typegpu/src/std/texture.ts
```diff
@@ -118,3 +134,300 @@ export const textureSample: TextureSampleOverload = createDualImpl(
...
+
+    if (maybeValue !== undefined) {
+      args.push(maybeValue);
+    }
+
+    return {
+      value: `textureStore(${args.map((v) => v.value).join(', ')})`,
+      dataType: Void,
+    };
+  },
```
mhawryluk:
```suggestion
  (texture, coords, arrayIndexOrValue, maybeValue) => ({
    value: `textureStore(${
      [texture, coords, arrayIndexOrValue, maybeValue].filter(
        (arg) => arg !== undefined,
      ).map((v) => v.value).join(', ')
    })`,
    dataType: Void,
  }),
```

packages/typegpu/src/std/texture.ts
```diff
@@ -118,3 +134,300 @@ export const textureSample: TextureSampleOverload = createDualImpl(
...
+    }
+
+    const textureInfo = texture.dataType as unknown as
+      | TgpuSampledTexture
+      | TgpuStorageTexture;
+
+    const is1D = textureInfo.dimension === '1d';
+    const is2D = ['2d', '2d-array', 'cube', 'cube-array'].includes(
+      textureInfo.dimension,
+    );
```
mhawryluk: is2D is evaluated even when is1D is true. I would remove both of these, that is inline them in the ternary expression below

apps/infra-benchmarks/deno.json
reczkok: It makes it work in Zed (otherwise it forces deno for both linting and formatting)

packages/typegpu/src/tgpuBindGroupLayout.ts
reczkok: Oopsies - this got merged under a different name in a separate pr. Will remove

mhawryluk (approved):
Awesome!! 🌊

apps/typegpu-docs/src/content/examples/simulation/stable-fluid/index.html
```diff
@@ -0,0 +1 @@
+<canvas></canvas>
```
aleksanderkatan, line 30: Maybe we could add a popup like in Cubemap Reflection to tell the users that this example can be interacted with using a mouse?

packages/typegpu/src/std/texture.ts
```diff
@@ -118,3 +134,282 @@ export const textureSample: TextureSampleOverload = createDualImpl(
...
+    coords: v2f,
+    level: number,
+  ): v4f;
+  <T extends TgpuSampledTexture<'2d'>>(
+    texture: T,
+    sampler: TgpuSampler,
+    coords: v2f,
+    level: number,
+    offset: v2i,
+  ): v4f;
```
aleksanderkatan:
```suggestion
  <T extends TgpuSampledTexture<'2d'>>(
    texture: T,
    sampler: TgpuSampler,
    coords: v2f,
    level: number,
    offset?: v2i,
  ): v4f;
```
Could we simplify it like this? (same goes for other overloads in this file)

packages/typegpu/src/std/texture.ts
```diff
@@ -118,3 +134,282 @@ export const textureSample: TextureSampleOverload = createDualImpl(
...
+      return {
+        value: `textureLoad(${args.map((v) => v.value).join(', ')})`,
+        dataType: textureInfo.texelDataType,
+      };
+    }
+
+    return {
+      value: `textureLoad(${args.map((v) => v.value).join(', ')})`,
+      dataType: channelDataToInstance[textureInfo.channelDataType.type],
+    };
```
aleksanderkatan:
```suggestion
    return {
      value: `textureLoad(${args.map((v) => v.value).join(', ')})`,
      dataType: 'texelDataType' in textureInfo
        ? textureInfo.texelDataType
        : channelDataToInstance[textureInfo.channelDataType.type],
    };
```

apps/typegpu-docs/src/content/examples/simulation/stable-fluid/index.ts
```diff
@@ -0,0 +1,486 @@
...
+canvas.addEventListener('mousedown', (e) => {
+  const x = e.offsetX * devicePixelRatio;
+  const y = e.offsetY * devicePixelRatio;
+  brushState = {
+    pos: toGrid(x, y),
+    delta: [0, 0],
+    isDown: true,
+  };
+});
+canvas.addEventListener('mouseup', () => {
```
aleksanderkatan:
```suggestion
window.addEventListener('mouseup', () => {
```

aleksanderkatan (approved):
Great example! 🟠 🟢 🔵

apps/typegpu-docs/src/content/examples/simulation/stable-fluid/index.ts
```diff
@@ -0,0 +1,496 @@
...
+  pipeline
+    .withColorAttachment({
+      view: context.getCurrentTexture().createView(),
+      loadOp: 'clear',
+      storeOp: 'store',
+    })
+    .with(renderLayout, renderBG)
+    .draw(6);
+
+  root['~unstable'].flush();
```
iwoplaza: Render pipelines should flush by default now 👀

apps/typegpu-docs/src/content/examples/simulation/stable-fluid/render.ts
```diff
@@ -0,0 +1,110 @@
...
+  const uvs = [
+    d.vec2f(0, 0),
+    d.vec2f(1, 0),
+    d.vec2f(0, 1),
+    d.vec2f(1, 0),
+    d.vec2f(1, 1),
+    d.vec2f(0, 1),
+  ];
+  return { pos: verts[i.idx], uv: uvs[i.idx] };
+});
```
iwoplaza: We can copy over the single-triangle screen "rect" vertex fn from the oklab example 👀

iwoplaza (approved):
Awesome example! 🌊

apps/typegpu-docs/src/content/examples/simulation/stable-fluid/render.ts
reczkok: I don't know... In my opinion it will make it more complex to handle the sampling (due to the weird uv space) for no obvious benefit. I think it will be a net gain in lines (with all the uv scaling logic) and a net loss in readability. I'm happy to switch to a triangle strip tho

iwoplaza: What about this?

```ts
export const fullScreenTriangle = tgpu['~unstable'].vertexFn({
  in: { idx: d.builtin.vertexIndex },
  out: { pos: d.builtin.position, uv: d.vec2f },
})((i) => {
  const pos = [d.vec2f(-1, -1), d.vec2f(3, -1), d.vec2f(-1, 3)];
  const uvs = [d.vec2f(0, 0), d.vec2f(2, 0), d.vec2f(0, 2)];

  return { pos: d.vec4f(pos[i.idx], 0, 1), uv: uvs[i.idx] };
});
```
We don't have to apply any scaling logic, we still can just write out the UVs

reczkok: I suppose it does interpolate the UVs correctly, but I still find the fullscreen triangle a bit over-engineered for example purposes. That said, this version is much more concise so I’ll yield and make the change.

PR #1189, opened 2025-04-24
feat: Texture std expansion and generator adjustments

reczkok:
Lib changes from #1181


packages/typegpu/tests/std/numeric/div.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  vec2i,
+  vec2u,
+  vec3f,
+  vec3i,
+  vec3u,
+  vec4f,
+  vec4i,
+  vec4u,
+} from '../../../src/data/vector.ts';
+import { div } from '../../../src/std/numeric.ts';
```
iwoplaza: Let's change these to imports of entry files, since we're testing the public API surface.
```suggestion
import * as d from '../../../src/data/index.ts';
import { div } from '../../../src/std/index.ts';
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -245,7 +246,9 @@ export function getTypeFromWgsl(resource: Wgsl): AnyData | UnknownData {
     }
   }
 
-  return isWgslData(resource) ? resource : UnknownData;
+  return isWgslData(resource) || isSelfResolvable(resource)
```
iwoplaza, line 250: Could you jot down a comment saying that this is temporary, until we figure out textures-as-data?

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -341,6 +341,12 @@ export type InferLayoutEntry<T extends TgpuLayoutEntry | null> =
               ? StorageTextureUsageForEntry<T>
               : never;
 
+export type BindGroupLayoutToEntries<
```
iwoplaza: Could we rename this to `ExtractBindGroupInputFromLayout`?

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -147,7 +147,9 @@ export function createFnCore(
           applyExternals(
             externalMap,
             Object.fromEntries(
-              ast.argNames.props.map(({ prop, alias }) => [alias, prop]),
+              ast.argNames.props
+                .filter(({ prop, alias }) => prop !== alias)
+                .map(({ prop, alias }) => [alias, prop]),
```
iwoplaza: Could you add a comment what the filtering does? 👀

reczkok: This was to avoid unnecessary cluttering of the external map since I don't think the alias can ever be namable. I did it mostly for easier debugging.

reczkok: Since I don't think it does much harm and could potentially cause a bug I think I will revert this.

iwoplaza (approved):
YIPPEEEE

PR #1194, opened 2025-04-26
Add Oklab color space

deluksic:
Adds `Oklab` color space as described in:
- https://bottosson.github.io/posts/oklab/
- https://bottosson.github.io/posts/gamutclipping/

Adds an example page to compare it against CSS implementation:

image

Unfortunately, the real CSS algorithm is quite a bit [more involved](https://drafts.csswg.org/css-color-4/#binsearch) and uses wider color space if monitor allows it, so out of gamut colors (checkered pattern area) will not match CSS. I found alpha of 0.4 to match somewhat close.

WIP:

- [x] naming and organization
- [x] ~add oklch~ will add in a later PR
- [x] consider something better than checker pattern


deluksic:
Added small indicator for the values and both sides  of the color space because it looks cool:
image

deluksic:
Added some additional example controls:
- ability to switch gamut clipping function between `L0 0.5` and `L0 cusp`
- ability to switch out of gamut pattern to projection lines

image

deluksic:
Although its fun to play with different out of gamut projection algorithms and see what they do, I believe we should have a single "blessed" `oklabToRgb` and `rgbToOklab`. I think it should directly go to `sRGB` instead of linear as well.

apps/typegpu-docs/src/content/examples/simple/color/index.ts
```diff
@@ -0,0 +1,285 @@
...
+
+const Uniforms = d.struct({
+  hue: d.f32,
+  alpha: d.f32,
+});
+
+const uniforms: d.Infer<typeof Uniforms> = {
+  hue: 0.7,
+  alpha: 0.05,
+};
```
iwoplaza: There's now another way this can be written, one that does not involve an explicit type. Let me know which one you prefer 🙌 

```suggestion
const uniforms = Uniforms({
  hue: 0.7,
  alpha: 0.05,
});
```

apps/typegpu-docs/src/content/examples/simple/color/index.ts
```diff
@@ -0,0 +1,285 @@
...
+      selections.outOfGamutPattern = outOfGamutPatternOptions[selected];
+      pipeline = root['~unstable']
+        .withVertex(mainVertex, {})
+        .withFragment(
+          mainFragment(selections.gamutClip, selections.outOfGamutPattern),
+          {
+            format: presentationFormat,
+          },
+        )
+        .createPipeline();
```
deluksic: reuse `.createPipeline` logic

apps/typegpu-docs/src/content/examples/simple/color/index.ts
```diff
@@ -0,0 +1,285 @@
...
+    { uv: d.vec2f, clipLab: d.vec3f },
+    d.f32,
+  )(
+    /* wgsl */ `{
+      let stripeX = modulo(uv.x, 0.1) / 0.1 > 0.5;
+      let stripeY = modulo(uv.y, 0.1) / 0.1 > 0.5;
+      return f32(stripeX != stripeY);
+    }`,
+  )
+  .$uses({ modulo });
```
iwoplaza: After our experiment with object arguments, we decided we're going somewhere in between and will implement a unified API for defining functions, with similar DX improvements that object arguments brought, but without the limitations.

For now, let's define functions using the array args, as it will make migrating to the new hybrid API easier.

```suggestion
const patternCheckers = tgpu['~unstable'].fn(
  [d.vec2f, d.vec3f],
  d.f32,
) /* wgsl */`(uv: vec2f, clipLab: vec3f) -> f32 {
  let stripeX = modulo(uv.x, 0.1) / 0.1 > 0.5;
  let stripeY = modulo(uv.y, 0.1) / 0.1 > 0.5;
  return f32(stripeX != stripeY);
}
`.$uses({ modulo });
```

apps/typegpu-docs/src/content/examples/simple/color/index.ts
```diff
@@ -0,0 +1,285 @@
...
+  let Lgrid = 0.02 - thickness - abs(clipLab.x % 0.04 - 0.02);
+  return select(clamp(Lgrid / fwidth(Lgrid), 0, 1), 1, thickness < 0.0002);
+}`);
+
+const patternSolid = tgpu['~unstable'].fn(
+  { uv: d.vec2f, clipLab: d.vec3f },
+  d.f32,
+)(() => 1);
+
+const mainFragment = (
```
iwoplaza: Cool pattern! (no pun intended)
By creating functions that accept resources and return other resources, we're able to write shaders using functions/resources that users can inject from the outside.

We have a primitive that should help with that.

**Before:**
```ts
function mainFragment(pattern: TgpuFn<[uv: d.Vec2f, clipLab: d.Vec3f], d.F32>) {
  return tgpu['~unstable'].fragmentFn({
      in: { uv: d.vec2f },
      out: d.vec4f,
    }) /* wgsl */`{
    let pat = pattern(...);
  }`.$uses({ pattern });
}

let pipeline = root['~unstable']
  .withVertex(mainVertex, {})
  .withFragment(mainFragment(patternCheckers), {
    format: presentationFormat,
  })
  .createPipeline();
```

**After:**
```ts
const pattern = tgpu['~unstable'].slot(patternSolid);

const mainFragment = tgpu['~unstable'].fragmentFn({
      in: { uv: d.vec2f },
      out: d.vec4f,
    }) /* wgsl */`{
    let pat = pattern(...);
  }`.$uses({ pattern });
}

let pipeline = root['~unstable']
  .with(pattern, patternCheckers) // replace the slot with something else
  .withVertex(mainVertex, {})
  .withFragment(mainFragment, {
    format: presentationFormat,
  })
  .createPipeline();
```

With **slots**, you can use resources that don't yet exist (or have a default value that you want to be able to swap).

deluksic: What is the  benefit over the (simpler?) functional approach?

apps/typegpu-docs/src/content/examples/simple/color/index.ts
deluksic: Ooh nice!

apps/typegpu-docs/src/content/examples/simple/color/index.ts
deluksic: Ok, will revert all usages

apps/typegpu-docs/src/content/examples/simple/color/index.ts
iwoplaza: No difference in behavior (in most cases), but:
- Parts of the shader can be swapped, while not requiring the end-user to explicitly provide the defaults. This can useful for libraries that assume a happy path or good defaults, but want to allow devs to customize.
- A function deep in the call stack can use a slot, and it can be provided way higher in the tree. This reduces the need to refactor every function in the call chain to pass down that requirement, at the cost of being less explicit.
- If a function is used in two places, and uses the same slot values, it will be reused. This leads to less duplicate WGSL code being generated.

deluksic: Thank you for the explanation. I think it makes sense, I will have to try it to get the feeling for it. Could deduplication be done anyway (without slots) or is it fundamental somehow?

deluksic: This could actually be used to have a single oklabToRgb while providing a way to customize out-of-gamut projection. Is this also a way to provide the alpha constant?

iwoplaza: > Could deduplication be done anyway (without slots) or is it fundamental somehow?

It could! The function could maintain a cache and verify if it has been called with the same inputs before, and instead of creating a new function, return a cached result. The difference is that the slot cache lives only during code generation, so it will clean itself up once the WGSL shader code is generated.

iwoplaza: > This could actually be used to have a single oklabToRgb while providing a way to customize out-of-gamut projection. Is this also a way to provide the alpha constant?

Possibly! I am still not 100% sure what `alpha` is responsible for, I will have to read up on the articles you provided.
But slots can be used to inject any value into a shader, including numeric constants:

```ts
const alpha = tgpu['~unstable'].slot(0.5);

const foo = tgpu['~unstable'].fn([], d.vec3f)`() -> vec3f {
  return vec3f(alpha);
}
`.$usage({ alpha });
```

Will by default generate:

```wgsl
fn foo() -> vec3f {
  return vec3f(0.5);
}
```

apps/typegpu-docs/src/content/examples/simple/color/index.ts
deluksic: `{ a: f32, b: f32, L1: f32, C1: f32, L0: f32, cusp: LC }` 😭

packages/typegpu-color/src/oklab.ts
```diff
@@ -0,0 +1,346 @@
...
+
+        const u_b = b1 / (b1 * b1 - 0.5 * b * b2);
+        let t_b = -b * u_b;
+
+        // @ts-expect-error select doesn't work on number?
+        t_r = select(FLT_MAX, t_r, u_r >= 0);
+        // @ts-expect-error
+        t_g = select(FLT_MAX, t_g, u_g >= 0);
+        // @ts-expect-error
+        t_b = select(FLT_MAX, t_b, u_b >= 0);
```
deluksic, line 261: what should I do here? It seems like a simple usage.

packages/typegpu-color/src/oklab.ts
```diff
@@ -0,0 +1,346 @@
...
+    linearRgbToOklab,
+    findCusp,
+  });
+
+export const gamutClipAdaptiveL0cusp = tgpu['~unstable']
+  .fn(
+    [vec3f, f32],
+    vec3f,
+  )(
+    // TGSL does not have sign and sqrt yet
```
deluksic: missing sign and sqrt to convert all functions to TGSL.

packages/typegpu-color/src/oklab.ts
iwoplaza: I've filed an issue, and prepared a pull request that fixes this problem. This seems to be a mistyped signature 👀
https://github.com/software-mansion/TypeGPU/issues/1204

deluksic: awesome, thank you!

packages/typegpu-color/src/oklab.ts
iwoplaza: `std.sqrt` just got merged, and `std.sign` is waiting for review #1205

packages/typegpu-color/src/oklab.ts
```diff
@@ -0,0 +1,346 @@
+import tgpu from 'typegpu';
+import { f32, struct, vec3f } from 'typegpu/data';
+import { clamp, length, max, min, mix, select } from 'typegpu/std';
+
+const cbrt = tgpu['~unstable'].fn(
+  [f32],
+  f32,
+)(/* wgsl */ `(x: f32) -> f32 {
+  return sign(x) * pow(abs(x), 1.0 / 3);
+}`);
```
iwoplaza: Could we use the template string variant here? It gets treated by out formatter a lot nicer 💜
```suggestion
const cbrt = tgpu['~unstable'].fn([f32], f32) /* wgsl */`(x: f32) -> f32 {
  return sign(x) * pow(abs(x), 1.0 / 3);
}`;
```

deluksic: I'd use TGSL when `sign` is merged, right?

deluksic: What do you think about adding`cbrt` to `std`? I know it's not in WGSL, but it is in `Math`. It is not a super common operation, I guess..

apps/typegpu-docs/src/content/examples/simple/color/index.ts
```diff
@@ -0,0 +1,274 @@
...
+const uniformsBindGroupLayout = tgpu.bindGroupLayout({
+  uniforms: { uniform: Uniforms },
+});
+
+const uniformsBuffer = root.createBuffer(Uniforms, uniforms).$usage('uniform');
+const uniformsBindGroup = root.createBindGroup(uniformsBindGroupLayout, {
+  uniforms: uniformsBuffer,
+});
+
+const modulo = tgpu['~unstable'].fn(
```
deluksic: this could also be added to `std`. I really hate that both `%` and `mod` are basically useless for graphics programming https://github.com/gpuweb/gpuweb/issues/3987

deluksic: maybe not `std` as it is not in actual WGSL, but I feel like this is going to be a very common function for me to copy paste around.

packages/typegpu-color/src/oklab.ts
iwoplaza: Should be on main 🫡

packages/typegpu-color/src/oklab.ts
iwoplaza: Merged! 🚀

packages/typegpu-color/src/oklab.ts
iwoplaza: I think it's a good idea! Sounds useful 💜
Would you like to add it to std as part of this PR?

apps/typegpu-docs/src/content/examples/simple/color/index.ts
deluksic: I converted the example yesterday to slots, I will work on making a canonical `oklabToRgb` next using slots.

packages/typegpu-color/src/oklab.ts
deluksic: It can be separate I think.

apps/typegpu-docs/src/content/examples/simple/color/index.ts
```diff
@@ -0,0 +1,278 @@
...
+});
+
+const patternCheckers = tgpu['~unstable'].fn(
+  [d.vec2f, d.vec3f],
+  d.f32,
+)((uv, _clipLab) => {
+  const stripeX = modulo(uv.x, 0.1) / 0.1 > 0.5;
+  const stripeY = modulo(uv.y, 0.1) / 0.1 > 0.5;
+  return d.f32(stripeX !== stripeY);
+});
```
iwoplaza: I remember I had a checkerboard floor in one of my apps, and I don't remember where I got it from, but here's an implementation that doesn't use `modulo`:

```ts
const patternCheckers = tgpu['~unstable'].fn(
  [d.vec2f, d.vec3f],
  d.f32,
)((uv, _clipLab) => {
  const suv = std.floor(std.mul(20, uv));
  return (suv.x + suv.y) - 2 * std.floor((suv.x + suv.y) * 0.5);
});
```

deluksic: took me a while, but this is a really smart way of doing it, no divisions.

apps/typegpu-docs/src/content/examples/simple/oklab/index.html
deluksic, line 1: Do note, I changed the name of the example folder to `oklab` from `color`.

iwoplaza (approved):
Thanks for yet another contribution, and an amazing example to showcase it! 👏👏👏
It's really fun to play around with 💜

PR #1205, opened 2025-04-29
feat: std.sign

iwoplaza:


reczkok (approved):
Yes please

packages/typegpu/src/std/numeric.ts
```diff
@@ -323,6 +323,18 @@ export const min = createDualImpl(
...
+  <T extends AnyNumericVecInstance | number>(e: T): T => {
+    if (typeof e === 'number') {
+      return Math.sign(e) as T;
+    }
+    return VectorOps.sign[e.kind](e) as T;
+  },
+  // GPU implementation
+  (e) => ({ value: `sign(${e.value})`, dataType: e.dataType }),
+);
+
```
aleksanderkatan: The signature allows for vecNu and vecNi, while there are no implementations for those. Please add implementation for vecNi, and update signature to only include vecNi, vecNf and vecNh (edit: and number ofc).

packages/typegpu/src/std/numeric.ts
```diff
@@ -323,6 +323,18 @@ export const min = createDualImpl(
...
+export const sign = createDualImpl(
+  // CPU implementation
+  <T extends AnyNumericVecInstance | number>(e: T): T => {
+    if (typeof e === 'number') {
+      return Math.sign(e) as T;
+    }
+    return VectorOps.sign[e.kind](e) as T;
+  },
+  // GPU implementation
+  (e) => ({ value: `sign(${e.value})`, dataType: e.dataType }),
```
aleksanderkatan, line 345: I'm not sure about the result type, from the documentation it seems that it would always be i32 or vecNi, could you check it?

iwoplaza: Good question! I was trying to mirror the WGSL spec signature, and from what it specified, the type of the input if the type of the output. I tested it empirically, and it looks to be true!

**u32 input:**
Screenshot 2025-04-29 at 12 20 25

**f32 input:**
Screenshot 2025-04-29 at 12 20 44

aleksanderkatan (approved):
Great addition!

PR #1211, opened 2025-04-30
test: Replace all occurrences of `toEqual` with `toStrictEqual`

aleksanderkatan:


aleksanderkatan:
I replaced all occurrences of `toEqual` with `toBe` when tested value was primitive, and with `toStrictEqual` otherwise.

I think we can just use `toBe` for primitives/reference checks and `toStrictEqual` for everything else by default, and to use `toEqual` only when it's difficult to provide an expected value that strictly equals (including prototype, undefined props etc.) the actual tested one. Let me know if you agree.

mhawryluk (approved):
Looks good to me! 👮‍♂️

iwoplaza (approved):
Definitely!

PR #1213, opened 2025-05-05
refactor: Use Deno for formatting

iwoplaza:
## Motivation
The formatting is more forgiving to our functions API.


deno.json
```diff
@@ -0,0 +1,6 @@
+{
+  "fmt": {
+    "singleQuote": true,
+    "exclude": ["pnpm-lock.yaml"]
+  }
+}
```
aleksanderkatan, line 6: Can we use Deno workspaces to omit having to write the defaults (right now only the "singleQuote" prop) in every package?

.vscode/extensions.json
```diff
@@ -0,0 +1,5 @@
+{
+  "recommendations": [
+    "emeraldwalk.runonsave"
+  ]
```
aleksanderkatan, line 4: Is it possible to also display a message to people cloning the repo asking them to install Deno cli? Edit: or we could add a note to README

.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "[astro]": {
+    "editor.defaultFormatter": "astro-build.astro-vscode"
+  },
+  "emeraldwalk.runonsave": {
+    "commands": [
+      {
+        "match": ".*",
+        "cmd": "echo \"Formatting ${relativeFile}\" && deno fmt ./${relativeFile}"
```
aleksanderkatan, line 10: Cmd+k, s should save without formatting, but in this setup, it formats anyway.  Here's an issue with a workaround: https://github.com/emeraldwalk/vscode-runonsave/issues/81

PR #1221, opened 2025-05-07
feat: Store labels of objects on `globalThis`

aleksanderkatan:


aleksanderkatan:
Changes:
- add `getName` and `setName` functions working on `globalThis`,
- remove `Labelled` interface and all label properties defined by us and replace them with `getName` and `setName` calls,
- add a symbol `$labelForward` -- when an object defines this symbol as a property, the `getName` and `setName` functions will consider the object behind that property instead (e.g. `tgpuFn` takes its label after `fnCore`),
- modify `LayoutMembership` from a raw object to a class that sets its own name in the constructor, because some objects derived its key as their label,
- change parameter type of `$name` methods to `string` (previously it was sometimes `string | undefined`).

Notes:
- raw gpu objects (e.g. GPUBuffer) still define a label property since we can't modify them, these objects are not saved on `globalThis`,
- there is no `setNameIfMissing` function because I think we should prefer checking manually and calling `$name` instead, as it often renames either more object than one (e.g. TgpuBufferImpl), or doesn't rename the caller but one of its properties instead (e.g. TgpuComputePipelineImpl).

.vscode/settings.json
```diff
@@ -11,5 +11,6 @@
       }
     ]
   },
-  "deno.enable": false
+  "deno.enable": false,
+  "typescript.preferences.importModuleSpecifier": "relative"
```
iwoplaza, line 21: 💜

packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -146,12 +147,8 @@ export class TgpuLaidOutSamplerImpl implements TgpuSampler, SelfResolvable {
     this[$internal] = {};
   }
 
-  get label(): string | undefined {
-    return this._membership.key;
-  }
-
   '~resolve'(ctx: ResolutionCtx): string {
-    const id = ctx.names.makeUnique(this.label);
+    const id = ctx.names.makeUnique(getName(this));
```
iwoplaza, line 152: We were using `this._membership.key` as the label before, but now that's lost.
I would recommend adding `setName(this, _membership.key);` to the constructor 👌

packages/typegpu/src/core/texture/externalTexture.ts
```diff
@@ -22,15 +24,14 @@ export function isExternalTexture<T extends TgpuExternalTexture>(
...
   implements TgpuExternalTexture, SelfResolvable {
   public readonly resourceType = 'external-texture';
+  public readonly [$labelForward]: LayoutMembership;
 
-  constructor(private readonly _membership: LayoutMembership) {}
-
-  get label(): string | undefined {
-    return this._membership.key;
+  constructor(private readonly _membership: LayoutMembership) {
+    this[$labelForward] = _membership;
```
iwoplaza: We can keep the name on this texture view, but do the same name assignment as in the sampler.

iwoplaza (changes requested):

packages/typegpu/src/core/sampler/sampler.ts
aleksanderkatan: ~~I forgot to add `[$labelForward]: _membership` here, I'll add this instead, ok?~~
The comment below

packages/typegpu/src/core/texture/externalTexture.ts
aleksanderkatan: Should I revert the changes to LayoutMembership and assume that it is immutable? I already kind of assume it

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,53 @@
...
+  globalThis.__TYPEGPU_META__ ??= new WeakMap();
+  const map = globalThis.__TYPEGPU_META__;
+  map.set(definition, { ...map.get(definition), name });
+}
+
+/**
+ * Can be assigned a name. Not to be confused with
+ * being able to HAVE a name.
+ * The `$name` function should use `setName` to rename the object itself,
+ * or rename the object `$labelForward` symbol points to instead if applicable.
```
iwoplaza:
```suggestion
 * or rename the object `$getNameForward` symbol points to instead if applicable.
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,53 @@
...
+  if (isForwarded(definition)) {
+    return getName(definition[$getNameForward]);
+  }
+  return globalThis.__TYPEGPU_META__?.get(definition)?.name;
+}
+
+export function setName(definition: object, name: string): void {
+  globalThis.__TYPEGPU_META__ ??= new WeakMap();
+  const map = globalThis.__TYPEGPU_META__;
+  map.set(definition, { ...map.get(definition), name });
```
iwoplaza:
```suggestion
  const map = (globalThis as unknown as GlobalWithMeta).__TYPEGPU_META__ ??= new WeakMap();
  map.set(definition, { ...map.get(definition), name });
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,53 @@
...
+function isObject(value: unknown): value is object {
+  return !!value && (typeof value === 'object' || typeof value === 'function');
+}
+
+function isForwarded(value: object): value is { [$getNameForward]: object } {
+  if (!($getNameForward in value)) {
+    return false;
+  }
+  return isObject(value[$getNameForward]);
+}
```
iwoplaza: Since we're checking if the value is an object again at the beginning of `getName` we can omit it.
```suggestion
function isForwarded(value: object): value is { [$getNameForward]: unknown } {
  return !!value?.[$getNameForward];
}
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,53 @@
+import { $getNameForward } from './shared/symbols.ts';
+
+interface MetaData {
+  name: string | undefined;
+}
+
+declare global {
+  var __TYPEGPU_META__: WeakMap<object, MetaData>;
+}
```
iwoplaza: Declaring this here causes any project using TypeGPU to be aware of this value. It will complicate our internals a bit without this block of code, but it won't pollute the global types 🙇‍♂️

image


```suggestion
interface GlobalWithMeta {
  __TYPEGPU_META__: WeakMap<object, MetaData>;
}
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export function getName(definition: unknown): string | undefined {
+  if (!isObject(definition)) {
+    return undefined;
+  }
+  if (isForwarded(definition)) {
+    return getName(definition[$getNameForward]);
+  }
+  return globalThis.__TYPEGPU_META__?.get(definition)?.name;
+}
```
iwoplaza: It turns out that using a primitive as a key when retrieving a value from a WeakMap does not cause a runtime error, meaning we can omit the check for object-ness.

image

```suggestion
export function getName(definition: unknown): string | undefined {
  if (isForwarded(definition)) {
    return getName(definition[$getNameForward]);
  }
  return (globalThis as unknown as GlobalWithMeta).__TYPEGPU_META__?.get(definition)?.name;
}
```

iwoplaza (changes requested):

packages/typegpu/src/name.ts
aleksanderkatan: Biome does not like this
image

aleksanderkatan:
I'm not sure if I like the number of casts in the `name.ts` file, should we really leave it like this?

packages/typegpu/src/core/function/tgpuComputeFn.ts
```diff
@@ -175,7 +175,7 @@ function createComputeFn<ComputeIn extends Record<string, AnyComputeBuiltin>>(
     },
 
     toString() {
-      return `computeFn:${this.label ?? '<unnamed>'}`;
+      return `computeFn:${getName(this) ?? '<unnamed>'}`;
```
iwoplaza: If we set name of core, let's also get it's name here (even though we could take advantage of `$getNameForward`)
```suggestion
      return `computeFn:${getName(core) ?? '<unnamed>'}`;
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -234,13 +237,10 @@ function createFn<
...
 
-  // Making the label available as a readonly property.
-  Object.defineProperty(fn, 'label', {
-    get: () => core.label,
-  });
-
   Object.defineProperty(fn, 'toString', {
-    value: () => `fn:${core.label ?? '<unnamed>'}`,
+    value() {
+      return `fn:${getName(this) ?? '<unnamed>'}`;
```
iwoplaza: Same here
```suggestion
      return `fn:${getName(core) ?? '<unnamed>'}`;
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -300,14 +303,9 @@ function createBoundFunction<
...
 
-  // Making the label available as a readonly property.
-  Object.defineProperty(fn, 'label', {
-    get: () => innerFn.label,
-  });
-
   Object.defineProperty(fn, 'toString', {
     value() {
-      const fnLabel = innerFn.label ?? '<unnamed>';
+      const fnLabel = getName(this) ?? '<unnamed>';
```
```suggestion
      const fnLabel = getName(innerFn) ?? '<unnamed>';
```

packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -242,7 +240,7 @@ function createFragmentFn(
     },
 
     toString() {
-      return `fragmentFn:${this.label ?? '<unnamed>'}`;
+      return `fragmentFn:${getName(this) ?? '<unnamed>'}`;
```
iwoplaza:
```suggestion
      return `fragmentFn:${getName(core) ?? '<unnamed>'}`;
```

packages/typegpu/src/core/function/tgpuVertexFn.ts
```diff
@@ -198,7 +198,7 @@ function createVertexFn(
     },
 
     toString() {
-      return `vertexFn:${this.label ?? '<unnamed>'}`;
+      return `vertexFn:${getName(this) ?? '<unnamed>'}`;
```
iwoplaza:
```suggestion
      return `vertexFn:${getName(core) ?? '<unnamed>'}`;
```

packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -108,10 +106,10 @@ class TgpuComputePipelineImpl implements TgpuComputePipeline {
     z?: number | undefined,
   ): void {
     const memo = this._core.unwrap();
-    const { branch, label } = this._core;
+    const { branch } = this._core;
 
     const pass = branch.commandEncoder.beginComputePass({
-      label: label ?? '<unnamed>',
+      label: getName(this) ?? '<unnamed>',
```
iwoplaza:
```suggestion
      label: getName(this._core) ?? '<unnamed>',
```

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -334,8 +332,9 @@ class TgpuRenderPipelineImpl implements TgpuRenderPipeline {
       colorAttachments,
     };
 
-    if (internals.core.label !== undefined) {
-      renderPassDescriptor.label = internals.core.label;
+    const label = getName(this);
```
iwoplaza:
```suggestion
    const label = getName(internals.core);
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,48 @@
...
+
+interface GlobalWithMeta {
+  __TYPEGPU_META__: WeakMap<object, MetaData>;
+}
+
+function isForwarded(value: unknown): value is { [$getNameForward]: unknown } {
+  return !!value &&
+    (typeof value === 'object' || typeof value === 'function') &&
+    $getNameForward in value;
```
iwoplaza:
```suggestion
  return !!(value as { [$getNameForward]?: unknown })?.[$getNameForward];
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,48 @@
...
+/**
+ * Can be assigned a name. Not to be confused with
+ * being able to HAVE a name.
+ * The `$name` function should use `setName` to rename the object itself,
+ * or rename the object `$getNameForward` symbol points to instead if applicable.
+ */
+export interface TgpuNamable {
+  $name(label: string): this;
+}
+export function isNamable(value: unknown): value is TgpuNamable {
```
iwoplaza:
```suggestion
}

export function isNamable(value: unknown): value is TgpuNamable {
```

packages/typegpu/src/name.ts
```diff
@@ -0,0 +1,48 @@
...
+  return (
+    !!value &&
+    (typeof value === 'object' || typeof value === 'function') &&
+    '$name' in value
+  );
```
iwoplaza:
```suggestion
  return !!(value as TgpuNamable)?.['~name'];
```

iwoplaza (changes requested):

iwoplaza (approved):
GLOBAL

PR #1223, opened 2025-05-08
impr: Throw error when trying to define tgpu.derived during resolution

### LINKED ISSUE #810: feat: Throw error when trying to define a `derived` value during resolution.
## Motivation
Allowing this pattern would introduce a lot of unnecessary complexity to how we handle slot bindings. It could theoretically be possible, but lets restrict it for now for our own sanity. If there's a valid use-case for this, and a sane behavior, we can re-enable it.

## Possible implementation
We can check if `getResolutionCtx()` returns a defined value, and throw a descriptive error if it does.

mhawryluk:
closes #810

### DIFF
--- a/packages/typegpu/src/core/slot/derived.ts
+++ b/packages/typegpu/src/core/slot/derived.ts
@@ -25,6 +25,12 @@ function stringifyPair([slot, value]: SlotValuePair): string {
 }
 
 function createDerived<T>(compute: () => T): TgpuDerived<T> {
+  if (getResolutionCtx()) {
+    throw new Error(
+      'Cannot create tgpu.derived objects at the resolution stage.',
+    );
+  }
+
   const result = {
     resourceType: 'derived' as const,
     '~compute': compute,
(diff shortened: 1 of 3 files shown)


packages/typegpu/src/core/slot/derived.ts
```diff
@@ -60,6 +66,12 @@ function createBoundDerived<T>(
   innerDerived: TgpuDerived<T>,
   pairs: SlotValuePair[],
 ): TgpuDerived<T> {
+  if (getResolutionCtx()) {
+    throw new Error(
+      'Cannot create tgpu.derived objects at the resolution stage.',
+    );
+  }
+
```
iwoplaza: Binding derived should be possible during resolution, only defining new derived should be outlawed 🔫

iwoplaza (changes requested):

reczkok (approved):
Nice 🛃

iwoplaza (approved):
YUP

PR #1224, opened 2025-05-08
fix: Handle single-line statements in TGSL if/for/while

### LINKED ISSUE #1047: bug: Properly parse single-statement code blocks to WGSL.
Currently, the following function parses to incorrect WGSL:

```typescript
export const f = tgpu['~unstable'].fn([], d.i32).does(() => {
  if (true) return 0;
  return 1;
});
```

This is because `if (true) return 0;` is not valid WGSL syntax, the body must be wrapped in curly braces, like `if (true) { return 0; }`

mhawryluk:
closes #1047


reczkok (approved):
Great addition!

PR #1226, opened 2025-05-09
feat: Allow for...of statement to work with vectors and matrices

### LINKED ISSUE #1106: impr: Implement `[Symbol.iterator]()` for vec instances

mhawryluk:
closes #1106


packages/typegpu/src/data/matrix.ts
```diff
@@ -143,6 +143,14 @@ abstract class mat2x2Impl<TColumn extends v2f>
     this.columns[1].y = value;
   }
 
+  *[Symbol.iterator]() {
+    yield this[0];
+    yield this[1];
+    yield this[2];
+    yield this[3];
+    return undefined;
```
aleksanderkatan: I think that `return undefined` is unnecessary, is there a reason for including this?

aleksanderkatan (approved):
LGTM!

deluksic:
This will indeed work, just fyi the current implementation of vector already supports iteration due to it extending from `Array`. I believe the only issue was typescript. It should be possible to `const [x, y] = vec2()` and have `x` and `y` be `number` (not `number | undefined`).

packages/typegpu/src/data/matrix.ts
mhawryluk: without it the function's return type is void instead of undefined and there's a type error (not assignable to the property in base class)

mhawryluk: nevermind. that's only for vectors, cause they extend Array

iwoplaza (approved):
Awesome change! 🚀

PR #1228, opened 2025-05-09
feat: Use accessor for OKLab alpha

iwoplaza:


packages/typegpu-color/src/index.ts
```diff
@@ -4,7 +4,7 @@ export { linearToSrgb, srgbToLinear } from './srgb.ts';
 export {
   linearRgbToOklab,
   oklabGamutClip,
-  oklabGamutClipAlphaSlot,
+  oklabGamutClipAlphaAccess,
```
mhawryluk, line 7: doesn't follow our Naming Convention guide, but okayy

mhawryluk (approved):
Great! 🌈

apps/typegpu-docs/src/content/examples/simple/oklab/index.ts
```diff
@@ -39,108 +39,109 @@ document.addEventListener(
...
-)((uv, _clipLab) => {
-  const suv = floor(mul(20, uv));
-  return suv.x + suv.y - 2 * floor((suv.x + suv.y) * 0.5);
-});
+// #region Patterns
+
+const patternFn = tgpu['~unstable'].fn([d.vec2f, d.vec3f], d.f32);
+
+const patternCheckers = patternFn(
+  (uv, _clipLab) => {
```
mhawryluk:
```suggestion
  (uv) => {
```

apps/typegpu-docs/src/content/examples/simple/oklab/index.ts
```diff
@@ -39,108 +39,109 @@ document.addEventListener(
...
-  return select(clamp(Lgrid / fwidth(Lgrid), 0, 1), 1, thickness < 0.0002);
-}`;
+const patternL0ProjectionLines =
+  patternFn /* wgsl */`(uv: vec2f, clipLab: vec3f) -> f32 {
+    let thickness = fwidth(clipLab.x);
+    let Lgrid = 0.02 - thickness - abs(clipLab.x % 0.04 - 0.02);
+    return select(clamp(Lgrid / fwidth(Lgrid), 0, 1), 1, thickness < 0.0002);
+  }`;
+
+const patternSolid = patternFn((_uv, _clipLab) => {
```
mhawryluk:
```suggestion
const patternSolid = patternFn(() => {
```

iwoplaza:
@deluksic I used something called an "accessor" in place of the slot in the case of the `alpha` value. There's a slight distinction between a slot and an accessor:
- slots are mechanisms of injecting compile-time resources/js values
- accessors are mechanisms of injecting runtime resources/wgsl values

What this means in practice is, at the use-site, we only care that we're getting a `f32` value, not the means of retrieving it. The call-site can decide if the value is computed by a function that returns `f32`, or if it's a reference to a uniform/buffer.

apps/typegpu-docs/src/content/examples/simple/oklab/index.ts
```diff
@@ -39,108 +39,100 @@ document.addEventListener(
...
-  pattern: patternSlot,
+  return d.vec4f(select(color, mul(patternScaled, color), outOfGamut), 1);
 });
 
-const alphaFromUniforms = tgpu['~unstable'].fn(
-  [],
-  d.f32,
-)(() => uniformsBindGroupLayout.bound.uniforms.value.alpha);
+const alphaFromUniforms = tgpu['~unstable'].fn([], d.f32)(
+  () => layout.$.uniforms.alpha,
```
deluksic, line 113: oh this is cool, I didn't know about this shorthand

deluksic:
Looks great, thanks!

PR #1231, opened 2025-05-12
docs: TGSL guide

### LINKED ISSUE #1167: docs: TGSL guide

mhawryluk:
closes #1167


apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,137 @@
...
+which is then used to generate equivalent WGSL.
+
+You can check the current state of supported JavaScript syntax in the [tinyest-for-wgsl repository](https://github.com/software-mansion/TypeGPU/blob/release/packages/tinyest-for-wgsl/src/parsers.ts).
+
+For the TGSL functions to work, you need to use either the dedicated build plugin -- [unplugin-typegpu](/TypeGPU/tooling/unplugin-typegpu), 
+or the Just-In-Time transpiler -- *tgpu-jit* (WIP).
+
+## Usage
+
+You can pass TGSL to tgpu function shell implementation calls, just as you would with the WGSL code strings.
```
aleksanderkatan:
```suggestion
Instead of using a WGSL code string, you can pass TGSL to the tgpu function shell as an argument instead.
```
Iirc, we discourage people passing the WGSL string as an argument and encourage the tagged function syntax instead

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,137 @@
...
+Functions from the WGSL standard library (*distance, arrayLength, workgroupBarrier...*) are accessible through the `typegpu/std` endpoint. 
+The package also includes functions for vector and matrix operators (*add, eq, lt...*).
```
aleksanderkatan, line 23: Do we plan on adding and maintaining a more comprehensive list, or do we just assume that users will see the IDE hints and that's enough? I guess we could also add a link like [this](https://github.com/software-mansion/TypeGPU/blob/release/packages/typegpu/src/std/index.ts) akin to the tinyest link in the paragraph above

reczkok (approved):
I like it! I think that it would be worth including the actual bind group layout code in the example that uses it. Also we could draw a parallel to CUDA __host__ __device__ with our kernel & js utility functions. Maybe that will be useful to some people (and some credibility for us). The second one is just a suggestion tho.

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,137 @@
...
+})(() => backgroundColorUniform.value);
+//                              ^?
+
+const fragmentWgsl = tgpu['~unstable'].fragmentFn({
+  out: d.location(0, d.vec4f),
+})`{ 
+  return backgroundColorUniform; 
+}`.$uses({ backgroundColorUniform });
+```
+
```
aleksanderkatan, line 136:
```suggestion

* **Operators** --
JavaScript does not support operator overloading. 
This means that, while you can still use operators for numbers, 
you have to use supplementary functions from `typegpu/std` (*add, mul, eq, lt, ge...*) for operations involving vectors and matrices.

```

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,137 @@
...
+  const color = oklabToRgb(lab);
+
+  const patternScaled = patternSlot.value(input.uv, clipLab) * 0.1 + 0.9;
+
+  return d.vec4f(select(color, mul(patternScaled, color), outOfGamut), 1);
+});
+```
+
+## What to keep in mind
+
```
aleksanderkatan, line 99:
```suggestion

* **TGSL limitations** --
For a function to be valid TGSL, it must consist only of supported JS syntax (again, see [tinyest-for-wgsl repository](https://github.com/software-mansion/TypeGPU/blob/release/packages/tinyest-for-wgsl/src/parsers.ts)), possibly including references to bound buffers, constant variables defined outside of the function, other TGSL functions etc.
This means that, for example, `console.log()` calls will not work on the GPU.

```
This point follows clearly from the paragraphs above, but I think it still may help users better organize their knowledge about TGSL.

aleksanderkatan (approved):
I have some suggestions, but it looks good to me!

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
mhawryluk: ultimately we will have the automatic API reference, but I will add the link to repo for now

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,159 @@
+---
+title: TGSL
+description: Guide on using JavaScript for WebGPU function definitions.
+---
+
+:::caution[Experimental]
+This feature is under heavy development and is yet to reach stability.
+:::
+
+**TGSL (TypeGPU Shading Language)** is a subset of JavaScript that can be used for defining functions run on the GPU via TypeGPU.
```
iwoplaza: Small suggestion
```suggestion
**TGSL (TypeGPU Shading Language)** is a subset of JavaScript used to define functions that run on the GPU via TypeGPU.
```

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,159 @@
...
+      d.i32(obs.center.y) + d.i32(obs.size.y) / 2,
+    );
+
+    if (x >= minX && x <= maxX && y >= minY && y <= maxY) {
+      return true;
+    }
+  }
+
+  return false;
+});
```
iwoplaza: Did a small round of clean-up to the "Fluid (double-buffering)" example, to make it look a bit more pleasant 🧽🫧

```suggestion
const isInsideObstacle = tgpu['~unstable'].fn([d.i32, d.i32], d.bool)(
  (x, y) => {
    for (let obsIdx = 0; obsIdx < MAX_OBSTACLES; obsIdx++) {
      const obs = obstaclesReadonly.value[obsIdx];

      if (obs.enabled === 0) {
        continue;
      }

      const minX = std.max(0, obs.center.x - d.i32(obs.size.x / 2));
      const maxX = std.min(gridSize, obs.center.x + d.i32(obs.size.x / 2));
      const minY = std.max(0, obs.center.y - d.i32(obs.size.y / 2));
      const maxY = std.min(gridSize, obs.center.y + d.i32(obs.size.y / 2));

      if (x >= minX && x <= maxX && y >= minY && y <= maxY) {
        return true;
      }
    }

    return false;
  },
);
```

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,159 @@
...
+    enabled: d.u32,
+  }), 4))
+  .$usage('storage')
+  .as('readonly');
+
+const isInsideObstacle = tgpu['~unstable'].fn(
+  [d.i32, d.i32],
+  d.bool,
+)((x, y) => {
+  for (let obsIdx = 0; obsIdx < MAX_OBSTACLES; obsIdx++) {
```
iwoplaza: Could we define `MAX_OBSTACLES` somewhere?

Maybe:
```ts
const MAX_OBSTACLES = 4;
```

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -0,0 +1,159 @@
...
+
+```ts
+import tgpu from 'typegpu';
+import * as d from 'typegpu/data';
+import * as std from 'typegpu/std';
+
+const obstaclesReadonly = root
+  .createBuffer(d.arrayOf(d.struct({
+    center: d.vec2u,
+    size: d.vec2u,
```
iwoplaza: Required by the changes to the function implementation
```suggestion
    center: d.vec2i,
    size: d.vec2i,
```

iwoplaza (approved):
T G S L

PR #1233, opened 2025-05-12
feat: Function to extract argument information from a wgsl function code string

aleksanderkatan:


aleksanderkatan:
Notes:
- some of the decisions, like making `lastParsed` a property instead of just returning it from other methods, and defining all of the character sets outside of the functions, were made with optimizations in mind,
- I considered using arrays instead of sets because of their small size, but some rudimentary benchmarks at benchjs.com showed that there is no clear winner,
- some invalid WGSL code becomes valid, for example putting a comment in the middle of an identifier does not throw an error, but I do not think that is an issue,
- range includes trailing whitespaces because it required less code,
- in WGSL not including a return type is actually valid syntax for void functions, so we might want to change the result of `extractArgs` accordingly.

packages/typegpu/tests/extractArgs.test.ts
```diff
@@ -0,0 +1,335 @@
...
+    expect(ret).toStrictEqual(createReturn([], 'f32'));
+    expect(range).toStrictEqual({ begin: 13, end: 85 });
+  });
+
+  it('extracts when missing argument types', () => {
+    const wgslFn = /* wgsl */ `
+      fn add(a, b: f32) -> f32 {
+        return a + b;
+      }
+    `;
```
reczkok, line 232: Why do we handle that? I don't think that is valid WGSL.

reczkok: >A [function](https://www.w3.org/TR/WGSL/#function-function) declaration appears at module-scope. A function declaration contains declarations for formal parameters, if it has any.

->

>A formal parameter [declaration](https://www.w3.org/TR/WGSL/#declaration) specifies an [identifier](https://www.w3.org/TR/WGSL/#identifier) name and a type for a value that [must](https://www.w3.org/TR/WGSL/#shader-creation-error) be provided when invoking the function.

iwoplaza: We want to accept a superset of WGSL, a more loose one in this regard. This is so that devs can omit data types in WGSL defined strings, and we'll generate them.

packages/typegpu/src/extractArgs.ts
```diff
@@ -0,0 +1,266 @@
...
+          return this.#pos;
+        }
+      }
+      this.#pos += 1;
+    }
+    throw new Error('Reached the end of the string without finding a match!');
+  }
+}
+
+// sets and tuples of characters defined here so they are not recreated on every call
```
mhawryluk:
```suggestion
```

packages/typegpu/src/extractArgs.ts
mhawryluk, line 1: let's move this file to core/function

packages/typegpu/tests/extractArgs.test.ts
```diff
@@ -0,0 +1,335 @@
...
+
+// for better formatting and less boilerplate
+function createArg(
+  identifier: string,
+  attributes: string[],
+  type: string | undefined,
+) {
+  return { identifier, attributes, type };
+}
+function createReturn(
```
mhawryluk, line 330:
```suggestion
}

function createReturn(
```

mhawryluk (approved):
that's beautiful, good job!!

packages/typegpu/tests/extractArgs.test.ts
```diff
@@ -0,0 +1,335 @@
...
+  attributes: string[],
+  type: string,
+) {
+  return { attributes, type };
+}
```
reczkok, line 335: Could we move them to the beginning of the file? I think it will make them more discoverable if we ever extend/modify these tests.

packages/typegpu/src/core/function/extractArgs.ts
```diff
@@ -0,0 +1,266 @@
...
+interface ArgInfo {
+  identifier: string;
+  attributes: string[];
+  type: string | undefined;
+}
+
+type ReturnInfo = {
+  attributes: string[];
+  type: string;
+} | undefined;
```
iwoplaza: I would prefer the definition be always defined, and mark it as optional on the use-site.

```suggestion
interface FunctionArgsInfo {
  args: ArgInfo[];
  ret: ReturnInfo | undefined;
  range: {
    begin: number;
    end: number;
  };
}

interface ArgInfo {
  identifier: string;
  attributes: string[];
  type: string | undefined;
}

type ReturnInfo = {
  attributes: string[];
  type: string;
};
```

reczkok (commented):
Nice 👍 One thing I would consider would be adding a utility method like:
```ts
consume(str: string): void {
  if (!this.isAt(str)) {
    throw new Error(
      `Expected '${str}' at position ${this.#pos}, but found '${
        this.str.slice(this.#pos, this.#pos + str.length)
      }'`,
    );
  }
  this.advanceBy(str.length);
}
```
This would mean we don't need the comment explaining the advancing and have free validation.

PR #1246, opened 2025-05-13
fix: Remove experimental object args API

iwoplaza:


packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -46,22 +43,18 @@ export function createFnCore(
...
-          implementation,
-          shell.returnType,
-          (externals) => externalsToApply.push(externals),
-        );
-      }
+      addArgTypesToExternals(
+        implementation,
+        Array.isArray(shell.argTypes)
+          ? shell.argTypes
+          : Object.values(shell.argTypes),
```
reczkok:
```suggestion
        shell.argTypes
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -355,15 +355,7 @@ export function generateExpression(
...
-        : Object.entries(argTypes).map(([key, type]) => {
-          const sn = (argSnippets[0]?.value as Record<string, Snippet>)[
-            key
-          ];
-          if (!sn) {
-            throw new Error(`Missing argument ${key} in function call`);
-          }
-          return [type, sn] as [wgsl.AnyWgslData, Snippet];
-        });
+        : [];
```
reczkok: Since now there are only 2 options we can simplify this a bit (couldn't drop a suggestion so here's the code). The filter is redundant since if the snippet is falsey the code will be invalid.
```ts
const pairs =
  (Array.isArray(argTypes) ? argTypes : (argTypes(...resolvedSnippets)))
    .map((type, i) => [type, resolvedSnippets[i]]) as [
      wgsl.AnyWgslData,
      Snippet,
    ][];
```

reczkok (approved):
Looks good 👍

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: @aleksanderkatan Do you want to take this one, or should I?

PR #1247, opened 2025-05-13
docs: Update tgpu.fn API

iwoplaza:


aleksanderkatan (approved):
nice!

aleksanderkatan (changes requested):
Actually, the `Build Plugin` section also uses the old function API, could you update it as well?

aleksanderkatan (approved):
Good job!

PR #1251, opened 2025-05-14
fix: Use a more compatible texture format for stable-fluids example

reczkok:
It seems safari on iOS does not have the `float32-filterable` feature. Since the simulation will be just fine with a little less precision this PR moves all textures to `rgba16float` which is guaranteed to be both filterable and bindable as storage on all devices.


mhawryluk (approved):
🌊

reczkok:
Sorry for the re-review but i just realized there was no touch controls handling 🤦‍♂️

iwoplaza (approved):
🪼🪼🪼

PR #1257, opened 2025-05-16
feat: Allow destructuring arguments in TGSL function implementations

### LINKED ISSUE #1242: feat: Allow destructuring arguments in entry function TGSL implementations

### LINKED ISSUE #1270: fix: Type cast error for unknown type
error resolves the types, fails with unknown and a resolution error is thrown instead

```ts
throw new Error(
  `Cannot convert ${ctx.resolve(sn.dataType)} to ${
    ctx.resolve(type)
  }`,
);
```

### LINKED ISSUE #1271: fix:  Automatic type cast with function arguments
in `typegpu-confetti`:

```ts
const centerGravity: GravityFn = (args) => {
  'kernel';
  return std.mul(2, d.vec2f(d.f32(-args.pos.x), d.f32(-args.pos.y)));
};
```
explicit casting is required, cause argument type is unknown ://///

mhawryluk:
closes #1242
closes #1270 
closes #1271


reczkok (approved):
Very nice feature to have! I was a little worried about shadowing but it works well 🙇‍♂️

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -178,12 +184,36 @@ export function createFnCore(
                 : undefined) ?? `arg_${i}`,
               dataType: arg as AnyWgslData,
             }))
-            : []
+            : [
+              {
+                value: 'in',
```
aleksanderkatan: I'm not sure on how our variable naming strategy works, could this generate a problem if the user creates a variable called `in` inside of the function? Maybe we should prepend this identifier with an underscore?

mhawryluk: I believe that because `in` is a reserved keyword in javascript and not in wgsl this shouldn't be a problem

mhawryluk: changed to _arg_0, _arg_1...

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -192,7 +195,13 @@ class ItemStateStackImpl implements ItemStateStack {
           return arg;
         }
 
+        const argAlias = layer.argAliases[id];
+        if (argAlias !== undefined) {
+          return argAlias;
+        }
```
reczkok:
```suggestion
        if (layer.argAliases[id]) {
          return layer.argAliases[id];
        }
```

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -119,52 +121,36 @@ export function createFnCore(
...
+                  alias,
+                  {
+                    value: `_arg_${i}.${name}`,
+                    dataType: (shell.argTypes[i] as AnyWgslStruct)
+                      .propTypes[name] as AnyWgslData,
+                  },
+                ])
+                : []
+            ),
+          ),
```
reczkok, line 153: I'm gonna hang this diff on my wall😍

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -452,48 +453,43 @@ export function extractFunctionParts(rootNode: JsNode): {
...
   return {
-    params: {
-      type: 'identifiers',
-      names: functionNode.params.flatMap((x) =>
-        x.type === 'Identifier' ? [x.name] : []
-      ),
-    },
+    params: functionNode.params.filter((param) =>
+      param.type === 'ObjectPattern' || param.type === 'Identifier'
+    ).map((param) =>
```
reczkok: Instead of filtering here we could check if there are any other and have a nice early error

reczkok (approved):
Amazing change! 💜

aleksanderkatan (approved):
Great!!

iwoplaza (approved):
Wooh, this is really nice! A pleasure to read 👏

PR #1258, opened 2025-05-16
feat: Migrate unplugin from using `tgpu.__assignAst` to `globalThis`

aleksanderkatan:


packages/typegpu/src/name.ts
```diff
@@ -1,9 +1,19 @@
...
 }
 
+export interface FunctionMetaData extends MetaData {
+  ast: {
+    argNames: ArgNames;
+    body: Block;
+    externalNames: string[];
+  };
+  externals?: Record<string, unknown> | undefined;
+}
```
iwoplaza: I think we can store all metadata on a single object, as optional keys & values. It's going to be easier to add existing metadata to new types of resources, and new types of metadata.

```ts

export interface MetaData {
  name?: string | undefined;
  ast?: {
    argNames: ArgNames;
    body: Block;
    externalNames: string[];
  } | undefined;
  externals?: Record<string, unknown> | undefined;
}
```

packages/typegpu/src/name.ts
```diff
@@ -1,9 +1,19 @@
+import { ArgNames, Block } from 'tinyest';
```
iwoplaza:
```suggestion
import type { ArgNames, Block } from 'tinyest';
```

packages/typegpu/src/index.ts
```diff
@@ -50,12 +49,6 @@ export const tgpu = {
 };
 export default tgpu;
 
-// Hidden API, used only by tooling (e.g., rollup plugin).
-Object.assign(tgpu, {
-  __assignAst: assignAst,
-  __removedJsImpl: removedJsImpl,
-});
-
```
iwoplaza: We can keep this API for now, and polyfill it for a few minor versions 🙏

packages/typegpu/src/name.ts
iwoplaza, line 1: Can we rename this file to `meta.ts` and move it to `src/shared`?

iwoplaza (changes requested):

aleksanderkatan:
Currently there is a small divergence between babel and rollup, namely in the following case:
```typescript
const sum = add(1, 2);
function add(a, b) {
  'kernel & js';
  return a + b;
};
```
Rollup throws a specialized error explaining that functions are no longer hoisted, and babel does not.

aleksanderkatan:
I have only tested rollup empirically

packages/typegpu/src/shared/meta.ts
```diff
@@ -1,7 +1,14 @@
-import { $getNameForward } from './shared/symbols.ts';
+import type { ArgNames, Block } from 'tinyest';
+import { $getNameForward } from './symbols.ts';
 
-interface MetaData {
-  name: string | undefined;
+export interface MetaData {
+  name?: string;
```
lursz: This may need to remain 'string | undefined'

packages/typegpu/src/core/function/astUtils.ts
```diff
@@ -11,23 +12,18 @@ export type AstInfo = {
...
-const functionToAstMap = new WeakMap<
-  (...args: unknown[]) => unknown,
-  AstInfo
->();
-
 export function getPrebuiltAstFor(
   fn: (...args: unknown[]) => unknown,
 ): AstInfo | undefined {
-  return functionToAstMap.get(fn);
+  return getMetaData(fn) as AstInfo;
```
mhawryluk: we can remove getPrebuiltAstFor and AstInfo

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -40,42 +40,68 @@ function getKernelDirective(
...
-      directive === 'kernel & js'
-        ? node
-        : template.expression`${tgpuAlias}.__removedJsImpl(${
-          name ? `"${name}"` : ''
-        })`(),
-      template.expression`${embedJSON({ argNames, body, externalNames })}`(),
-      types.objectExpression(
-        externalNames.map((name) =>
-          types.objectProperty(types.identifier(name), types.identifier(name))
+  const newNode = types.callExpression(
```
mhawryluk: you can just return it without assigning to a constant

packages/typegpu/src/shared/meta.ts
iwoplaza:
```suggestion
  name?: string | undefined;
```

packages/typegpu/src/shared/meta.ts
```diff
@@ -1,7 +1,14 @@
...
+  ast?: {
+    argNames: ArgNames;
+    body: Block;
+    externalNames: string[];
+  };
+  externals?: Record<string, unknown>;
```
iwoplaza:
```suggestion
  ast?: {
    argNames: ArgNames;
    body: Block;
    externalNames: string[];
  } | undefined;
  externals?: Record<string, unknown> | undefined;
```

packages/typegpu/src/shared/meta.ts
```diff
@@ -1,7 +1,14 @@
-import { $getNameForward } from './shared/symbols.ts';
+import type { ArgNames, Block } from 'tinyest';
+import { $getNameForward } from './symbols.ts';
 
-interface MetaData {
-  name: string | undefined;
+export interface MetaData {
```
iwoplaza, line 4: Could we also store versioning information in the `MetaData` type, to be injected by `unplugin-typegpu`?

```ts
interface MetaData {
  ast?: {
    v: number, // e.g. 11
    // ...
  } | undefined,
  // ...
}
```
The value in `ast.v` would have to be a constant exported from `tinyest` that can be used to differentiate between breaking changes in the format.

iwoplaza (changes requested):

mhawryluk (approved):
cool!

packages/typegpu/src/shared/meta.ts
aleksanderkatan: Is that really what we want? Both in babel and in rollup, if we pass externals and ast, they are defined. In this approach, we need to check whether `'prop' in meta && !!meta.prop`, instead of just `'prop' in meta`.

iwoplaza: Not really, we can check just `!!meta.prop`, since the property can exist on the type, and TS knows it.
Reproduction: https://www.typescriptlang.org/play/?exactOptionalPropertyTypes=true#code/C4TwDgpgBAshwEMAmDFQLxQN4CgpQQGdgB+ALmwIoEYoBfKAHygFcA7JCAMwEs2IkAGhx0A3DhycAxgBsEAJ2hSA9m2JQAtvAQU4iFInE4eXKAAotiAHRFgASmx4oKtcpkQrM5QHML2m8Q2dqJQAPShUIA8G4CR+yI4QA

PR #1259, opened 2025-05-19
impr: Add proper restrictions for Vertex stage builtins

### LINKED ISSUE #827: Add proper restrictions for Vertex and Fragment stage builtins
Currently `vertexFn` and `fragmentFn` accept any builtins. We should restrict them based on stage and direction.

mhawryluk:
closes #827


reczkok (approved):
Looks good 👍 The errors are unfortunately not the most readable but I don't think we can do much about that at the moment.

iwoplaza (approved):
Awesome change! 👏

Issue #1087, opened 2025-03-27
Improve `std.arrayLength` to work on non-dynamic arrays as well


iwoplaza:
`std.arrayLength(...)` currently wraps the passed in Snippet in a call to `arrayLength` in WGSL. This however breaks for arrays of known-size (schema.length > 0), so we should branch out for static arrays and just resolve to the length of schema.

This has already been solved for the `.length` property in arrays in the `wgslGenerator.ts` file:

```ts
if (wgsl.isWgslArray(target.dataType)) {
  if (property === 'length') {
    if (target.dataType.elementCount === 0) {
      // Dynamically-sized array
      return {
        value: `arrayLength(&${ctx.resolve(target.value)})`,
        dataType: d.u32,
      };
    }

    return {
      value: String(target.dataType.elementCount),
      dataType: abstractInt,
    };
  }
}
```

PR #1261, opened 2025-05-19
feat: improved std.arrayLength

lursz:
Fix to #1087 
std.arrayLength now works on non-dynamic arrays as well


iwoplaza (approved):
Awesome! ❇️❇️❇️

PR #1281, opened 2025-05-21
fix: "Cannot determine size" error in Benchmarks

aleksanderkatan:
The symbol tag for structs was bundled into two different files and therefore some checks on equality with that symbol were failing.


iwoplaza (approved):
Yessir 🫡

PR #1292, opened 2025-05-25
feat(typegpu/noise): 2D/3D Perlin Noise

iwoplaza:


apps/typegpu-docs/src/content/examples/rendering/perlin-noise/index.html
```diff
@@ -0,0 +1 @@
+<canvas></canvas>
```
aleksanderkatan, line 1: I think this example could be made to use the fit-to-container canvas, but that may be done in a different PR. This, and a speed slider

aleksanderkatan (commented):
The example is great visually and the perlin noise lib looks very useful!

Issue #1296, opened 2025-05-26
bug: Resolve struct names when `tgpu.fn` is created from raw WGSL.


aleksanderkatan:
Currently, in functions we resolve the struct name either from externals or from the return type, for example:
```typescript
const Point = d.struct({ a: d.u32, b: d.u32 }).$name('P');

const func = tgpu['~unstable']
  .fn([Point], Point)(
    /* wgsl */ `(a: MyArbitraryNameForAStruct) -> MyArbitraryNameForAStruct {
    return MyArbitraryNameForAStruct;
  }`,
  );
```
In generated wgsl, `MyArbitraryNameForAStruct` will be replaced by a name based on the label `P`.
This behavior is not present when the struct type is used as a parameter, and the following code is resolved incorrectly:
```typescript
const func = tgpu['~unstable']
  .fn([d.arrayOf(Point, 4)])(
    /* wgsl */ `(a: MyStruct) : u32 {
    return a.a;
  }`,
  );
```
Of course, the same goes for more complex uses, like `param : array<MyStruct, 4>`.

Right now, an incorrect wgsl code is generated. I think that we should either expand on this functionality, or get rid of it all together and force the users to name their structs in externals. What do you think @reczkok @mhawryluk @lursz @iwoplaza?

PR #1306, opened 2025-05-27
feat: Allow omitting types in wgsl header

aleksanderkatan:


packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -95,12 +97,35 @@ export function createFnCore(
...
+            throw new Error(
+              `WGSL implementation has ${providedArgs.args.length} arguments, while the shell has ${shell.argTypes.length} arguments!`,
+            );
+          }
+
+          const input = providedArgs.args.map((argInfo, i) =>
+            `${argInfo.identifier}: ${
+              argInfo.type ?? ctx.resolve(shell.argTypes[i])
+            }`
+          ).join(', ');
```
reczkok: Couldn't we get some early errors here? I assume structs might pose an issue but we should have enough info at this point? We could "apply externals" to the argInfo.type and then compare with the resolved type.

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -95,12 +97,35 @@ export function createFnCore(
               }`
             : '';
           header = `${input} ${output} `;
+          body = implementation;
+        } else {
+          const providedArgs = extractArgs(implementation);
+
+          if (providedArgs.args.length !== shell.argTypes.length) {
+            throw new Error(
+              `WGSL implementation has ${providedArgs.args.length} arguments, while the shell has ${shell.argTypes.length} arguments!`,
```
reczkok: Let's not shout at the user (I'm guilty too)
```suggestion
              `WGSL implementation has ${providedArgs.args.length} arguments, while the shell has ${shell.argTypes.length} arguments.`,
```

packages/typegpu/src/core/function/fnCore.ts
aleksanderkatan: I will add a check that omits structs, and add an issue for the structs case

aleksanderkatan: I added the check and it turns out it works with structs (except structs not present in externals, that are nested in some other type, see #1296)

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -176,3 +215,28 @@ export function createFnCore(
...
 }
+
+function checkAndReturnType(
+  name: string,
+  wgslType: string | undefined,
+  jsType: unknown,
+  ctx: ResolutionCtx,
+  externals: ExternalMap,
+) {
+  const resolvedJsType = ctx.resolve(jsType).replace(/\s/g, '');
```
iwoplaza, line 226: Why are we removing whitespace here? 👀

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -176,3 +215,28 @@ export function createFnCore(
...
+  }
+
+  const resolvedWgslType = replaceExternalsInWgsl(ctx, externals, wgslType)
+    .replace(/\s/g, '');
+
+  if (resolvedWgslType !== resolvedJsType) {
+    throw new Error(
+      `Type mismatch between TGPU shell and WGSL code string: ${name}, JS type "${resolvedJsType}", WGSL type "${resolvedWgslType}".`,
+    );
+  }
```
iwoplaza, line 238: Awesome, great messaging! 🙌

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -176,3 +215,28 @@ export function createFnCore(
...
+  ctx: ResolutionCtx,
+  externals: ExternalMap,
+) {
+  const resolvedJsType = ctx.resolve(jsType).replace(/\s/g, '');
+
+  if (!wgslType) {
+    return resolvedJsType;
+  }
+
+  const resolvedWgslType = replaceExternalsInWgsl(ctx, externals, wgslType)
```
iwoplaza: If we could run `replaceExternalsInWgsl` on the implementation before doing any of the header substitution, we would reduce the amount of times it needs to be called to `1`.

iwoplaza (changes requested):

packages/typegpu/src/core/function/fnCore.ts
aleksanderkatan: In one of the tests, one of these parsed to `array<P, 4>` and the other parsed to `array<P,4>`. Just in case, I removed whitespaces from both of these.

packages/typegpu/src/core/function/fnCore.ts
aleksanderkatan: From what I see, we need to call it a second time for header in entry point function case. Is there a way to easily extract what `In` and `Out` types should be, or is calling the following the easiest option?
```typescript
header = replaceExternalsInWgsl(
  ctx,
  externalMap,
  `${input} ${output} `,
);
```

packages/typegpu/src/core/function/fnCore.ts
iwoplaza: Ah, I see. Let's leave it then

packages/typegpu/src/core/function/fnCore.ts
iwoplaza: > Is there a way to easily extract what In and Out types should be (...)

There is!
```diff
- const input = isWgslStruct(shell.argTypes[0]) ? '(in: In)' : '()';
+ const input = isWgslStruct(shell.argTypes[0]) ? `(in: ${ctx.resolve(shell.argTypes[0])})` : '()';
```

```diff
  const output = shell.returnType !== Void
    ? isWgslStruct(shell.returnType)
-     ? '-> Out'
+     ? `-> ${ctx.resolve(shell.returnType)}`
      : `-> ${attributes !== '' ? attributes : '@location(0)'} ${
        ctx.resolve(shell.returnType)
      }`
    : '';
```

aleksanderkatan: Nice!

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -176,3 +218,27 @@ export function createFnCore(
 
   return core;
 }
+
+function checkAndReturnType(
+  name: string,
+  wgslType: string | undefined,
+  jsType: unknown,
+  ctx: ResolutionCtx,
+  externals: ExternalMap,
```
iwoplaza: Let's remove `externals` as it's not used anymore, and put the `ctx` at the top, for consistency.
```suggestion
  ctx: ResolutionCtx,
  name: string,
  wgslType: string | undefined,
  jsType: unknown,
```

PR #1311, opened 2025-05-28
feat: Cast elements to correct type when creating vectors

aleksanderkatan:


packages/typegpu/tests/vector.test.ts
```diff
@@ -834,13 +850,13 @@ describe('v3i', () => {
 describe('v4f', () => {
   describe('(v3f, number) constructor', () => {
     it('works in JS', () => {
-      const red = d.vec3f(0.9, 0.2, 0.1);
+      const red = d.vec3f(0.125, 0.25, 0.375);
```
aleksanderkatan, line 870: There is a little side effect of the float cast we do with the `Float32Array`
image

aleksanderkatan:
All those checks seem to have affected the performance a bit

image

packages/typegpu/src/data/numeric.ts
```diff
@@ -22,17 +22,47 @@ export const abstractFloat = {
...
+ * Can also be called to cast a value to an bool in accordance with WGSL casting rules.
+ *
+ * @example
+ * const value = bool(); // false
+ * @example
+ * const value = bool(0); // false
+ * @example
+ * const value = bool(-0); // false
+ * @example
+ * const value = bool(21.37); // true
```
reczkok, line 53: xd

packages/typegpu/src/data/numeric.ts
```diff
@@ -22,17 +22,47 @@ export const abstractFloat = {
...
+  // GPU implementation
+  (v) => {
+    return snip(`bool(${v?.value ?? ''})`, bool);
+  },
+);
+
 /**
  * A schema that represents a boolean value. (equivalent to `bool` in WGSL)
+ *
+ * Can also be called to cast a value to an bool in accordance with WGSL casting rules.
```
mhawryluk:
```suggestion
 * Can also be called to cast a value to a bool in accordance with WGSL casting rules.
```

packages/typegpu/src/data/numeric.ts
```diff
@@ -46,7 +76,9 @@ const u32Cast = createDualImpl(
     return Math.max(0, Math.min(0xffffffff, Math.floor(v)));
   },
   // GPU implementation
-  (v) => snip(`u32(${v.value})`, u32),
+  (v) => {
+    return snip(`u32(${v?.value ?? ''})`, u32);
+  },
```
mhawryluk: why `=> { return ...; }` instead of `=> ...,`?

mhawryluk (approved):
Well done!! 🚀

reczkok (commented):
I'm really on the fence about this — I get that the consistency between JS and WGSL is nice, but is it really worth the performance hit (which, as we've established, is kind of important)?

reczkok:
> I'm really on the fence about this — I get that the consistency between JS and WGSL is nice, but is it really worth the performance hit (which, as we've established, is kind of important)?

Maybe we could leave this in and figure out a way to provide users with a very fast mass vector constructor

aleksanderkatan:
I think we could modify the schema casts for scalar values, so that `f32` and `f16` are identity, `i32` is `Math.floor(n)`, `u32` is `Math.max(0, Math.floor(n))`, and `bool` is left as is (all plus the undefined check). This would probably increase the performance and leave the most important part of the cast in place. What do you think?

reczkok:
I think we should keep the casting logic to stay compliant with WGSL's behavior. I’m actually curious how much of the performance hit comes from the casting itself versus the overhead of the function calls.

packages/typegpu/src/data/numeric.ts
aleksanderkatan: mental missclick

aleksanderkatan:
@reczkok seems like most of the overhead is in Math functions and array writes.

Before any changes:
image

When calling schema constructors, but all they do is return the value passed in:
image

When calling schema constructors, and casting all values properly:
image

aleksanderkatan:
The benchmark results are now wildly inconsistent. For one test, I get 30 ops/s five times in a row, then I get 11 ops/s five times in a row...

I added the coercion to the vector element assignment (thgough `[0]` and through `.x`). I also removed the float value casts, and I simplified a bit the integer casts.

packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -1,11 +1,14 @@
 import { $internal } from '../shared/symbols.ts';
 import type { SelfResolvable } from '../types.ts';
+import { bool, f16, f32, i32, u32 } from './numeric.ts';
 import type { VecKind } from './wgslTypes.ts';
 
 // deno-fmt-ignore
 export abstract class VecBase<S> extends Array implements SelfResolvable {
   public readonly [$internal] = true;
   abstract get kind(): VecKind;
+  abstract getElementSchema(): (v?: S) => S;
```
iwoplaza, line 10: Why is this a function? Can be a getter.

```suggestion
  abstract get elementSchema(): (v?: S) => S;
```

packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -375,15 +378,30 @@ type Tuple4<S> = [S, S, S, S];
...
   declare readonly length = 2;
-  abstract getDefaultValue(): S;
 
-  0: S;
-  1: S;
+  e0: S;
+  e1: S;
 
-  constructor(x?: S, y?: S) {
+  constructor(x?: S, y?: S, z?: S) {
```
iwoplaza: z? 👀 
```suggestion
  constructor(x?: S, y?: S) {
```

packages/typegpu/src/data/vectorImpl.ts
aleksanderkatan: It was meant to be a getter but I don't think this is possible, unless I am missing something
image

iwoplaza: Aah, I see, I did not know about this limitation. It's good then!

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -704,7 +704,8 @@ export interface Bool {
 
   // Type-tokens, not available at runtime
   readonly [$repr]: boolean;
-  // ---
```
iwoplaza, line 707: Why remove this?

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -762,7 +763,17 @@ export interface U32 {
...
   // ---
 
-  (v: number | boolean): number;
+  (v?: number | boolean): number;
+}
+
+/**
+ * Unsigned 16-bit integer schema used exclusively for index buffer schemas.
+ */
+
```
iwoplaza: Why the space here?

packages/typegpu/tests/numeric.test.ts
```diff
@@ -48,3 +50,37 @@ describe('f16', () => {
...
+  it('works for default constructors', () => {
+    const main = tgpu['~unstable']
+      .fn([])(() => {
+        const f = d.f32();
+        const h = d.f16();
+        const i = d.i32();
+        const u = d.u32();
+        const b = d.bool();
+      })
+      .$name('main');
```
iwoplaza: Not necessary anymore
```suggestion
      });
```

packages/typegpu/src/data/wgslTypes.ts
aleksanderkatan: Botched merge

iwoplaza (approved):
This is a great step forward 👏

PR #1317, opened 2025-05-30
feat(typegpu/std): implement cosh, acosh, exp2

lursz:
- cosh
- acosh
- exp2


packages/typegpu/src/data/vectorOps.ts
```diff
@@ -307,6 +307,23 @@ export const VectorOps = {
...
+    vec3f: unary3f(Math.acosh),
+    vec3h: unary3h(Math.acosh),
+    vec3i: unary3i(Math.acosh),
+    vec3u: unary3u(Math.acosh),
+
+    vec4f: unary4f(Math.acosh),
+    vec4h: unary4h(Math.acosh),
+    vec4i: unary4i(Math.acosh),
+    vec4u: unary4u(Math.acosh),
+  } as Record<VecKind, <T extends vBase>(v: T) => T>,
```
aleksanderkatan: `acosh` should only accept floats

packages/typegpu/src/std/numeric.ts
```diff
@@ -276,6 +276,28 @@ export const acos = createDualImpl(
...
+  <T extends AnyFloatVecInstance | number>(value: T): T => {
+    if (typeof value === 'number') {
+      return Math.acosh(value) as T;
+    }
+    return VectorOps.acosh[(value as AnyFloatVecInstance).kind](
+      value as never,
+    ) as T;
+  },
+  // GPU implementation
+  (value) => ({ value: `acosh(${value.value})`, dataType: value.dataType }),
```
```suggestion
  (value) => snip(`acosh(${value.value})`, value.dataType),
```

packages/typegpu/src/std/numeric.ts
```diff
@@ -343,6 +365,22 @@ export const cos = createDualImpl(
...
+export const cosh = createDualImpl(
+  // CPU implementation
+  <T extends AnyFloatVecInstance | number>(value: T): T => {
+    if (typeof value === 'number') {
+      return Math.cosh(value) as T;
+    }
+    return VectorOps.cosh[value.kind](value) as T;
+  },
+  // GPU implementation
+  (value) => ({ value: `cosh(${value.value})`, dataType: value.dataType }),
```
```suggestion  
  (value) => snip(`cosh(${value.value})`, value.dataType),
```

packages/typegpu/tests/std/numeric/acosh.test.ts
```diff
@@ -0,0 +1,38 @@
...
+describe('acosh', () => {
+  it('computes acosh of a number', () => {
+    expect(acosh(1)).toBeCloseTo(0);
+    expect(acosh(Math.cosh(1))).toBeCloseTo(1);
+    expect(acosh(Math.cosh(-1))).toBeCloseTo(1);
+  });
+
+  it('computes acosh of vec2f', () => {
+    const input = vec2f(1, Math.cosh(1));
+    const expected = vec2f(Math.acosh(1), Math.acosh(Math.cosh(1)));
```
aleksanderkatan, line 20:
```suggestion
    const expected = vec2f(Math.acosh(1), 1));
```

packages/typegpu/tests/std/numeric/acosh.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    expect(isCloseTo(acosh(input), expected)).toBe(true);
+  });
+
+  it('computes acosh of vec4f', () => {
+    const input = vec4f(1, Math.cosh(1), Math.cosh(-1), Math.cosh(2));
+    const expected = vec4f(
+      Math.acosh(1),
+      Math.acosh(Math.cosh(1)),
+      Math.acosh(Math.cosh(-1)),
+      Math.acosh(Math.cosh(2)),
```
aleksanderkatan:
```suggestion
      1,
      -1,
      2,
```

packages/typegpu/tests/std/numeric/acosh.test.ts
```diff
@@ -0,0 +1,38 @@
...
+    const expected = vec2f(Math.acosh(1), Math.acosh(Math.cosh(1)));
+    expect(isCloseTo(acosh(input), expected)).toBe(true);
+  });
+
+  it('computes acosh of vec3f', () => {
+    const input = vec3f(1, Math.cosh(1), Math.cosh(-1));
+    const expected = vec3f(
+      Math.acosh(1),
+      Math.acosh(Math.cosh(1)),
+      Math.acosh(Math.cosh(-1)),
```
aleksanderkatan:
```suggestion
      1,
      -1,
```

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -1025,6 +1053,17 @@ export const VectorOps = {
...
+    vec2f: unary2f((val) => 2 ** val),
+    vec2h: unary2h((val) => 2 ** val),
+
+    vec3f: unary3f((val) => 2 ** val),
+    vec3h: unary3h((val) => 2 ** val),
+
+    vec4f: unary4f((val) => 2 ** val),
+    vec4h: unary4h((val) => 2 ** val),
+  } as Record<VecKind, <T extends vBase>(v: T) => T>,
+
```
aleksanderkatan, line 1060: We could just use `pow` instead of creating a new entry in the vectorops

aleksanderkatan (approved):
Gj!

packages/typegpu/src/data/vectorOps.ts
lursz: Good catch!

packages/typegpu/tests/std/numeric/acosh.test.ts
lursz: Good point, I calculated actual results and inserted them. I nonetheless left one test to verify acosh(cosh()) works

aleksanderkatan: I feel like this is testing the js standard library, but I guess it can stay as a sanity check

packages/typegpu/src/data/vectorOps.ts
lursz: Good idea, but it's a bit more complicated than that. We'd have to get around implementing a version for each vector size

aleksanderkatan: I guess we could do something along the lines of:
```typescript
export const exp2 = createDualImpl(
  // CPU implementation
  <T extends AnyFloatVecInstance | number>(value: T): T => {
    if (typeof value === 'number') {
      return (2 ** value) as T;
    }
    return VectorOps.pow[value.kind](
      vecTypeToConstructor[value.kind](2),
      value,
    ) as T;
  },
  // GPU implementation
  (value) => snip(`exp2(${value.value})`, value.dataType),
);
```
with `vecTypeToConstructor` exported from `../data/vector.ts`, but this may be a job for an another ticket.

Edit: #1336

packages/typegpu/src/std/numeric.ts
```diff
@@ -276,6 +276,28 @@ export const acos = createDualImpl(
...
+ */
+export const acosh = createDualImpl(
+  // CPU implementation
+  <T extends AnyFloatVecInstance | number>(value: T): T => {
+    if (typeof value === 'number') {
+      return Math.acosh(value) as T;
+    }
+    return VectorOps.acosh[(value as AnyFloatVecInstance).kind](
+      value as never,
+    ) as T;
```
```suggestion
    return VectorOps.acosh[value.kind](value) as T;
```
value as never? 😆 
Could you try to remove as many casts from the trigo functions as you can?

aleksanderkatan (approved):
LGTM!

mhawryluk (approved):
Great 🎉

PR #1327, opened 2025-06-03
docs: Fix control panel height on mobile

mhawryluk:
before: 
Zrzut ekranu 2025-06-3 o 11 24 55

after:
Zrzut ekranu 2025-06-3 o 11 24 17


aleksanderkatan (approved):
Nice fix!

PR #1335, opened 2025-06-04
fix: Small thumbnails not appearing

aleksanderkatan:
To my understanding, the query `w=512;1024` only creates two result images if the width of input image is at least 513, because the only allowed operation by default is downscaling.


apps/typegpu-docs/src/utils/examples/exampleContent.ts
```diff
@@ -83,12 +83,14 @@ const thumbnailFiles = R.pipe(
...
-  R.mapValues((value): ThumbnailPair => ({
-    small: value[0],
-    large: value[1],
-  })),
+  R.mapValues((
+    value,
+  ): ThumbnailPair => (typeof value === 'string'
+    ? { small: value, large: value }
+    : { small: value[0], large: value[1] })
+  ),
```
reczkok: I would probably prefer having a minimum size requirement for the thumbnail size and throw a descriptive error if it's too small. This would guarantee consistent look for the thumbnails

aleksanderkatan: image

iwoplaza (approved):
Let's go 🎉

PR #1338, opened 2025-06-05
fix: Remove redundant InferReturn<T>

iwoplaza:


mhawryluk (approved):
well yes! thanks 🫧

packages/typegpu/src/core/function/fnTypes.ts
```diff
@@ -40,26 +40,13 @@ export type InferArgs<T extends unknown[]> = {
...
-export type InferReturn<T> = T extends undefined
-  // biome-ignore lint/suspicious/noConfusingVoidType: <void is used as a return type>
-  ? void
-  : Infer<T>;
-
 export type JsImplementation<
-  Args extends unknown[] | Record<string, unknown> =
-    | unknown[]
-    | Record<string, unknown>,
+  Args extends unknown[] | [] = unknown[] | [],
```
aleksanderkatan, line 44: Isn't `unknown[] | []` equivalent to `unknown[]`? There is no type error when I change the type to `Args extends unknown[] = unknown[]`

iwoplaza: It's different from TS's point of view in specific scenarios, e.g., when inferring a tuple type. The change here is done for consistency reasons, as removing `| []` in the definition of `Implementation` instead does cause a type error.

PR #1339, opened 2025-06-05
fix: Page scroll on touch events in examples

### LINKED ISSUE #1337: fix: Prevent default in fish-example and function-visualizer mobile event handler
Currently when panning the view on mobile we scroll the entire example

mhawryluk:
closes #1337


reczkok (approved):
Great 👍 I would prevent the default regardless of touches length on the canvas.

PR #1343, opened 2025-06-06
fix: 3D Fish and Gravity improper rendering

aleksanderkatan:


reczkok (approved):
Necessary bloat I guess. Maybe we could create a separate function that loads the fish data? That way we will just need to change one thing back when it's fixed. (If it's difficult without running into the bug it's fine the way it is)

PR #1344, opened 2025-06-09
feat: Automatically assign names to resources in Unplugin

aleksanderkatan:
The high-level rule for autonaming is like this: if there is an assignment to a variable, we try to find one of the `resourceConstructors` among the identifiers in the assigned expression; if we succeed (and we don't find the `$name` identifier), we wrap the expression in a call to `__TYPEGPU_AUTONAME__`. If typegpu is initialized, this function checks if the expression has the `$internal` symbol, and names it (using `$name`) if it isn't already named.

I also needed to add `[$internal]: true` to some of the resources.

Let me know if I missed any of the resource generating functions.


aleksanderkatan:
At least +500, -100 of these changes are tests :)

The high-level rule for autonaming is like this: if there is an assignment to a variable, we try to find one of the `resourceConstructors` among the identifiers in the assigned expression; if we succeed (and we don't find the `$name` identifier), we wrap the expression in a call to `__TYPEGPU_AUTONAME__`. If typegpu is initialized, this function checks if the expression has the `$internal` symbol, and names it if it isn't already named.

I also needed to add `[$internal]: true` to some of the resources.

Let me know if I missed any of the resource generating functions.

aleksanderkatan:
As of right now, functions marked with our directives are not assigned names (no `$name` method, we need to save the names in the metadata). I decided to delegate that in another ticket: #1358

reczkok:
> At least +500, -100 of these changes are tests :)
> 
> The high-level rule for autonaming is like this: if there is an assignment to a variable, we try to find one of the `resourceConstructors` among the identifiers in the assigned expression; if we succeed (and we don't find the `$name` identifier), we wrap the expression in a call to `__TYPEGPU_AUTONAME__`. If typegpu is initialized, this function checks if the expression has the `$internal` symbol, and names it (using `$name`) if it isn't already named.
> 
> I also needed to add `[$internal]: true` to some of the resources.
> 
> Let me know if I missed any of the resource generating functions.

Please add that to the PR description

packages/typegpu/tests/unplugin/autoname.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  it("autonames resources created using root['~unstable']", ({ root }) => {
+    const myPipeline = root['~unstable']
+      .withCompute(
+        tgpu['~unstable'].computeFn({ workgroupSize: [1] })(() => {}),
+      )
+      .createPipeline();
+    const myMutableBuffer = root['~unstable'].createMutable(d.u32);
+    const myReadonlyBuffer = root['~unstable'].createReadonly(d.u32);
+    const myUniformBuffer = root['~unstable'].createUniform(d.u32);
+    const myTexture = root['~unstable'].createTexture({
```
mhawryluk, line 56: add a test for `sampler` and `comparisonSampler` too, and add it to the list of recognized constructors

packages/unplugin-typegpu/test/aliasing.test.ts
```diff
@@ -14,13 +14,13 @@ describe('[BABEL] tgpu alias gathering', () => {
 
     expect(babelTransform(code)).toMatchInlineSnapshot(`
       "import hello from 'typegpu';
-      const increment = hello['~unstable'].fn([])(($ => (globalThis.__TYPEGPU_META__ ??= new WeakMap()).set($.f = () => {
+      const increment = (globalThis.__TYPEGPU_AUTONAME__ ?? (a => a))(hello['~unstable'].fn([])(($ => (globalThis.__TYPEGPU_META__ ??= new WeakMap()).set($.f = () => {
```
mhawryluk: maybe we should turn off plugin autonaming for the unplugin tests that aren't meant for testing this particular feature (turned off by default in tests and enabled in auto-naming.test.ts)? what do you think?

packages/unplugin-typegpu/test/auto-naming.test.ts
```diff
@@ -0,0 +1,427 @@
...
+      console.log(myStruct1, myStruct2, bait);"
+    `);
+  });
+
+  it('works with root items', async () => {
+    const code = `\
+      import tgpu from 'typegpu';
+
+      const root = await tgpu.init();
+      const myBuffer = root.createBuffer(d.u32, 2);
```
mhawryluk, line 90: i'd add test for when there are multiple chained methods. so like when creating buffer usages with .$usage, or setting externals on a function

mhawryluk (approved):
Great job!! 🦾

packages/typegpu/tests/unplugin/autoname.test.ts
aleksanderkatan: Those don't have the `$name` method, should I add it?

mhawryluk: the fixed ones do, laid-out don't

packages/unplugin-typegpu/test/aliasing.test.ts
aleksanderkatan: Done 👍

packages/unplugin-typegpu/src/index.ts
```diff
@@ -53,6 +54,33 @@ function removeKernelDirective(node: FunctionNode) {
...
+  magicString: MagicStringAST,
+  node: acorn.AnyNode,
+  metadata: string,
+) {
+  magicString.prependLeft(
+    node.start,
+    '(($ => (globalThis.__TYPEGPU_META__ ??= new WeakMap()).set($.f = (',
+  ).appendRight(
+    node.end,
+    `) , ${metadata}) && $.f)({}))`,
```
iwoplaza: Why this space here? 👀 
```suggestion
    `), ${metadata}) && $.f)({}))`,
```

PR #1352, opened 2025-06-10
feat: Add support for timestamp queries in pipelines and introduce TgpuQuerySet

### LINKED ISSUE #78: Create a GPU compute time measuring example
Extend matrix multiplication example or create a brand new one containing measuring and displaying time it took to compute on GPU and compare it with doing the same computations on CPU.

something like in this example: https://webgpu.github.io/webgpu-samples/?sample=bitonicSort

### LINKED ISSUE #1354: feat: Introduce TgpuQuerySet

### LINKED ISSUE #1355: feat: Add a high level timing API to pipelines

reczkok:
It came out a little beefy. It includes:
- `TgpuQuerySet` implementation
- `.withPerformanceCallback` API
- `.withTimestampWrites` API
- A new example based on matrix multiplication that uses high level API
- MNIST example timing using the low level API
- A doc about enabling features in root
- A doc about timing pipelines

closes #78 
references #1104 
closes #1354 
closes #1355


apps/typegpu-docs/astro.config.mjs
```diff
@@ -116,6 +116,14 @@ export default defineConfig({
...
               slug: 'fundamentals/vertex-layouts',
             },
+            {
+              label: 'Enabling Features',
+              slug: 'fundamentals/enabling-features',
+            },
+            {
+              label: 'Timing Your Pipelines',
+              slug: 'fundamentals/timestamp-queries',
+            },
```
aleksanderkatan:
```suggestion
            {
              label: 'Enabling Features',
              slug: 'fundamentals/enabling-features',
              badge: { text: 'new' },
            },
            {
              label: 'Timing Your Pipelines',
              slug: 'fundamentals/timestamp-queries',
              badge: { text: 'new' },
            },
```
Let's bring people's attention

apps/typegpu-docs/src/content/docs/fundamentals/timestamp-queries.mdx
aleksanderkatan, line 1: Do we consider this feature experimental? If I'm not mistaken, the pipelines are still unstable.

apps/typegpu-docs/src/content/docs/fundamentals/timestamp-queries.mdx
```diff
@@ -0,0 +1,144 @@
...
+```
+
+Omit `beginningOfPassWriteIndex` or `endOfPassWriteIndex` to skip writing at the start or end of the pass, respectively.
+
+### Attaching to a Raw WebGPU Command Encoder
+
+You can also attach timestamp queries directly to a raw WebGPU command encoder by specifying `timestampWrites` in the pass descriptor:
+
+```ts
+const encoder = device.createCommandEncoder();
```
aleksanderkatan, line 77:
```suggestion
const querySet = root.createQuerySet('timestamp', 2);
const encoder = device.createCommandEncoder();
```
I think it would add clarity

apps/typegpu-docs/src/content/docs/fundamentals/timestamp-queries.mdx
```diff
@@ -0,0 +1,144 @@
...
+// timestamps[0] is start, timestamps[1] is end
+```
+
+:::caution
+If you don't resolve the query set, the read operation will throw an error or return stale data (if the query set was previously resolved).
+:::
+
+### Continuous Timestamp Queries
+
+When profiling in a loop, check availability before resolving and reading:
```
aleksanderkatan: Maybe we could include a one-sentence explanation of why the the resolve takes more time than the pass?

apps/typegpu-docs/src/content/examples/algorithms/matrix-multiplication/meta.json
```diff
@@ -1,5 +1,5 @@
 {
   "title": "Matrix Multiplication",
   "category": "algorithms",
-  "tags": ["compute"]
+  "tags": ["compute", "timestamp query"]
```
aleksanderkatan: I don't think this example was modified to include the queries

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.html
```diff
@@ -0,0 +1,126 @@
+<div class="timing-display">
+  <div class="timing-content">Press Compute to time</div>
+</div>
```
aleksanderkatan: Sometimes this div jumps from the right of the matrices to the top after pressing `Compute`. Maybe it would be better to always keep it above?

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.ts
```diff
@@ -0,0 +1,328 @@
...
+    },
+  },
+  '#2 columns': {
+    initial: state.dimensions.secondColumnCount,
+    ...paramSettings,
+    onSliderChange: (value: number) => {
+      state.dimensions.secondColumnCount = value;
+      generateMatrices();
+    },
+  },
```
aleksanderkatan, line 333: Right now it is kind of difficult to select a specific value. Also, I'm not sure if it is really necessary to provide 3 sliders. Maybe we could replace those with just a select of values like 8, 64, 512, and let all the matrices have the same dimensions?

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.ts
aleksanderkatan, line 1: I think the comparison between CPU and GPU is really important for this example. How about we let the user select only the gpu strategy, and always run both CPU and GPU computation, and show both times?

apps/typegpu-docs/src/content/examples/algorithms/mnist-inference/index.ts
aleksanderkatan, line 1: This example could have "timestamp query" in tags

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.ts
```diff
@@ -0,0 +1,328 @@
...
+    return;
+  }
+
+  while (state.bufferCapacity < requiredSize) {
+    state.bufferCapacity *= 2;
+  }
+
+  for (const buffer of [buffers.first, buffers.second, buffers.result]) {
+    buffer.destroy();
+  }
```
aleksanderkatan, line 89: Is there a chance for a read from these destroyed buffers before the new ones are created?

packages/typegpu/src/core/querySet/querySet.ts
```diff
@@ -0,0 +1,177 @@
...
+}
+
+export function INTERNAL_createQuerySet<T extends GPUQueryType>(
+  group: ExperimentalTgpuRoot,
+  type: T,
+  count: number,
+  rawQuerySet?: GPUQuerySet,
+): TgpuQuerySet<T> {
+  return new TgpuQuerySetImpl(group, type, count, rawQuerySet);
+}
```
aleksanderkatan, line 31: Could we inline this function? it is called in only one place

packages/typegpu/src/core/querySet/querySet.ts
```diff
@@ -0,0 +1,177 @@
...
+
+export function isQuerySet<T extends GPUQueryType>(
+  value: unknown,
+): value is TgpuQuerySet<T> {
+  return (value as TgpuQuerySet<T>)?.resourceType === 'query-set';
```
aleksanderkatan:
```suggestion
  const maybe = value as TgpuQuerySet<T>;
  return maybe?.resourceType === 'query-set' && !!maybe[$internal];
```
we probably should check for $internal

packages/typegpu/src/core/pipeline/timable.ts
aleksanderkatan, line 1: I think it's `timeable`, not `timable`

packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -108,9 +147,13 @@ class TgpuComputePipelineImpl implements TgpuComputePipeline {
     const memo = this._core.unwrap();
     const { branch } = this._core;
 
-    const pass = branch.commandEncoder.beginComputePass({
-      label: getName(this._core) ?? '<unnamed>',
-    });
+    const label = getName(this._core) ?? '<unnamed>';
+    const passDescriptor: GPUComputePassDescriptor = {
+      label,
```
aleksanderkatan:
```suggestion
    const passDescriptor: GPUComputePassDescriptor = {
      label: getName(this._core) ?? '<unnamed>',
```

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -329,15 +364,16 @@ class TgpuRenderPipelineImpl implements TgpuRenderPipeline {
       return attachment;
     }) as GPURenderPassColorAttachment[];
 
+    const label = getName(internals.core) ?? '<unnamed>';
     const renderPassDescriptor: GPURenderPassDescriptor = {
+      label,
```
aleksanderkatan:
```suggestion
    const renderPassDescriptor: GPURenderPassDescriptor = {
      label: getName(internals.core) ?? '<unnamed>',
```

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.html
```diff
@@ -0,0 +1,126 @@
+<div class="timing-display">
+  <div class="timing-content">Press Compute to time</div>
```
mhawryluk: I don't think we need the nested div

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.html
```diff
@@ -0,0 +1,126 @@
...
+  .matrices [class*='matrix'] > div {
+    background: white;
+    padding: 0.3rem;
+    text-align: center;
+    border-radius: 3px;
+    min-width: 2rem;
+    border: 1px solid #dee2e6;
+  }
+
+  .initial-message {
```
mhawryluk, line 77: a lot of these styles don't apply cause they lose in specificity

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.html
```diff
@@ -0,0 +1,126 @@
...
+    border-radius: 0.25rem;
+    font-family: monospace;
+  }
+
+  .timing-content {
+    color: #666;
+    font-style: italic;
+  }
+
+  .matrices {
```
mhawryluk, line 38: maybe we could stop timing-display from scrolling with the matrices by setting overflow auto and max-height 100% on the matrices section

apps/typegpu-docs/src/content/docs/fundamentals/enabling-features.mdx
```diff
@@ -0,0 +1,53 @@
...
+description: How to declare and activate WebGPU device features in TypeGPU.
+---
+
+Some GPU capabilities, such as timestamp queries or subgroup operations, must be explicitly requested before use. TypeGPU passes your configuration directly to the WebGPU API, with a convenient extension for marking features as optional.
+
+:::tip
+For a complete list of feature names, consult the [WebGPU specification](https://www.w3.org/TR/webgpu/#gpufeaturename).
+:::
+
+## Requesting Features
```
mhawryluk: we've been using *Sentence case* for the section headers in other docs, let's stay consistent

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/types.ts
```diff
@@ -0,0 +1,29 @@
...
+  firstMatrix: {
+    storage: (n: number) => createMatrixData(n),
+    access: 'readonly',
+  },
+  secondMatrix: {
+    storage: (n: number) => createMatrixData(n),
+    access: 'readonly',
+  },
+  resultMatrix: {
+    storage: (n: number) => createMatrixData(n),
```
mhawryluk:
```suggestion
    storage: createMatrixData,
    access: 'readonly',
  },
  secondMatrix: {
    storage: createMatrixData,
    access: 'readonly',
  },
  resultMatrix: {
    storage: createMatrixData,
```

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -30,37 +30,46 @@ import {
...
-  extends TgpuNamable {
+  extends TgpuNamable, Timeable<TgpuRenderPipeline> {
   readonly [$internal]: RenderPipelineInternals;
   readonly resourceType: 'render-pipeline';
 
   with<TData extends WgslArray | Disarray>(
     vertexLayout: TgpuVertexLayout<TData>,
     buffer: TgpuBuffer<TData> & VertexFlag,
-  ): TgpuRenderPipeline<IOLayout>;
+  ): TgpuRenderPipeline;
```
mhawryluk, line 60: shouldn't these all be `TgpuRenderPipeline<Output>` to not lose the type parameter information? like if someone calls `with` before `withColorAttachment`

packages/typegpu/src/core/pipeline/timable.ts
```diff
@@ -0,0 +1,182 @@
...
+  root: ExperimentalTgpuRoot,
+): T {
+  if (!root.enabledFeatures.has('timestamp-query')) {
+    throw new Error(
+      'Timestamp writes require the "timestamp-query" feature to be enabled on GPU device.',
+    );
+  }
+
+  if (currentPriors.hasAutoQuerySet) {
+    if (currentPriors.timestampWrites) {
```
mhawryluk: &&

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.ts
reczkok: Hmmm. Maybe if someone is very fast with the mouse. I will add a safety check.

packages/typegpu/src/core/querySet/querySet.ts
reczkok: Aren't we? `!!maybe[$internal]`

packages/typegpu/src/core/pipeline/renderPipeline.ts
reczkok: I think they probably should but I would say this is out of scope of this PR. I just removed those as more of a cleanup (the result is the same). We could create a ticket to address that.

packages/typegpu/src/core/querySet/querySet.ts
reczkok: No we arent 🤣

apps/typegpu-docs/src/content/examples/algorithms/matrix-next/index.ts
```diff
@@ -0,0 +1,340 @@
...
+    state.kernelTime = Number(end - start) / 1_000_000;
+  };
+
+  const optimized = root['~unstable']
+    .withCompute(computeSharedMemory)
+    .createPipeline();
+  const simple = root['~unstable'].withCompute(computeSimple).createPipeline();
+
+  return {
+    'gpu-optimized': performanceCallback
```
iwoplaza: But the performance callback is always defined... 👀

iwoplaza (changes requested):

mhawryluk (approved):
oui

PR #1359, opened 2025-06-11
feat: std.log and std.log2

lursz:


reczkok (approved):
Looks good. If Math.log ever breaks we have it covered.

PR #1365, opened 2025-06-16
docs: A port of XorDev's "Runner" to TypeGPU

iwoplaza:


apps/typegpu-docs/src/components/design/ColorPicker.tsx
```diff
@@ -7,7 +7,7 @@ function hexToRgb(hex: string) {
 }
 
 function componentToHex(c: number) {
-  const hex = (c * 255).toString(16);
+  const hex = (Math.floor(c * 255)).toString(16);
```
iwoplaza, line 10: It was not setting a proper initial value in the UI if there was a fractional part 👀

apps/typegpu-docs/src/content/examples/rendering/xor-dev-runner/index.ts
```diff
@@ -0,0 +1,178 @@
...
+draw();
+
+// #region Example controls and cleanup
+
+export const controls = {
+  scale: {
+    initial: 2,
+    min: -100,
+    max: 100,
+    step: 0.001,
```
aleksanderkatan: Not sure if we need this much precision, how about 0.1?

mhawryluk (approved):
So cool!

PR #1366, opened 2025-06-16
fix: Replace unplugin hoisting error with a warning

### LINKED ISSUE #1353: fix: Unplugin hoisting error
Currently, rollup unplugin throws if I define tests like these:

```typescript
it('names function expression', () => {
  const myFun = function () {
    'kernel & js';
    return 0;
  };

  const myGpuFun = tgpu['~unstable'].fn([], d.u32)(myFun);

  expect(getName(myFun)).toBe('myFun');
  expect(getName(myGpuFun)).toBe('myGpuFun');
});

it('names function definition', () => {
  function myFun() {
    'kernel & js';
    return 0;
  }

  const myGpuFun = tgpu['~unstable'].fn([], d.u32)(myFun);

  expect(getName(myFun)).toBe('myFun');
  expect(getName(myGpuFun)).toBe('myGpuFun');
});
```

`function "myFun", containing kernel & js directive, is referenced before its usage. Function statements are no longer hoisted after being transformed by the plugin.`

This is because we search using a regex, without regard for context names.

One possible solution to the problem is to just replace the throw with a warn. 
It is also worth noting, that babel version of the plugin does not detect this as of right now.

mhawryluk:
closes #1353


aleksanderkatan (approved):
Nice!

PR #1376, opened 2025-06-17
docs: Add ASCII (not really) camera filter example

reczkok:


apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -0,0 +1,320 @@
...
+const device = root.device;
+
+const layout = tgpu.bindGroupLayout({
+  inputTexture: { texture: 'float' },
+});
+
+const useExtendedCharacters = root['~unstable'].createUniform(d.u32);
+const displayModeBuffer = root['~unstable'].createUniform(d.u32);
+const gammaCorrectionBuffer = root['~unstable'].createUniform(d.f32);
+const glyphSizeBuffer = root['~unstable'].createUniform(d.u32, 8);
```
iwoplaza: Small nitpick, a more consistent naming 👀 
```suggestion
const charsetExtendedUniform = root['~unstable'].createUniform(d.u32);
const displayModeUniform = root['~unstable'].createUniform(d.u32);
const gammaCorrectionUniform = root['~unstable'].createUniform(d.f32);
const glyphSizeUniform = root['~unstable'].createUniform(d.u32, 8);
```

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -0,0 +1,321 @@
...
+    shaderSampler,
+    blockCoord,
+  );
+
+  const rawGray = 0.3 * color.x + 0.59 * color.y + 0.11 * color.z;
+  const gray = std.pow(rawGray, gammaCorrectionBuffer.value);
+
+  let n = d.u32(4096);
+  if (useExtendedCharacters.value === 0) {
+    if (gray > 0.2) n = 65600; // :
```
aleksanderkatan, line 99: These seem to be close to linear thresholds, how about instead of this giant chain of `if`s we create an array?
```typescript
const chars = [
  4096 /* . */,
  4096 /* . */,
  65600 /* : */,
  163153 /* o */,
  ..., 
  1512810 /* # */
];
const n = chars[d.u32(gray*10)];
```
Do we have a way to define WGSL const variables?

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -0,0 +1,321 @@
...
+    ((pix.x / halfCell) % 2) - 1,
+    ((pix.y / halfCell) % 2) - 1,
+  );
+
+  const charValue = characterFn(n, p);
+
+  const colorMode = displayModeBuffer.value;
+  let resultColor = d.vec3f(1);
+  // Color mode
+  if (colorMode === 0) {
```
aleksanderkatan, line 161: How about using this instead of raw numbers?
```typecsript
const colorModes = {
  color: 0,
  grayscale: 1,
  white: 2,
} as const;
```

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -0,0 +1,321 @@
...
+const shaderSampler = tgpu['~unstable'].sampler({
+  magFilter: 'linear',
+  minFilter: 'linear',
+});
+
+/**
+ * Adapted from the original Shadertoy implementation by movAX13h:
+ * https://www.shadertoy.com/view/lssGDj
+ */
+const characterFn = tgpu['~unstable'].fn([d.u32, d.vec2f], d.f32)((n, p) => {
```
aleksanderkatan, line 37: It is a little unclear to me what this function does or what the parameters are, could you do some refactor or add some comments clarifying this?

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
reczkok: We do via `tgpu.const(dType, val);` but unfortunately it's broken at the moment. I will do a fix in another PR (the type is not properly set for consts) but I wouldn't like for this PR to wait for another release if it does not have to. I will create a ticket so we can change it to they way you proposed since it's nicer.

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
reczkok: This is already after one refactor - see the original 😅 I will try

PR #1382, opened 2025-06-18
fix: Add explicit type annotation to constant declarations

reczkok:
For example if we did:
```ts
const nums = tgpu['~unstable'].const(d.arrayOf(d.u32, 4), [1,2,3,4]);
```
It will get resolved to:
```
const nums = array(1, 2, 3, 4);
```
So we get an array of `i32` instead. This PR makes the types explicit so we don't have to deal with that.


iwoplaza (approved):
Awesome change, yes please! 👏

PR #1386, opened 2025-06-20
feat: Better TgpuFn type (0.6.0)

iwoplaza:
Changes:
- Argument names are maintained by the type system (if implemented using TGSL)
- `TgpuFn` without generic instantiation is wide-enough that it represents any function created using `tgpu.fn`

https://github.com/user-attachments/assets/3f79bf10-9d03-4564-8d05-35dd6f5e7248

### DIFF
--- a/packages/typegpu/src/core/function/fnTypes.ts
+++ b/packages/typegpu/src/core/function/fnTypes.ts
@@ -23,6 +23,8 @@ import type {
 } from '../../data/wgslTypes.ts';
 import type { Infer } from '../../shared/repr.ts';
 
+export type AnyFn = (...args: never[]) => unknown;
+
 /**
  * Information extracted from transpiling a JS function.
  */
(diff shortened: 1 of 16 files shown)


packages/typegpu/src/core/function/fnTypes.ts
```diff
@@ -40,15 +42,35 @@ export type InferArgs<T extends unknown[]> = {
...
+        ...args: InheritTupleValues<Parameters<From>, Parameters<T>>
+      ) => ReturnType<T>)
+      & T
+    >
+  ) => ReturnType<T>;
+};
+
+export type InferImplSchema<ImplSchema extends AnyFn> = (
+  ...args: InferArgs<Parameters<ImplSchema>>
+) => Infer<ReturnType<ImplSchema>>;
```
reczkok, line 69: The implementation looks solid and I didn't manage to break it in TS Playground but we could cover it with tests to make sure we don't break it in the future

iwoplaza: Tests added 🫡

PR #1294, opened 2025-05-26
feat: Branded snippets

iwoplaza:

PR #1387, opened 2025-06-20
feat: Hide more internal functionality under symbols

iwoplaza:


packages/typegpu/src/shared/symbols.ts
```diff
@@ -13,3 +13,30 @@ export const $wgslDataType = Symbol(`typegpu:${version}:$wgslDataType`);
...
+ */
+export const $repr = Symbol(`typegpu:${version}:$repr`);
+/**
+ * Type token for the inferred (GPU-side) representation of a resource
+ * If present, it shadows the value of `$repr` for GPU-side inference.
+ */
+export const $gpuRepr = Symbol(`typegpu:${version}:$gpuRepr`);
+/**
+ * Type token for the inferred partial representation of a resource.
+ * If present, it shadows the value of `$repr`
```
reczkok: Only for partial IO

aleksanderkatan:
I'm still not really convinced on using so many symbols. As we discussed earlier, they don't really provide privacy, the only advantage I see is that there are no possible name collisions. Why not hide these potential props behind `[$internal]` instead of creating a new symbol for each one?
Could we include some UUIDs in these symbols like I mentioned in #1294 so that multiple imports are easier to debug?

iwoplaza:
> I'm still not really convinced on using so many symbols. As we discussed earlier, they don't really provide privacy, the only advantage I see is that there are no possible name collisions. Why not hide these potential props behind `[$internal]` instead of creating a new symbol for each one?

I was leaning towards creating a unique symbol for each, rather than putting them under `$internal`, so that it's:
1. Easier to track what their purpose is (they can hold documentation, which does not have to be copied at each use site),
2. Easier to find their uses in the codebase.
3. We don't have to fear of accidentally using the same name for a different purpose across objects. This was the case for $internal.dataType, which meant different things for different resources, causing unnecessarily complicated and bug prone resolution logic.

> Could we include some UUIDs in these symbols like I mentioned in #1294 so that multiple imports are easier to debug?

That's definitely something we could do 👍. But maybe as part of a separate PR.

PR #1388, opened 2025-06-20
docs: Port of XorDev's "Centrifuge 2" example

iwoplaza:
https://github.com/user-attachments/assets/824dbb11-a618-4954-83ba-7e892d33ed32


apps/typegpu-docs/src/content/examples/rendering/xor-dev-centrifuge-2/index.ts
```diff
@@ -0,0 +1,196 @@
...
+  alphaMode: 'premultiplied',
+});
+
+const pipeline = root['~unstable']
+  .withVertex(vertexMain, {})
+  .withFragment(fragmentMain, { format: presentationFormat })
+  .createPipeline();
+
+function draw() {
+  aspectRatioUniform.write(canvas.clientWidth / canvas.clientHeight);
```
reczkok, line 120: We could do that only on resize, but it probably won’t make a difference.

reczkok (approved):
No clue what the shader actually does but looks 🆒

PR #1407, opened 2025-06-26
refactor: Remove name method calls from examples and tests

aleksanderkatan:
- Removed `$name` calls from examples,
- removed `$name` calls from some tests.

After trying to remove the `$name` calls from all tests, I opted to leave it in where it was at least somewhat relevant to the test, so for example when there was a `parseResolved` call or expected errors included the name.

### DIFF
--- a/apps/typegpu-docs/src/content/examples/tests/wgsl-resolution/index.ts
+++ b/apps/typegpu-docs/src/content/examples/tests/wgsl-resolution/index.ts
@@ -8,21 +8,19 @@ const root = await tgpu.init();
 const TRIANGLE_SIZE = 0.03;
 
 // data structures
-const Params = d
-  .struct({
-    separationDistance: d.f32,
-    separationStrength: d.f32,
-    alignmentDistance: d.f32,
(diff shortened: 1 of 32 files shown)


apps/typegpu-docs/src/content/examples/tests/wgsl-resolution/index.ts
```diff
@@ -54,11 +50,11 @@ const rotate = tgpu['~unstable'].fn([d.vec2f, d.f32], d.vec2f)((v, angle) => {
   const cos = std.cos(angle);
   const sin = std.sin(angle);
   return d.vec2f(v.x * cos - v.y * sin, v.x * sin + v.y * cos);
-});
+}).$name('rotate util');
```
iwoplaza, line 53: Why are names added here?

aleksanderkatan: I thought it would be nice to show that the names added have an impact on the resolved WGSL

PR #1409, opened 2025-06-26
docs: Stable tgpu.fn guide

iwoplaza:


apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,152 @@ title: Functions
...
 :::note[Recommended reading]
 We assume that you are familiar with the following concepts:
 - <a href="https://webgpufundamentals.org/webgpu/lessons/webgpu-fundamentals.html" target="_blank" rel="noopener noreferrer">WebGPU Fundamentals</a>
 - <a href="https://webgpufundamentals.org/webgpu/lessons/webgpu-wgsl.html" target="_blank" rel="noopener noreferrer">WebGPU Shading Language</a>
 :::
 
-TypeGPU allows writing shaders by composing typed functions, which are special wrappers around WGSL code. 
-These functions can reference outside resources, like other user-defined or helper functions, buffers, bind group layouts etc. 
+WGSL (WebGPU Shading Language) lacks many quality-of-life features that help with building complex shaders, like the ability to split functionality across
+different files, or to import a shader utility function from a library. TypeGPU offers a more composable and familiar approach:
```
reczkok: I wouldn't phrase it like that - we don't want to throw shade at WSGL. Instead of `WGSL lacks x` -> `TypeGPU offers ~x`. People who know will know why that is good and the other won't care either way.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,152 @@ title: Functions
...
 
 ## Entry functions
 
-Defining entry functions is similar to regular ones, but is done through dedicated constructors:
+:::caution[Experimental]
+Entry functions are an *unstable* feature. The API may be subject to change in the near future.
+:::
+
+Entry point functions can be passed to root-defined pipelines so that TypeGPU creates shader modules with them under the hood.
+Defining entry functions is similar to regular ones, but is done through dedicated shell constructors:
```
reczkok: I would maybe start with what entry functions are to the user - an abstraction which aims to simplify and provide type safety to the shader creation and dispatching process (we can just mention this and say that it will be explained later).

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,152 @@ title: Functions
...
+
 - `tgpu['~unstable'].vertexFn`
 - `tgpu['~unstable'].fragmentFn`
 - `tgpu['~unstable'].computeFn`
 
-They can be passed to root-defined pipelines and they accept special arguments like builtins (`d.builtin`) and decorated data (`d.location`).
+Shell constructors accept objects with special values like builtins (`d.builtin`) and decorated data (`d.location`).
```
reczkok: We should also mention how this integrates with vertex buffers via attributes

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,152 @@ title: Functions
...
+    time += deltaTime;
+  }
+  let phase = (time / 300) + particleData[index].seed;
+  particleData[index].position += particleData[index].velocity * deltaTime / 20 + vec2f(sin(phase) / 600, cos(phase) / 500);
+}`.$uses({ particleData: particleDataStorage, deltaTime, time });
+```
+
+In the case of entry functions, you should not include the WGSL function header in the code string. Instead, the WGSL implementations need to access these arguments as passed in via structs.
+TypeGPU schemas for these structs are created automatically by the library and their definitions are included when resolving the functions.
+Input values are accessible through the struct named `in`, while the automatically created structs for input and output shall be referenced in implementation as `In` and `Out` respectively.
```
reczkok: We should show some resolved code

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,159 @@ title: Functions
...
-The following code defines a function that accepts one argument and returns one value.
+- An array of schemas representing argument types,
+- (Optionally) a schema representing the return type.
 
-```ts
-const getGradientColor = tgpu.fn([d.f32], d.vec4f)(/* wgsl */ `(ratio: f32) -> vec4f {
-  let color = mix(vec4f(0.769, 0.392, 1.0, 1), vec4f(0.114, 0.447, 0.941, 1), ratio);
-  return color;
-}`);
+Then the actual WGSL implementation is passed in to a shell invocation using the tagged function syntax (no parentheses).
```
reczkok: nit
```suggestion
Then the actual WGSL implementation is passed in to a shell invocation using the tagged template literal.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,159 @@ title: Functions
...
+  var blue = vec4f(0.114, 0.447, 0.941, 1);
+  return mix(purple, blue, ratio);
+}`;
+```
 
 ## External resources
 
-Functions can use external resources passed inside a record via the `$uses` method. 
-Externals can be any value or TypeGPU resource that can be resolved to WGSL (functions, buffer usages, slots, accessors, constants, variables, declarations, vectors, matrices, textures, samplers etc.).
+Functions can use external resources passed via the `$uses` method during the function creation process.
```
reczkok: `Function creation process` for me would be more resolution than definition
```suggestion
Functions can use external resources passed via the `$uses` method.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,159 @@ title: Functions
...
+Externals can include anything that can be resolved to WGSL by TypeGPU (numbers, vectors, matrices, constants, functions, buffer usages, textures, samplers, slots, accessors etc.).
```
reczkok: I know you didn't change that but now that I read it it could be confusing for someone (especially because passing a js function to externals will silently just not do anything - which we should fix #1469)
```suggestion
Externals can include anything that can be resolved to WGSL by TypeGPU (numbers, vectors, matrices, constants, TypeGPU functions, buffer usages, textures, samplers, slots, accessors etc.).
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,159 @@ title: Functions
...
+On the other hand, defining entry functions is not done by annotating `TgpuFn` with attributes, but is achieved through dedicated shell constructors instead:
 
-```ts
+- `tgpu['~unstable'].vertexFn`,
+- `tgpu['~unstable'].fragmentFn`,
+- `tgpu['~unstable'].computeFn`.
+
+Shell constructors of entry functions accept an object that describes the IO as well as other relevant parameters, such as workgroup size.
+
+Builtin attributes are available on the `d.builtin` object.
```
reczkok:
```suggestion
Builtin inter-stage inputs and outputs are available on the `d.builtin` object. 
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,159 @@ title: Functions
...
+TypeGPU has split the wgsl attribute syntax into different places to provide type safety and to simplify the shader creation and dispatching process.
+For example, all groups and bindings are assigned automatically.
+On the other hand, defining entry functions is not done by annotating `TgpuFn` with attributes, but is achieved through dedicated shell constructors instead:
 
-```ts
+- `tgpu['~unstable'].vertexFn`,
+- `tgpu['~unstable'].fragmentFn`,
+- `tgpu['~unstable'].computeFn`.
+
+Shell constructors of entry functions accept an object that describes the IO as well as other relevant parameters, such as workgroup size.
```
reczkok: I would split the declarations into separate paragraphs since they are unique. We should be particular about what the entry point parameters can be. We can do a short 1/2 sentence introduction to what tgpu entry functions are and then explain each one.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
 - <a href="https://webgpufundamentals.org/webgpu/lessons/webgpu-wgsl.html" target="_blank" rel="noopener noreferrer">WebGPU Shading Language</a>
 :::
 
-TypeGPU allows writing shaders by composing typed functions, which are special wrappers around WGSL code. 
-These functions can reference outside resources, like other user-defined or helper functions, buffers, bind group layouts etc. 
+TypeGPU offers many quality-of-life features that help with building complex WGSL shaders, like the ability to split functionality across
+different files, or to import a shader utility function from a library. The approach that TypeGPU takes aims for composability:
 
-## Creating a function
+- Each definition in WGSL is represented as its own TypeGPU resource -- a typed wrapper representing some kind of WGSL resource (e.g., `TgpuFn`, `TgpuBuffer`, `TgpuVar`).
```
reczkok: I would say it's the other way around - each TypeGPU resource represents a WGSL construct that is later used to compile the complete shader code.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+- Resources can reference outside resources, like buffers, bind group layouts, or user-defined TypeGPU functions.
```
reczkok: `Resources can reference outside resources` 🧐

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
 :::note[Recommended reading]
 We assume that you are familiar with the following concepts:
 - <a href="https://webgpufundamentals.org/webgpu/lessons/webgpu-fundamentals.html" target="_blank" rel="noopener noreferrer">WebGPU Fundamentals</a>
 - <a href="https://webgpufundamentals.org/webgpu/lessons/webgpu-wgsl.html" target="_blank" rel="noopener noreferrer">WebGPU Shading Language</a>
 :::
 
-TypeGPU allows writing shaders by composing typed functions, which are special wrappers around WGSL code. 
-These functions can reference outside resources, like other user-defined or helper functions, buffers, bind group layouts etc. 
+TypeGPU offers many quality-of-life features that help with building complex WGSL shaders, like the ability to split functionality across
+different files, or to import a shader utility function from a library. The approach that TypeGPU takes aims for composability:
```
reczkok: Maybe something more along this lines? More concrete info about functions themselves and what they allow you to do and less about what the library is about.  
```suggestion
TypeGPU functions let you define shader logic in a modular and type-safe way. Their signatures are fully visible to TypeScript, enabling tooling and static checks. Dependencies, including GPU resources or other functions, are resolved automatically, with no duplication or name clashes. This also supports distributing shader logic across multiple modules or packages. Imported functions from external sources are automatically resolved and embedded into the final shader when referenced.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+- Each definition in WGSL is represented as its own TypeGPU resource -- a typed wrapper representing some kind of WGSL resource (e.g., `TgpuFn`, `TgpuBuffer`, `TgpuVar`).
+- Resources can reference outside resources, like buffers, bind group layouts, or user-defined TypeGPU functions.
+- Resources can be imported and exported like any other JavaScript value.
 
-Functions are constructed by first defining their shells, which specify their inputs and outputs.
-Then the actual WGSL implementation is passed in as an argument to a shell invocation. If the code string is a template literal, you can omit the parentheses, which may result in a more compact Biome/Prettier formatting.
+## Defining a function
 
-The following code defines a function that accepts one argument and returns one value.
+In order to construct a TypeGPU function, you need to start with defining its shell -- an object holding only the input and output types of a TypeGPU function.
```
reczkok:
```suggestion
In order to construct a TypeGPU function, you need to start by defining its shell, an object holding only the input and output types.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+The shell constructor `tgpu.fn` makes use of [TypeGPU schemas](/TypeGPU/fundamentals/data-schemas) -- objects representing different WGSL data types, that help us with generating WGSL at runtime.
```
reczkok:
```suggestion
The shell constructor `tgpu.fn` relies on [TypeGPU schemas](/TypeGPU/fundamentals/data-schemas), objects that represent WGSL data types and assist in generating shader code at runtime.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+  return mix(purple, blue, ratio);
+}`;
 ```
 
-If you're using Visual Studio Code, you can use an [extension](https://marketplace.visualstudio.com/items?itemName=ggsimm.wgsl-literal) that brings syntax highlighting to the code fragments marked with `/* wgsl */` comments.
+:::tip
+If you're using Visual Studio Code, you can use [this extension](https://marketplace.visualstudio.com/items?itemName=ggsimm.wgsl-literal) that brings syntax highlighting to the code fragments marked with `/* wgsl */` comments.
+:::
+
+Since the information about types is already present in the shell, we allow for the WGSL header to be simplified to just include the argument names.
```
reczkok:
```suggestion
Since type information is already present in the shell, the WGSL header can be simplified to include only the argument names.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
-- `tgpu['~unstable'].vertexFn`
-- `tgpu['~unstable'].fragmentFn`
-- `tgpu['~unstable'].computeFn`
+:::caution[Experimental]
+Entry functions are an *unstable* feature. The API may be subject to change in the near future.
+:::
+
+TypeGPU has split the WGSL attribute syntax into different places to provide type safety and to simplify the shader creation and dispatching process.
+For example, all groups and bindings are assigned automatically.
+On the other hand, defining entry functions is not done by annotating `TgpuFn` with attributes, but is achieved through dedicated shell constructors instead:
```
reczkok:
```suggestion
Instead of annotating a `TgpuFn` with attributes, entry functions are defined using dedicated shell constructors:
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+- `tgpu['~unstable'].computeFn`,
+- `tgpu['~unstable'].vertexFn`,
+- `tgpu['~unstable'].fragmentFn`.
 
-They can be passed to root-defined pipelines and they accept special arguments like builtins (`d.builtin`) and decorated data (`d.location`).
+### Entry point function I/O
+
+For describing the I/O of an entry point function, we use `IORecord`s --
+JS objects, where keys represent the names of the arguments,
+and values represent the types of the arguments.
```
reczkok:
```suggestion
To describe the input and output of an entry point function, we use `IORecord`s, JavaScript objects that map argument names to their types.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+}
+```
+
+As you may note, builtin inter-stage inputs and outputs are available on the `d.builtin` object,
+and require no further type clarification.
+
+Another thing to note is that there is no need to specify locations of the arguments,
+as TypeGPU tries to assign locations automatically.
+However, if you wish to, you can still wrap any schema in a `d.location` decorator.
+Other potentially useful decorators include `d.size` and `d.align`, corresponding to `@size` and `@align` WGSL attributes.
```
reczkok: We could note that they are not unique to entry-points but are usable here too since the IO will be wrapped in a struct. (so we could move it one sentece down and say something like - `since the IO is wrapped in automatically generated structs you can also use struct specific attributes like @size and @align`

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+
+During WGSL generation, TypeGPU automatically generates structs corresponding to the passed `IORecord`s.
+In WGSL implementation, input and output structs of the given function shall be referenced as `In` and `Out` respectively.
```
reczkok:
```suggestion
In WGSL implementation, input and output structs of the given function can be referenced as `In` and `Out` respectively.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+Headers in WGSL implementations should be omitted, all input values are accessible through the struct named `in`.
```
reczkok: Can they not be ommited?

aleksanderkatan: I don't think so

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
aleksanderkatan: I removed those bullet points in favor of the change in the comment right below

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,84 +3,213 @@ title: Functions
...
+Another thing to note is that there is no need to specify locations of the arguments,
+as TypeGPU tries to assign locations automatically.
+
+During WGSL generation, TypeGPU automatically generates structs corresponding to the passed `IORecord`s.
+In WGSL implementation, input and output structs of the given function can be referenced as `In` and `Out` respectively.
+Headers in WGSL implementations must be omitted, all input values are accessible through the struct named `in`.
+
+:::note
+Schemas used in `d.struct` can be wrapped in `d.location`, `d.size` and `d.align` decorators, 
+corresponding to `@location`, `@size` and `@align` WGSL attributes.
```
reczkok: One last nit I swear 🙏 While `@size` and `@align` are exclusive to structs, `@location` isn't. Using it outside an entry point IO struct will result in an error.

PR #1411, opened 2025-06-26
feat: Stabilize tgpu.fn & remove deprecated .does API

iwoplaza:


apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -47,18 +45,17 @@ Functions can use external resources passed inside a record via the `$uses` meth
...
 
 const purple = d.vec4f(0.769, 0.392, 1.0, 1);
 
-const getGradientColor = tgpu['~unstable']
-  .fn([d.f32], d.vec4f)`(ratio: f32) -> vec4f {
-    let color = mix(purple, getBlue(), ratio);
-    return color;
-  }
-  `.$uses({ purple, getBlue });
+const getGradientColor = tgpu.fn([d.f32], d.vec4f)`(ratio: f32) -> vec4f {
```
mhawryluk, line 54: we should mention in the guide that you can skip the types in the function header, if we don't already

iwoplaza: That gonna be another PR 🫡

iwoplaza: #1409

reczkok (approved):
Let's gooo 🥳

PR #1415, opened 2025-06-26
feat: Autoname assignment expressions and properties

aleksanderkatan:
- Extract some of autonaming logic to common.ts,
- autoname in AssignmentExpressions,
- autoname in Properties, 
- autoname query sets.

I tried to somehow simplify the Babel visitor but I couldn't because of the path node replacement. Let me know if you find a way to clean this up.

### DIFF
--- a/packages/unplugin-typegpu/src/common.ts
+++ b/packages/unplugin-typegpu/src/common.ts
@@ -176,6 +176,7 @@ function containsResourceConstructorCall(
 
 type ExpressionFor<T extends acorn.AnyNode | babel.Node> = T extends
   acorn.AnyNode ? acorn.Expression : babel.Expression;
+
 /**
  * Checks if `node` contains a label and a tgpu expression that could be named.
  * If so, it calls the provided callback. Nodes selected for naming include:


packages/unplugin-typegpu/src/common.ts
```diff
@@ -172,6 +173,55 @@ export function containsResourceConstructorCall(
...
+ *
+ * `property: tgpu.bindGroupLayout({})` (Property/ObjectProperty)
+ *
+ * Since it is mostly for debugging and clean WGSL generation,
+ * some false positives and false negatives are admissible.
+ *
+ * @privateRemarks
+ * When adding new checks, you need to call this method in the corresponding node in Babel.
+ */
+export function findNameableExpression<T extends acorn.AnyNode | babel.Node>(
```
mhawryluk: I'm not a fan of the name of this function. It doesn't really find the expression, but rather performs the autonaming. I would also change Nameable to Namable for consistency with the rest of the codebase.

mhawryluk (approved):
Awesome!! 🤖

packages/unplugin-typegpu/src/common.ts
aleksanderkatan: I can't really come up with a good name, is `handleExpressionNaming` good enough? `performExpressionNaming`?

packages/unplugin-typegpu/src/common.ts
```diff
@@ -173,6 +174,55 @@ export function containsResourceConstructorCall(
   return false;
 }
 
+type ExpressionFor<T extends acorn.AnyNode | babel.Node> = T extends
+  acorn.AnyNode ? acorn.Expression : babel.Expression;
+/**
```
iwoplaza: Tiny nit
```suggestion
type ExpressionFor<T extends acorn.AnyNode | babel.Node> = T extends
  acorn.AnyNode ? acorn.Expression : babel.Expression;

/**
```

iwoplaza (approved):
Great change! 🎉

PR #1419, opened 2025-06-27
chore: Remove JIT transpiler

iwoplaza:
## Why?
Since we no longer use the JIT Transpiler internally (apart from minor testing), it would become a maintenance burder.
An alternative for devs wanting to use the JIT Transpiler, is using standalone Babel.

## Updated error message
image


apps/typegpu-docs/src/content/docs/tooling/unplugin-typegpu.mdx
```diff
@@ -62,18 +62,12 @@ The package includes the following functionalities:
   ```
 
   Besides transpiling JS into AST, the plugin also collects external references, so it is not necessary to pass them to the `$uses` method anymore.
-  This is not possible when using a JIT transpiler.
 
-- **[WIP] Automatic naming of tgpu objects**
+- Automatic naming of tgpu objects**
 
   Naming gpu objects via the `$name` method is very helpful for debugging. Soon it will not be necessary to do that explicitly.
   Instead, the plugin will be able to name the objects based on the variable names that they are assigned to.
```
aleksanderkatan: These two paragraphs can be updated as well

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -159,7 +159,13 @@ export function createFnCore(
 
           applyExternals(externalMap, missing);
         }
-        const ast = pluginData?.ast ?? ctx.transpileFn(String(implementation));
+
+        const ast = pluginData?.ast;
+        if (!ast) {
+          throw new Error(
+            "Missing metadata for tgpu.fn function body (either missing 'kernel' directive, or misconfigured `unplugin-typegpu`)",
+          );
```
reczkok, line 167: I think we could be more specific with the error here almost for free by checking if we are missing the ast or pluginData

iwoplaza: Hmm, I don't think there's a situation where (for functions) `pluginData` exists, but `ast` does not

Issue #1254, opened 2025-05-15
feat: Make array schemas callable, both for creating n-sized uninitialized arrays in TGSL, and for casting


iwoplaza:
We want the equivalent of:
```
var result: array<f32, 8>;
```
... in TGSL. I would love to see suggestions from everyone 🙏

aleksanderkatan:
If I'm reading the docs correctly, `var result: array<f32, 8>;` is not uninitialized:
[The docs](https://www.w3.org/TR/WGSL/#var-and-value) say that if there is no initializer, the variable is [default initialized](https://www.w3.org/TR/WGSL/#default-initial-value), which I think we can achieve on JS side if we also give our structs default constructors (#1480).
(I'm mentioning this because the PR title says `uninitialized`, my goal right now is to parse `d.arrayOf(d.f32, 8)()` to `array<f32, 8>();`)

PR #1421, opened 2025-06-27
feat: Apply copy semantics to function calls in JS

aleksanderkatan:
This PR introduces a convention where schema calls can create deep copies of corresponding objects. For example,
```typescript
const Schema = d.struct({ prop: d.vec2f });
const struct = Schema({ prop: d.vec2f(1, 2) });
const clone = Schema(struct);
expect(struct).toStrictEqual(clone);
expect(struct).not.toBe(clone);
expect(struct.prop).not.toBe(clone.prop);
```
Changes:
- deep copy structs in struct schema calls,
- deep copy tgpuFn arguments on calls in JS,
- updated docs and tests.

Arrays do not follow this convention yet, I left it as a part of #1254


package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "DEV=true pnpm --parallel -r dev",
     "fix": "biome lint --write . && deno fmt",
-    "test": "pnpm run test:circular-deps && pnpm run test:types && pnpm run test:style && pnpm run test:unit-and-attest",
+    "test": "pnpm run test:types && pnpm run test:style && pnpm run test:unit-and-attest && pnpm run test:circular-deps",
```
aleksanderkatan, line 10: I included this change because circular deps is more of a sanity check, no need to wait for it before we fail on other tests

apps/typegpu-docs/src/content/docs/reference/data-schema-cheatsheet.mdx
```diff
@@ -46,3 +46,26 @@ const a = vec2f(1, 2);  // 1, 2
...
+
+Scalar, vector and matrix schemas can be called with no arguments to get a default value of the type represented by the schema.
+
+<Code code="
+import { f32, bool, vec3f } from 'typegpu/data';
+
+const f32Default = f32(); // 0.0
+const boolDefault = bool(); // false
+const vecDefault = vec3f(); // vec3f(0.0, 0.0, 0,0)
+" lang="js" />
```
iwoplaza: I think, we can use normal code blocks here (like \```js bla bla bla ```)

packages/typegpu/src/data/struct.ts
```diff
@@ -20,7 +21,14 @@ import type { AnyWgslData, WgslStruct } from './wgslTypes.ts';
...
 ): WgslStruct<TProps> {
-  const struct = <T>(props: T) => props;
+  // in the schema call, create and return a deep copy
+  // by wrapping all the values in corresponding schema calls
+  const struct = <T extends {}>(instanceProps: T) =>
+    Object.fromEntries(
+      Object.entries(instanceProps).map((
+        [key, value],
+      ) => [key, schemaCallWrapper(props[key], value)]),
+    );
```
iwoplaza: Let's iterate over the `props` instead of `instanceProps`, to automatically remove any unwanted keys without causing an issue.

iwoplaza (approved):
Like a tiny nit, looks awesome! 🚀

iwoplaza (approved):
YES

PR #1428, opened 2025-06-29
feat: @typegpu/sdf package

iwoplaza:


apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
```diff
@@ -0,0 +1,261 @@
...
+    4,
+  );
+});
+
+const vertexMain = tgpu['~unstable'].vertexFn({
+  in: { idx: d.builtin.vertexIndex },
+  out: { pos: d.builtin.position, uv: d.vec2f },
+})((input) => {
+  const pos = [d.vec2f(1, 1), d.vec2f(-1, 1), d.vec2f(1, -1), d.vec2f(-1, -1)];
+  const uv = [d.vec2f(1, 1), d.vec2f(0, 1), d.vec2f(1, 0), d.vec2f(0, 0)];
```
deluksic: Doing a single triangle which covers the screen would be tiiiiiny bit more performant due to duplicated fragment work along the diagonal.

Makes it a bit simpler as well due to not needing `triangle-strip`.

More info https://wallisc.github.io/rendering/2021/04/18/Fullscreen-Pass.html

iwoplaza: We're doing that in a couple examples already, I'll pull it in from those 🫡

packages/typegpu-sdf/src/3d.ts
```diff
@@ -0,0 +1,61 @@
+import tgpu from 'typegpu';
+import { f32, vec3f } from 'typegpu/data';
+import { abs, add, dot, length, max, min, sub } from 'typegpu/std';
+
+/**
+ * Signed distance function for a sphere
+ * @param p Point to evaluate
+ * @param r Radius of the sphere
+ */
+export const sdSphere = tgpu.fn([vec3f, f32], f32)((p, r) => {
```
deluksic: for ease of use, maybe full parameter names would be better: `center`, `radius`. Also in other functions

packages/typegpu-sdf/src/3d.ts
```diff
@@ -0,0 +1,61 @@
...
+/**
+ * Signed distance function for an infinite plane
+ * @param p Point to evaluate
+ * @param n Normal vector of the plane (must be normalized)
+ * @param h Height/offset of the plane along the normal
+ */
+export const sdPlane = tgpu.fn([vec3f, vec3f, f32], f32)((p, n, h) => {
+  'kernel & js';
+  return dot(p, n) + h;
+});
```
reczkok: We could just normalize it - then we don't really care whether it's normalized or not.

packages/typegpu-sdf/src/3d.ts
reczkok: Yeah - I would either JS doc all the functions or give all parameters meaningful names

packages/typegpu-sdf/src/2d.ts
```diff
@@ -0,0 +1,20 @@
...
+  'kernel & js';
+  const d = sub(abs(p), b);
+  return length(max(d, vec2f(0))) + min(max(d.x, d.y), f32(0));
+});
+
+export const sdRoundedBox = tgpu.fn([vec2f, vec2f, f32], f32)((p, b, r) => {
+  'kernel & js';
+  const d = add(sub(abs(p), b), vec2f(r));
+  return length(max(d, vec2f(0))) + min(max(d.x, d.y), f32(0)) - r;
+});
```
reczkok: Same here - I would either add JSdoc or make the parameter names more meaningful

packages/typegpu-sdf/src/3d.ts
deluksic: this can be unnecessarily slow, but I see the appeal of not caring

packages/typegpu-sdf/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "license": "MIT",
+  "peerDependencies": {
+    "typegpu": "^0.5.9"
+  },
+  "devDependencies": {
+    "@typegpu/tgpu-dev-cli": "workspace:*",
+    "@webgpu/types": "catalog:",
+    "typegpu": "workspace:*",
+    "typescript": "catalog:",
+    "unbuild": "catalog:",
```
reczkok:
```suggestion
    "@webgpu/types": "catalog:types",
    "typegpu": "workspace:*",
    "typescript": "catalog:types",
    "unbuild": "catalog:build",
```

apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
```diff
@@ -0,0 +1,257 @@
...
+
+// Structure to hold both distance and color
+const Shape = d.struct({
+  color: d.vec3f,
+  dist: d.f32,
+});
+
+const checkerBoard = tgpu.fn([d.vec2f], d.f32)((uv) => {
+  const suv = std.floor(std.mul(20, uv));
+  return suv.x + suv.y - 2 * std.floor((suv.x + suv.y) * 0.5);
```
reczkok:
```suggestion
  return (suv.x + suv.y) % 2;
```

apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
```diff
@@ -0,0 +1,257 @@
...
+  device: root.device,
+  format: presentationFormat,
+  alphaMode: 'premultiplied',
+});
+
+const time = root.createUniform(d.f32);
+const resolution = root.createUniform(d.vec2f);
+
+const MAX_STEPS = 1000;
+const MAX_DIST = 500;
```
reczkok: I think we can reduce it to something like 20 - anything beyond that is just aliased checkerboard. I also like the "small planet" appearance this gives along with performance benefits.
| Version   | Visual Output                                                                                      | Performance Logs                                                                                      |
|-----------|----------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------------|
| **Before** | Before Visual | Before Logs       |
| **After**  | After Visual  | After Logs        |

reczkok (approved):
Great work!

apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
iwoplaza: Does not have the same behavior for negative coordinates
image

apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
iwoplaza: Nice! The spherical world effect messes with my head a bit though, since the checkerboard is not distorted to go around a spherical world. What if we added a fog instead?

https://github.com/user-attachments/assets/4a207540-0b24-42b6-a8de-8bc5b5e4b956

apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
reczkok: I guess not
```suggestion
  return std.abs(suv.x + suv.y) % 2;
```

apps/typegpu-docs/src/content/examples/rendering/ray-marching/index.ts
deluksic: if you're not aware of this masterpiece, highly recommend reading https://iquilezles.org/articles/checkerfiltering/

PR #1441, opened 2025-07-01
feat: Measure performance of resolution and shader compilation

iwoplaza:
https://github.com/user-attachments/assets/26e2e798-f53e-478b-8a2f-17e4850b1e4b

Benchmark results as generated HTML & JSON:

image


aleksanderkatan:
I get an error when running the tests, any ideas?

image

apps/typegpu-docs/.gitignore
```diff
@@ -22,3 +22,7 @@ pnpm-debug.log*
 
 # autogenerated by typedoc
 src/content/docs/api
+
+# tests
+tests/artifacts
+!tests/artifacts/README.md
```
aleksanderkatan, line 28:
```suggestion
!tests/artifacts/README.md

```

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -565,35 +565,27 @@ const createSampleScene = () => {
...
-    loop(dt);
-    requestAnimationFrame(runner);
-  };
-  requestAnimationFrame(runner);
-};
-
-onFrame((deltaTime: number) => {
-  msSinceLastTick += deltaTime;
+let animationFrame: number;
+let lastTime = Date.now();
```
aleksanderkatan, line 570: Would `performance.now()` be more fitting?

apps/typegpu-docs/tests/examples.test.browser.ts
```diff
@@ -0,0 +1,93 @@
+import { expect, test } from 'vitest';
+import { examples as exampleRecord } from '../src/utils/examples/exampleContent.ts';
+import { server } from '@vitest/browser/context';
+import * as Plot from '@observablehq/plot';
+import type { INTERNAL_GlobalExt } from 'typegpu';
+import { pipe } from 'remeda';
+
+const globalExt = globalThis as unknown as
```
aleksanderkatan, line 8:
```suggestion
const globalExt = globalThis as
```

packages/typegpu/src/shared/meta.ts
```diff
@@ -12,9 +13,20 @@ export interface MetaData {
   externals?: Record<string, unknown> | undefined;
 }
 
-interface GlobalWithMeta {
+/**
+ * Don't use or you WILL get fired from your job.
```
aleksanderkatan, line 17: :flushed:

packages/typegpu/src/shared/meta.ts
```diff
@@ -25,6 +37,33 @@ Object.assign(globalThis, {
       : exp,
 });
 
+const globalWithMeta = globalThis as unknown as
```
aleksanderkatan, line 40:
```suggestion
const globalWithMeta = globalThis as
```

packages/typegpu/src/shared/meta.ts
```diff
@@ -25,6 +37,33 @@ Object.assign(globalThis, {
...
+        let entries = records.get(name);
+        if (!entries) {
+          entries = [];
+          records.set(name, entries);
+        }
+        entries.push(data);
+      },
+    };
+  })()
+  : undefined;
```
aleksanderkatan, line 65:
```suggestion
export const PERF = (DEV || TEST)
  ? (() => ({
    get enabled() {
      return !!globalWithMeta.__TYPEGPU_MEASURE_PERF__;
    },
    record(name: string, data: unknown) {
      // biome-ignore lint/suspicious/noAssignInExpressions: it's fine
      const records = (globalWithMeta.__TYPEGPU_PERF_RECORDS__ ??= new Map());
      let entries = records.get(name);
      if (!entries) {
        entries = [];
        records.set(name, entries);
      }
      entries.push(data);
    },
  }))()
  : undefined;
```
Why is this wrapped in a lambda? This object is not created if the ternary operator goes into false branch even without the lambda, right?

packages/typegpu/src/shared/meta.ts
```diff
@@ -25,6 +37,33 @@ Object.assign(globalThis, {
...
+
+/**
+ * Performance measurements are only enabled in dev & test environments for now
+ */
+export const PERF = (DEV || TEST)
+  ? (() => {
+    return {
+      get enabled() {
+        return !!globalWithMeta.__TYPEGPU_MEASURE_PERF__;
+      },
```
aleksanderkatan, line 52: Could we simplify this to remove this ternary operator and just return `(DEV || TEST) && ...` in `get enabled()`? We always do `if (PERF?.enabled)` anyway

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -565,35 +565,27 @@ const createSampleScene = () => {
...
+function run() {
+  const now = Date.now();
+  const dt = now - lastTime;
+  lastTime = now;
+  msSinceLastTick += dt;
```
reczkok: If we are already cleaning it up i would move to requestAnimationFrame timestamp callback

packages/typegpu/src/shared/env.ts
```diff
@@ -7,3 +7,6 @@
  */
 // biome-ignore lint/suspicious/noExplicitAny: the types are not important here
 export const DEV = (globalThis as any).process.env.NODE_ENV === 'development';
```
aleksanderkatan, line 9:
```suggestion
export const DEV = globalThis.process.env.NODE_ENV === 'development';
```
There is no intellisense for `NODE_ENV`, but the types seem to be fine if I do this

apps/typegpu-docs/tests/examples.test.browser.ts
```diff
@@ -0,0 +1,93 @@
...
+  onCleanup(): void;
+}
+
+/**
+ * The amount of time after Vitest decides that the test
+ * has gone rogue. Adjust as necessary.
+ */
+const timeout = 120 * 1000;
+
+const iterations = 10;
```
aleksanderkatan, line 29: Could we make it a little smaller, or could we make the number of iterations passable via the command line?

apps/typegpu-docs/tests/examples.test.browser.ts
```diff
@@ -0,0 +1,93 @@
...
+
+interface ExampleRuntime {
+  onCleanup(): void;
+}
+
+/**
+ * The amount of time after Vitest decides that the test
+ * has gone rogue. Adjust as necessary.
+ */
+const timeout = 120 * 1000;
```
aleksanderkatan, line 27: I think this timeout may be a little to close to how long the test actually takes, it actually timed out on me once even though there was no lag nor wait for camera permission. Can we also make it depend on the number of iterations?

packages/typegpu/src/globals.d.ts
aleksanderkatan, line 1: What exactly is this file and what is it used for?

aleksanderkatan (commented):
Great job, super useful!

packages/typegpu/src/shared/meta.ts
iwoplaza: Doing `(DEV || TEST) &&` before creating the object allows it to be stripped from the bundle altogether when an app is built for production. If we were checking for that inside of `enabled`, bundlers would not be able to strip both PERF, and code using it.

PR #1442, opened 2025-07-02
feat: Update "kernel" and "kernel & js" directives

aleksanderkatan:
~Blocked by #1415~
~Blocked by #1443~

Changes:
- remove "kernel & js" directive (now we only use "kernel"),
- always keep JS implementation of functions transformed by unplugin,
- cleanup.


iwoplaza:
> Is this a good moment to delete this? (typegpu/src/core/function/astUtils.ts) image

Sure, let's remove it!

packages/unplugin-typegpu/test/kernel-directive.test.ts
```diff
@@ -12,14 +12,16 @@ describe('[BABEL] "kernel" directive', () => {
       };
 
       const addCPU = (a, b) => {
-        return a + b;
+          return a + b;
```
iwoplaza: Why the indent? 👀 
```suggestion
        return a + b;
```

iwoplaza (approved):
Yup yup! 💯

packages/unplugin-typegpu/test/kernel-directive.test.ts
aleksanderkatan: This probably happened because this file is actually modified kernel-and-js-directive.test.ts that I renamed to kernel-directive.test.ts

PR #1446, opened 2025-07-02
fix: division operator resolution

lursz:


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -148,6 +148,16 @@ export function generateExpression(
     const rhsStr = ctx.resolve(convRhs.value);
     const type = operatorToType(convLhs.dataType, op, convRhs.dataType);
 
+    if (
+      op === '/' && !['f32', 'f16'].includes(convLhs.dataType.type) &&
+      !['f32', 'f16'].includes(convRhs.dataType.type)
+    ) {
+      return snip(
+        `f32(${lhsStr}) / f32(${rhsStr})`,
+        type,
```
aleksanderkatan: Type should probably change to f32 here

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -148,6 +148,16 @@ export function generateExpression(
     const rhsStr = ctx.resolve(convRhs.value);
     const type = operatorToType(convLhs.dataType, op, convRhs.dataType);
 
+    if (
+      op === '/' && !['f32', 'f16'].includes(convLhs.dataType.type) &&
+      !['f32', 'f16'].includes(convRhs.dataType.type)
```
aleksanderkatan: does `f32(1)/2` get wrapped by this? How about `f32(1)/u32(2)`

packages/typegpu/tests/tgslFn.test.ts
aleksanderkatan, line 1: Add tests with more complex expressions like `(1+2)/2`, or `f()/2`, `(1/2)/2`

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -148,6 +148,16 @@ export function generateExpression(
...
+    ) {
+      return snip(
```
aleksanderkatan: I would probably log a warning here saying what we are doing, since this is not the most obvious way of how to handle division

lursz: Good idea!

packages/typegpu/src/tgsl/wgslGenerator.ts
lursz: Nice catch!

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -148,6 +148,19 @@ export function generateExpression(
     const rhsStr = ctx.resolve(convRhs.value);
     const type = operatorToType(convLhs.dataType, op, convRhs.dataType);
 
+    if (
+      op === '/' && !['f32', 'f16'].includes(convLhs.dataType.type) &&
+      !['f32', 'f16'].includes(convRhs.dataType.type)
```
iwoplaza: A tiny suggestion. It's a bit more verbose, but since we want to minimize the performance impact of resolution, we can reduce the array instantiations and function invocations here.
```suggestion
      op === '/' &&
      convLhs.dataType.type !== 'f32' &&
      convLhs.dataType.type !== 'f16' &&
      convRhs.dataType.type !== 'f32' &&
      convRhs.dataType.type !== 'f16'
```

iwoplaza (changes requested):

packages/typegpu/src/tgsl/wgslGenerator.ts
lursz: True, good catch

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -156,6 +156,38 @@ export function generateExpression(
...
 
+    if (op === '/') {
+      const lhsIsFloat = convLhs.dataType.type === 'f32' ||
+        convLhs.dataType.type === 'f16';
+      const rhsIsFloat = convRhs.dataType.type === 'f32' ||
+        convRhs.dataType.type === 'f16';
+
+      if (!lhsIsFloat && !rhsIsFloat) {
+        console.warn(
+          'In division we cast both sides to the f32. This may not be the expected behaviour. Consider using explicit conversions instead.',
```
aleksanderkatan: Do we use GB English (behaviour)?

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -156,6 +156,38 @@ export function generateExpression(
...
+        return snip(
+          `f32(${lhsStr}) / f32(${rhsStr})`,
+          d.f32,
+        );
+      }
+
+      if (!lhsIsFloat || !rhsIsFloat) {
+        console.warn(
+          !lhsIsFloat
+            ? 'In division we cast left side to the f32. This may not be the expected behaviour. Consider using explicit conversions instead.'
```
aleksanderkatan: You could just select the side in a ternary operator since the rest of the warn is the same

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -156,6 +156,38 @@ export function generateExpression(
...
+            : 'In division we cast right side to the f32. This may not be the expected behaviour. Consider using explicit conversions instead.',
```
aleksanderkatan: I think that this warn may be misinterpreted as "In any division, we always cast the right side to f32", I would rewrite it to maybe \`A value of type ${type}\ on the ${side} of a division operator was automatically cast to f32. -||-`

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan, line 1: We could use `divInteger` from vectorOps.ts to create a dualImpl of integer division, and mention it in the warn as a way to still perform the integer division, what do you think?

aleksanderkatan, line 1: In 3D Fish, Box Raytracing, Ray Marching, Centrifuge, Runner, Oklab, both Boids... many examples there are warns thrown. I think it is worth investigating them and wrapping what we can in `f32`.

Something weird happens to Fluid Double Buffering.

Cubemap Reflection and Fluid With Atomics do not work anymore.

aleksanderkatan (changes requested):

packages/typegpu/src/tgsl/wgslGenerator.ts
lursz: We might have to decide upon that in the future. This does not matter here, as those warnings are deemed unnecessary (we aim to mimic JS) and will be removed.

packages/typegpu/src/tgsl/wgslGenerator.ts
lursz: Good catch, thanks for reporting this!

lursz: Something to consider, but js always returns division results as a float. Our goal is to align with that, so for now I'd leave it as it is.

reczkok (changes requested):
It breaks `Matrix Multiplication (next)` (we should also minimize the implicit conversions that were created)

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -184,6 +184,39 @@ export function generateExpression(
     const rhsStr = ctx.resolve(convRhs.value);
     const type = operatorToType(convLhs.dataType, op, convRhs.dataType);
 
+    if (op === '/') {
+      const lhsIsFloat = convLhs.dataType.type === 'f32' ||
+        convLhs.dataType.type === 'f16';
+      const rhsIsFloat = convRhs.dataType.type === 'f32' ||
+        convRhs.dataType.type === 'f16';
```
reczkok: I think we could also check for abstractFloat

reczkok (approved):
Yes

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -207,7 +207,10 @@ const decideWaterLevel = tgpu.fn([d.u32, d.u32])((x, y) => {
   if (!isWall(x - 1, y)) {
     const flowRaw = d.i32(waterLevelBefore) - d.i32(getWaterLevel(x - 1, y));
     if (flowRaw > 0) {
-      const change = std.max(std.min(4, remainingWater), d.u32(flowRaw) / 4);
+      const change = std.max(
+        std.min(4, remainingWater),
+        d.u32(d.u32(flowRaw) / d.f32(4)),
```
iwoplaza: Why not this?
```suggestion
        d.u32(flowRaw / 4),
```

iwoplaza: Let's not show warnings on int / int division, and simplify these operations.

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -222,7 +225,10 @@ const decideWaterLevel = tgpu.fn([d.u32, d.u32])((x, y) => {
   if (!isWall(x + 1, y)) {
     const flowRaw = d.i32(waterLevelBefore) - d.i32(getWaterLevel(x + 1, y));
     if (flowRaw > 0) {
-      const change = std.max(std.min(4, remainingWater), d.u32(flowRaw) / 4);
+      const change = std.max(
+        std.min(4, remainingWater),
+        flowRaw / 4,
```
iwoplaza, line 230: I think we want integer division here
```suggestion
        d.u32(flowRaw / 4),
```

PR #1450, opened 2025-07-03
Feat: std-smoothstep

lursz:


packages/typegpu/src/data/vectorOps.ts
```diff
@@ -1,3 +1,4 @@
+// import { smoothstep } from '../std/numeric.ts';
```
aleksanderkatan:
```suggestion
```

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -27,6 +28,49 @@ type v4 = wgsl.v4f | wgsl.v4h | wgsl.v4i | wgsl.v4u;
 
 type MatKind = 'mat2x2f' | 'mat3x3f' | 'mat4x4f';
 
+const smoothstepScalar = (edge0: number, edge1: number, x: number): number => {
+  if (x <= edge0) return 0;
+  if (x >= edge1) return 1;
+  const t = clamp((x - edge0) / (edge1 - edge0), 0.0, 1.0);
```
aleksanderkatan: I think that either this clamp is unnecessary, or the two ifs above it are

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -27,6 +28,49 @@ type v4 = wgsl.v4f | wgsl.v4h | wgsl.v4i | wgsl.v4u;
...
+  return t * t * (3 - 2 * t);
+};
+
+type TernaryOp = (a: number, b: number, c: number) => number;
```
aleksanderkatan: Could you move this entire section to be where UnaryOp and BinaryOp are?

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -27,6 +28,49 @@ type v4 = wgsl.v4f | wgsl.v4h | wgsl.v4i | wgsl.v4u;
 
 type MatKind = 'mat2x2f' | 'mat3x3f' | 'mat4x4f';
 
+const smoothstepScalar = (edge0: number, edge1: number, x: number): number => {
```
aleksanderkatan: Can you move this to `NumberOps`?

packages/typegpu/src/std/numeric.ts
```diff
@@ -719,3 +719,28 @@ export const tanh = createDualImpl(
...
+  <T extends AnyFloatVecInstance | number>(edge0: T, edge1: T, x: T): T => {
+    if (typeof x === 'number') {
+      const e0 = edge0 as number;
+      const e1 = edge1 as number;
+
+      if (x <= e0) return 0 as T;
+      if (x >= e1) return 1 as T;
+
+      const t = clamp((x - e0) / (e1 - e0), 0.0, 1.0);
+      return (t * t * (3 - 2 * t)) as T;
```
aleksanderkatan: Duplicated code, you can use the function defined in VectorOps.ts

packages/typegpu/tests/std/numeric/smoothstep.test.ts
```diff
@@ -0,0 +1,101 @@
...
+    expect(result2).toStrictEqual(vec2f(1, 1));
+
+    // components between edge0 and edge1
+    const result3 = smoothstep(
+      vec2f(0.3, 0.5),
+      vec2f(0.8, 0.9),
+      vec2f(0.55, 0.7),
+    );
+    expect(result3.x).toBeCloseTo(0.5);
+    expect(result3.y).toBeCloseTo(0.5);
```
aleksanderkatan: you could use `expect(isCloseTo(result3, vec2f(0.5, 0.5))).toBe(true)`

packages/typegpu/tests/std/numeric/smoothstep.test.ts
```diff
@@ -0,0 +1,101 @@
...
+  it('works with arbitrary edge values', () => {
+    expect(smoothstep(10, 20, 5)).toBe(0);
+    expect(smoothstep(10, 20, 25)).toBe(1);
+    expect(smoothstep(10, 20, 15)).toBeCloseTo(0.5);
+    expect(smoothstep(-5, 5, 0)).toBeCloseTo(0.5);
+  });
+
+  it('works with vec2f', () => {
+    // all components less than edge0
+    const result1 = smoothstep(
```
aleksanderkatan, line 25: I think it is enough to include one of these in this test

aleksanderkatan (commented):
Some nits 🌮

packages/typegpu/src/data/vectorOps.ts
deluksic: Note that `smoothstep` is defined for `edge0 > edge1`. This means that the two ifs are actually incorrect.

https://www.w3.org/TR/WGSL/#smoothstep-builtin

packages/typegpu/src/data/numberOps.ts
```diff
@@ -0,0 +1,11 @@
+export const smoothstepScalar = (
+  edge0: number,
+  edge1: number,
+  x: number,
+): number => {
+  const t = clamp((x - edge0) / (edge1 - edge0), 0.0, 1.0);
+  return t * t * (3 - 2 * t);
+};
+
+export const clamp = (value: number, low: number, high: number) =>
```
deluksic, line 13: can this be reused from std?

packages/typegpu/src/data/vectorOps.ts
aleksanderkatan: I didn't mean a new file named numberOps.ts, but the object `NumberOps` that already exists in vectorOps.ts. I guess it may be located in another file, but then please move the `divInteger` function to that file as well. You could also move `clamp` to the same place.

packages/typegpu/src/data/numberOps.ts
aleksanderkatan: Using clamp from std would create a circular dependency numeric.ts -> vectorOps.ts -> numberOps.ts -> numeric.ts. I think that if we want to keep this file, we should move all existing JS definitions of such functions here (`divInteger`, `clamp`, `smoothstep` etc.), and then use them in vectorOps.ts and numeric.ts

packages/typegpu/tests/std/numeric/smoothstep.test.ts
```diff
@@ -0,0 +1,86 @@
...
+  });
+
+  it('works with vec4f', () => {
+    const result = smoothstep(
+      vec4f(0.0, 0.1, 0.2, 0.3),
+      vec4f(1.0, 0.9, 0.8, 0.7),
+      vec4f(0.0, 0.5, 0.8, 0.7),
+    );
+    expect(isCloseTo(result, vec4f(0, 0.5, 1, 1))).toBe(true);
+  });
```
reczkok, line 73: We should test for more than the 3 most obvious cases

aleksanderkatan (approved):
🥤🚶‍♂️

reczkok (approved):
Oke

PR #1453, opened 2025-07-04
feat: Add @typegpu/three package and Three.js integration example

reczkok:
**Disclaimer: Very WIP**


packages/typegpu-three/src/typegpu-node.ts
```diff
@@ -0,0 +1,263 @@
...
+  node: TSL.ShaderNodeObject<TNode>,
+  options: { type: T } | { type: (length: number) => T },
+): TSLAccessor<T, TNode> {
+  return new TSLAccessor<T, TNode>(
+    node,
+    d.isData(options.type)
+      ? options.type as T
+      : (options.type as (length: number) => T)(0),
+  );
+}
```
aleksanderkatan: type is required, can we make it a parameter and potentially re-add options as a third argument later on?

iwoplaza: It would be nice to write out some more examples of @typegpu/three to see if one is more readable than the other. My gut was telling me that an object with named params would be easier to read, especially when applied to more complex inline TSL nodes, not just references to already defined nodes.

Could you extract uses of the fromTSL API, and create a side-by-side comparison of both APIs?

aleksanderkatan: image

I may be biased, because in example that I rewrote, there were so many simple buffors

apps/typegpu-docs/src/examples/threejs/attractors/index.html
aleksanderkatan, line 1: All examples are missing thumbnails

aleksanderkatan, line 1: Can we add a new tag as well for three examples?

apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
```diff
@@ -0,0 +1,341 @@
...
+let timeSinceLastStep = 0;
+let timestamp = 0;
+
+const clock = new THREE.Clock();
+
+const params = {
+  wireframe: false,
+  sphere: true,
+  wind: 1,
+};
```
aleksanderkatan, line 47: Right now, these are readonly. Can you add some controls for these?

apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
```diff
@@ -0,0 +1,341 @@
...
+
+const API = {
+  color: 0x204080, // sRGB
+  sheenColor: 0xffffff, // sRGB
+};
+
+// TODO: Fix example with WebGL backend
```
aleksanderkatan:
```suggestion
```

apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
```diff
@@ -0,0 +1,341 @@
...
+  while (timeSinceLastStep >= timePerStep) {
+    // run a verlet system simulation step
+    timestamp += timePerStep;
+    timeSinceLastStep -= timePerStep;
+    updateSphere();
+    await verletSim.update(renderer);
+  }
+
+  await renderer.renderAsync(scene, camera);
+}
```
aleksanderkatan, line 322: There probably should be some cleanup with `renderer.dispose()` here

apps/typegpu-docs/src/examples/threejs/compute-cloth/output.wgsl
aleksanderkatan, line 1: These .wgsl files are probably leftover debug

apps/typegpu-docs/src/examples/threejs/compute-cloth/triNoise.ts
```diff
@@ -0,0 +1,46 @@
...
+export const triNoise3D = tgpu.fn([d.vec3f, d.f32, d.f32], d.f32)(
+  (position, speed, time) => {
+    'use gpu';
+
+    let nodeVar0 = d.vec3f(position);
+    let nodeVar1 = d.f32(1.4);
+    let nodeVar2 = d.f32();
+    let nodeVar3 = d.vec3f(nodeVar0);
+    let nodeVar4 = d.vec3f();
+    let nodeVar5 = d.f32();
```
aleksanderkatan: How did these names come to be? 👀 
image

packages/typegpu-three/src/typegpu-material.ts
```diff
@@ -0,0 +1,109 @@
+import * as THREE from 'three/webgpu';
+import tgpu, { type TgpuFn } from 'typegpu';
+
+class FragmentNode extends THREE.CodeNode {
+  private tgslFn: TgpuFn;
+  private functionName: string | null | undefined;
+  private threeVars: THREE.TSL.ShaderNodeObject<THREE.Node>[] | undefined;
```
aleksanderkatan: Unused variables
```suggestion
  private functionName: string | null | undefined;
```

packages/typegpu/tests/tgsl/shellless.test.ts
```diff
@@ -55,7 +55,7 @@ describe('shellless', () => {
         return dot(a, a);
       }
 
-      fn foo() -> f32 {
+      fn item_0() -> f32 {
```
aleksanderkatan: What happened here?

aleksanderkatan (commented):
Great changes! 3️⃣☕📝

iwoplaza:
> I had a situation like the following:
> 
> ```ts
> const helper = () => { 'use gpu'; ... }
> 
> const anotherHelper = () => { 'use gpu'; helper(); ... }
> 
> const tslUpdate = Fn(() => { toTSL(helper); toTSL(anotherHelper); ... }
> ```
> 
> This resolves `helper` function twice. Aside from code duplication, this also made my code incorrect, because I used an injected `Math.random()` and expected different calls to return the same result. I'm not sure if we can do something about it, if we can't, we should definitely mention it in docs.
> 
> By the way, a nice side effect of console logs being not supported is that we can check exactly how many times a function is resolved.

I tried to reproduce this behavior with the following code snippet:

```ts
const signal = tgpu['~unstable'].comptime(() => {
  console.log('comptime');
  return 1;
});

const helper = () => {
  'use gpu';
  return 0.4 * signal();
};

const lava = t3.toTSL(() => {
  'use gpu';
  const coords = t3.uv().$.mul(2);
  const pattern = perlin3d.sample(d.vec3f(coords, t3.time.$ * 0.2));
  return d.vec4f(tanh(pattern * 5), 0.2, helper(), 1);
});

const lava2 = t3.toTSL(() => {
  'use gpu';
  const coords = t3.uv().$.mul(3);
  const pattern = perlin3d.sample(d.vec3f(coords, t3.time.$ * 0.2));
  return d.vec4f(0.1, tanh(pattern * 5), helper(), 1);
});

material.colorNode = TSL.Fn(() => {
  return lava.add(lava2);
})();
```

But the result was the following:
 
image

Am I missing something?

aleksanderkatan:
> Am I missing something?

Sorry, it was my fault for using this function in compute and in render at the same time.
While trying to repro this, I came across another issue: #1973

PR #1456, opened 2025-07-04
Use bool for builtin.frontFacing

deluksic:
https://www.w3.org/TR/WGSL/#front-facing-builtin-value


iwoplaza (approved):
Great catch! Thanks a lot 💜

PR #1457, opened 2025-07-04
feat: Perlin Noise derivatives

iwoplaza:


packages/typegpu-noise/src/perlin-2d/algorithm.ts
```diff
@@ -0,0 +1,79 @@
...
+  const grad = add(
+    ga,
+    add(
+      add(
+        add(mul(u.x, sub(gb, ga)), mul(u.y, sub(gc, ga))),
+        mul(u.x, mul(u.y, add(sub(sub(ga, gb), gc), gd))),
+      ),
+      mul(du, sub(add(mul(u.yx, va - vb - vc + vd), d.vec2f(vb, vc)), va)),
+    ),
+  );
```
reczkok, line 77: Could we split it into some variables to make it a little more readable?

packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
```diff
@@ -0,0 +1,252 @@
...
+  const cleanValuesSlot = tgpu['~unstable'].derived(() => {
+    return {
+      get size() {
+        // biome-ignore lint/suspicious/noExplicitAny: TS is mad at us
+        return (valuesSlot.value as any)[`${prefix}size`] as d.v2u;
+      },
+      get memory() {
+        // biome-ignore lint/suspicious/noExplicitAny: TS is mad at us
+        return (valuesSlot.value as any)[`${prefix}memory`] as d.v2f[];
+      },
```
reczkok: We could make TS a little happier
```suggestion
      get size() {
        return (valuesSlot.value as Record<`${Prefix}size`, d.v2u>)[
          `${prefix}size`
        ];
      },
      get memory() {
        return (valuesSlot.value as Record<`${Prefix}memory`, d.v2f[]>)[
          `${prefix}memory`
        ];
      },
```

packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
```diff
@@ -0,0 +1,252 @@
...
+    return cleanValuesSlot.value.memory[x + y * size.x] as d.v2f;
+  });
+
+  const computeLayout = tgpu.bindGroupLayout({
+    size: { uniform: d.vec2u },
+    memory: { storage: MemorySchema, access: 'mutable' },
+  });
+
+  const mainCompute = tgpu['~unstable'].computeFn({
+    workgroupSize: [1, 1, 1],
```
reczkok, line 164: We should consider making this more than one - if we look at our examples (game of life and fluid with atomics) there are significant performance gains when using higher workgroup sizes

reczkok: We should benchmark this but I think having workgroups at least a wavefront size (64 - which is 2 warps) with boundary checks inside the shader will be universally more performant

iwoplaza: Let's try to make this a separate PR. Could you try this out @reczkok?

packages/typegpu-noise/src/perlin-2d/algorithm.ts
iwoplaza: It's hard to find names for the variables for the sake of splitting it up. I added the equivalent GLSL expression/equation, does it help?

packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
reczkok: I will create a task for it

reczkok (approved):
Great work!

packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
reczkok: #1460

PR #1466, opened 2025-07-09
docs: Create a shader compilation tool

### LINKED ISSUE #1465: dev: Create automated tooling for shader compilation result analysis
We can use [naga](https://github.com/gfx-rs/wgpu/tree/trunk/naga) and [spirv-dis](https://github.com/KhronosGroup/SPIRV-Tools) (which is a part of the Vulkan SDK) to create SPIR-V disassembly which we can interpret to see what the compiler can optimize.

Edit: I also experimented with just using naga to go WGSL -> SPIR-V -> WGSL but the last translation seems weird

reczkok:
EDIT: I think requiring rust to be able to run the dev script is kinda stupid - I think it's better to create a separate repo for the rust tool and publish the built js package for us to use in our monorepo

closes #1465

### DIFF
--- a/apps/typegpu-docs/astro.config.mjs
+++ b/apps/typegpu-docs/astro.config.mjs
@@ -32,7 +32,11 @@ export default defineConfig({
       typegpu({ include: [/\.m?[jt]sx?/] }),
       /** @type {any} */ imagetools(),
     ],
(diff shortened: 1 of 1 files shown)


iwoplaza (approved):
Yes please! 👏👏👏

PR #1468, opened 2025-07-09
chore: Clarifying purpose of execution modes

iwoplaza:
These changes are mostly to prepare us for a potential another mode: “SIMULATE”, which will allow properly propagate slot values when running tgpu.fn on the CPU, and allow access to variables on the CPU

### DIFF
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -11,6 +11,7 @@ import {
 import * as d from '../data/index.ts';
 import { abstractInt } from '../data/numeric.ts';
 import * as wgsl from '../data/wgslTypes.ts';
+import { ResolutionError } from '../errors.ts';
 import { getName } from '../shared/meta.ts';
 import { $internal } from '../shared/symbols.ts';
 import { type FnArgsConversionHint, isMarkedInternal } from '../types.ts';
@@ -24,7 +25,6 @@ import {
   getTypeForPropAccess,
   numericLiteralToSnippet,
 } from './generationHelpers.ts';
-import { ResolutionError } from '../errors.ts';
 
 const { NodeTypeCatalog: NODE } = tinyest;
 
@@ -313,7 +313,17 @@ export function generateExpression(
     }
 
     if (wgsl.isWgslStruct(id.value)) {
-      const resolvedId = ctx.resolve(id.value);
+      // There are two ways a struct can be called that we support:
+      // - with an objectExpr `Struct({ x: 1, y: 2 })`,
+      // - with another struct `Struct(otherStruct)`.
+      // In the second case, we assume the `otherStruct` is defined on TGSL side
+      // and we just strip the constructor to let the assignment operator clone it.
+      const resolvedId = args.length === 1 &&
+          Array.isArray(args[0]) &&
+          args[0].length > 0 &&
+          args[0][0] === NODE.objectExpr
+        ? ctx.resolve(id.value)
+        : '';
 
       return snip(
         `${resolvedId}(${argValues.join(', ')})`,
--- a/packages/typegpu/tests/tgsl/wgslGenerator.test.ts
+++ b/packages/typegpu/tests/tgsl/wgslGenerator.test.ts
@@ -615,6 +615,60 @@ describe('wgslGenerator', () => {
     expect(res.dataType).toStrictEqual(d.arrayOf(TestStruct, 2));
   });
 
+  it('generates correct code when struct clone is used', () => {
+    const TestStruct = d.struct({
+      x: d.u32,
+      y: d.f32,
+    });
+
+    const testFn = tgpu.fn([])(() => {
+      const myStruct = TestStruct({ x: 1, y: 2 });
+      const myClone = TestStruct(myStruct);
+      return;
+    });
+
+    expect(parseResolved({ testFn })).toBe(
+      parse(`
+      struct TestStruct {
+        x: u32,
+        y: f32,
+      }
+
+      fn testFn() {
+        var myStruct = TestStruct(1, 2);
+        var myClone = (myStruct);
+        return;
+      }`),
+    );
+  });
+
+  it('generates correct code when complex struct clone is used', () => {
+    const TestStruct = d.struct({
+      x: d.u32,
+      y: d.f32,
+    });
+
+    const testFn = tgpu.fn([])(() => {
+      const myStructs = [TestStruct({ x: 1, y: 2 })] as const;
+      const myClone = TestStruct(myStructs[0]);
+      return;
+    });
+
+    expect(parseResolved({ testFn })).toBe(
+      parse(`
+      struct TestStruct {
+        x: u32,
+        y: f32,
+      }
+
+      fn testFn() {
+        var myStructs = array<TestStruct, 1>(TestStruct(1, 2));
+        var myClone = (myStructs[0]);
+        return;
+      }`),
+    );
+  });
+
   it('generates correct code for array expressions with derived elements', () => {
     const testFn = tgpu.fn([], d.f32)(() => {
       const arr = [derivedV2f.$, std.mul(derivedV2f.$, d.vec2f(2, 2))];


packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -79,7 +79,7 @@ class TgpuConstImpl<TDataType extends AnyWgslData>
   }
 
   get value(): InferGPU<TDataType> {
-    if (!inGPUMode()) {
+    if (!inCodegenMode()) {
```
aleksanderkatan: If we plan on adding more modes, maybe it would be clearer if there was no negation in such conditions? I think the negation is fine if we throw inside of the if (for functions that e.g. only make sense in GPU/codegen mode), but here we really are just checking if it is the CPU/eval mode.

packages/typegpu/src/execMode.ts
```diff
@@ -3,15 +3,13 @@ import type { ResolutionCtx } from './types.ts';
...
-  CPU: CPUMode,
-  GPU: GPUMode,
+export const ExecMode = {
+  COMPTIME: 0,
+  CODEGEN: 1,
 } as const;
 
-const resolutionModeStack: (typeof CPUMode | typeof GPUMode)[] = [];
+const resolutionModeStack:
+  (typeof ExecMode.COMPTIME | typeof ExecMode.CODEGEN)[] = [];
```
aleksanderkatan: How about adding a type like this?

```ts
export const ExecModes = {
  COMPTIME: 0,
  CODEGEN: 1,
} as const;

export type ExecMode = typeof ExecModes[keyof typeof ExecModes];
```
Or is it for better type hints? 

Could we use string literals instead of this object?

packages/typegpu/src/execMode.ts
```diff
@@ -28,17 +26,25 @@ export function getResolutionCtx(): ResolutionCtx | null {
...
     invariant(mode === expected, 'Unexpected mode');
   }
 }
 
-export const inGPUMode = () =>
+export const inCodegenMode = () =>
+  resolutionModeStack.length > 0 &&
+  resolutionModeStack[resolutionModeStack.length - 1] === ExecMode.CODEGEN;
+
+export const inComptimeMode = () =>
```
aleksanderkatan, line 40: I don't understand the name of this mode, is it compile time, compute time? If so, why?

aleksanderkatan:
I'm not sure if I understand what this new mode would be for. How would that differ from the old CPU mode, and why is it not possible to just be a modification to the CPU mode?

packages/typegpu/src/std/texture.ts
```diff
@@ -107,17 +107,19 @@ type TextureSampleOverload = {
 };
 
 export const textureSample: TextureSampleOverload = createDualImpl(
-  // CPU implementation
+  // JS Implementation
```
aleksanderkatan: There are many cases of this nomenclature being used that were left out (search the project for `// CPU implementation`)

iwoplaza:
> I'm not sure if I understand what this new mode would be for. How would that differ from the old CPU mode, and why is it not possible to just be a modification to the CPU mode?

We were mixing the terminology and purpose of these modes a bit. Since we're going towards being able to run more tgpu.fns on the CPU (ones accessing slots, using privateVars, etc.), I feel like it would be productive to delineate "simulating" GPU behavior on the CPU (e.g. executing tgpu.fns top-level), and running JS as part of the resolution process or at "compile time" (derived). We could keep only two modes (GPU and CPU), but I think it's productive to distinguish between JS that has access to GPU resources, and JS that shouldn't. That way, we can throw descriptive errors when somebody erroneously accesses a variable in a derived.


> will `dualImpl` become `tripleImpl`?

Nope! DualImpl will define one behavior for the CPU (in which case it will inherit the mode from the calling scope, and appropriately throw if it accesses resources in COMPTIME mode), and a CODEGEN behavior.

packages/typegpu/src/execMode.ts
iwoplaza: This mode applies when CODEGEN is interrupted with having to run some JS code at resolution time, and calling vec3f() should not return WGSL, but a v3f instance. That's why it's called COMPTIME (compile time), borrowed from Zig's comptime code execution. I welcome different name suggestions 🙏

packages/typegpu/src/execMode.ts
```diff
@@ -3,15 +3,9 @@ import type { ResolutionCtx } from './types.ts';
...
-const GPUMode = Symbol('GPU');
+export type ExecMode = 'comptime' | 'codegen';
 
-export const RuntimeMode = {
-  CPU: CPUMode,
-  GPU: GPUMode,
-} as const;
-
-const resolutionModeStack: (typeof CPUMode | typeof GPUMode)[] = [];
+const resolutionModeStack: ExecMode[] = [];
```
reczkok, line 8: Why the decision to move back to string literals? (I honestly don't remember why we moved to symbols in the first place)

packages/typegpu/src/execMode.ts
reczkok: I kinda like `COMPTIME` because both interpretations (compile time and compute time) make sense to me 😅 I would just consider dropping the `TIME` - compilation time mode sounds off to me

packages/typegpu/src/std/texture.ts
```diff
@@ -370,10 +378,10 @@ export const textureDimensions: TextureDimensionsOverload = createDualImpl(
   // CPU implementation
   (_texture: TgpuSampledTexture | TgpuStorageTexture, _level?: number) => {
     throw new Error(
-      'Texture dimensions are not supported outside of GPU mode.',
+      'Texture dimensions are not supported outside of CODEGEN mode.',
```
reczkok: Maybe we could make these errors more readable while we're at it. I don't think this message would be helpful to the user. I would go for something along the lines of:
`XYZ relies on GPU resources and cannot be executed outside of a compute dispatch or draw call`

packages/typegpu/src/execMode.ts
iwoplaza: I wanted to ask why we were using Symbols, haha. Might've been preemptive optimization, hard to say, but it's they were not being used anywhere apart from this stack. I changed it for the sake of simplicity

reczkok: Yeah - I agree

packages/typegpu/src/execMode.ts
iwoplaza: I think it's supposed to be an opposition to "runtime". I haven't heard "compilation time" be used ever, but I thought "compile time" was a widely used term.

iwoplaza: Would the alternative proposal be:
- CODEGEN
- COMPILE
- (and eventually) SIMULATE
?

packages/typegpu/src/std/texture.ts
```diff
@@ -370,10 +378,10 @@ export const textureDimensions: TextureDimensionsOverload = createDualImpl(
   // CPU implementation
   (_texture: TgpuSampledTexture | TgpuStorageTexture, _level?: number) => {
     throw new Error(
-      'Texture dimensions are not supported outside of GPU mode.',
+      '`textureDimensions` relies on GPU resources and cannot be executed outside of a compute dispatch or draw call',
```
iwoplaza:
```suggestion
      '`textureDimensions` relies on GPU resources and cannot be executed outside of a draw call',
```

packages/typegpu/src/std/texture.ts
```diff
@@ -317,9 +323,11 @@ export const textureStore: TextureStoreOverload = createDualImpl(
     _arrayIndexOrValue?: number | TexelData,
     _maybeValue?: TexelData,
   ) => {
-    throw new Error('Texture storing is not supported outside of GPU mode.');
+    throw new Error(
+      '`textureStore` relies on GPU resources and cannot be executed outside of a compute dispatch or draw call',
```
iwoplaza:
```suggestion
      '`textureStore` relies on GPU resources and cannot be executed outside of a draw call',
```

packages/typegpu/src/std/texture.ts
```diff
@@ -261,9 +265,11 @@ export const textureLoad: TextureLoadOverload = createDualImpl(
     _coords: number | v2i | v2u | v3i | v3u,
     _levelOrArrayIndex?: number,
   ) => {
-    throw new Error('Texture loading is not supported outside of GPU mode.');
+    throw new Error(
+      '`textureLoad` relies on GPU resources and cannot be executed outside of a compute dispatch or draw call',
```
iwoplaza:
```suggestion
      '`textureLoad` relies on GPU resources and cannot be executed outside of a draw call',
```

packages/typegpu/src/std/texture.ts
```diff
@@ -175,9 +177,11 @@ export const textureSampleLevel: TextureSampleLevelOverload = createDualImpl(
     _level: number,
     _offsetOrArrayIndex?: v2i | v3i | number,
   ) => {
-    throw new Error('Texture sampling is not supported outside of GPU mode.');
+    throw new Error(
+      'Texture sampling is not supported outside of CODEGEN mode.',
```
iwoplaza:
```suggestion
      'Texture sampling relies on GPU resources and cannot be executed outside of a draw call',
```

packages/typegpu/src/std/texture.ts
```diff
@@ -115,9 +115,11 @@ export const textureSample: TextureSampleOverload = createDualImpl(
     _offsetOrArrayIndex?: v2i | v3i | number,
     _maybeOffset?: v2i | v3i,
   ) => {
-    throw new Error('Texture sampling is not supported outside of GPU mode.');
+    throw new Error(
+      'Texture sampling is not supported outside of CODEGEN mode.',
```
iwoplaza:
```suggestion
      'Texture sampling relies on GPU resources and cannot be executed outside of a draw call',
```

packages/typegpu/src/std/discard.ts
```diff
@@ -5,7 +5,7 @@ import { createDualImpl } from '../shared/generators.ts';
 export const discard = createDualImpl(
   // CPU
   (): never => {
-    throw new Error('discard() can only be used on the GPU.');
+    throw new Error('`discard` relies on GPU resources and cannot be executed outside of a compute dispatch or draw call');
```
iwoplaza:
```suggestion
    throw new Error('`discard` relies on GPU resources and cannot be executed outside of a draw call');
```

reczkok (approved):
Looks good 👍

PR #1470, opened 2025-07-09
fix: Struct clone constructors on wgsl side

aleksanderkatan:
In WgslGenerator.ts, if a `WgslStruct` is called with an object expression, the behaviour is as it was before. Otherwise, we strip the struct call in order to just use the properties of wgsl assignment operator and to clone the struct value this way.

```ts
const myStruct = TestStruct({ x: 1, y: 2 });
const myClone = TestStruct(myStruct);
```

The above now resolves to:

```wgsl
var myStruct = TestStruct(1, 2);
var myClone = (myStruct);
```

### DIFF
--- a/apps/typegpu-docs/src/components/stackblitz/openInStackBlitz.ts
+++ b/apps/typegpu-docs/src/components/stackblitz/openInStackBlitz.ts
@@ -7,6 +7,9 @@ import typegpuColorPackageJson from '@typegpu/color/package.json' with {
 import typegpuNoisePackageJson from '@typegpu/noise/package.json' with {
   type: 'json',
 };
+import typegpuSdfPackageJson from '@typegpu/sdf/package.json' with {
+  type: 'json',
+};
 import typegpuPackageJson from 'typegpu/package.json' with { type: 'json' };
 import unpluginPackageJson from 'unplugin-typegpu/package.json' with {
   type: 'json',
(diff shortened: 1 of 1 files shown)


reczkok:
> The following does not resolve (neither before nor after the changes):
> 
> ```ts
> const myStruct = { x: 1, y: 2 };
> const testFn = tgpu.fn([])(() => {
>   const myClone = TestStruct(myStruct);
>   return;
> });
> ```
> 
> Do we consider this intended behaviour?
> 
> Let me know if I missed anything, I still don't understand the wgslGenerator too well.

We could make it work by treating plain js object in context of struct constructors the same way as object literals and coercing them.
```ts
// `generationHelpers.ts` add at line 610
  if (
    typeof value === 'object' &&
    value !== null
  ) {
    const ctx = getResolutionCtx() as GenerationCtx | undefined;
    if (!ctx) {
      throw new Error('Tried to coerce object without a context');
    }

    const structType = ctx.callStack[ctx.callStack.length - 1];
    if (isWgslStruct(structType)) {
      const snippetString = Object
        .keys(structType.propTypes)
        .map((key) =>
          coerceToSnippet((value as Record<string, unknown>)[key]).value
        )
        .join(', ');

      return snip(snippetString, structType);
    }
  }
```
The reason why we didn't do it is that it becomes non-trivial when it comes to nested structs - it's certainly doable, just not easy. We could create a separate ticket for that if we decide it's worth the additional logic.

reczkok (approved):
Great work 👍 I don't think there is a better way to do that without refactoring the entire system (which is long overdue #1471)

PR #1480, opened 2025-07-11
feat: Struct default constructor

aleksanderkatan:
Structs can now be called with no arguments in order to get a default value of the struct (as long as there are no nested arrays inside).

This is a prerequisite for #1254 (`d.arrayOf(myStruct, 4)()`)

### DIFF
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -313,23 +313,28 @@ export function generateExpression(
     }
 
     if (wgsl.isWgslStruct(id.value)) {
-      // There are two ways a struct can be called that we support:
+      const resolvedId = ctx.resolve(id.value);
+      // There are three ways a struct can be called that we support:
+      // - with no arguments `Struct()`,
       // - with an objectExpr `Struct({ x: 1, y: 2 })`,
       // - with another struct `Struct(otherStruct)`.
-      // In the second case, we assume the `otherStruct` is defined on TGSL side
+      // In the last case, we assume the `otherStruct` is defined on TGSL side
       // and we just strip the constructor to let the assignment operator clone it.
-      const resolvedId = args.length === 1 &&
-          Array.isArray(args[0]) &&
-          args[0].length > 0 &&
-          args[0][0] === NODE.objectExpr
-        ? ctx.resolve(id.value)
-        : '';
+      if (args.length === 0) {
+        return snip(`${resolvedId}()`, id.value);
+      }
 
-      return snip(
-        `${resolvedId}(${argValues.join(', ')})`,
-        // Unintuitive, but the type of the return value is the struct itself
-        id.value,
-      );
+      if (
+        args.length === 1 &&
+        Array.isArray(args[0]) &&
+        args[0].length > 0 &&
+        args[0][0] === NODE.objectExpr
+      ) {
+        return snip(`${resolvedId}(${argValues.join(', ')})`, id.value);
+      }
+
+      return snip(`(${argValues.join(', ')})`, id.value);
+      // The type of the return value is the struct itself
     }
 
     if (!isMarkedInternal(id.value)) {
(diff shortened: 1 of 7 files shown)


packages/typegpu/tests/struct.test.ts
```diff
@@ -277,4 +279,48 @@ describe('struct', () => {
...
+    });
+  });
+
+  // TODO: uncomment this once this starts working
+  // it('can be called to create a default value with nested array', () => {
+  //   const schema = struct({ arr: arrayOf(u32, 1) });
+
+  //   const defaultStruct = schema();
+
+  //   expect(defaultStruct).toStrictEqual({ arr: [1] });
```
reczkok: I know it's commented but why would it equal `[1]`?

packages/typegpu/tests/struct.test.ts
```diff
@@ -277,4 +279,48 @@ describe('struct', () => {
...
+
+        struct outerStruct {
+          nested: nestedStruct,
+        }
+
+        fn testFunction() {
+          var defaultValue = outerStruct();
+        }
+      `));
+  });
```
iwoplaza:
```suggestion
    const Nested = struct({ prop1: vec2f, prop2: u32 });
    const Outer = struct({
      nested: Nested,
    });

    const testFunction = tgpu.fn([])(() => {
      const defaultValue = Outer();
    });

    expect(parseResolved({ testFunction })).toBe(parse(`
        struct Nested {
          prop1: vec2f,
          prop2: u32,
        }

        struct Outer {
          nested: Nested,
        }

        fn testFunction() {
          var defaultValue = Outer();
        }
      `));
  });
```

packages/typegpu/tests/struct.test.ts
```diff
@@ -254,7 +256,7 @@ describe('struct', () => {
     });
 
     // @ts-expect-error
-    TestStruct({ x: 1, z: 2 });
+    (() => TestStruct({ x: 1, z: 2 }));
```
iwoplaza, line 259: Why this change?

aleksanderkatan: I think we shouldn't run snippets that are invalid on type level. For example, in this case, the value `undefined` was passed in to the `vec3u` constructor, which throws because it tries to interpret this as either a number, a boolean or a vector.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -313,23 +313,28 @@ export function generateExpression(
...
+      if (
+        args.length === 1 &&
+        Array.isArray(args[0]) &&
+        args[0].length > 0 &&
+        args[0][0] === NODE.objectExpr
+      ) {
+        return snip(`${resolvedId}(${argValues.join(', ')})`, id.value);
+      }
+
+      return snip(`(${argValues.join(', ')})`, id.value);
```
iwoplaza: Why is there no ID being called? Am I missing something, or is there missing test coverage for this case?
```suggestion
      return snip(`${resolvedId}(${argValues.join(', ')})`, id.value);
```

aleksanderkatan: This logic was already there before, as a part of #1470:
```ts
const resolvedId = args.length === 1 &&
    Array.isArray(args[0]) &&
    args[0].length > 0 &&
    args[0][0] === NODE.objectExpr
  ? ctx.resolve(id.value)
  : '';
```
`Struct(otherStruct)` is parsed to `(otherStruct)`, so that it can be cloned in assignments (wgsl has no copy constructor for structs)

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -313,23 +313,28 @@ export function generateExpression(
...
+      // The type of the return value is the struct itself
```
reczkok:
```suggestion
      // The type of the return value is the struct itself
      return snip(`(${argValues.join(', ')})`, id.value);
```

packages/typegpu/tests/struct.test.ts
aleksanderkatan: It shouldn't 🤷‍♀️

PR #1482, opened 2025-07-12
fix: Proper type for .value and .$ in TypeGPU variables

iwoplaza:


packages/typegpu/src/core/variable/tgpuVariable.ts
```diff
@@ -100,18 +101,28 @@ class TgpuVarImpl<TScope extends VariableScope, TDataType extends AnyData>
...
-  get value(): Infer<TDataType> {
+  get $(): InferGPU<TDataType> {
+    if (inCodegenMode()) {
+      return this[$gpuValueOf]();
+    }
+
+    throw new Error(
+      '`tgpu.var` relies on GPU resources and cannot be accessed outside of a compute dispatch or draw call',
+    );
+  }
```
aleksanderkatan: Could you remind me what the difference between `.value` and `.$` is? Isn't this the exact same code?

iwoplaza: `.value` and `.$` are the same, we can discuss a plan for keeping just one of them going forward and deprecating the other.

aleksanderkatan: OK, in this case, can we eliminate code duplication by changing this to the following?
```ts
get $(): InferGPU<TDataType> {
  return this.value;
}
```

PR #1486, opened 2025-07-14
feat: Fluent operators interface

aleksanderkatan:


packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -20,16 +20,54 @@ import type { Prettify } from '../shared/utilityTypes.ts';
 
 type DecoratedLocation<T extends BaseData> = Decorated<T, Location[]>;
 
+export interface BaseData {
+  readonly [$internal]: true;
+  readonly type: string;
+  readonly [$repr]: unknown;
+}
```
iwoplaza, line 27:
```suggestion
}

```

packages/typegpu/src/shared/generators.ts
```diff
@@ -21,7 +21,7 @@ export function* naturalsExcept(
   }
 }
 
-type MapValueToSnippet<T> = { [K in keyof T]: Snippet };
+export type MapValueToSnippet<T> = { [K in keyof T]: Snippet };
```
iwoplaza: I would move this close the the `Snippet` interface definition, to avoid any circular dependencies in the future.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -14,6 +15,7 @@ import * as wgsl from '../data/wgslTypes.ts';
 import { ResolutionError } from '../errors.ts';
 import { getName } from '../shared/meta.ts';
 import { $internal } from '../shared/symbols.ts';
+import * as std from '../std/index.ts';
```
iwoplaza: Let's not import the std functions from the barrel file internally, but instead the individual files that export them.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -51,6 +53,33 @@ const parenthesizedOps = [
 
 const binaryLogicalOps = ['&&', '||', '==', '!=', '<', '<=', '>', '>='];
 
+const infixableKind = [
```
iwoplaza: Cannot be fixed kind 😭

packages/typegpu/src/data/dataTypes.ts
```diff
@@ -195,6 +197,17 @@ export const UnknownData = {
...
+export class InfixDispatch {
+  constructor(
+    private lhs: Snippet,
+    private operator: (lhs: Snippet, rhs: Snippet) => Snippet,
+  ) {}
+
+  applyWith(rhs: Snippet): Snippet {
+    return this.operator(this.lhs, rhs);
+  }
+}
```
iwoplaza: Since this is an internal construct, we can keep this just a data container, meaning `lhs` and `operator` can be public, and we use both in `wgslGenerator.ts` directly. I would also include the name of the operator (`mul`, `add`, etc.) so we can provide a better error message down the line.

iwoplaza:
```suggestion
export class InfixDispatch {
  constructor(
    readonly name: string,
    readonly lhs: Snippet,
    readonly operator: (lhs: Snippet, rhs: Snippet) => Snippet,
  ) {}
}
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -332,6 +374,13 @@ export function generateExpression(
       );
     }
 
+    if (id.value instanceof InfixDispatch) {
+      if (!argSnippets[0]) {
+        throw new Error('An infix operator was called without any arguments!');
```
iwoplaza: No need to shout 🤫 (we can remove the exclamation point). Also, we can tell the user which infix operator broke.
```suggestion
        throw new Error(`An infix operator '${id.value.name}' was called without any arguments`);
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -332,6 +374,13 @@ export function generateExpression(
...
+      }
+      return id.value.applyWith(argSnippets[0]);
```
iwoplaza:
```suggestion
      return id.value.operator(id.value.operator.lhs, argSnippets[0]);
```

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: Is leaving the `d` import fine?

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: I renamed this to `infixKinds`

packages/typegpu/src/data/index.ts
```diff
@@ -2,6 +2,32 @@
...
+  object: typeof VecBase | typeof MatBase,
+  operator: InfixOperator,
+) {
+  // @ts-ignore
+  // biome-ignore lint/suspicious/noExplicitAny: <shh>
+  object.prototype[operator] = function (this: any, other: any) {
+    // @ts-ignore
+    return std[operator][$internal].jsImpl(this, other);
+  };
+}
```
reczkok:
```suggestion
function assignInfixOperator<T extends typeof VecBase | typeof MatBase>(
  object: T,
  operator: InfixOperator,
) {
  type Instance = InstanceType<T>;

  const proto = object.prototype as {
    [K in InfixOperator]?: (this: Instance, other: unknown) => unknown;
  };
  const opImpl = std[operator][$internal].jsImpl as (
    lhs: Instance,
    rhs: unknown,
  ) => unknown;

  proto[operator] = function (this: Instance, other: unknown): unknown {
    return opImpl(this, other);
  };
}
```
Functionally the same but I would sleep better at night

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: It's fine, because `wgslTypes.ts` is not a barrel file (reexporting things from other files), but the source of what we're importing.

PR #1495, opened 2025-07-15
feat: concurrent sum (for huge numbers)

lursz:


deluksic:
Is this a concurrent prefix sum? This is exactly what I would need to implement line dashing on the GPU. Basically, you get an array of points and you need to compute total distance up to that point so you can color the dashes in the fragment shader.

If you let users pass a custom (must be associative) operator, you could also get prefix max, prefix min, etc.

packages/typegpu-concurrent-sum/src/compute/cacheOld.ts
```diff
@@ -0,0 +1,42 @@
...
+  private readIndex = 0;
+  private writeIndex = 0;
+  private count = 4;
+
+  constructor(private root: TgpuRoot, private n: number, private capacity = 4) {
+    this.buffers = {
+      0: root.createBuffer(d.arrayOf(d.u32, n)).$usage('storage'),
+      1: root.createBuffer(d.arrayOf(d.u32, n)).$usage('storage'),
+      2: root.createBuffer(d.arrayOf(d.u32, n)).$usage('storage'),
+      3: root.createBuffer(d.arrayOf(d.u32, n)).$usage('storage'),
```
lursz, line 18: Scan should require only 3 buffers

packages/typegpu-concurrent-sum/src/concurrentSum.ts
```diff
@@ -0,0 +1,192 @@
...
+  function recursiveScan(
+    n: number,
+    inputBuffer: TgpuBuffer<d.WgslArray<d.U32>> & StorageFlag,
+  ): TgpuBuffer<d.WgslArray<d.U32>> & StorageFlag {
+    const itemsPerWorkgroup = workgroupSize * 2;
+    depthCount++;
+
+    // Up-pass
+    const outputBuffer = cache.pop();
+    const upPassSumsBuffer = cache.pop();
```
lursz, line 77: Introduce double-buffering

packages/typegpu-concurrent-sum/src/compute/computeUpPass.ts
```diff
@@ -0,0 +1,69 @@
...
+  // copy input data to shared memory
+  const idx0 = gId * 2;
+  const idx1 = gId * 2 + 1;
+  if (idx0 < totalInputLength) {
+    sharedMem.value[lid.x * 2] = layout.$.inputArray[idx0] as number;
+  }
+  if (idx1 < totalInputLength) {
+    sharedMem.value[lid.x * 2 + 1] = layout.$.inputArray[idx1] as number;
+  }
+  std.workgroupBarrier();
```
lursz, line 37: Don't initially copy the whole buffer, just fill it out as the scan progresses

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,191 @@
...
+    );
+
+    buffer = this.recursiveScan(buffer, buffer.dataType.elementCount);
+    this.root['~unstable'].flush();
+
+    this.qs?.resolve();
+    this.qs?.read().then((timestamps) => {
+      const diff = Number(timestamps[1]! - timestamps[0]!) / 1_000_000;
+      console.log(`Prefix sum computed in ${diff} ms`);
+    });
```
lursz: During the PR review, please do ignore all code designated for time calculation, it will be eventually removed.

Also ignore all code associated with example itself. I placed it there only for testing purposes. It will get removed.

packages/typegpu-concurrent-sum/src/compute/singleScan.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  // Upsweep
+  for (let d_val = d.u32(workgroupSize / 2); d_val > 0; d_val >>= 1) {
+    std.workgroupBarrier();
+    if (localIdx < d_val) {
+      const ai = offset * (2 * localIdx + 1) - 1;
+      const bi = offset * (2 * localIdx + 2) - 1;
+      workgroupMemory.$[bi] = operatorSlot.$(
+        workgroupMemory.$[bi],
+        workgroupMemory.$[ai] as number,
```
aleksanderkatan: If the op is matrix multiplication, this will give wrong result

apps/typegpu-docs/src/examples/simple/triangle-next/thumbnail.png
aleksanderkatan, line 1: Change this thumbnail, it is unrelated to the example

packages/typegpu-concurrent-sum/README.md
aleksanderkatan, line 1: I think a simple usage example of the library would suffice to remove the **Under Construction**

packages/typegpu-concurrent-sum/src/compute/singleScan.ts
lursz: Yup, fair

packages/typegpu-concurrent-sum/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@typegpu/concurrent-sum",
+  "type": "module",
+  "version": "0.0.8",
+  "description": "A concurrent sum module.",
```
aleksanderkatan: Do we want to keep this name if we allow for all kinds of operations, and we do not only the sum, but also the entire prefix sum? concurrent-reduce, parallel-reduce, gpu-scan, kernel-fold?

apps/typegpu-docs/src/content/examples/algorithms/concurrent-sum/index.ts
aleksanderkatan, line 1: The example does not work for me:
```
- computePipelineCore
- computeFn:uniformAdd: Missing value for 'slot:operatorSlot'
    at ResolutionError.appendToTrace (errors.ts:75:12)
    at ResolutionCtxImpl._getOrInstantiate (resolutionCtx.ts:604:19)
    at resolutionCtx.ts:625:54
    at provideCtx (execMode.ts:42:12)
    at ResolutionCtxImpl.resolve (resolutionCtx.ts:625:26)
    at resolve (resolutionCtx.ts:717:11)
    at ComputePipelineCore.unwrap (computePipeline.ts:243:28)
    at TgpuComputePipelineImpl.dispatchWorkgroups (computePipeline.ts:157:29)
    at PrefixScanComputer.recursiveScan (prefixScan.ts:143:8)
    at PrefixScanComputer.recursiveScan (prefixScan.ts:130:23)Caused by: MissingSlotValueError: Missing value for 'slot:operatorSlot'
    at ResolutionCtxImpl.readSlot (resolutionCtx.ts:455:13)
    at ResolutionCtxImpl.unwrap (resolutionCtx.ts:497:30)
    at [typegpu:0.7.0:$gpuValueOf] (slot.ts:51:44)
    at get value (slot.ts:60:28)
    at get $ (slot.ts:64:17)
    at generateExpression (wgslGenerator.ts:266:33)
    at generateExpression (wgslGenerator.ts:358:20)
    at generateExpression (wgslGenerator.ts:193:21)
    at generateStatement (wgslGenerator.ts:696:35)
    at wgslGenerator.ts:125:7
```

apps/typegpu-docs/src/content/examples/algorithms/concurrent-sum/index.html
```diff
@@ -0,0 +1,6 @@
+<div>
+  <canvas></canvas>
+  <button id="runButton" style="display: block; margin: 1em 0">
```
aleksanderkatan: Why not use the example controls instead?

apps/typegpu-docs/src/content/examples/algorithms/concurrent-sum/index.ts
```diff
@@ -0,0 +1,95 @@
...
+  'Array length': {
+    initial: 2 ** 22,
+    ...paramSettings,
+    onSliderChange: (value: number) => {
+      arraySize = value;
+      // Update the inputArray when the slider changes
+      inputArray = Array.from({ length: arraySize }, (_, i) => 1);
+    },
+  },
+};
```
aleksanderkatan: Please make it a select instead of a slider

packages/typegpu-concurrent-sum/src/compute/scan.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (baseIdx + i < arrayLength) {
+      elements[i] = scanLayout.$.input[baseIdx + i] as number;
+    }
+  }
+
+  const partialSums = [d.f32(0), 0, 0, 0, 0, 0, 0, 0];
+  partialSums[0] = elements[0] as number;
+  for (let i = d.u32(1); i < 8; i++) {
+    partialSums[i] = partialSums[i - 1] as number + (elements[i] as number);
+  }
```
lursz: Merge this into single loop

packages/typegpu-concurrent-sum/src/compute/scan.ts
```diff
@@ -0,0 +1,98 @@
...
+  const totalSum = partialSums[7];
```
reczkok:
```suggestion
  const partialSums = [d.f32(), 0, 0, 0, 0, 0, 0, 0];
  for (let i = d.u32(); i < 8; i++) {
    if (baseIdx + i < arrayLength) {
      partialSums[i] = scanLayout.$.input[baseIdx + i] as number;
      if (i > 0) {
        (partialSums[i] as number) += partialSums[i - 1] as number;
      }
    }
  }

  // copy to shared memory
  workgroupMemory.$[localIdx] = partialSums[7] as number;
```

packages/typegpu-concurrent-sum/src/compute/scan.ts
```diff
@@ -0,0 +1,98 @@
...
+  for (let i = d.u32(0); i < 8; i++) {
+    if (baseIdx + i < arrayLength) {
+      elements[i] = scanLayout.$.input[baseIdx + i] as number;
+    }
+  }
+
+  const partialSums = [d.f32(0), 0, 0, 0, 0, 0, 0, 0];
+  partialSums[0] = elements[0] as number;
+  for (let i = d.u32(1); i < 8; i++) {
+    partialSums[i] = partialSums[i - 1] as number + (elements[i] as number);
```
aleksanderkatan: No operatorSlot here?

packages/typegpu-concurrent-sum/sum.md
aleksanderkatan, line 1: This file should be deleted, right?

packages/typegpu-concurrent-sum/src/schemas.ts
```diff
@@ -0,0 +1,16 @@
+import tgpu, { type TgpuFn } from 'typegpu';
+import * as d from 'typegpu/data';
+
+export const workgroupSize = 256;
```
aleksanderkatan: 256 is the worst-case-scenario workgroup size limit, I think it might be worth it to check the device limits (`root.device.limits.maxComputeWorkgroupSizeX`, `...Y`, `...Z`) and make this variable dependent on it.

packages/typegpu-concurrent-sum/src/schemas.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const scanLayout = tgpu.bindGroupLayout({
+  input: { storage: (n: number) => d.arrayOf(d.f32, n), access: 'mutable' },
+  sums: { storage: (n: number) => d.arrayOf(d.f32, n), access: 'mutable' },
```
aleksanderkatan: nit:
```suggestion
  input: { storage: d.arrayOf(d.f32), access: 'mutable' },
  sums: { storage: d.arrayOf(d.f32), access: 'mutable' },
```

packages/typegpu-concurrent-sum/src/schemas.ts
```diff
@@ -0,0 +1,16 @@
...
+});
+
+export const uniformAddLayout = tgpu.bindGroupLayout({
+  input: { storage: (n: number) => d.arrayOf(d.f32, n), access: 'mutable' },
+  sums: { storage: (n: number) => d.arrayOf(d.f32, n), access: 'readonly' },
```
aleksanderkatan: nit:
```suggestion
  input: { storage: d.arrayOf(d.f32), access: 'mutable' },
  sums: { storage: d.arrayOf(d.f32), access: 'readonly' },
```

packages/typegpu-concurrent-sum/src/schemas.ts
```diff
@@ -0,0 +1,16 @@
+import tgpu, { type TgpuFn } from 'typegpu';
+import * as d from 'typegpu/data';
+
+export const workgroupSize = 256;
+
+export const scanLayout = tgpu.bindGroupLayout({
+  input: { storage: (n: number) => d.arrayOf(d.f32, n), access: 'mutable' },
```
aleksanderkatan: Should input be mutable?

packages/typegpu-concurrent-sum/src/schemas.ts
aleksanderkatan: Why is input mutable, and sums readonly? Shouldn't it be the other way around?

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  uniformAddLayout,
+  workgroupSize,
+} from './schemas.ts';
+import { scanBlock } from './compute/scan.ts';
+import { uniformAdd } from './compute/applySums.ts';
+import { scanGreatestBlock } from './compute/singleScan.ts';
+
+let computer: PrefixScanComputer | null = null;
+
+export class PrefixScanComputer {
```
aleksanderkatan: Should this be exported?

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+}
+
+export function initConcurrentSum(
+  root: TgpuRoot,
+  operatorFn: (x: number, y: number) => number,
+  identity: number,
+): void {
+  // allocate all resources ahead of demand
+  computer = new PrefixScanComputer(root, operatorFn, identity, false);
+}
```
aleksanderkatan: Is this function necessary? `concurrentScan` seems to do this and more.

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  computer ??= new PrefixScanComputer(
+    root,
+    operatorFn,
+    identity,
+    onlyGreatestElement,
+  );
+  const result = computer.compute(buffer);
+
+  return result;
+}
```
aleksanderkatan: If I want to sum two different arrays, it will just compute the sum of the first one again. I'd keep a cache of `PrefixScanComputer`s indexed by `{ root, buffer, operatorFn, identity }`.

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+): void {
+  // allocate all resources ahead of demand
+  computer = new PrefixScanComputer(root, operatorFn, identity, false);
+}
+
+export function concurrentScan(
+  root: TgpuRoot,
+  buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
+  operatorFn: (x: number, y: number) => number,
+  identity: number,
```
aleksanderkatan: nit: I'd rename this to `identityElement` or `neutralElement`

packages/typegpu-concurrent-sum/package.json
aleksanderkatan: When we decide on a name, please make sure to replace all occurrences of the "synonyms" with it, like `concurrentScan`, `PrefixSumComputer` etc.

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+export function initConcurrentSum(
+  root: TgpuRoot,
+  operatorFn: (x: number, y: number) => number,
+  identity: number,
+): void {
+  // allocate all resources ahead of demand
+  computer = new PrefixScanComputer(root, operatorFn, identity, false);
+}
+
+export function concurrentScan(
```
aleksanderkatan: Please add jsdocs here with a simple example usage

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  root: TgpuRoot,
+  buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
+  operatorFn: (x: number, y: number) => number,
```
aleksanderkatan: Do we plan on generalizing this to structs in this PR?

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  private qs: TgpuQuerySet<'timestamp'> | null = null;
+  private first = true;
+
+  constructor(
+    private root: TgpuRoot,
+    private operatorFn: (x: number, y: number) => number,
+    private identity: number,
+    private onlyGreatestElement: boolean,
+  ) {
+    this.root = root;
```
aleksanderkatan:
```suggestion
```

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+    this.qs = root.createQuerySet('timestamp', 2);
+  }
+
+  private get ScanPipeline(): TgpuComputePipeline {
```
aleksanderkatan: Getters/setters should probably be in camelCase

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+        this.root.createBuffer(d.arrayOf(d.f32, size)).$usage('storage'),
+      );
+    }
+    return this.scratchBuffers.get(key) as
+      & TgpuBuffer<d.WgslArray<d.F32>>
+      & StorageFlag;
+  }
+
+  private recursiveScan(
+    buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
```
aleksanderkatan: Please try to name variables more descriptively

packages/typegpu-concurrent-sum/src/prefixScan.ts
iwoplaza: I suggest a similar change, but to cache using `root` and '{ operator, neutral }`. The operator and neutral element are usually coupled together, so we can group them into a BinaryOp type. If we were to cache against the neutral element, we would have to do it in a non-weak map.

We don't have to cache against the buffer, as we don't have to recompile the shader on buffer change

packages/typegpu-concurrent-sum/src/schemas.ts
lursz: This is intended, in this case buffer operations are being done in-place

packages/typegpu-concurrent-sum/src/prefixScan.ts
lursz: This allows you to allocate all resources ahead of demand

packages/typegpu-concurrent-sum/src/schemas.ts
aleksanderkatan: Is this what we want to do though? Is it possible to reverse the upsweep so that the buffer is still usable after the greatest sum is calculated?

packages/typegpu-concurrent-sum/src/compute/scan.ts
```diff
@@ -0,0 +1,93 @@
...
+export const scanBlock = tgpu['~unstable'].computeFn({
+  workgroupSize: [workgroupSize],
+  in: {
+    gid: d.builtin.globalInvocationId,
+    lid: d.builtin.localInvocationId,
+    nwg: d.builtin.numWorkgroups,
+    wid: d.builtin.workgroupId,
+  },
+})(({ gid, lid, nwg, wid }) => {
+  const globalIdx = gid.x + gid.y * nwg.x + gid.z * nwg.x * nwg.y;
```
aleksanderkatan: Maybe a helper for this?
```ts
const calculateIndex = tgpu.fn([d.vec3u, d.vec3u], d.u32)((id, nwg) =>
  id.x + id.y * nwg.x + id.z * nwg.x * nwg.y
);
```

packages/typegpu-concurrent-sum/src/compute/scan.ts
```diff
@@ -0,0 +1,93 @@
...
+  const globalWid = wid.x + wid.y * nwg.x + wid.z * nwg.x * nwg.y;
+  const localIdx = lid.x;
+  const arrayLength = scanLayout.$.input.length;
+  let offset = d.u32(1);
+
+  // 8 elements per thread
```
aleksanderkatan: Have you tried experimenting with this value to check which one will be the most efficient? I'd extract this to a constant. The array below can be initialized with `d.arrayOf(d.f32, ELEMENTS_PER_THREAD)()`.

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  private scratchBuffers: Map<
+    number,
+    TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag
+  > = new Map();
+  private qs: TgpuQuerySet<'timestamp'> | null = null;
+  private first = true;
+
+  constructor(
+    private root: TgpuRoot,
+    private operatorFn: (x: number, y: number) => number,
```
aleksanderkatan: This should probably be a TgpuFn, right? Or a union of types/new type, like TgpuCallable if we want to allow std functions here.

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  ) {
+    this.root = root;
+    this.qs = root.createQuerySet('timestamp', 2);
+  }
+
+  private get ScanPipeline(): TgpuComputePipeline {
+    if (!this.scanPipeline) {
+      this.scanPipeline = this.root['~unstable'].with(
+        operatorSlot,
+        this.operatorFn as unknown as TgpuFn,
```
aleksanderkatan: You can't do that. An arrow function passed in here will just throw an error.

packages/typegpu-concurrent-sum/src/prefixScan.ts
aleksanderkatan, line 1: It does not work for different ops. For example, the result of multiplying an array of ten 2s is `[1, 3, 5, 7, 9, 11, 13, 15, 16, 18]`, and for max (neutral=-10000) it's `[-10000, -9998, -9996, -9994, -9992, -9990, -9988, -9986, 16, 18]`

aleksanderkatan, line 1: It returns a wrong result for the onlyGreatestElement when the array size is not a multiple of 8, e.g. 28 for `[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]`.

packages/typegpu-concurrent-sum/src/prefixScan.ts
```diff
@@ -0,0 +1,197 @@
...
+  computer = new PrefixScanComputer(root, operatorFn, identity, false);
+}
+
+export function concurrentScan(
+  root: TgpuRoot,
+  buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
+  operatorFn: (x: number, y: number) => number,
+  identity: number,
+  onlyGreatestElement = false,
+): TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag {
```
aleksanderkatan: Can you try to return a buffer of d.f32 (instead of an array of length 1) when onlyGreatestElement is true?

aleksanderkatan (changes requested):

aleksanderkatan:
Could you add a test example (distinct from the performance example) purely for testing this library?

packages/typegpu-concurrent-sum/src/compute/scan.ts
lursz: It only needs to be calculated once, I don't think creating helper function is necessary here

aleksanderkatan: This snippet appears 4 times in the code

packages/typegpu-concurrent-sum/src/schemas.ts
reczkok: I think there is no benefit to doing that. Threads operate in groups of up to 64 anyway so there is no parallelisation benefit. As long as it's a multiple of 64 we should be dandy.

packages/typegpu-concurrent-sum/src/schemas.ts
reczkok: It's possible but if you want that then just copy the buffer - i think the cost will be smaller

packages/typegpu-concurrent-sum/src/schemas.ts
reczkok: Nope - the sums in the pass are used to increment the input buffer in the correct places. They are already calculated in this pass.

packages/typegpu-concurrent-sum/src/compute/scan.ts
reczkok: The number 8 was based on [this article](https://developer.nvidia.com/gpugems/gpugems3/part-vi-gpu-computing/chapter-39-parallel-prefix-sum-scan-cuda) and in my limited testing (1,2,4,8,16,32) was the fastest. (very close with 4)

packages/typegpu-concurrent-sum/src/schemas.ts
aleksanderkatan: Can we implement this buffer copy as a concurrentSum mode? I think this may be a common enough scenario to warrant implementing it

packages/typegpu-concurrent-sum/src/schemas.ts
aleksanderkatan: Makes sense 👍

packages/typegpu-concurrent-sum/src/schemas.ts
lursz: Current implementation (one returning only the greatest element) does not modify the input buffer, so 👍🏿

packages/typegpu-concurrent-sum/src/prefixScan.ts
lursz: Aye aye

packages/typegpu-concurrent-sum/src/compute/scan.ts
lursz: True

packages/typegpu-concurrent-sum/src/prefixScan.ts
lursz: Resolved, introduced `TgpuCallable` type

packages/typegpu-concurrent-sum/src/prefixScan.ts
lursz: We can, but it would require a write into the new buffer, which hampers the performance. I agree, however, that intuitively this feels like a correct default behaviour

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/calculator.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+  );
+
+  root['~unstable'].flush();
+  await root.device.queue.onSubmittedWorkDone();
+  const gpuTime = performance.now() - gpuStartTime;
+
+  // Compare results
+  const gpuResult = await calcResult.read();
+  const gpuShaderTime = await timePromise;
```
reczkok: This could be a little simpler 
```suggestion
let timestampPromise: Promise<bigint[]> = Promise.resolve([]);
  const calcResult = prefixScan(
    root,
    inputBuffer,
    { operation: std.add, identityElement: 0 },
    (timeTgpuQuery) => {
      timestampPromise = timeTgpuQuery.read();
    },
  );

  root['~unstable'].flush();
  await root.device.queue.onSubmittedWorkDone();
  const gpuTime = performance.now() - gpuStartTime;

  // Compare results
  const gpuResult = await calcResult.read();
  const timestamps = await timestampPromise;
  const gpuShaderTime = Number(timestamps[1] - timestamps[0]) / 1_000_000;
```

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/calculator.ts
```diff
@@ -0,0 +1,75 @@
...
+import * as d from 'typegpu/data';
+import type { TgpuRoot } from 'typegpu';
+import * as std from 'typegpu/std';
+
+type SumResult = {
+  success: boolean;
+  jsTime?: number;
+  gpuTime?: number;
+  gpuShaderTime?: number;
+  expectedSum?: number;
```
reczkok: Never used 
```suggestion
```

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/calculator.ts
```diff
@@ -0,0 +1,75 @@
...
+    };
+  }
+
+  return {
+    success: true,
+    jsTime,
+    gpuTime,
+    gpuShaderTime,
+    expectedSum: jsResult[jsResult.length - 1],
+  };
```
reczkok:
```suggestion
  const isEqual = compareArrayWithBuffer(jsResult, gpuResult);

  return {
    success: isEqual,
    jsTime,
    gpuTime,
    gpuShaderTime,
  };
```

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export function compareArrayWithBuffer(
+  arr1: number[],
+  arr2: number[],
+): boolean {
+  if (arr1.length !== arr2.length) {
+    return false;
+  }
+  for (let i = 0; i < arr1.length - 1; i++) {
+    if (arr1[i] !== arr2[i + 1]) {
```
reczkok: Could you add a comment explaining why we are checking different indices?

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/utils.ts
```diff
@@ -0,0 +1,35 @@
+export function concurrentSumOnJS(arr: number[]) {
+  arr.reduce((accumulator, currentValue, index) => {
+    if (index > 0) {
+      arr[index] = arr[index - 1] + currentValue;
+    }
+    return arr[index];
+  }, 0);
+  return arr;
+}
```
reczkok: I think a good old for loop may be the better option here (same for multiplication)
```suggestion
export function concurrentSumOnJS(arr: number[]) {
  for (let i = 1; i < arr.length; i++){ 
    arr[i] += arr[i - 1];
  }
  return arr;
}
```

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/calculator.ts
lursz: True

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/calculator.ts
```diff
@@ -0,0 +1,60 @@
+import {
+  initCache,
+  prefixScan,
+} from '../../../../../../../packages/typegpu-concurrent-scan/src/index.ts';
```
reczkok: import from package not relative

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/thumbnail.png
reczkok, line 1: please update the thumbnail

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+
+export function concurrentMultiplyOnJS(arr: number[]) {
+  arr.reduce((accumulator, currentValue, index) => {
+    if (index > 0) {
+      arr[index] = arr[index - 1] * currentValue;
+    }
+    return arr[index];
+  }, 1);
+  return arr;
```
reczkok: update like sum

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/utils.ts
```diff
@@ -0,0 +1,46 @@
...
+    return [0, 0, 0, 0, 0];
+  }
+
+  const labels: number[] = [];
+  for (let i = 0; i < 5; i++) {
+    labels.push((i / 4) * maxValue);
+  }
+
+  return labels;
+}
```
reczkok: We can remove this helper since it's only used once and inline it
```ts
yAxisTicks = overallMax <= 0
  ? [0, 0, 0, 0, 0]
  : Array.from({ length: 5 }, (_, i) => (i / 4) * overallMax);
```

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,231 @@
...
+  private _addPipeline?: TgpuComputePipeline;
+  private scratchBuffers: Map<
+    number,
+    TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag
+  > = new Map();
+  private querySet: TgpuQuerySet<'timestamp'> | null = null;
+  private first = true;
+  private _timeCallback?:
+    | ((timeTgpuQuery: TgpuQuerySet<'timestamp'>) => void)
+    | undefined;
```
reczkok: Use [proper private variables](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,231 @@
...
+    if (!this.scratchBuffers.has(key)) {
+      this.scratchBuffers.set(
+        key,
+        this.root.createBuffer(d.arrayOf(d.f32, size)).$usage('storage'),
+      );
+    }
+    return this.scratchBuffers.get(key) as
+      & TgpuBuffer<d.WgslArray<d.F32>>
+      & StorageFlag;
+  }
```
reczkok: I think this is pointless since there is not a single cache hit in my testing - either simplify or (better) come up with a better way for sharing buffers

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,231 @@
...
+  private recursiveScan(
+    buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
+    actualLength: number,
+    level = 0,
+  ): TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag {
+    const numWorkgroups = Math.ceil(actualLength / (workgroupSize * 8));
+
+    // Base case: single workgroup
+    if (numWorkgroups === 1) {
+      const dummySums = this.getScratchBuffer(1, level);
```
reczkok: This name is misleading since they are not dummy at all
```suggestion
      const sums = this.getScratchBuffer(1, level);
```

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/index.ts
aleksanderkatan, line 1: image

Why is the first GPU time so long sometimes? Is WebGPU warming up or something?

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,231 @@
...
+  }
+
+  if (timeCallback) {
+    (computer as PrefixScanComputer).updateTimeCallback(timeCallback);
+  }
+
+  return computer.compute(buffer);
+}
+
+export function scan(
```
aleksanderkatan: Please add some basic jsdocs to the exported stuff

packages/typegpu-concurrent-sum/src/prefixScan.ts
aleksanderkatan: This is resolved but this was not fixed
```ts
const buffer = root
  .createBuffer(d.arrayOf(d.f32, 10), [0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
  .$usage('storage');
const result = scan(root, buffer, {
  identityElement: 0,
  operation: std.add,
});
console.log(await result.read());
```

Reply: Could you please provide more context? In current state this indeed should be working (for me returns what seems to be correct result)
image

packages/typegpu-concurrent-sum/src/prefixScan.ts
aleksanderkatan: So, do we? You gave a like, resolved the conversation and didn't do it.

Reply: Unfortunately no, this feature is not going in as a part of this PR. It will get added in the future in another PR.

packages/typegpu-concurrent-sum/src/prefixScan.ts
aleksanderkatan: There are no jsdocs to be seen, yet this conversation has been resolved.

Reply: Same as with other docs, will do, just as the very end. My mistake for resolving

packages/typegpu-concurrent-scan/src/prefixScan.ts
aleksanderkatan, line 1: Here is a free test case with a function that has a neutral element (0), is associative but is not communicative. This would be cleaner if we allowed other numbers than `d.f32`.

```ts
/**
 * Concats two numbers. Loses precision when the result has more than 7 digits.
 * @example
 * concat(123, 456); // 123456
 */
const concat10 = tgpu.fn([d.f32, d.f32], d.f32)((a, b) => {
  if (a === 0) return b;
  if (b === 0) return a;
  if (b === 1) return a * 10 + b;
  const digits = std.ceil(std.log(b) / std.log(10));
  const result = std.pow(10, digits) * a + b;
  const roundedResult = std.ceil(result - 0.4);
  return roundedResult;
});

// deno-fmt-ignore
const buffer = root
  .createBuffer(d.arrayOf(d.f32, 16), [0, 0, 1, 0, 2, 0, 3, 0, 4, 0, 5, 6, 0, 0, 7, 0])
  .$usage('storage');
const result = prefixScan(root, buffer, {
  identityElement: 0,
  operation: concat10,
});

// deno-fmt-ignore
console.log("Expected:", [0, 0, 0, 1, 1, 12, 12, 123, 123, 1234, 1234, 12345, 123456, 123456, 123456, 1234567])
console.log('Actual:', await result.read());
```

aleksanderkatan (changes requested):
Please add some tests

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/index.ts
lursz: Could you please provide a corresponding performance graph? In my case it happens pretty seldom and is correlated to garbage collection

packages/typegpu-concurrent-scan/src/prefixScan.ts
lursz: Will do, just waiting till everyone approves the current impl, so I don't have to do it twice.

packages/typegpu-concurrent-scan/src/prefixScan.ts
lursz: Good point, I'll remove the caching mechanism, but will keep the buffer getter (as a convenient buffer pool rather than an actual cache). Unfortunately, reusing buffers and creating actual cache isn't frivolous in this case, because:
- we need to keep the result of each tree level, until the final sum is calculated. Otherwise, we wouldn't have any values to offset in the final stage of the scan
- with each tree level, buffers of different length are needed, and we do not yet support buffer slicing

apps/typegpu-docs/src/content/examples/algorithms/concurrent-chart/index.ts
aleksanderkatan: I can't right now, something is throwing when I open up the example.
Failed to resolve import "@typegpu/concurrent-scan" from "src/content/examples/algorithms/concurrent-chart/calculator.ts". Does the file exist?

aleksanderkatan: It seems to be quick, why is the 21037 value so skewed?
Not sure if I'm looking at the correct values

image

apps/typegpu-docs/package.json
aleksanderkatan, line 1: We need to commit to one name and stay consistent. Right now, `concurrent scan` and `prefix scan` are still used interchangeably.

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,255 @@
...
+ * @param root - The TypeGPU root/context used to create pipelines, bind groups and buffers.
+ * @param buffer - A storage buffer (`TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag`) containing
+ *                 the input values to reduce.
+ * @param binaryOp - Binary operation defining the associative operation and identity element
+ *                   used for the reduction e.g. {std.add, 0} for sum.
+ * @param timeCallback - Optional callback invoked with a timestamp `TgpuQuerySet` after the
+ *                       GPU dispatch has been submitted; useful for measuring execution time.
+ * @returns A buffer containing the aggregated reduction result (single-element buffer).
+ */
+export function scan(
```
aleksanderkatan: Not sure if I like the name of this, this is not a scan, but rather reduce.
Wiki: In computer science, the prefix sum, cumulative sum, inclusive scan, or simply scan of a sequence of numbers x0, x1, x2, ... is a second sequence of numbers y0, y1, y2, ..., the [sums](https://en.wikipedia.org/wiki/Summation) of [prefixes](https://en.wikipedia.org/wiki/Prefix_(computer_science)) ([running totals](https://en.wikipedia.org/wiki/Running_total)) of the input sequence

this may also imply that prefixScan should not be named like that, since it can be used with functions other than addition

packages/typegpu-concurrent-scan/src/prefixScan.ts
aleksanderkatan, line 1: Right now, these cannot be used without enabling timestamp queries in device, even if no callback is passed.

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,255 @@
...
+ * provided `binaryOp`) over the values in `buffer`. For instance, this can be used to
+ * compute a prefix sum over an array of numbers.
+ *
+ * @param root - The TypeGPU root/context used to create pipelines, bind groups and buffers.
+ * @param buffer - A storage buffer (`TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag`) containing
+ *                 the input values to scan.
+ * @param binaryOp - Binary operation defining the associative operation and identity element
+ *                   used for the scan e.g. addition for prefix-sum.
+ * @param timeCallback - Optional callback invoked with a timestamp `TgpuQuerySet` after the
+ *                       GPU dispatch has been submitted; useful for measuring execution time.
```
aleksanderkatan: I think this could use a jsdocs example, to show that either std functions or tgpu functions can be used, and to show that the last sum is not included.

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,255 @@
...
+ * @returns The `buffer` instance which contains the scanned values.
+ */
+export function prefixScan(
+  root: TgpuRoot,
+  buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
+  binaryOp: BinaryOp,
+  timeCallback?: (timeTgpuQuery: TgpuQuerySet<'timestamp'>) => void,
+): TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag {
```
aleksanderkatan: The current interface does not allow passing in buffer shorthands, do we care? 
```ts
const readonly = root
    .createReadonly(d.arrayOf(d.f32, 8), [10, 11, 12, 13, 14, 15, 16, 17]);

// invalid  
prefixScan(root, readonly, addFn);
```

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
aleksanderkatan, line 1: Do we plan on keeping this example? If so, this means that the timestamp query would stay, right?

packages/typegpu-concurrent-scan/src/prefixScan.ts
```diff
@@ -0,0 +1,255 @@
...
+ *                 the input values to scan.
+ * @param binaryOp - Binary operation defining the associative operation and identity element
+ *                   used for the scan e.g. addition for prefix-sum.
+ * @param timeCallback - Optional callback invoked with a timestamp `TgpuQuerySet` after the
+ *                       GPU dispatch has been submitted; useful for measuring execution time.
+ * @returns The `buffer` instance which contains the scanned values.
+ */
+export function prefixScan(
+  root: TgpuRoot,
+  buffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag,
```
aleksanderkatan: Still not a fan of the buffer being destroyed here. Can we have an optional `outputBuffer` param here, that defaults to the input buffer if not provided?

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
lursz: Yup, we're keeping `concurrent-chart` example.  Timestamp query is essential for it to work, so it's staying as well 👍🏿

packages/typegpu-concurrent-scan/src/prefixScan.ts
lursz: Nice catch, fixed

apps/typegpu-docs/package.json
lursz: In current version naming seems to be accurate 👍🏿

packages/typegpu-concurrent-scan/src/prefixScan.ts
lursz: Name `prefixScan` doesn't imply that it can be used only with addition

packages/typegpu-concurrent-scan/src/prefixScan.ts
lursz: This was planned for another PR

packages/typegpu-concurrent-scan/src/prefixScan.ts
lursz: Solved. The whole api has changed, new one accounts for different output buffer

packages/typegpu-concurrent-scan/src/prefixScan.ts
aleksanderkatan: Will you create a milestone for prefixScan? We should also support operations on props in some way, like adding boid.pos, this one and whatever else you think we could use

packages/typegpu-concurrent-sum/src/prefixScan.ts
aleksanderkatan: This comment is now outdated, iirc we now want to keep the example

apps/typegpu-docs/src/examples/tests/prefix-scan/index.ts
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// running the tests
+
+async function runTests(): Promise<boolean> {
+  let result = true;
+
+  result = await testAdd8() && result;
+  result = await testAdd123() && result;
+  // result = await testMul() && result; // fails, returns 0
```
aleksanderkatan: This test now works
```suggestion
  result = await testMul() && result;
```

apps/typegpu-docs/src/examples/tests/prefix-scan/index.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  result = await testPrefixAdd8() && result;
+  result = await testPrefixAdd123() && result;
+  result = await testPrefixMul() && result;
+  result = await testPrefixStdMax() && result;
+  result = await testPrefixConcat() && result;
+  result = await testPrefixLength1() && result;
+  result = await testPrefixLength65537() && result;
+  result = await testPrefixLength16777217() && result;
+  // result = await testPrefixDoesNotDestroyBuffer() && result; // fails
```
aleksanderkatan: This one still doesn't. I think that we should talk the API through

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+    return false;
+  }
+  // due to how the Blelloch scan works, the first element of the result is the identity element, so we compare with the offset
+  for (let i = 0; i < arr1.length - 1; i++) {
+    if (arr1[i] !== arr2[i + 1]) {
+      console.log(`Mismatch at index ${i}: ${arr1[i]} !== ${arr2[i]}`);
+      return false;
+    }
+  }
+  return true;
```
aleksanderkatan:
```suggestion
  for (let i = 0; i < arr1.length; i++) {
    if (arr1[i] !== arr2[i]) {
      console.log(`Mismatch at index ${i}: ${arr1[i]} !== ${arr2[i]}`);
      return false;
    }
  }
  return true;
```
combined with the following suggestion

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/utils.ts
```diff
@@ -0,0 +1,23 @@
+export function concurrentSumOnJS(arr: number[]) {
+  for (let i = 1; i < arr.length; i++) {
+    arr[i] += arr[i - 1];
+  }
+  return arr;
+}
```
aleksanderkatan:
```suggestion
export function concurrentSumOnJS(arr: number[]) {
  for (let i = 1; i < arr.length; i++) {
    arr[i] += arr[i - 1];
  }
  // In Blelloch scan, the result starts with identity element
  arr.unshift(0);
  arr.pop();
  return arr;
}
```

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
```diff
@@ -0,0 +1,145 @@
...
+
+const lengthMap: Record<
+  string,
+  { jsTime: number; gpuTime: number; gpuShaderTime: number }
+> = {
+  '21037': { jsTime: 0, gpuTime: 0, gpuShaderTime: 0 },
+  '131072': { jsTime: 0, gpuTime: 0, gpuShaderTime: 0 },
+  '1048576': { jsTime: 0, gpuTime: 0, gpuShaderTime: 0 },
+  '4194304': { jsTime: 0, gpuTime: 0, gpuShaderTime: 0 },
+  '8388608': { jsTime: 0, gpuTime: 0, gpuShaderTime: 0 },
```
aleksanderkatan: We now have tests for non-round numbers, so we can choose some more eye-pleasing preset here

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
```diff
@@ -0,0 +1,145 @@
...
+  const keys = Object.keys(lengthMap);
+  for (let i = 0; i < bars.length / 3; i++) {
+    const value = lengthMap[keys[i]];
+    speedupLabels[i].textContent = `${
+      (Number(value.jsTime) / Number(value.gpuShaderTime)).toFixed(1)
+    }x`;
+    xAxisLabels[i].textContent = keys[i];
+
+    // CPU
+    const normalizedJsHeight = maxJsTime > 0 ? value.jsTime / overallMax : 0;
```
aleksanderkatan: `maxJsTime > 0` can this be false? Draw is always after an await.

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
```diff
@@ -0,0 +1,145 @@
...
+    bars[3 * i].style.setProperty(
+      '--bar-height',
+      `${normalizedJsHeight}`,
+    );
+    bars[3 * i].style.setProperty('--highlight-opacity', '1');
+    tooltips[3 * i].textContent = `JS time: ${
+      Number(value.jsTime).toFixed(2)
```
aleksanderkatan: `value.jsTime` already is a number, right? And so are other things that are casted here

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
```diff
@@ -0,0 +1,145 @@
+import tgpu from 'typegpu';
+import { performCalculationsWithTime } from './calculator.ts';
+
+const bars = Array.from(document.querySelectorAll<HTMLDivElement>('.bar'));
```
aleksanderkatan: I think this file could be cleaned up a little, for example this could help

```suggestion
const barElements = document.querySelectorAll<HTMLDivElement>('.bar');
const bars = Array.from({ length: 5 }, (_, i) => ({
  jsBar: barElements[i * 3 + 0],
  gpuTotalBar: barElements[i * 3 + 1],
  gpuBar: barElements[i * 3 + 2],
}));
```

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
```diff
@@ -0,0 +1,145 @@
...
+      Number(value.jsTime).toFixed(2)
+    }ms  |  Array size: ${keys[i]}`;
+
+    // GPU
+    const normalizedGpuHeight = maxGpuTime > 0 ? value.gpuTime / overallMax : 0;
+    bars[3 * i + 1].style.setProperty(
+      '--bar-height',
+      `${normalizedGpuHeight}`,
+    );
+    bars[3 * i + 1].style.setProperty('--highlight-opacity', '1');
```
aleksanderkatan: This doesn't seem to change during the execution, why re-set it in the loop?

Sanity check, we still plan on keeping this example, right?

packages/typegpu-concurrent-scan/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+# @typegpu/concurrent-scan
+
+</div>
+
+A concurrent scan module. For use in WebGPU/TypeGPU apps.\
+Example usage:
+
+```ts
+const calcResult = scan(
```
aleksanderkatan: The following snippet's comment suggests that this one wasn't only the greatest element
```suggestion
const calcResult = prefixScan(
```

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+    // Base case: single workgroup
+    if (numWorkgroups === 1) {
+      const finalSums = this.getScratchBuffer(1);
+      const bg = this.root.createBindGroup(scanLayout, {
+        input: buffer,
+        sums: finalSums,
+      });
+      if (this.#timeCallback) {
+        this.scanPipeline
+          .with(scanLayout, bg)
```
aleksanderkatan: Deprecated api (there are more instances of this)
```suggestion
          .with(bg)
```

packages/typegpu-concurrent-scan/src/compute/applySums.ts
```diff
@@ -0,0 +1,31 @@
+import tgpu from 'typegpu';
+import * as d from 'typegpu/data';
+import {
+  calculateIndex,
+  operatorSlot,
+  uniformAddLayout,
+  workgroupSize,
+} from '../schemas.ts';
+
+export const uniformAdd = tgpu['~unstable'].computeFn({
```
aleksanderkatan: Can we rename the identifiers with `add` to something else? (they apply the op, which may be something other than addition)

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+              ? 0
+              : (undefined as unknown as number),
+            endOfPassWriteIndex: 1,
+          })
+          .dispatchWorkgroups(1);
+      } else {
+        this.scanPipeline
+          .with(scanLayout, bg)
+          .dispatchWorkgroups(1);
+      }
```
aleksanderkatan: This seems like a common pattern, did the pipeline.configure proposal go through? I don't see any api like this right now...
Or maybe we could allow passing in undefined queryset as a noop?

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+): TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag {
+  const binaryOp: BinaryOp = {
+    operation: options.operation,
+    identityElement: options.identityElement,
+  };
+
+  let computer = cache.get(root)?.get(binaryOp);
+  if (!computer) {
+    computer = initCache(root, binaryOp, true);
+  }
```
aleksanderkatan:
```suggestion
  const computer = initCache(root, binaryOp, true);
```
We could also rename `initCache` to maybe `getOrInitialize`

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+ */
+export function scan(
+  root: TgpuRoot,
+  options: {
+    inputBuffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag;
+    outputBuffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag;
+    operation: BinaryOp['operation'];
+    identityElement: BinaryOp['identityElement'];
+  },
+  timeCallback?: (timeTgpuQuery: TgpuQuerySet<'timestamp'>) => void,
```
aleksanderkatan: Why is this not a part of options?

apps/typegpu-docs/src/examples/tests/prefix-scan/index.ts
```diff
@@ -0,0 +1,281 @@
...
+
+  prefixScan(root, {
+    inputBuffer: input,
+    outputBuffer: input,
+    operation: addFn,
+    identityElement: 0,
+  });
+
+  return isArrayEqual(await input.read(), [1, 2, 3, 4, 5, 6, 7, 8]);
+}
```
aleksanderkatan, line 199: This test is faulty
```suggestion
async function testPrefixDoesNotDestroyBuffer(): Promise<boolean> {
  const input = root
    .createBuffer(d.arrayOf(d.f32, 8), [1, 2, 3, 4, 5, 6, 7, 8])
    .$usage('storage');

  const output = root.createBuffer(d.arrayOf(d.f32, 8)).$usage('storage');

  prefixScan(root, {
    inputBuffer: input,
    outputBuffer: output,
    operation: addFn,
    identityElement: 0,
  });

  return isArrayEqual(await input.read(), [1, 2, 3, 4, 5, 6, 7, 8]);
}
```

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+ *     identity: 1,
+ *   },
+ * );
+ * ```
+ */
+export function scan(
+  root: TgpuRoot,
+  options: {
+    inputBuffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag;
+    outputBuffer: TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag;
```
aleksanderkatan: `outputBuffer` does not work in scan (there is always a new one-element buffer returned). I think we can remove it from this signature.

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  if (timeCallback) {
+    (computer as PrefixScanComputer).updateTimeCallback(timeCallback);
+  }
+
+  if (options.inputBuffer !== options.outputBuffer) {
+    options.outputBuffer.copyFrom(options.inputBuffer);
+    return computer.compute(options.outputBuffer);
+  }
+  return computer.compute(options.inputBuffer);
```
aleksanderkatan: Code duplication with `scan` (only difference is `true` vs `false`), this body can be extracted to a helper function.

packages/typegpu-concurrent-scan/src/compute/singleScan.ts
```diff
@@ -0,0 +1,78 @@
...
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+    d.f32(identitySlot.$),
+  ];
```
aleksanderkatan: Why is this array 12 elements ling instead of 8?

packages/typegpu-concurrent-scan/src/compute/scan.ts
aleksanderkatan, line 1: There is a giant piece of duplicated code from `singleScan.ts` here. I see two differences, the existence of `lastIdx` variable, and the existence of a downsweep. If the former is not an issue, then this is an ideal usecase for slots+branch pruning. We can just use `if (performDownsweepSlot.$)`, and create two functions, one `.with(performDownsweepSlot, true)`, one `.with(performDownsweepSlot, false)` and remove like 100 lines of code.

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,363 @@
...
+    }
+    return this.#addPipeline;
+  }
+
+  private getScratchBuffer(
+    size: number,
+  ): TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag {
+    return this.root.createBuffer(d.arrayOf(d.f32, size)).$usage('storage') as
+      & TgpuBuffer<d.WgslArray<d.F32>>
+      & StorageFlag;
```
aleksanderkatan: Is this cast necessary?
```suggestion
    return this.root.createBuffer(d.arrayOf(d.f32, size)).$usage('storage');
```

packages/typegpu-concurrent-scan/src/compute/scan.ts
```diff
@@ -0,0 +1,108 @@
...
+  const partialSums = [
+    d.f32(),
+    identitySlot.$,
+    identitySlot.$,
+    identitySlot.$,
+    identitySlot.$,
+    identitySlot.$,
+    identitySlot.$,
+    identitySlot.$,
+  ];
```
aleksanderkatan: We override all elements anyway
```suggestion
  const partialSums = d.arrayOf(d.f32, 8)();
```

apps/typegpu-docs/src/examples/tests/prefix-scan/index.ts
aleksanderkatan:
```suggestion
  result = await testPrefixDoesNotDestroyBuffer() && result;
```

packages/typegpu-concurrent-scan/README.md
```diff
@@ -0,0 +1,29 @@
+<div align="center">
+
+# @typegpu/concurrent-scan
+
+</div>
+
+A concurrent scan module. For use in WebGPU/TypeGPU apps.\
```
iwoplaza: Why the backslash?
```suggestion
A concurrent scan module. For use in WebGPU/TypeGPU apps.
```

packages/typegpu-concurrent-scan/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "sideEffects": false,
+  "scripts": {
+    "build": "unbuild",
+    "test:types": "pnpm tsc --p ./tsconfig.json --noEmit",
+    "prepublishOnly": "tgpu-dev-cli prepack"
+  },
+  "keywords": [],
+  "license": "MIT",
+  "peerDependencies": {
+    "typegpu": "^0.8.0"
```
iwoplaza:
```suggestion
    "typegpu": "^0.9.0"
```

packages/typegpu-concurrent-scan/package.json
```diff
@@ -0,0 +1,44 @@
+{
+  "name": "@typegpu/concurrent-scan",
+  "type": "module",
+  "version": "0.0.8",
```
iwoplaza:
```suggestion
  "version": "0.9.0",
```

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -775,6 +777,9 @@ export type mBaseForVec<T extends AnyVecInstance> = T extends v2f ? m2x2f
 // #endregion
 
 // #region WGSL Schema Types
+export type TgpuCallable<
+  ImplSchema extends AnyFn = (...args: never[]) => unknown,
+> = TgpuFn<ImplSchema> | DualFn<ImplSchema>;
```
iwoplaza: Is this necessary now that we can use regular JS callbacks?

aleksanderkatan: ~~I wouldn't like shellful users to be excluded~~ nvm, if we just require a function, then TgpuFn satisfies that

packages/typegpu-concurrent-scan/src/compute/singleScan.ts
lursz: It's a leftover, will fix

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
lursz: Suggestions? 🤔

packages/typegpu-concurrent-scan/src/concurrentScan.ts
lursz: Just a design choice, it's a performance measuring feature that is going to be used pretty seldom, so it makes sense to put it separately

packages/typegpu-concurrent-scan/src/concurrentScan.ts
lursz: I'm on the fence with this one, I'd like to make sure people understand it's a cache after all. Will discuss irl

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
aleksanderkatan: Well anything really.  `[1_000, 10_000, 100_000, 1_000_000]` Or `[2**8, 2**12, 2**16, 2**20, 2**24]`

lursz: Yeah okay I'll change that, but I'm pretty sure somewhere above there is a requested change to scramble the numbers, as they were all 2^x

Edit: We need to consult this. Those numbers ([1_000, 10_000, 100_000, 1_000_000]) are waaaaay to low to give GPU a fighting chance. Also we cannot use anything above 2**24, as the buffer gets too large to keep

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
lursz: It's just a guard against division by 0. Theoretically it is possible for `maxJsTime` to be 0 and thus for `overallMax` to be 0. Besides, it's a good practice.

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
lursz: Seems so. Fixed

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/index.ts
lursz: Done

packages/typegpu-concurrent-scan/src/concurrentScan.ts
lursz: How the time flies. Fixed them all.

packages/typegpu-concurrent-scan/src/concurrentScan.ts
lursz: done

packages/typegpu/src/data/wgslTypes.ts
lursz: Yup. Fixed

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,341 @@
...
+  },
+  onlyGreatestElement: boolean,
+  timeCallback?: (timeTgpuQuery: TgpuQuerySet<'timestamp'>) => void,
+): TgpuBuffer<d.WgslArray<d.F32>> & StorageFlag {
+  const binaryOp: BinaryOp = {
+    operation: options.operation,
+    identityElement: options.identityElement,
+  };
+
+  let computer = cache.get(root)?.get(binaryOp);
```
reczkok: I'm pretty sure this guarantees the cache is never hit (we create binaryOp each run so the object identity check won't pass). I would propose a two level map (first level has operation as key and the second uses a string like ``${binaryOp.identityElement}:${onlyGreatestElement}``)

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,341 @@
...
+import { uniformOp } from './compute/applySums.ts';
+import { scanGreatestBlock } from './compute/singleScan.ts';
+
+const cache = new WeakMap<TgpuRoot, WeakMap<BinaryOp, PrefixScanComputer>>();
+
+class PrefixScanComputer {
+  #scanPipeline?: TgpuComputePipeline;
+  #addPipeline?: TgpuComputePipeline;
+  #querySet: TgpuQuerySet<'timestamp'> | null = null;
+  #first = true;
```
reczkok: This is a likely bug that only works because we never cache the computers but the flag is never reset to true

apps/typegpu-docs/src/examples/tests/prefix-scan/index.ts
```diff
@@ -0,0 +1,276 @@
...
+
+// running the tests
+
+async function runTests(): Promise<boolean> {
+  let result = true;
+
+  result = await testAdd8() && result;
+  result = await testAdd123() && result;
+  // result = await testMul() && result; // fails, returns 0
```
aleksanderkatan:
```suggestion
  result = await testMul() && result;
```

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/calculator.ts
aleksanderkatan, line 1: There are no resolution tests for the new examples

packages/typegpu-concurrent-scan/src/compute/applySums.ts
```diff
@@ -0,0 +1,25 @@
...
+    gid: d.builtin.globalInvocationId,
+    wid: d.builtin.workgroupId,
+  },
+})(({ gid, wid }) => {
+  const globalIdx = gid.x;
+  const workgroupId = wid.x;
+  const baseIdx = globalIdx * 8;
+  const sumValue = uniformAddLayout.$.sums[workgroupId];
+
+  for (let i = d.u32(0); i < 8; i++) {
```
aleksanderkatan, line 18: Just so we don't forget.
```suggestion
  // TODO: use `tgpu.unroll(8)`
  for (let i = d.u32(0); i < 8; i++) {
```
I guess this applies to other fixed size fors as well.

packages/typegpu-concurrent-scan/src/compute/singleScan.ts
```diff
@@ -0,0 +1,72 @@
...
+  let lastIdx = d.u32(0);
+  for (let i = d.u32(); i < 8; i++) {
+    if (baseIdx + i < arrayLength) {
+      partialSums[i] = operatorSlot.$(
+        partialSums[i - 1] as number,
```
aleksanderkatan: There is no guard for negative index here

packages/typegpu-concurrent-scan/src/compute/singleScan.ts
aleksanderkatan, line 1: Can we use slots in `scanBlock` to create a version with and without downsweep, to get rid of this file?

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,337 @@
...
+
+const cache = new WeakMap<
+  TgpuRoot,
+  WeakMap<BinaryOp['operation'], PrefixScanComputer>
+>();
+
+class PrefixScanComputer {
+  #scanPipeline?: TgpuComputePipeline;
+  #reducePipeline?: TgpuComputePipeline;
+  #addPipeline?: TgpuComputePipeline;
```
aleksanderkatan: can you rename this variable? Scan supports all types of operators

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,337 @@
...
+}
+
+/**
+ * Create or retrieve a cached `PrefixScanComputer` for the given `root` and `binaryOp`.
+ *
+ * @param root - The TypeGPU root/context to associate with the cached computer.
+ * @param binaryOp - The binary operation used by the computer.
+ * @returns A `PrefixScanComputer` instance associated with the provided `root` and `binaryOp`.
+ */
+export function initCache(
```
aleksanderkatan, line 324: If we export this, can we export `PrefixScanComputer` (or only its type) as well?

packages/typegpu-concurrent-scan/src/compute/singleScan.ts
reczkok: It's actually guarded by the fact this is a `u32` (-1 will be max int so > arrayLength) but I agree it's a little cursed

packages/typegpu-concurrent-scan/src/compute/singleScan.ts
reczkok: see #2182

packages/typegpu-concurrent-scan/src/concurrentScan.ts
reczkok: it's internal so I didn't bother but will do

packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -0,0 +1,337 @@
...
+  let rootCache = cache.get(root);
+  if (!rootCache) {
+    rootCache = new WeakMap();
+    cache.set(root, rootCache);
+  }
+  let computer = rootCache.get(binaryOp.operation);
+  if (!computer) {
+    computer = new PrefixScanComputer(
+      root,
+      binaryOp.operation,
```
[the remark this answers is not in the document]
reczkok: I doubt that any usable prefix scan operation will ever have >1 identity element

PR #1499, opened 2025-07-16
feat: Make array schemas callable

aleksanderkatan:
Changes:
- array schemas can be called with no arguments to create an array filled with default arguments (both JS and WGSL),
- array schemas can be called with another array to create a deep clone on JS side, and to "do nothing" on WGSL side (just like structs, again, we assume the assignment will do the job),
- some minor changes to struct and struct tests.


apps/typegpu-docs/src/content/docs/reference/data-schema-cheatsheet.mdx
```diff
@@ -49,7 +49,7 @@ const c = vec4f(b.xz, a.xx);  // 0, 2, 1, 1
 
 ## Constructor conventions
 
-Scalar, vector, matrix and struct schemas can be called with no arguments to get a default value of the type represented by the schema. Arrays are a WIP.
+Scalar, vector, matrix, struct and fixed-size array schemas can be called with no arguments to get a default value of the type represented by the schema. Arrays are a WIP.
```
reczkok:
```suggestion
Scalar, vector, matrix, struct and fixed-size array schemas can be called with no arguments to get a default value of the type represented by the schema. 
```

iwoplaza: I don't know if runtime-sized arrays will ever be callable. They should only exist to mark a runtime-sized array in a bind group layout, and won't be part of a real schema or tgsl program. I think we can remove the "Arrays are a WIP" bit, and refer to fixed-size array schemas as just array schemas. What do you think?

packages/typegpu/src/data/array.ts
```diff
@@ -33,43 +22,52 @@ export function arrayOf<TElement extends AnyWgslData>(
...
+
+  if (!Number.isInteger(elementCount) || elementCount < 0) {
+    throw new Error(
+      `Cannot create array schema with invalid element count: ${elementCount}.`,
+    );
+  }
+  arraySchema.elementCount = elementCount;
+
+  // @ts-ignore
+  return arraySchema as WgslArray<TElement>;
```
iwoplaza: I would feel more comfortable if we did the following:
```suggestion
  return arraySchema as unknown as WgslArray<TElement>;
```

packages/typegpu/src/data/array.ts
```diff
@@ -33,43 +22,52 @@ export function arrayOf<TElement extends AnyWgslData>(
...
-    idx: number;
-    value: InferPartial<TElement>;
-  }[];
-  declare readonly [$memIdent]: WgslArray<MemIdentity<TElement>>;
-  // ---
+const WgslArrayImpl = {
+  [$internal]: true,
+  type: 'array',
+  elementCount: undefined,
+  elementType: undefined,
```
iwoplaza: I don't think we need these
```suggestion
```

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1189,6 +1189,8 @@ export interface Mat4x4f {
  * the `byteAlignment` requirement of its elementType.
  */
 export interface WgslArray<TElement extends BaseData = BaseData> {
+  <T extends TElement>(elements: Infer<T>[]): Infer<T>[];
```
iwoplaza, line 1192: Do we need the generic param?
```suggestion
  (elements: readonly Infer<TElement>[]): Infer<TElement>[];
```

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1214,7 +1216,7 @@ export interface WgslArray<TElement extends BaseData = BaseData> {
 export interface WgslStruct<
   TProps extends Record<string, BaseData> = Record<string, BaseData>,
 > extends TgpuNamable {
-  (props: Prettify<InferRecord<TProps>>): Prettify<InferRecord<TProps>>;
+  <T extends TProps>(props: Prettify<InferRecord<T>>): Prettify<InferRecord<T>>;
```
iwoplaza: Same here... do the types break somewhere?

packages/typegpu/src/data/wgslTypes.ts
aleksanderkatan: We get a type error without this generic, that I cannot pinpoint:
image
Also, I think the tooltip is nicer:
```ts
const ArraySchema = d.arrayOf(d.u32, 1);
const myArray = ArraySchema([d.u32(10)]);
//              ^?
```
generic: `const ArraySchema: d.WgslArray<d.U32>(elements: number[]) => number[]`
non-generic: `const ArraySchema: d.WgslArray(elements: readonly number[]) => number[]`

packages/typegpu/src/data/wgslTypes.ts
aleksanderkatan: In this case, types are fine, I did it for consistency sake. I can revert this one
Edit: done

packages/typegpu/src/data/wgslTypes.ts
reczkok: I also quite like the tooltip

reczkok (approved):
Amazing🙇

packages/typegpu/src/data/wgslTypes.ts
iwoplaza: Niiice, okay, let's leave it 💜
Could you add an `attest` test for snapshotting this tooltip? I wouldn't want to lose it when refactoring this part of the code.

aleksanderkatan: I tried for a bit but I don't think attest is capable of testing it. The type of the struct/array is `WgslStruct<{prop:Vec2f;}>` and does not involve the call signature. You won't lose this signature on accident, since the intuitive way of refactoring this gives a type error as mentioned before.

PR #1507, opened 2025-07-17
fix(docs): Use the same base URL when importing from esm.sh

iwoplaza:


apps/typegpu-docs/src/pages/benchmark/modules.ts
```diff
@@ -28,19 +44,25 @@ export function importTypeGPUData(
...
+    ]);
   }
 
   if (locator.type === 'pr') {
-    return import(
-      /* @vite-ignore */ `https://esm.sh/pr/typegpu@${locator.commit}/data/?bundle=false`
-    );
+    return Promise.race([
+      import(
+        /* @vite-ignore */ `https://esm.sh/pr/software-mansion/TypeGPU/typegpu@${locator.commit}/data/?bundle=false`
```
iwoplaza, line 62: The actual fix is right here

aleksanderkatan (approved):
🍽️

PR #1511, opened 2025-07-17
feat: API for simulating variable and buffer access on the CPU

iwoplaza:


packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -17,6 +17,7 @@ export interface TgpuConst<TDataType extends AnyWgslData = AnyWgslData>
   readonly $: InferGPU<TDataType>;
 
   readonly [$internal]: {
+    // Makes it differentiable on the type level
     readonly dataType: TDataType;
```
aleksanderkatan: Is this only for type level checks? If so, can we make it into a declare?

packages/typegpu/tests/simulate.test.ts
aleksanderkatan, line 1: Can you add a test for buffers as well?

packages/typegpu/tests/variable.test.ts
```diff
@@ -162,4 +162,51 @@ describe('var', () => {
...
+
+      const result1 = tgpu['~unstable'].simulate(fn);
+      const result2 = tgpu['~unstable'].simulate(fn);
+      const result3 = tgpu['~unstable'].simulate(fn);
+
+      expect(result1.value).toEqual(1);
+      expect(result2.value).toEqual(1);
+      expect(result3.value).toEqual(1);
+    });
+  });
```
aleksanderkatan, line 211: Do we plan on keeping simulation tests inside of simulate.test.ts, near the tested resource or both?

packages/typegpu/src/types.ts
```diff
@@ -107,13 +108,38 @@ export interface ItemStateStack {
...
+export type ExecState = ComptimeState | CodegenState | SimulationState;
+
 /**
  * Passed into each resolvable item. All items in a tree share a resolution ctx,
  * but there can be layers added and removed from the item stack when going down
  * and up the tree.
  */
 export interface ResolutionCtx {
   readonly names: NameRegistry;
+  readonly mode: ExecState | undefined;
```
aleksanderkatan: Can we consider not using `undefined` here and introducing a fourth mode, or is it too much of a hustle with the stack?

packages/typegpu/src/core/simulate/tgpuSimulate.ts
```diff
@@ -0,0 +1,88 @@
...
+  const simStates = Array.from(
+    { length: threads[0] },
+    (_, i) =>
+      Array.from(
+        { length: threads[1] },
+        (_, j) =>
+          Array.from({ length: threads[2] }, (_, k) => {
+            const wi = Math.floor(i / workgroupSize[0]);
+            const wj = Math.floor(j / workgroupSize[0]);
+            const wk = Math.floor(k / workgroupSize[0]);
```
aleksanderkatan:
```suggestion
            const wj = Math.floor(j / workgroupSize[1]);
            const wk = Math.floor(k / workgroupSize[2]);
```
shouldn't it be this?

packages/typegpu/src/core/variable/tgpuVariable.ts
```diff
@@ -106,23 +119,85 @@ class TgpuVarImpl<TScope extends VariableScope, TDataType extends AnyData>
...
   get $(): InferGPU<TDataType> {
-    if (inCodegenMode()) {
+    const mode = getExecMode();
+    const insideTgpuFn = isInsideTgpuFn();
+
+    if (!mode) {
+      throw new IllegalVarAccessError(
+        insideTgpuFn
+          ? `Cannot access ${
+            String(this)
```
aleksanderkatan: This is namable, maybe use `getName` in the error?

packages/typegpu/src/types.ts
```diff
@@ -107,13 +108,38 @@ export interface ItemStateStack {
   defineBlockVariable(id: string, type: AnyWgslData | UnknownData): Snippet;
 }
 
+export type ExecMode = 'comptime' | 'codegen' | 'simulate';
```
aleksanderkatan: Seing this list made me think that it is exhaustive and misinterpret 'comptime' again.
I needed to spend like 15 minutes to convince myself that that:
- 'comptime' is something else than just "not in 'codegen', not in 'simulate'" mode,
- 'comptime' refers to typegpu resolve time and not javascript bundle time.

Could you explain the modes in detail somewhere in this file?

iwoplaza:
> Sanity check, is the idea of simulate to emulate GPU while producing absolutely no side effects?

Essentially yes! The values of updated variables and buffers is available as a return value of `tgpu.simulate()`, as well as whatever is returned from the callback, so it's not totally useless, haha.

> If a zero-filled buffer is referenced, after the simulate will nothing happen to it?
That's correct. In theory (I have to check this), if these buffers are only referenced in the context of a simulation, then no VRAM allocations will happen at all! This is particularly useful for unit-testing.

> Will simulate be able to see an initial value of a buffer? What about buffers that had been filled with data by shaders?

packages/typegpu/tests/variable.test.ts
iwoplaza: Tests in `simulate.test.ts` are meant to test it's API and general behavior, and tests near each resource are supposed to verify their behavior across contexts.

packages/typegpu/src/types.ts
iwoplaza: While trying to explain the difference between the `undefined` or "top-level" execution mode, I also realized that the difference between them is too negligible. I simplified them a lot, so now we have `normal`, `codegen` and `simulate`

iwoplaza: Thanks for pushing me to think more about this!

packages/typegpu/src/core/simulate/tgpuSimulate.ts
```diff
@@ -0,0 +1,88 @@
...
+
+interface SimulationResult<T> {
+  value: T;
+
+  buffers: Map<TgpuBuffer<AnyData>, unknown>;
+  privateVars: Map<TgpuVar<'private', AnyData>, unknown>[][][];
+  workgroupVars: Map<TgpuVar<'workgroup', AnyData>, unknown>[][][];
+}
+
+export function simulate<T>(callback: () => T): SimulationResult<T> {
```
aleksanderkatan, line 39: Maybe a JS docs here?

packages/typegpu/src/types.ts
```diff
@@ -107,13 +108,93 @@ export interface ItemStateStack {
   defineBlockVariable(id: string, type: AnyWgslData | UnknownData): Snippet;
 }
 
+/**
```
aleksanderkatan, line 111: ❤️

packages/typegpu/src/core/buffer/bufferUsage.ts
```diff
@@ -138,18 +141,71 @@ class TgpuFixedBufferImpl<
...
     }
 
-    throw new Error(
-      'Direct access to buffer values is possible only as part of a compute dispatch or draw call. Try .read() or .write() instead',
-    );
+    if (mode.type === 'simulate') {
+      if (!mode.buffers.has(this.buffer)) { // Not initialized yet
+        mode.buffers.set(
+          this.buffer,
+          this.buffer.initial ?? schemaDefaultWrapper(this.buffer.dataType),
```
aleksanderkatan: Initial should be wrapped in the schema clone, right?
Does this work with runtime sized buffers?

iwoplaza: All buffers are runtime-sized 👀. We can write bind groups that are length-agnostic though, is that what you meant?

aleksanderkatan: Nvm, for a second I thought that `(n: number) => d.arrayOf(ComputeVertex, n)` can be a buffer data type instead of a bind group layout entry description

packages/typegpu/src/execMode.ts
```diff
@@ -1,42 +1,79 @@
...
-  const mode = resolutionModeStack.pop();
-  if (expected !== undefined) {
-    invariant(mode === expected, 'Unexpected mode');
+export function getSimulationState(): SimulationState | undefined {
+  const mode = resolutionCtx?.mode;
+  if (mode?.type === 'simulate') {
+    return mode;
   }
+  return undefined;
 }
```
reczkok: Seems unused?

Issue #1336, opened 2025-06-04
impr: Clean up VectorOps


aleksanderkatan:
Many of the entries in the VectorOps look like this:
```typescript
abs: {
  vec2f: unary2f(Math.abs),
  vec2h: unary2h(Math.abs),
  vec2i: unary2i(Math.abs),
  vec2u: unary2u(Math.abs),

  vec3f: unary3f(Math.abs),
  vec3h: unary3h(Math.abs),
  vec3i: unary3i(Math.abs),
  vec3u: unary3u(Math.abs),

  vec4f: unary4f(Math.abs),
  vec4h: unary4h(Math.abs),
  vec4i: unary4i(Math.abs),
  vec4u: unary4u(Math.abs),
} as Record<VecKind, <T extends vBase>(v: T) => T>,
```
I suggest implementing a group of `generate` functions (or one function that also accepts parameter `mode`) that allows simplifying the above to:

```typescript
abs: generateUnaryOps([vec2f, (...), vec4u], Math.abs), // the array can be extracted to a variable since it is reused 
```

Try to:
- make the types work so that `as Record<VecKind, <T extends vBase>(v: T) => T>` is no longer necessary,
- simplify ops that can be simplified (e.g. `clamp`),
- remove unnecessary VectorOps (for example, `exp` and `exp2` can be replaced with a call to pow with the use of `vecTypeToConstructor` exported from `../data/vector.ts`),
- possibly also include in the signature the appropriate return type (`lt`, `gt` etc. take numeric vectors and return boolean vectors, it would be nice to clean up those as well).

aleksanderkatan:
Do we need to pre-create all vector ops? Or could we just compute results of the more generic component wise ones on the fly?

PR #1513, opened 2025-07-18
feat: std mod

lursz:


packages/typegpu/tests/std/numeric/mod.test.ts
```diff
@@ -0,0 +1,324 @@
...
+  });
+
+  it('computes modulo of a scalar and vec4u', () => {
+    expect(VectorOps.modMixed.vec4u(20, vec4u(3, 4, 5, 6))).toEqual(
+      vec4u(2, 0, 0, 2),
+    );
+    expect(VectorOps.modMixed.vec4u(25, vec4u(6, 7, 8, 9))).toEqual(
+      vec4u(1, 4, 1, 7),
+    );
+  });
```
lursz, line 325: No parseResolve test

packages/typegpu/src/std/numeric.ts
```diff
@@ -566,6 +566,49 @@ export const sin = createDualImpl(
...
+      return VectorOps.mod[a.kind](a, b) as T;
+    }
+    throw new Error(
+      `Mod called with invalid arguments. Received types: a=${typeof a}, b=${typeof b}. Expected types: number or vector.`,
+    );
+  },
+  // GPU implementation
+  (a, b) => {
+    const returnType = isSnippetNumeric(a)
+      ? b.dataType
```
lursz: organize this

aleksanderkatan:
Where do we put the line on which operations have an infix variant? Right now it is only `+ - / *`, the `== ! < ...` don't. Should modulo be infixable? (if so, it would be another ticket anyway)

packages/typegpu/src/data/vectorOps.ts
aleksanderkatan, line 1: Could you fix the division op and add a test to make sure it's fixed?

packages/typegpu/src/data/vectorOps.ts
```diff
@@ -162,6 +162,57 @@ const binaryComponentWise4x4f =
...
 
+const createMixedComponentWise2 = <T extends { x: number; y: number }>(
+  vecConstructor: (x: number, y: number) => T,
+) => {
+  return (op: BinaryOp) => (a: T | number, b: T | number) =>
+    vecConstructor(
+      op(typeof a === 'number' ? a : a.x, typeof b === 'number' ? b : b.x),
+      op(typeof a === 'number' ? a : a.y, typeof b === 'number' ? b : b.y),
+    );
+};
```
aleksanderkatan: The way this is written makes it usable for two vector operations as well. We could replace binaryComponentWise with this and simplify the ifs in std operators a little. Probably a job for #1336

packages/typegpu/src/std/numeric.ts
```diff
@@ -566,6 +566,57 @@ export const sin = createDualImpl(
...
+      // vector % scalar
+      return VectorOps.modMixed[a.kind](a, b) as T;
+    }
+
+    if (isVecInstance(a) && isVecInstance(b)) {
+      // vector % vector
+      return VectorOps.mod[a.kind](a, b) as T;
+    }
+    throw new Error(
+      `Mod called with invalid arguments. Received types: a=${typeof a}, b=${typeof b}. Expected types: number or vector.`,
```
aleksanderkatan: I like the idea of a detailed error message, but `typeof a` will just be `object` or `undefined` most of the time. I'd consider using the kind of the object if present, or just simplifying this error to one sentence.

packages/typegpu/src/std/numeric.ts
```diff
@@ -566,6 +566,57 @@ export const sin = createDualImpl(
...
+  (a, b) => {
+    const type = isSnippetNumeric(a)
+      ? b.dataType
+      : isSnippetNumeric(b)
+      ? a.dataType
+      : isVecInstance(a.dataType)
+      ? a.dataType
+      : isVecInstance(b.dataType)
+      ? b.dataType
+      : a.dataType;
```
```suggestion
    const type = isSnippetNumeric(a)
      ? b.dataType
      : a.dataType;
```

packages/typegpu/tests/std/numeric/mod.test.ts
```diff
@@ -0,0 +1,348 @@
...
+  vec2u,
+  vec3f,
+  vec3h,
+  vec3i,
+  vec3u,
+  vec4f,
+  vec4h,
+  vec4i,
+  vec4u,
+} from '../../../src/data/index.ts';
```
aleksanderkatan, line 15: I've seen this in some of our tests, is there a reason why we don't always use a namespace import? (`import * as d`)

packages/typegpu/tests/std/numeric/mod.test.ts
```diff
@@ -0,0 +1,348 @@
...
+});
+
+describe('VectorOps.modMixed vec and scalar', () => {
+  it('computes modulo of vec2f and a scalar', () => {
+    expect(VectorOps.modMixed.vec2f(vec2f(5, 10), 3)).toEqual(vec2f(2, 1));
+    expect(
+      isCloseTo(VectorOps.modMixed.vec2f(vec2f(7.5, 3.5), 2), vec2f(1.5, 1.5)),
+    ).toBe(true);
+  });
+
```
aleksanderkatan: I don't think we ever test VectorOps directly, I think it suffices to test the mod operator

aleksanderkatan (commented):
Here's your sticker

image

packages/typegpu/src/std/numeric.ts
lursz: Nice catch, not an obvious find

packages/typegpu/tests/std/numeric/mod.test.ts
lursz: I think it's just the default way the IDE handles it

packages/typegpu/src/data/vectorOps.ts
lursz: Working on it #1520

aleksanderkatan (approved):
All gucci

reczkok (changes requested):
Jak się człowiek śpieszy to się diabeł cieszy

aleksanderkatan (approved):
🧹🧹🧹 it is so much cleaner now!

PR #1516, opened 2025-07-20
chore: Upgrade unplugin-typegpu dependencies

iwoplaza:


reczkok (approved):
Yes

apps/typegpu-docs/src/content/docs/tooling/unplugin-typegpu.mdx
```diff
@@ -120,6 +120,7 @@ interface Options {
   exclude?: FilterPattern;
   enforce?: 'post' | 'pre' | undefined;
   forceTgpuAlias?: string;
+  autoNamingEnabled?: boolean;
```
aleksanderkatan, line 128: Nice catch. While we're at it, could you include the default options here as well?

packages/unplugin-typegpu/src/index.ts
```diff
@@ -100,9 +100,7 @@ const typegpu: UnpluginInstance<Options, false> = createUnplugin(
             autoNamingEnabled: options.autoNamingEnabled,
           };
 
-          const ast = this.parse(code, {
-            allowReturnOutsideFunction: true,
```
aleksanderkatan, line 104: Why was this option here, is it safe to remove this?

aleksanderkatan (commented):
🟦🟦🟦
Could you explain shortly in which case this change matters?

packages/unplugin-typegpu/src/index.ts
iwoplaza: This option makes sense when parsing CommonJS code. I guess there are cases where we could support it, no harm in keeping it there

iwoplaza:
> 🟦🟦🟦
> Could you explain shortly in which case this change matters?

This change matters when the bundler gives us raw TypeScript code, instead of stripping the types first before handing it to us. Rolldown did this, and this change makes it possible to use unplugin-typegpu with rolldown/tsdown.

reczkok (approved):
Yep

PR #1522, opened 2025-07-21
fix: TypeGPU functions using TGSL dependencies declaration order

aleksanderkatan:
The change allows for a function to reference another function defined later on.
This could potentially lead to people attempting to write recursive functions.

TypeScript screams when either of the following happens (something something _any_):
```ts
const bar = tgpu.fn([], d.f32)(() => foo() + 2);
const foo = tgpu.fn([], d.f32)(() => bar() - 2);
```
```ts
const bar = () => foo() + 2;
const foo = () => bar() - 2;
```

However, the following is fine:
```ts
let bar: TgpuFn; 
let foo: TgpuFn;
bar = tgpu.fn([], d.f32)(() => foo() + 2);
foo = tgpu.fn([], d.f32)(() => bar() - 2);
```
and in this case we get a stack overflow:
```ts
Resolution of the following tree failed: 
- <root>
- fn:bar
- foo: Resolution of the following tree failed: 
- call:foo
- fn:foo
- bar: Resolution of the following tree failed: 
- call:bar
- fn:bar
...
```
Should we try to detect this situation? This would probably require resolve to hold a set of items it is currently trying to resolve, it seems pretty doable. On the other hand, this is already illegal in wgsl, and it is not too difficult to know what is going on when seeing an error like this.

### DIFF
--- a/packages/unplugin-typegpu/test/kernel-directive.test.ts
+++ b/packages/unplugin-typegpu/test/kernel-directive.test.ts
@@ -419,6 +419,21 @@ describe('[ROLLUP] "kernel" directive', () => {
     `);
   });
 
+  it('throws when hoisting was meant to be used', async () => {
+    const code = `\
+      const sum = add(1, 2);
+      function add(a, b) {
+        'kernel';
+        return a + b;
+      };
+    `;
+
+    await rollupTransform(code);
+    expect(consoleWarnSpy).toHaveBeenCalledWith(
+      `File  virtual:code: function "add" might have been referenced before its usage. Function statements are no longer hoisted after being transformed by the plugin.`,
+    );
+  });
+
   it('parses when no typegpu import', async () => {
     const code = `\
       function add(a, b) {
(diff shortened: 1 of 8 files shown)


iwoplaza:
> Should we try to detect this situation? This would probably require resolve to hold a set of items it is currently trying to resolve, it seems pretty doable. On the other hand, this is already illegal in wgsl, and it is not too difficult to know what is going on when seeing an error like this.

People coming from a non-WGSL background, or people that don't yet know recursion is not permitted in WGSL could look at that error and think it's a problem with the library. I think we should throw a descriptive error in this case. I'll make an issue for this.

packages/unplugin-typegpu/test/kernel-directive.test.ts
```diff
@@ -407,21 +419,6 @@ describe('[ROLLUP] "kernel" directive', () => {
...
-        'kernel';
-        return a + b;
-      };
-    `;
-
-    await rollupTransform(code);
-    expect(consoleWarnSpy).toHaveBeenCalledWith(
-      `File  virtual:code: function "add" might have been referenced before its usage. Function statements are no longer hoisted after being transformed by the plugin.`,
-    );
-  });
```
iwoplaza: Why remove this test?

iwoplaza (approved):
Yes please! 💜

packages/unplugin-typegpu/test/kernel-directive.test.ts
reczkok: I think it was just moved

aleksanderkatan: No, sorry, this one and the warn throwing logic were removed as I didn't look closely enough and misunderstood what they did, I'll bring them back

PR #1526, opened 2025-07-22
fix: Fix pointer value assignment type conversion

reczkok:
Previously we always made the type conversion system pick the lhs type regardless of all rules which resulted in trying to create a reference for the rhs instead of dereferencing lhs.
```ts
const increment = tgpu.fn([d.ptrFn(d.f32)])((val) => {
  val += 1;
});
```
Before:
```wgsl
val += &1;
```
After:
```wgsl
*val += 1;
```
(I also removed the redundant whitespace before the newline in the resolution error, it was driving me crazy. My formatter kept stubbornly removing trailing whitespace in inline snapshots.)

### DIFF
--- a/packages/typegpu/tests/tgsl/wgslGenerator.test.ts
+++ b/packages/typegpu/tests/tgsl/wgslGenerator.test.ts
@@ -4,7 +4,7 @@ import { snip } from '../../src/data/dataTypes.ts';
 import * as d from '../../src/data/index.ts';
 import { abstractFloat, abstractInt } from '../../src/data/numeric.ts';
 import { Void } from '../../src/data/wgslTypes.ts';
-import tgpu, { StrictNameRegistry } from '../../src/index.ts';
+import tgpu from '../../src/index.ts';
 import { ResolutionCtxImpl } from '../../src/resolutionCtx.ts';
 import { getMetaData } from '../../src/shared/meta.ts';
 import { $internal } from '../../src/shared/symbols.ts';
@@ -12,6 +12,7 @@ import * as std from '../../src/std/index.ts';
 import * as wgslGenerator from '../../src/tgsl/wgslGenerator.ts';
 import { it } from '../utils/extendedIt.ts';
 import { parse, parseResolved } from '../utils/parseResolved.ts';
+import { StrictNameRegistry } from '../../src/nameRegistry.ts';
 import { CodegenState } from '../../src/types.ts';
 import { provideCtx } from '../../src/execMode.ts';
 
--- a/apps/typegpu-docs/astro.config.mjs
+++ b/apps/typegpu-docs/astro.config.mjs
@@ -235,7 +235,7 @@ export default defineConfig({
               label: 'Naming Convention',
               slug: 'reference/naming-convention',
             },
-            DEV && typeDocSidebarGroup,
+            typeDocSidebarGroup,
           ]),
         },
       ]),
--- a/packages/typegpu/src/index.ts
+++ b/packages/typegpu/src/index.ts
@@ -69,7 +69,6 @@ export {
   NotUniformError,
   ResolutionError,
 } from './errors.ts';
-export { RandomNameRegistry, StrictNameRegistry } from './nameRegistry.ts';
 export { isBuffer, isUsableAsVertex } from './core/buffer/buffer.ts';
 export { isDerived, isSlot } from './core/slot/slotTypes.ts';
 export { isComparisonSampler, isSampler } from './core/sampler/sampler.ts';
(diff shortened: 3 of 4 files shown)


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -168,8 +168,10 @@ export function generateExpression(
     const rhsExpr = generateExpression(ctx, rhs);
 
     const forcedType = expression[0] === NODE.assignmentExpr
-      ? [lhsExpr.dataType as AnyData]
-      : [];
+      ? lhsExpr.dataType.type === 'ptr'
+        ? [lhsExpr.dataType.inner as AnyData]
+        : [lhsExpr.dataType as AnyData]
+      : undefined;
```
reczkok, line 174: This was actually a significant bug that completely disabled the type conversion for logical and binary expressions.
```ts
const forcedType = expression[0] === NODE.assignmentExpr
    ? [lhsExpr.dataType as AnyData]
    : [];
```

reczkok: Passing an empty array to `convertToCommonType` as the `forcedType` parameter says that the converted types have to be one of the ones in the array - since it was empty conversions were disabled.

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -958,11 +958,25 @@ describe('wgslGenerator', () => {
...
 - fn:testFn
-- vec4f: Resolution of the following tree failed: 
+- vec4f: Resolution of the following tree failed:
 - vec4f: Cannot convert argument of type 'array' to 'f32' for function vec4f]
 `);
   });
+
+  it('generates correct code for pointer value assignment', () => {
+    const increment = tgpu.fn([d.ptrFn(d.f32)])((val) => {
+      // biome-ignore  lint/style/noParameterAssign: <go away>
```
iwoplaza: We don't actually need these brackets, they're just in the original message to note that it's a placeholder
```suggestion
      // biome-ignore  lint/style/noParameterAssign: go away
```

reczkok:
Should we add a warning (or maybe throw even)?
```ts
console.warn(
  'convertToCommonType was called with an empty restrictTo array, which prevents any conversions from being matched. If you want to allow all conversions, pass undefined instead. If this is intentional, consider calling the function conditionally, as the result will always be undefined.',
);
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -519,19 +519,26 @@ export function convertToCommonType(
...
   restrictTo?: AnyData[],
+  verbose = true,
 ): Snippet[] | undefined {
   const types = values.map((value) => value.dataType);
 
   if (types.some((type) => type === UnknownData)) {
     return undefined;
   }
 
+  if (verbose && Array.isArray(restrictTo) && restrictTo.length === 0) {
```
iwoplaza: Maybe we should show it only in DEV mode? 👀 
In this way, the warnings code is going to be stripped in production builds
```suggestion
  if (DEV && verbose && Array.isArray(restrictTo) && restrictTo.length === 0) {
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -519,19 +519,26 @@ export function convertToCommonType(
...
+    );
+  }
+
   const conversion = getBestConversion(types as AnyData[], restrictTo);
   if (!conversion) {
     return undefined;
   }
 
-  if (conversion.hasImplicitConversions) {
+  if (verbose && conversion.hasImplicitConversions) {
```
iwoplaza: Maybe same here?
```suggestion
  if (DEV && verbose && conversion.hasImplicitConversions) {
```

apps/typegpu-docs/src/content/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -264,17 +264,15 @@ const vertex = tgpu['~unstable'].vertexFn({
...
-    d.f32(h) /
-    d.f32(std.max(w, h));
+  const gridX = input.idx % w;
+  const gridY = (input.idx - gridX) / w;
+  const maxDim = d.f32(std.max(w, h));
+
+  const x = ((d.f32(gridX) + input.squareData.x) / d.f32(w) - 0.5) * 2 *
+    d.f32(w) / maxDim;
+  const y = ((d.f32(gridY) + input.squareData.y) / d.f32(h) - 0.5) * 2 *
+    d.f32(h) / maxDim;
```
iwoplaza: We could simplify it to this
```suggestion
  const gridX = input.idx % w;
  const gridY = d.u32(input.idx / w);
  const maxDim = std.max(w, h);

  const x = (2 * (d.f32(gridX) + input.squareData.x) - d.f32(w)) / maxDim;
  const y = (2 * (d.f32(gridY) + input.squareData.y) - d.f32(h)) / maxDim;
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -168,13 +168,16 @@ export function generateExpression(
...
 
     const converted = convertToCommonType(
       ctx,
       [lhsExpr, rhsExpr],
       op === '/' ? [f32, f16] : forcedType,
+      op !== '/',
```
iwoplaza: Could we add a comment here saying which parameter it is?
```suggestion
      /* verbose */ op !== '/',
```

packages/typegpu/src/tgsl/generationHelpers.ts
aleksanderkatan: What do you think about introducing our own logger with selectable warn levels etc?

PR #1529, opened 2025-07-23
docs: Refactor the translator and add TGSL mode

reczkok:


apps/typegpu-docs/src/components/translator/lib/useAutoCompile.ts
```diff
@@ -0,0 +1,59 @@
...
+      ) {
+        lastCompiledRef.current = { tgsl: tgslCode, wgsl: wgslCode, format };
+        handleCompile();
+      }
+    }, 1000);
+  }, [canCompile, handleCompile, mode, tgslCode, wgslCode, format]);
+
+  useEffect(() => {
+    const currentCode = mode === TRANSLATOR_MODES.TGSL ? tgslCode : wgslCode;
+    if (currentCode.trim() && canCompile) {
```
iwoplaza: Tiny nitpick, let's make this more explicit
```suggestion
    if (currentCode.trim().length > 0 && canCompile) {
```

apps/typegpu-docs/src/components/translator/lib/useAutoCompile.ts
```diff
@@ -0,0 +1,59 @@
...
+
+      if (
+        canCompile &&
+        (currentCode !== lastCode || format !== lastCompiledRef.current.format)
+      ) {
+        lastCompiledRef.current = { tgsl: tgslCode, wgsl: wgslCode, format };
+        handleCompile();
+      }
+    }, 1000);
+  }, [canCompile, handleCompile, mode, tgslCode, wgslCode, format]);
```
iwoplaza: As far as I can tell, the debounce-timer's lifetime is limited to the useEffect below. This means we can do the debouncing logic there, and change `debouncedCompile` to just `compile` hopefully it will simplify this react instrumentation a bit

reczkok: Maybe I’ll just drop the comparison altogether. It’s not a big deal if we recompile the same code when the user reverts to a previous state within the debounce window, and removing it would simplify the logic quite a bit.

iwoplaza (approved):
Awesome change!

Issue #1508, opened 2025-07-17
impr: Better type inference for arrays


aleksanderkatan:
#1499 introduces callable array schemas. 
We concretisize array type (AbstractInt -> i32) a little too early.
For example, in the following code:
```ts
const ArraySchema = d.arrayOf(d.u32, 1);

const testFn = tgpu.fn([])(() => {
  const myArray = ArraySchema([10]);
});
```
One would expect `myArray` to be of type `array<u32, 1>`, however, the resolution process makes it into `(array<i32, 1>(10))`. We could pass the expected type down in the context so that the following piece of code can make decisions based on that.
```ts
const type = targetType.type === 'abstractFloat'
  ? d.f32
  : targetType.type === 'abstractInt'
  ? d.i32
  : targetType;
```

PR #1532, opened 2025-07-23
feat: Expected type stack in WgslGenerator

### LINKED ISSUE #1510: (RFC) feat: Passing the intended type down the expression chain in WGSL generator
Currently, passing the expected type down the chain of happens for structs via the `callStack` property on the resolution context. A new value is pushed when:

a. A function's body is being resolved, in which case it's return type is added onto the stack
b. A function is being called, in which case it itself is added onto the stack

I think we can rework this system a bit to be a more powerful and reliable tool for expecting expressions to be of a certain type.

## Structs
With the current system, the following is possible:

Image

Creating an object literal that is unrelated to the return type still gets coerced to the return type, becasue it's the value on top of the `callStack`.

> [!NOTE]
> I believe the proper behavior in this case would be to throw a descriptive error, where we cannot infer the type of the struct, and to surround the object literal with a call to that schema

The following breaks entirely

Image

With the error:

Image

## Integer division

Unfortunetelly, compiling WGSL to SPIR-V and back does not get rid of the addtional `f32()` we add to each division:

Image

Let's ponder about the following TGSL
```ts
d.u32(1/2)
```
If during resolution of the expression `1/2` we knew that the intended type was a `u32`, we could omit the `d.f32()` casts entirely, and generate `1u/2u` WGSL code, instead of `u32(f32(1)/f32(2))`

## Possible implementation
Similarly to the `callStack`, we can track what's the intended type for the expression we're currently generating. This solution (maybe `intendedTypeStack`?) is meant to supersede `callStack` in the task of signaling about the intended type, and the task of signaling about the currently generated function's returnType is already being tracked by the ItemStateStack (pushFunctionScope), so we can fully remove `callStack`.

When user-defined functions are called (tgpu.fn-s), the WGSL generator can push each arg type onto the `intendedTypeStack` when the expression corresponding to that argument is being generated, then pop it.
Similarly in the case of casts, they can push and pop the type they expect from the inner resolution, and if the type of the snippet matches, they can skip the casting altogether!

aleksanderkatan:
Closes #1510, #1508.

Changes:
- added `expectedTypeStack` that lets us infer the struct types of JS objects and array types of JS arrays,
- added `generateTypedExpression` function, that pushes to the `expectedTypeStack`, calls `generateExpression`, pops from `expectedTypeStack` and tries to coerce the result to the correct type,
- removed `callStack`,
- [!] removed the feature that wrapped JS objects in entry functions into `Out` struct by default, 
- [!] removed some segments of WGSL generator that seemed to me like they did nothing (marked with github comments),
- renamed the internal prop of functions that held type conversion hints from `argTypes` to `argConversionHint`, renamed `'coerce'` to `'convert-arguments-to-common-type'`,
- rewritten or removed some pre-existing tests.

Expressions that use `generateTypedExpression`:
- returns,
- if/while/for conditions,
- array constructors,
- struct constructors,
- TGSL function calls.

Expressions that do not:
- vectors and matrices (there are multiple possible expected types),
- infix operators (-||-),
- functions that do not provide a conversion hint (like std functions, -||-),
- other things that I did not think about.

I did not change how JS variables defined outside of TGSL are parsed, I think it can be separated to another ticket. Same goes for optimizing number casting.

I am not convinced about the current implementation of the `expectedTypeStack`:
- Firstly, (at least at this moment) it theoretically suffices to just store the top of the stack in the context and let `generateTypedExpression` swap it, since the stack is exclusively modified in this function.
- Secondly, in its current form it is not applicable in many cases, especially when the type of the current expression is unknown, or when the expression could be of multiple types.
- Finally, it is not a reliable way of knowing what should the type of the current expression be:
```ts
let a = d.u32();
a = (1+2);
```
when generating the code for `(1+2)`, the stack is empty, and the existence of the stack might mislead developers to think that it would hold `u32` in this case. If we allowed unknown type and multiple types to be on the stack, we would be able to update this stack at all times. It would require many more changes, especially to std functions and vec/mat constructors that currently have no `argTypes` (as most of them cannot be given a single signature, they have overloads).
Let me know what direction would you like this taken to.


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -662,32 +632,6 @@ ${alternate}`;
...
-        entries[key] = generateExpression(ctx, value);
-      }
-
-      const convertedValues = convertStructValues(ctx, structType, entries);
-      const resolvedStruct = ctx.resolve(structType);
-      return `${ctx.pre}var ${id} = ${resolvedStruct}(${
-        convertedValues.map((sn) => ctx.resolve(sn.value)).join(', ')
-      });`;
-    }
-
```
aleksanderkatan, line 689: One of the deleted segments of logic.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -327,69 +343,42 @@ export function generateExpression(
...
-              : getName(id.value) ?? '<unnamed>'
-          }()'`,
+          'Type error: Array and struct schemas should always be called with at most 1 argument.',
         );
       }
-    });
 
-    if (typeof id.value === 'string') {
-      return snip(`${id.value}(${argValues.join(', ')})`, id.dataType);
-    }
```
aleksanderkatan, line 362: One of the deleted segments of logic.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -457,92 +443,86 @@ export function generateExpression(
...
+          );
         }
+        const result = generateTypedExpression(ctx, val, value as AnyData);
+        return [key, result];
+      }),
+    );
 
-        return snip(snippets, UnknownData);
-      }
-    }
```
aleksanderkatan: One of the deleted segments of logic.
It didn't seem to do anything, so I removed it.
I would be a terrible surgeon.

aleksanderkatan (commented):
image
🙄

packages/typegpu/src/std/numeric.ts
```diff
@@ -95,7 +95,7 @@ export const add = createDualImpl(
       `(${lhs.value} + ${rhs.value})`,
       isSnippetNumeric(lhs) ? rhs.dataType : lhs.dataType,
     ),
-  'coerce',
+  'convert-arguments-to-common-type',
```
reczkok: I think that's a little too verbose for me. Maybe `unify`? I guess it does not tell you explicitly that it will be unifying types but if we want to solve that issue i would move from positional arguments to object argument with descriptive field names.
```suggestion
  'unify',
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -327,69 +343,42 @@ export function generateExpression(
   }
 
   if (expression[0] === NODE.call) {
-    // Function Call
+    // Function Call.
```
reczkok: I don't think we need punctuation outside of jsdoc (applies to all short comments)
```suggestion 
    // Function Call
```

packages/typegpu/src/tgsl/wgslGenerator.ts
reczkok: It was the only way to construct the struct of an entry-point. I would argue it cannot be deleted until we provide another way.

packages/typegpu/src/types.ts
```diff
@@ -278,8 +279,17 @@ export function isWgsl(value: unknown): value is Wgsl {
...
+export type DefaultConversionStrategy =
+  | 'keep'
+  | 'convert-arguments-to-common-type';
 
+/**
+ * Optional hints for converting function argument types during resolution.
+ * In case of tgpu functions, this is just the array of argument schemas.
+ * In case of raw dualImpls (e.g. in std), this is either a function that converts the snippets appropriately,
+ * or a string defining a conversion strategy.
+ * The strategy 'keep' is equivalent to undefined strategy.
```
reczkok: Now that I read it we should have the argConversionHint not be optional but default to `keep`.

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: Yup, seems like this would handle functions that don't come from externals, which is impossible in WGSL

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: Not anymore 🫡

iwoplaza (approved):
This is spectacular, makes things so much more robust now 👏👏👏

packages/typegpu/tests/tgsl/typeInference.test.ts
```diff
@@ -0,0 +1,299 @@
...
+        return Boid(vec2f(1, 1), vec2f());
+      }
+    `));
+  });
+
+  it('coerces return value to a nested structs', () => {
+    const Inner = d.struct({ prop: d.vec2f });
+    const Outer = d.struct({ inner: Inner });
+
+    const myFn = tgpu.fn([], Outer)(() => {
```
reczkok, line 64: not used
```suggestion
    const myFn = tgpu.fn([])(() => {
```

packages/typegpu/tests/tgsl/typeInference.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    expect(parseResolved({ myFn })).toBe(parse(`
+      struct Struct {
+        prop: vec2f,
+      }
+      
+      fn myFn() {
+        var myStructArray = array<Struct, 1>(Struct(vec2f(1, 2)));
+      }
+    `));
+  });
```
reczkok: Is this intended? The length of our `StructArray` schema gets ignored.

packages/typegpu/tests/tgsl/typeInference.test.ts
```diff
@@ -0,0 +1,299 @@
...
+    const myFn = tgpu.fn([], d.u32)(() => {
+      return 1.1;
+    });
+
+    expect(() => parseResolved({ myFn })).toThrowErrorMatchingInlineSnapshot(`
+      [Error: Resolution of the following tree failed:
+      - <root>
+      - fn:myFn: Cannot convert value of type 'abstractFloat' to type 'u32']
+    `);
+  });
```
reczkok, line 241: Can't it? This should work atm I would think 🤔

iwoplaza: Nope! Invalid in WGSL as well

packages/typegpu/tests/tgsl/typeInference.test.ts
iwoplaza: Good catch, fixed!

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -332,225 +353,195 @@ export function generateExpression(
...
-              }`,
-              [{
-                function: id.value,
-                callStack: ctx.callStack,
-                error:
-                  `Cannot convert argument of type '${sn.dataType.type}' to '${type.type}'`,
-                toString: () => getName(id.value),
-              }],
+    const argConversionHint = callee.value[$internal]
+      ?.argConversionHint as FnArgsConversionHint ?? 'keep';
```
reczkok, line 406: Maybe we create something like:
```ts
export function isDualImpl<T extends (...args: unknown[]) => unknown>(
  value: unknown | TgpuDualFn<T>,
): value is TgpuDualFn<T> {
  const internal = (value as TgpuDualFn<T>)?.[$internal];
  return !!(value as TgpuDualFn<T>)?.[$internal] &&
    'jsImpl' in internal &&
    'gpuImpl' in internal &&
    'argConversionHint' in internal;
}
```
This sound useful and would clean up our types here (and potentially in other places - I didn't check)

```diff lang=ts
    if (!isDualImpl(callee.value)) {
      throw new Error(
        `Function ${String(callee.value)} ${
          getName(callee.value)
        } has not been created using TypeGPU APIs. Did you mean to wrap the function with tgpu.fn(args, return)(...) ?`,
      );
    }

    // Other, including tgsl functions, std and vector/matrix schema calls.
-    const argConversionHint = callee.value[$internal]
-      ?.argConversionHint as FnArgsConversionHint ?? 'keep';
+    const argConversionHint = callee.value[$internal].argConversionHint;
...
-      const fnRes =
-      (callee.value as unknown as (...args: unknown[]) => unknown)(
-        ...convertedArguments,
-      ) as Snippet;
+      const fnRes = (callee.value)(
+        ...convertedArguments,
+      ) as Snippet;
```

iwoplaza: Dual impls are not the only callables (e.g., tgpu.fn), so this was breaking a lot of tests. I guess we'd have to check for all callables, but we can do that in another pull request.

PR #1533, opened 2025-07-24
chore(docs): Update dependencies

iwoplaza:


aleksanderkatan (approved):
Nice! Do we have any tool that helps us find deprecated api calls?

iwoplaza:
> Nice! Do we have any tool that helps us find deprecated api calls?
Good question, I haven't seen one, but there has to be a tool like that

PR #1537, opened 2025-07-25
docs: Complete "Functions" in roadmap

iwoplaza:


aleksanderkatan (approved):
Nice! Do we plan on adding another point for entry point functions?

iwoplaza:
> Nice! Do we plan on adding another point for entry point functions?

If we choose to publish those separately from pipelines, then we will!

PR #1539, opened 2025-07-25
docs: Better loading state for examples page

iwoplaza:
Before:

https://github.com/user-attachments/assets/28a047c6-ebed-4a14-878f-2c1889a460ed

After:

https://github.com/user-attachments/assets/816634f4-afc2-44e5-a875-3b5600263b8c


apps/typegpu-docs/src/components/ExampleLayout.tsx
```diff
@@ -10,16 +11,16 @@ import {
...
   menuShownMobileAtom,
 } from '../utils/examples/menuShownAtom.ts';
-import ExamplePage from './ExamplePage.tsx';
 import { SearchableExampleList } from './SearchableExampleList.tsx';
 import { Button } from './design/Button.tsx';
 import { Toggle } from './design/Toggle.tsx';
 import { experimentalExamplesShownAtom } from '../utils/examples/showExperimentalExamplesAtom.ts';
 
-// biome-ignore lint/suspicious/noExplicitAny: it exists, I swear
-(globalThis as any).__TYPEGPU_MEASURE_PERF__ = true;
```
iwoplaza, line 20: We can always turn this back on if we want to benchmark the examples outside of our browser tests

apps/typegpu-docs/src/components/ExampleLayout.tsx
```diff
@@ -1,5 +1,6 @@
 import cs from 'classnames';
 import { useAtom, useAtomValue } from 'jotai';
+import type React from 'react';
```
reczkok: We could import just `ReactNode`

reczkok (approved):
Looks much better 🛠️

PR #1547, opened 2025-07-28
fix: Rework how schemas propagate usage validity

iwoplaza:
Changes: 
- Propagating validity the same as other properties, rather than checking recursively
- Buffer shorthands are validated the same way that buffers are
- Better intellisense for .$usage
- The reason as to why the schema is invalid is propagated, and information about where the error originates is also propagated on the type level (in the array element, in struct prop 'foo', etc.)
- `WgslStruct` without type parameters is now the same type as `AnyWgslStruct`, rendering it redundant (same for `Unstruct` and `AnyUnstruct`)

### Better intellisense for .$usage

Before:

https://github.com/user-attachments/assets/9a62fced-21b0-4dcc-99f8-3dbea94d48a3

After:

https://github.com/user-attachments/assets/4d4cfba3-1a47-4746-8910-0f10f81149e7


packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1345,6 +1428,12 @@ export interface Decorated<
   readonly [$memIdent]: TAttribs extends Location[]
     ? MemIdentity<TInner> | Decorated<MemIdentity<TInner>, TAttribs>
     : Decorated<MemIdentity<TInner>, TAttribs>;
+  readonly [$validStorageSchema]: IsValidStorageSchema<TInner>;
+  readonly [$validUniformSchema]: IsValidUniformSchema<TInner>;
+  readonly [$validVertexSchema]: IsValidVertexSchema<TInner>;
+  readonly [$invalidSchemaReason]: `decorated — ${ExtractInvalidSchemaError<
+    TInner
+  >}`;
```
reczkok: Image

```suggestion
  readonly [$invalidSchemaReason]: `in decorated inner type — ${ExtractInvalidSchemaError<
    TInner
  >}`;
```

packages/typegpu/src/shared/repr.ts
```diff
@@ -52,13 +64,58 @@ export type InferGPURecord<
...
   [Key in keyof T]: MemIdentity<T[Key]>;
 };
+
+export type IsValidStorageSchema<T> =
+  (T extends { readonly [$validStorageSchema]: true } ? true : false) extends //
+  // The check above can eval to `boolean` if parts of the union
+  // are valid, but some aren't. We only treat schemas invalid if all
+  // union elements are invalid.
+  false ? false
+    : true;
```
reczkok, line 83: Is this supposed to work? 🤔
Image

packages/typegpu/src/data/wgslTypes.ts
iwoplaza: What if we just skip prefixing at all in case of decorated? 👀

packages/typegpu/src/shared/repr.ts
iwoplaza: Generics work different unfortunately 😞. You have to use ValidateBufferSchema in the parameter

packages/typegpu/tests/buffer.test.ts
```diff
@@ -531,7 +540,100 @@ describe('TgpuBuffer', () => {
...
+
+  it('it treats union schemas as valid (even if they contain booleans)', () => {
+    expectTypeOf<IsValidUniformSchema<d.U32 | d.Bool>>()
+      .toEqualTypeOf<true>();
+    expectTypeOf<IsValidUniformSchema<d.U32 | d.WgslArray<d.Bool>>>()
+      .toEqualTypeOf<true>();
+    expectTypeOf<IsValidUniformSchema<d.WgslArray<d.Bool | d.U32>>>()
+      .toEqualTypeOf<true>();
+  });
+});
```
aleksanderkatan, line 568: Why do we consider this valid?

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1126,12 +1178,14 @@ export interface Vec4b {
...
-export interface Mat2x2f {
-  readonly [$internal]: true;
+export interface Mat2x2f extends BaseData {
   readonly type: 'mat2x2f';
 
   // Type-tokens, not available at runtime
   readonly [$repr]: m2x2f;
+  readonly [$validStorageSchema]: true;
+  readonly [$validUniformSchema]: true;
+  readonly [$validVertexSchema]: true;
```
reczkok: I don't think any of the matrix types will work atm. We could theoretically handle them but we currently don't.

reczkok (approved):
Amazing change! The DX is so much better 🙇

packages/typegpu/src/shared/symbols.ts
```diff
@@ -40,3 +40,28 @@ export const $reprPartial = Symbol(`typegpu:${version}:$reprPartial`);
...
+ */
+export const $validVertexSchema = Symbol(
+  `typegpu:${version}:$validVertexSchema`,
+);
+/**
+ * Type token, containing a reason for why the schema is invalid (if it is).
+ */
+export const $invalidSchemaReason = Symbol(
+  `typegpu:${version}:$invalidSchemaReason`,
+);
```
aleksanderkatan, line 67: So, to my understanding, this is a symbol for all kinds of errors, right? If a schema is an invalid uniform schema, invalid storage schema or invalid vertex schema, the same error will be used? Can we potentially want more flexibility with this error in the future?

aleksanderkatan (approved):
Nice! 🚌

packages/typegpu/src/shared/symbols.ts
iwoplaza: Yes, the error is for all invalid schema-related errors. We could split this in the future, if we need to.

packages/typegpu/tests/buffer.test.ts
iwoplaza: If we're not dealing with concrete types, it means the developer is trying to cover a more general case. We just want to be less annoying in those cases.

packages/typegpu/src/data/wgslTypes.ts
iwoplaza: Wdym? 👀

iwoplaza: Aaah, for vertex schemas? Right

iwoplaza: Fixed it, good catch!

PR #1553, opened 2025-07-31
feat: typescript support for Babel.transform() in translateTGSL()

GCyganek:
Adding TypeScript support for `Babel.transform()` function in  `translateTGSL()`.

This way Babel won't break if there are any types in TGSL:

https://github.com/user-attachments/assets/6417f7a2-cc61-45cc-9e59-05f2090884b2

With TypeScript support:

https://github.com/user-attachments/assets/01e97a18-ad46-4335-8e42-1740a014c655


iwoplaza (approved):
Awesome, thanks! 🚀

PR #1555, opened 2025-08-01
impr: Rework examples that use video input and publish them again on iOS safari

reczkok:
This PR:
- Adds a new script `dev:host` that exposes the dev server to the local network over https (which is necessary to use WebGPU)
- Moves examples that use cameras to `requestVideoFrameCallback` API instead of `requestAnimationFrame` which ensures our shaders don't run faster than the camera can provide frames and streamlines video dimension logic (the callback provides metadata about the frame size so we don't need to observe the HTMLVideoElement).
- Unified the used web API (moved all examples to HTMLVideoElement + external texture) for stability and easier maintenance 
- Removed the non-next Chroma Keying example since it was outdated and removed `next` from Chroma Keying (next) since no experimental features are used (that weren't in the original)
- Added and refined some std fn's to be able to use external textures in TGSL


apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -57,7 +57,7 @@ const fullScreenTriangle = tgpu['~unstable'].vertexFn({
   out: { pos: d.builtin.position, uv: d.vec2f },
 })((input) => {
   const pos = [d.vec2f(-1, -1), d.vec2f(3, -1), d.vec2f(-1, 3)];
-  const uv = [d.vec2f(0, 0), d.vec2f(2, 0), d.vec2f(0, 2)];
+  const uv = [d.vec2f(0, 1), d.vec2f(2, 1), d.vec2f(0, -1)];
```
reczkok, line 53: I think this (blue on the image) is the correct uv order to not `flip` our output

Image

apps/typegpu-docs/src/content/examples/image-processing/chroma-keying-next/index.ts
```diff
@@ -122,10 +122,20 @@ if (!samplingContext) {
...
+let reader: ReadableStreamDefaultReader<VideoFrame> | undefined;
+let worker: Worker | undefined;
+
+if (typeof MediaStreamTrackProcessor !== 'undefined') {
+  const mediaProcessor = new MediaStreamTrackProcessor({
+    track: stream.getVideoTracks()[0],
+  });
+  reader = mediaProcessor.readable.getReader();
+} else {
+  worker = new Worker(new URL('./processor.worker.ts', import.meta.url), {
```
reczkok: I have my doubts that this will work in prod

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -1,23 +1,19 @@
...
-  inputTexture: { texture: 'float' },
+  externalTexture: { externalTexture: {} },
 });
 
 const charsetExtended = root.createUniform(d.u32);
 const displayMode = root.createUniform(d.u32);
 const gammaCorrection = root.createUniform(d.f32);
 const glyphSize = root.createUniform(d.u32, 8);
+const uvTransformBuffer = root
+  .createUniform(d.mat2x2f, d.mat2x2f(1, 0, 0, 1));
```
aleksanderkatan: nit 
```suggestion
  .createUniform(d.mat2x2f, d.mat2x2f.identity());
```

packages/typegpu/src/std/texture.ts
```diff
@@ -395,3 +408,33 @@ export const textureDimensions: TextureDimensionsOverload = createDualImpl(
...
 );
+
+type TextureSampleBaseClampToEdge = {
+  <T extends TgpuSampledTexture<'2d', F32>>(
+    texture: T,
+    sampler: TgpuSampler,
+    coords: v2f,
+  ): v4f;
+  (texture: TgpuExternalTexture, sampler: TgpuSampler, coords: v2f): v4f;
+};
```
aleksanderkatan: Do we have a convention for texture types? I think this one could be simplified
```suggestion
type TextureSampleBaseClampToEdge = (
  texture: TgpuSampledTexture<'2d', F32> | TgpuExternalTexture,
  sampler: TgpuSampler,
  coords: v2f,
) => v4f;
```

apps/typegpu-docs/src/content/examples/image-processing/chroma-keying/index.ts
aleksanderkatan, line 1: After opening this example and then switching to another, the camera does not stop being actively used by the tab, there is probably some cleanup missing.

packages/typegpu/src/std/texture.ts
reczkok: It sure could. Good catch

apps/typegpu-docs/src/content/examples/image-processing/chroma-keying/index.ts
reczkok: Yes, I need to add back `track.stop()` to all the examples, thanks

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
```diff
@@ -192,99 +183,101 @@ const pipeline = root['~unstable']
...
-    size: [video.videoWidth, video.videoHeight],
-    format: presentationFormat as 'rgba8unorm' | 'bgra8unorm',
-  }).$usage('render', 'sampled');
+  const frameWidth = metadata.width;
+  const frameHeight = metadata.height;
+
+  if (
+    !lastFrameSize ||
+    lastFrameSize.width !== frameWidth ||
+    lastFrameSize.height !== frameHeight
```
lursz, line 219: nit
```suggestion
  lastFrameSize?.width !== frameWidth ||
  lastFrameSize?.height !== frameHeight
```

apps/typegpu-docs/src/content/examples/image-processing/camera-thresholding/index.ts
```diff
@@ -143,17 +169,30 @@ const renderPassDescriptor: GPURenderPassDescriptor = {
...
   }
 
-  // Updating the target render texture
+  const frameWidth = metadata.width;
+  const frameHeight = metadata.height;
+
+  if (
+    !lastFrameSize ||
+    lastFrameSize.width !== frameWidth ||
+    lastFrameSize.height !== frameHeight
```
lursz, line 190:
```suggestion
    lastFrameSize?.width !== frameWidth ||
    lastFrameSize?.height !== frameHeight
```

aleksanderkatan (approved):
LGTM! The screen rotations work fine on Android

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
reczkok: It's not the same but works here (I think always? Unless metadata fields can be undefined which I don't think is possible) so I will use it

lursz (approved):

apps/typegpu-docs/src/content/examples/image-processing/ascii-filter/index.ts
cieplypolar: Nice observation, I didn't know the order of the vertices matters.

cieplypolar:
Great work!!! Everything works, even on my old phone with Android 12.

PR #1557, opened 2025-08-01
fix: concretize abstract types when performing division

### LINKED ISSUE #1551: impr: Add concretisation to `convertToCommonType`
Case:
```ts
const buf = root.createBuffer(d.arrayOf(d.f32, 1)).$usage('storage');
const val = buf.as('mutable');

const f = tgpu['~unstable'].computeFn({
  in: { gid: d.builtin.globalInvocationId },
  workgroupSize: [1],
})((input) => {
  val.$[0] = 1 / 5;
});

const p = root['~unstable'].withCompute(f).createPipeline();

p.dispatchWorkgroups(1);

const x = await buf.read();
console.log('Div on gpu: ', x);
console.log('Div in js: ', 1 / 5);
```
We pass `1 / 5` to `convertToCommonType` with forced type of `f32 | f16` but since both `1` and `5` are abstract we decide to do nothing. This breaks because WGSL treats `1` and `5` as abstract integers and reduces the division to `0`. We want to add a flag to `convertToCommonType` that forces concretisation of abstract types before the proper conversion.
After we call `concretise` on `1` nad `5` our conversion system will treat them as `i32` and cast to `f32` properly.

cieplypolar:
Changes:
- `convertToCommonType` now has an additional parameter, `concretizeTypes`, which makes the function concretize the types before looking for the best conversion
- `generateExpression` now concretizes types when it encounters division
- tests have been updated accordingly

Closes #1551.

### DIFF
--- a/packages/typegpu/src/tgsl/generationHelpers.ts
+++ b/packages/typegpu/src/tgsl/generationHelpers.ts
@@ -46,6 +46,7 @@ import {
   isVec,
   isVecInstance,
   isWgslArray,
+  isWgslData,
   isWgslStruct,
   type U32,
 } from '../data/wgslTypes.ts';
@@ -539,7 +540,12 @@ export function convertToCommonType({
   }
 
   if (concretizeTypes) {
-    types = (types as AnyWgslData[]).map(concretize);
+    types = types.map((type) => {
+      if (isWgslData(type)) {
+        return concretize(type);
+      }
+      return type;
+    });
   }
 
   if (DEV && verbose && Array.isArray(restrictTo) && restrictTo.length === 0) {
--- a/packages/typegpu/src/tgsl/generationHelpers.ts
+++ b/packages/typegpu/src/tgsl/generationHelpers.ts
@@ -46,7 +46,6 @@ import {
   isVec,
   isVecInstance,
   isWgslArray,
-  isWgslData,
   isWgslStruct,
   type U32,
 } from '../data/wgslTypes.ts';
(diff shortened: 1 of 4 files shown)


packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -517,18 +518,36 @@ function applyActionToSnippet(
   }
 }
 
-export function convertToCommonType(
-  ctx: GenerationCtx,
-  values: Snippet[],
-  restrictTo?: AnyData[],
+export type ConvertToCommonTypeInfo = {
```
iwoplaza: Tiny nitpick, we commonly use the `Options` suffix for the named function parameters type
```suggestion
export type ConvertToCommonTypeOptions = {
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -517,18 +518,36 @@ function applyActionToSnippet(
...
 
+  if (concretizeTypes) {
+    types = types.map((type) => {
+      if (isWgslData(type)) {
+        return concretize(type);
+      }
+      return type;
+    });
+  }
+
```
reczkok: Since we know concretize is an identity function for non-abstract types (the only prerequisite is that the input has a `type` field) we can simplify this a little.
```suggestion
  const types = values.map((value) =>
    concretizeTypes ? concretize(value.dataType as AnyData) : value.dataType
  );

  if (types.some((type) => type === UnknownData)) {
    return undefined;
  }

```

cieplypolar: I have to cast `value.dataType` to `AnyWgslData`, but we are fine, since `AnyData` will fall to identity.

PR #1560, opened 2025-08-01
fix: Add support for packed vertex formats in compiled writer and improve the writer selection logic

reczkok:
Up until now loose data was not supported in compiled writers (despite vaguely looking like it was). This PR adds support for them as well as introducing a fallback in case we fail to compile the writer.

### DIFF
--- a/packages/typegpu/src/data/compiledIO.ts
+++ b/packages/typegpu/src/data/compiledIO.ts
@@ -4,6 +4,7 @@ import { alignmentOf } from './alignmentOf.ts';
 import { isDisarray, isUnstruct } from './dataTypes.ts';
 import { offsetsForProps } from './offsets.ts';
 import { sizeOf } from './sizeOf.ts';
+import { formatToWGSLType, isPackedData } from './vertexFormatData.ts';
 import * as wgsl from './wgslTypes.ts';
 
 export const EVAL_ALLOWED_IN_ENV: boolean = (() => {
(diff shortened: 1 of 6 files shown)


packages/typegpu/src/data/compiledIO.ts
```diff
@@ -69,11 +132,8 @@ export function buildWriter(
 
   if (wgsl.isWgslStruct(node) || isUnstruct(node)) {
     const propOffsets = offsetsForProps(node);
-    const sortedProps = Object.entries(propOffsets).sort(
-      (a, b) => a[1].offset - b[1].offset,
-    );
```
reczkok, line 74: No need to sort since the offsets don’t depend on it. I would still test whether writing the bytes in order gives any performance benefits.

packages/typegpu/src/data/vertexFormatData.ts
```diff
@@ -250,3 +250,12 @@ export type PackedData =
   | sint32x4
   | unorm10_10_10_2
   | unorm8x4_bgra;
+
+export function isPackedData(
+  value: unknown,
+): value is PackedData {
+  return (value as PackedData)?.[$internal] &&
+    Object.keys(formatToWGSLType).includes(
```
iwoplaza: Let's create a set out of `Object.keys(formatToWGSLType)` in the global scope, and use that instead

packages/typegpu/src/data/compiledIO.ts
```diff
@@ -142,6 +227,57 @@ export function buildWriter(
...
+        i * componentSize
+      }), ${value}, littleEndian);\n`;
+    }
+
+    return code;
+  }
+
+  if (!Object.hasOwn(typeToPrimitive, node.type)) {
+    throw new Error(
+      `Unsupported primitive encounter when compiling writer: ${node.type}`,
```
iwoplaza: Imo too verbose, maybe something like this?
```suggestion
      `Primitive ${node.type} is unsupported by compiled writer`,
```

iwoplaza (approved):
Awesome, this is a big fix! 🚀

packages/typegpu/src/data/vertexFormatData.ts
reczkok: Turns out we have the set at home

PR #1561, opened 2025-08-03
feat: Bun plugin

iwoplaza:


reczkok (approved):
Oh, the implementation is much simpler than I anticipated. Great work! 🙌

PR #1567, opened 2025-08-05
feat: Add all numeric and derivative std (some with no cpu implementation)

reczkok:


packages/typegpu/src/std/numeric.ts
```diff
@@ -444,6 +606,80 @@ export const fract = createDualImpl(
...
+    );
+  },
+  // GPU implementation
+  (value) =>
+    snip(
+      `frexp(${value.value})`,
+      builtinStruct({
+        fract: f32,
+        exp: i32,
+      }, '__frexp_result_f32'),
```
iwoplaza: Could we define this globally, and reuse the struct on each call?

packages/typegpu/src/std/numeric.ts
```diff
@@ -783,3 +1134,115 @@ export const smoothstep = createDualImpl(
...
+    return snip(`transpose(${e.value})`, e.dataType);
+  },
+  'transpose',
+);
+
+export const trunc = createDualImpl(
+  // CPU implementation
+  <T extends AnyFloatVecInstance | number>(value: T): T => {
+    throw new Error(
+      'CPU implementation for transpose not implemented yet. Please submit an issue at https://github.com/software-mansion/TypeGPU/issues',
```
iwoplaza:
```suggestion
      'CPU implementation for trunc not implemented yet. Please submit an issue at https://github.com/software-mansion/TypeGPU/issues',
```

packages/typegpu/src/data/struct.ts
```diff
@@ -38,12 +38,23 @@ export function struct<TProps extends Record<string, AnyWgslData>>(
...
+
 // --------------
 // Implementation
 // --------------
 
 const WgslStructImpl = {
-  [$internal]: true,
+  [$internal]: {
+    builtinName: undefined,
+  },
```
aleksanderkatan: How about we make this property optional and just put `{}` here? Could we make `{}` a default instead of `true` for internals in general, or is it too much overhead?

packages/typegpu/tests/struct.test.ts
aleksanderkatan, line 1: Could you also add a test to make sure the following resolves correctly?
```ts
const testFn = tgpu.fn([])(() => {
  const result = std.frexp(1);
});
```

packages/typegpu/src/std/numeric.ts
aleksanderkatan, line 1: How about we categorize these (as well as boolean.ts) into files the same way they are listed in the WGSL docs, and create a separate file for all operators?

packages/typegpu/src/std/numeric.ts
```diff
@@ -406,12 +442,122 @@ export const dot = createDualImpl(
...
+  (e1: number, e2: number): number => {
+    throw new Error(
+      'CPU implementation for dot4U8Packed not implemented yet. Please submit an issue at https://github.com/software-mansion/TypeGPU/issues',
+    );
+  },
+  // GPU implementation
+  (e1, e2) => snip(`dot4U8Packed(${e1.value}, ${e2.value})`, u32),
+  'dot4U8Packed',
+  [u32, u32],
+);
```
aleksanderkatan: ~~Nit: these two are not following the alphabetical order~~
They are in alphabetical order! It is just WGSL spec that has them in wrong order:
image

packages/typegpu/src/std/numeric.ts
aleksanderkatan, line 1: Some of the functions are missing the private remarks. I wonder, are they even necessary? I would either opt to remove them entirely, or to make them regular jsdocs so that users can easily check which functions are WGSL builtins and which are not (`allEq`, operators, `isCloseTo`, potentially `ulog2` in the future etc.).

packages/typegpu/src/std/numeric.ts
```diff
@@ -608,38 +841,102 @@ export const mod: ModOverload = createDualImpl(
...
- */
-export const exp = createDualImpl(
+export const modf = createDualImpl(
   // CPU implementation
-  <T extends AnyFloatVecInstance | number>(value: T): T => {
-    if (typeof value === 'number') {
-      return Math.exp(value) as T;
-    }
-    return VectorOps.exp[value.kind](value) as T;
+  <T extends AnyFloatVecInstance | number>(value: T) => {
```
aleksanderkatan, line 793:
```suggestion
  <T extends AnyFloat32VecInstance | number>(value: T) => {
```

packages/typegpu/src/std/numeric.ts
```diff
@@ -708,58 +1002,115 @@ export const reflect = createDualImpl(
...
+  },
+  // GPU implementation
+  (e1, e2, e3) =>
+    snip(`refract(${e1.value}, ${e2.value}, ${e3.value})`, e1.dataType),
+  'refract',
+  (
+    e1,
+    e2,
+    e3,
+  ) => [
```
```suggestion
  (e1, e2, e3) => [
```

packages/typegpu/src/data/struct.ts
reczkok: I think I like the verbosity in this particular case - but now that I think about it it should be `null` instead of `undefined`. Also I don't see a reason to change it to `{}` at least for this particular pr.

packages/typegpu/src/std/numeric.ts
reczkok: I agree, it will be easier to maintain

reczkok: I think turning them into regular jscocs is a good idea

packages/typegpu/tests/struct.test.ts
reczkok: It will just resolve to:
```
fn testFn(x) {
  var result = frexp(1);
}
```
But I can

packages/typegpu/src/std/numeric.ts
```diff
@@ -381,292 +220,439 @@ export const cosh = createDualImpl(
...
+  },
+  // GPU implementation
+  (a) => snip(`fract(${a.value})`, a.dataType),
+  'fract',
+);
+
+const FrexpResultF32 = builtinStruct(
+  { fract: f32, exp: i32 },
+  '__frexp_result_f32',
+);
```
reczkok: I don't export it since it cannot be used in user WGSL code:

Image

packages/typegpu/src/std/numeric.ts
```diff
@@ -381,292 +220,439 @@ export const cosh = createDualImpl(
...
+  },
+  // GPU implementation
+  (value) =>
+    snip(
+      `frexp(${value.value})`,
+      FrexpResultF32,
+    ),
+  'frexp',
+  [f32],
+);
```
aleksanderkatan: Frexp overloads actually include f16, f32, abstract float and vectors of those

packages/typegpu/src/std/numeric.ts
```diff
@@ -381,292 +220,439 @@ export const cosh = createDualImpl(
...
+  (value) => snip(`inverseSqrt(${value.value})`, value.dataType),
+  'inverseSqrt',
+);
+
+export const ldexp = createDualImpl(
+  // CPU implementation
+  <T extends AnyFloatVecInstance | number>(
+    e1: T,
+    e2: AnyIntegerVecInstance | number,
+  ): T => {
```
aleksanderkatan, line 623: "I is a vector if and only if T is a vector"
Here, we allow for all combinations

packages/typegpu/src/std/numeric.ts
```diff
@@ -699,6 +685,80 @@ export const mix: MixOverload = createDualImpl(
   'mix',
 );
 
+export const modf = createDualImpl(
+  // CPU implementation
+  <T extends AnyFloat32VecInstance | number>(value: T) => {
+    throw new Error(
```
aleksanderkatan: modf can accept f16

reczkok: I think the return type is also busted

packages/typegpu/src/std/matrix.ts
```diff
@@ -2,8 +2,7 @@ import type { Snippet } from '../data/snippet.ts';
 import { mat4x4f } from '../data/matrix.ts';
 import type { m4x4f, v3f } from '../data/wgslTypes.ts';
 import { createDualImpl } from '../core/function/dualImpl.ts';
-import { mul } from './numeric.ts';
-
+import { mul } from './operators.ts';
```
iwoplaza, line 5: Tiny nitpick, missing empty line
```suggestion
import { mul } from './operators.ts';

```

PR #1569, opened 2025-08-05
fix: Fix infix operator value resolution for bindings

### LINKED ISSUE #1570: bug: Infix operators don't work when applied directly to fixed entries.
Example:
```ts
const someUniform = root.createUniform(d.vec4f);
const someBuffer = root.createBuffer(d.vec2f).$usage('storage');
const someFixedUsage = someBuffer.as('mutable');

const testFn = tgpu.fn([])(() => {
  const v2 = someUniform.$.mul(2);
  const v3 = someUniform.$.add(d.vec3f().zxzx);
  const v4 = someFixedUsage.$.add(d.vec2f(1, 2));
});
```
Expected:
```wgsl
@group(0) @binding(0) var<uniform> someUniform_1: vec4f;
@group(0) @binding(1) var<storage, read_write> someBuffer_2: vec2f;

fn testFn_0() {
  var v2 = (someUniform_1 * 2);
  var v3 = (someUniform_1 + vec3f().zxzx);
  var v4 = (someBuffer_2 + vec2f(1, 2));
}
```
Actual:
```wgsl
fn testFn_0() {
  var v2 = (.value:someUniform * 2);
  var v3 = (.value:someUniform + vec3f().zxzx);
  var v4 = (.value:someBuffer + vec2f(1, 2));
}
```

reczkok:
This PR ~~fixes~~ (_does not fix_ as it turns out) an issue when fixed entries wouldn't get properly resolved in combination with infix dispatches. Also adds some more tests.

closes #1570

### DIFF
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -381,6 +381,8 @@ export function generateExpression(
     }
 
     if (callee.value instanceof InfixDispatch) {
+      const lhs = callee.value.lhs;
+
       // Infix operator dispatch.
       if (!argNodes[0]) {
         throw new WgslTypeError(
@@ -388,7 +390,7 @@ export function generateExpression(
         );
       }
       return callee.value.operator(
-        callee.value.lhs,
+        snip(ctx.resolve(lhs.value), lhs.dataType),
         generateExpression(ctx, argNodes[0]),
       );
     }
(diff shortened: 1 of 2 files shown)


aleksanderkatan (approved):
Nice! ⛵

iwoplaza:
This is unfortunately just a temporary fix for a bigger issue, the "snippets not being resolved in dualImpls" issue.
For example, using buffers as right-hand-side of an infix operator still has the same problem:
```ts
const v4 = someFixedUsage.$.add(someFixedUsage.$);
// resolves to: var v4 = (someBuffer_2 + .value:someBuffer);
```
The same fix would have to be applied to the `std.mul` function, which at that point, we could remove the fix specific to infix operators

iwoplaza:
I wanted to wait for https://github.com/software-mansion/TypeGPU/pull/1567 to merge, as the fix would touch the same files

PR #1575, opened 2025-08-06
fix: More accurate inference for slots & derived

iwoplaza:


packages/typegpu/src/shared/repr.ts
```diff
@@ -64,6 +65,9 @@ export type InferGPURecord<
   [Key in keyof T]: InferGPU<T[Key]>;
 };
 
+export type GPUValueOf<T> = T extends
+  { [$gpuValueOf](...args: never[]): infer TValue } ? TValue : T;
```
reczkok, line 69: Is there a difference between `never[]` and `unknown[]` here?

reczkok (approved):
Yes🫡

packages/typegpu/src/core/slot/derived.ts
```diff
@@ -42,14 +37,12 @@ function createDerived<T>(compute: () => T): TgpuDerived<T> {
   const result = {
     resourceType: 'derived' as const,
     '~compute': compute,
-    [$repr]: undefined as Infer<T>,
-    [$gpuRepr]: undefined as InferGPU<T>,
 
-    [$gpuValueOf](ctx: ResolutionCtx): InferGPU<T> {
+    [$gpuValueOf](ctx: ResolutionCtx): GPUValueOf<T> {
```
aleksanderkatan, line 41: What is the general idea behind the `$gpuValueOf` symbol? Here, it is used internally in `value`, is it ever used anywhere else? If not, why do we keep it in the interface?

packages/typegpu/src/shared/repr.ts
iwoplaza: `...args: never[]` is just wider than `...args: unknown[]`

packages/typegpu/src/core/slot/derived.ts
iwoplaza: It's used by slots to unwrap the gpu value of what they hold, so when we do `.$` on a slot, it goes into the value it hold and also does a `.$` on it until it can't anymore. If we were to literally access `.$` on it, then objects with a `$` property would also be unwrapped, so we hold the value under a symbol.

PR #1584, opened 2025-08-07
feat: Callable unstructs and disarrays

aleksanderkatan:
Changes:
- merge `schemaDefaultWrapper` and `schemaCloneWrapper` into `schemaCallWrapper` that also handles vertex formats,
- add copy semantics to unstruct call,
- implement disarray call.

The wgslGenerator is left as is.


packages/typegpu/src/data/utils.ts
```diff
@@ -1,31 +1,27 @@
...
-    return (schema as unknown as ((item: T) => T))(item);
-  } catch {
-    const maybeType = (schema as { type: string })?.type;
+    // TgpuVertexFormatData are not callable
+    const callSchema = (maybeType in formatToWGSLType
+      ? formatToWGSLType[maybeType as keyof typeof formatToWGSLType]
+      : schema) as unknown as ((item?: T) => T);
+    if (item === undefined) {
+      return callSchema();
+    }
```
aleksanderkatan, line 19: This `if` is necessary because vector constructors actually distinguish between `vec2f()` and `vec2f(undefined)` (the second throws an error, I'm not sure, do we consider this an issue?)

reczkok: I would say the error is correct. Calling `vec2f(undefined)` is not valid imo.

aleksanderkatan: It's just that in most cases, `f()` is equivalent to `f(undefined)`.
image
Though, now that I researched some more, I see that there are more exceptions to this, like destructuring a default parameter, so I guess we may just conclude `vec2f(undefined)` as invalid.

packages/typegpu/src/data/struct.ts
```diff
@@ -1,6 +1,7 @@
 import { getName, setName } from '../shared/meta.ts';
 import { $internal } from '../shared/symbols.ts';
-import { schemaCloneWrapper, schemaDefaultWrapper } from './utils.ts';
+import type { AnyData } from './index.ts';
```
iwoplaza:
```suggestion
import type { AnyData } from './dataTypes.ts';
```

iwoplaza (approved):
Merge! Merge! Merge! (after nit fix)

PR #1585, opened 2025-08-07
feat: GPU recursion error

### LINKED ISSUE #1523: feat: Detect recursion and throw descriptive error
We can hold the set of items currently being resolved in the ResolutionCtx, and throw a new `GPURecursionError` that says something along the lines of "Recursion is not supported on GPUs, `foo` references `bar`".

cieplypolar:
Changes:
- In `ResolutionCtxImpl.resolve` we keep track of items that are currently being resolved. If we try to resolve item for the second time, it is a recursion and we throw error.
- Uncommented appropriate test in `tgslFn.test.ts`.

Condition looks like this:
```ts
if (this._currentlyResolvedItems.has(item) &&
   !this._memoizedResolves.has(item)) {
  throw ...
}
```

Closes #1523.

### DIFF
--- a/packages/typegpu/src/core/resolve/externals.ts
+++ b/packages/typegpu/src/core/resolve/externals.ts
@@ -95,7 +95,7 @@ export function replaceExternalsInWgsl(
       !externalRegex.test(wgsl)
     ) {
       console.warn(
-        `The external '${externalName}' was unused in the resolved template.`,
+        `The external '${externalName}' wasn't used in the resolved template.`,
       );
       // continue anyway, we still might need to resolve the external
     }
--- a/packages/typegpu/tests/resolve.test.ts
+++ b/packages/typegpu/tests/resolve.test.ts
@@ -39,11 +39,8 @@ describe('tgpu resolve', () => {
       },
       names: 'strict',
     });
-    expect(parse(resolved)).toBe(
-      parse(`
-          fn foo() { 
-            var g = 1000; 
-          }`),
+    expect(resolved).toMatchInlineSnapshot(
+      `"fn foo() { var g = 1000; }"`,
     );
   });
 
@@ -533,23 +530,18 @@ describe('tgpu resolveWithContext', () => {
       .mockImplementation(() => {});
 
     tgpu.resolveWithContext({
-      template: `
-          fn testFn() {
-            return;
-          }
-        `,
+      template: 'fn testFn() { return; }',
       externals: {
         ArraySchema: d.arrayOf(d.u32, 4),
         JavaScriptObject: { field: d.vec2f() },
       },
-      names: 'strict',
     });
 
     expect(consoleWarnSpy).toHaveBeenCalledWith(
-      "The external 'ArraySchema' was unused in the resolved template.",
+      "The external 'ArraySchema' wasn't used in the resolved template.",
     );
     expect(consoleWarnSpy).toHaveBeenCalledWith(
-      "The external 'JavaScriptObject' was unused in the resolved template.",
+      "The external 'JavaScriptObject' wasn't used in the resolved template.",
     );
   });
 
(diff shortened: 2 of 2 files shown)


packages/typegpu/src/resolutionCtx.ts
```diff
@@ -332,6 +332,7 @@ export class ResolutionCtxImpl implements ResolutionCtx {
   readonly #modeStack: ExecState[] = [];
   private readonly _declarations: string[] = [];
   private _varyingLocations: Record<string, number> | undefined;
+  private _currentlyResolvedItems: WeakSet<object> = new WeakSet();
```
iwoplaza: For new private properties, we can use the recently added ECMAScript standard "private elements".
Can be also made `readonly`, as we never reassign it to anything.
```suggestion
  readonly #currentlyResolvedItems = new WeakSet<object>();
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -620,6 +621,18 @@ export class ResolutionCtxImpl implements ResolutionCtx {
   }
 
   resolve(item: unknown, schema?: AnyData | undefined): string {
+    if (item && typeof item === 'function') {
```
reczkok: Could we limit what we store in the set? Nearly all our objects are of type `function` and if I understand correctly we only care about `TgpuFn`'s
```suggestion
    if (isTgpuFn(item)) {
```

reczkok (approved):
Works great! Here are two free test cases:
```ts
it('throws when it detects a cyclic dependency (when using slots)', () => {
  // biome-ignore lint/style/useConst: one has to be assigned later
  let one: TgpuFn;
  // biome-ignore lint/style/useConst: fnSlot has to be assigned later
  let fnSlot: TgpuSlot<TgpuFn<() => d.F32>>;
  // biome-ignore lint/style/useConst: three has to be assigned later
  let three: TgpuFn;
  // biome-ignore lint/style/useConst: two has to be assigned later
  let two: TgpuFn;
  one = tgpu.fn([], d.f32)(() => two() + 2);
  fnSlot = tgpu.slot(tgpu.fn([], d.f32)(() => one() + 2).$name('inner'));
  three = tgpu.fn([], d.f32)(() => fnSlot.$() + 1);
  two = tgpu.fn([], d.f32)(() => three() + 2);
  expect(() => parseResolved({ one })).toThrowErrorMatchingInlineSnapshot(`
    - <root>
    - fn:one
    - call:two
    - fn:two
    - call:three
    - fn:three
    - call:inner
    - fn:inner
    - call:one: Recursive function fn:one detected. Recursion is not allowed on the GPU.]
  `);
});

it('throws when it detects a cyclic dependency (when using derived)', () => {
  // biome-ignore lint/style/useConst: one has to be assigned later
  let one: TgpuFn;

  const flagSlot = tgpu.slot(false);
  const fnSlot = tgpu.slot<TgpuFn<() => d.F32>>();
  const mainFn = tgpu.fn([], d.f32)(() => 1000);
  const fallbackFn = tgpu.fn([], d.f32)(() => one());

  const derivedFn = tgpu['~unstable'].derived(() => {
    if (flagSlot.$) {
      return fnSlot.$;
    }
    return fallbackFn;
  }).with(fnSlot, mainFn);

  one = tgpu.fn([], d.f32)(() => derivedFn.$() + 2);

  expect(() => parseResolved({ one })).toThrowErrorMatchingInlineSnapshot(`
    - <root>
    - fn:one
    - call:fallbackFn
    - fn:fallbackFn
    - call:one: Recursive function fn:one detected. Recursion is not allowed on the GPU.]
  `);

  const boundOne = one.with(flagSlot, true);

  expect(tgpu.resolve({ externals: { boundOne } })).toMatchInlineSnapshot(`
    "fn mainFn_1() -> f32 {
      return 1000;
    }

    fn one_0() -> f32 {
      return (mainFn_1() + 2);
    }"
  `);
});
```

cieplypolar:
@reczkok, thank you for these two additional tests! ❤️

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -620,6 +622,18 @@ export class ResolutionCtxImpl implements ResolutionCtx {
   }
 
   resolve(item: unknown, schema?: AnyData | undefined): string {
+    if (isTgpuFn(item)) {
+      if (
+        this.#currentlyResolvedItems.has(item) &&
+        !this._memoizedResolves.has(item)
+      ) {
+        throw new Error(
+          `Recursive function ${item.toString()} detected. Recursion is not allowed on the GPU.`,
```
iwoplaza: A call to `.toString()` should happen automatically here

```suggestion
          `Recursive function ${item} detected. Recursion is not allowed on the GPU.`,
```

PR #1588, opened 2025-08-07
fix: Resolving in dualImpl

iwoplaza:
**Changes:**
- Resolving function arguments is deferred to the actual function, meaning a function can choose to resolve its passed in arguments, or act on their un-resolved values (useful for precomputing).
- All std functions should now resolve their arguments via the new `stitch` utility, which just resolves the given parameters and injects them into a template string.
- The `add`, `sub`, `mul` and `div` operators are now capable of precomputing values if they're known at resolution time.

### DIFF
--- a/apps/typegpu-docs/src/content/docs/fundamentals/bind-groups.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/bind-groups.mdx
@@ -142,6 +142,26 @@ const fooLayout = tgpu.bindGroupLayout({
   // ...
 });
 ```
+
+You don't have to write arrow functions every time, `d.arrayOf(schema)` is a convenient shorhand for that.
+
+```diff lang="ts"
+// main.ts
+const Filter = (n: number) =>
+  d.struct({
+    clamp: d.f32,
+    // Unfortunately, in structs, you cannot simply write d.arrayOf(d.f32)
+    values: d.arrayOf(d.f32, n),
+  });
+
+const fooLayout = tgpu.bindGroupLayout({
+-  factors: { storage: (n: number) => d.arrayOf(d.f32, n) },
++  factors: { storage: d.arrayOf(d.f32) },
+  filter: { storage: Filter },
+  // ...
+});
+```
+
 Matching WGSL code:
 ```wgsl
 // shader.wgsl
--- a/apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
@@ -32,7 +32,7 @@ d.u32(31.1); // 31
 d.u32(-1); // 4294967295
 d.f32(true); // 1
 d.i32(); // 0 (default value)
-d.bool(0.01); // true 
+d.bool(0.01); // true
 ```
 
 :::caution
@@ -203,17 +203,17 @@ const Circle = d.struct({
 
 // ---cut---
 const redCircle1 = Circle({
-  centerPos: d.vec2i(2, 4), 
+  centerPos: d.vec2i(2, 4),
   radius: 0.2,
 });
 
 const redCircle2 = Circle({
-  centerPos: d.vec3i(2, 4, 0), 
+  centerPos: d.vec3i(2, 4, 0),
   radius: "0.2",
 });
 
 const redCircle3 = Circle({
-  centerPos: d.vec3i(2, 4, 0), 
+  centerPos: d.vec3i(2, 4, 0),
 });
 
 const diam = redCircle1.rad * 2;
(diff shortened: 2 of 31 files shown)


packages/typegpu/src/core/resolve/stitch.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { ResolutionCtx } from '../../types.ts';
+
+type ValueOrArray<T> = T | T[];
+
+/**
+ * The reverse of snipping
+ * @param strings T
+ * @param ctx
+ * @param snippets
+ * @returns
```
aleksanderkatan: This JSDocs is outdated

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -82,7 +82,7 @@ describe('wgslGenerator', () => {
       scientificLiteral: {
         value: '12e10',
         wgsl: '12e10',
-        dataType: abstractFloat,
+        dataType: abstractInt,
```
lursz: Isn't `abstractInt` going to overflow in this situation? 
(test passes though)

aleksanderkatan: WebGPU actually throws a shader creation error for `i32(1e12)` and not for `f32(1e12)`

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -170,6 +171,15 @@ export function generateTypedExpression(
...
 }
 
+const opCodeToCodegen = {
+  '+': add[$internal].gpuImpl,
+  '-': sub[$internal].gpuImpl,
+  '*': mul[$internal].gpuImpl,
+  '/': div[$internal].gpuImpl,
+} satisfies Partial<
+  Record<tinyest.BinaryOperator, (...args: never[]) => unknown>
+>;
```
aleksanderkatan, line 181: Can this somehow be merged into `infixOperators` in the same file, or do we want to keep those separate?

packages/typegpu/src/std/operators.ts
```diff
@@ -79,13 +55,27 @@ function cpuAdd(lhs: number | NumVec | Mat, rhs: number | NumVec | Mat) {
...
-    ),
-  'unify',
+  // CODEGEN implementation
+  (lhs, rhs) => {
+    const resultType = isSnippetNumeric(lhs) ? rhs.dataType : lhs.dataType;
+
+    if (
+      (typeof lhs.value === 'number' ||
+        isVecInstance(lhs.value) ||
+        isMatInstance(lhs.value)) &&
```
aleksanderkatan, line 66: Can we check here if neither are a string instead, or can e.g. buffer usages still appear here?

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -187,33 +187,22 @@ export function getTypeForIndexAccess(
...
+  if (/^0x[0-9a-f]+$/i.test(str)) {
+    return Number.parseInt(str);
   }
 
   // Binary literals
-  if (/^0b[01]+$/i.test(value)) {
-    return snip(`${Number.parseInt(value.slice(2), 2)}`, abstractInt);
+  if (/^0b[01]+$/i.test(str)) {
+    return Number.parseInt(str.slice(2), 2);
   }
```
aleksanderkatan: Should we parse octal as well?

aleksanderkatan (approved):
Great! ✂️🪡✂️🪡

I see some places where there is a `snip` with no explicit resolve or stitch, could you check whether those are all fine?
- fnCore.ts:171
- fnCore.ts:183
- tgpuFn.ts:236
- tgpuFn.ts:300
- vector.ts:336
- texture.ts:174
- texture.ts:305

packages/typegpu/src/std/operators.ts
iwoplaza: Buffer usages less so, but proxies can (e.g. `bufferUsage.$`). I'll write a test for it to confirm though

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: Not too much overlap between them

packages/typegpu/src/tgsl/generationHelpers.ts
iwoplaza: We didn't support it before, and I plan on altering the tinyest format a bit so we don't have to worry about it at all

packages/typegpu/src/std/operators.ts
iwoplaza: Confirmed, does not work and causes proxies to fail

packages/typegpu/src/core/resolve/stitch.ts
```diff
@@ -0,0 +1,37 @@
...
+      : ctx.resolve(stringOrSnippet.value, stringOrSnippet.dataType);
+  }
+
+  let result = '';
+  for (let i = 0; i < strings.length; ++i) {
+    result += strings[i];
+    const snippet = snippets[i];
+    if (Array.isArray(snippet)) {
+      result += snippet
+        .filter((s) => s !== undefined)
```
reczkok, line 29: In what situation would we encounter undefined here?

reczkok (approved):
Amazing work! 🙇

packages/typegpu/src/core/resolve/stitch.ts
reczkok: Oh I answered myself by the end of the review😅

PR #1591, opened 2025-08-08
fix: Pointers for reference types

aleksanderkatan:
Blocked by #1584

### DIFF
--- a/apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/pointers.ts
+++ b/apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/pointers.ts
@@ -1,0 +1,36 @@
+import tgpu from 'typegpu';
+import * as d from 'typegpu/data';
+import * as std from 'typegpu/std';
+
+const SimpleStruct = d.struct({ vec: d.vec2f });
+
+const modifyNum = tgpu.fn([d.ptrFn(d.u32)])((ptr) => {
+  ptr += 1;
+});
+
+const modifyVec = tgpu.fn([d.ptrFn(d.vec2f)])((ptr) => {
+  ptr.x += 1;
+});
+
+const modifyStruct = tgpu.fn([d.ptrFn(SimpleStruct)])((ptr) => {
+  ptr.vec.x += 1;
+});
+
+// TODO: replace `s = s &&` with `s &&=` when implemented
+export const pointersTest = tgpu.fn([], d.bool)(() => {
+  let s = true;
+
+  let num = d.u32(0);
+  modifyNum(num);
+  s = s && (num === 1);
+
+  const vec = d.vec2f(1, 2);
+  modifyVec(vec);
+  s = s && std.allEq(vec, d.vec2f(2, 2));
+
+  const myStruct = SimpleStruct({ vec: d.vec2f(3, 4) });
+  modifyStruct(myStruct);
+  s = s && std.allEq(myStruct.vec, d.vec2f(4, 4));
+
+  return s;
+});
--- a/apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/index.ts
+++ b/apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/index.ts
@@ -4,6 +4,7 @@ import { arrayAndStructConstructorsTest } from './array-and-struct-constructors.
 import { infixOperatorsTests } from './infix-operators.ts';
 import { logicalExpressionTests } from './logical-expressions.ts';
 import { matrixOpsTests } from './matrix-ops.ts';
+import { pointersTest } from './pointers.ts';
 
 const root = await tgpu.init();
 const result = root.createMutable(d.i32, 0);
@@ -15,6 +16,7 @@ const computeRunTests = tgpu['~unstable']
     s = s && matrixOpsTests();
     s = s && infixOperatorsTests();
     s = s && arrayAndStructConstructorsTest();
+    s = s && pointersTest();
 
     if (s) {
       result.value = 1;
--- a/packages/typegpu/src/data/ptr.ts
+++ b/packages/typegpu/src/data/ptr.ts
@@ -1,76 +1,93 @@
 import { $internal } from '../shared/symbols.ts';
 import type { AnyData } from './dataTypes.ts';
-import type { Ptr } from './wgslTypes.ts';
+import type { Access, AddressSpace, Ptr } from './wgslTypes.ts';
 
 export function ptrFn<T extends AnyData>(
   inner: T,
 ): Ptr<'function', T, 'read-write'> {
-  return {
-    [$internal]: true,
-    type: 'ptr',
-    inner,
-    addressSpace: 'function',
-    access: 'read-write',
-  } as Ptr<'function', T, 'read-write'>;
+  return INTERNAL_createPtr('function', inner, 'read-write');
 }
 
 export function ptrPrivate<T extends AnyData>(
   inner: T,
 ): Ptr<'private', T, 'read-write'> {
-  return {
-    [$internal]: true,
-    type: 'ptr',
-    inner,
-    addressSpace: 'private',
-    access: 'read-write',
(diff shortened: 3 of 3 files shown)


packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1687,7 +1721,7 @@ export function isWgslStruct<T extends WgslStruct>(
 }
 
 /**
- * Checks whether passed in value is a pointer ('function' scope) schema.
+ * Checks whether passed in value is a pointer schema.
  *
  * @example
  * isPtrFn(d.ptrFn(d.f32)) // true
```
iwoplaza: I guess we also have to update the example 👀

apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
```diff
@@ -331,6 +330,34 @@ const myDefaultArray = MyArray(); // [0, 0, 0, 0]
...
+// ---cut---
+const AtomicI32 = d.atomic(d.i32);
+//    ^?
+```
+
+## Pointers
+
+To create a schema corresponding to a pointer data type, wrap the inner schema with a pointer constructor corresponding to the appropriate address space.
+
+For address spaces `function`, `private`, `workgroup`, `storage`, `uniform`, `handle`, use `d.ptrFn`, `d.ptrPrivate`, `d.ptrWorkgroup`, `d.ptrStorage`, `d.ptrUniform`, `d.ptrHandle` respectively.
```
reczkok: We should mention that only function and private are actually usable without an extension. (Maybe we could only mention those two here and add a note that if you use the `unrestricted_pointer_parameters` extensions you can also use the rest)

apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
```diff
@@ -331,6 +330,34 @@ const myDefaultArray = MyArray(); // [0, 0, 0, 0]
...
+
+```ts twoslash
+import * as d from 'typegpu/data';
+// ---cut---
+const VecPtrFn = d.ptrFn(d.vec3f);
+//    ^?
+
+const VecPtrUniform = d.ptrUniform(d.vec2f);
```
reczkok: I also wouldn't use a uniform ptr here

apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/pointers.ts
reczkok, line 1: We could also test private pointer

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1339,7 +1341,7 @@ export type Access = 'read' | 'write' | 'read-write';
 
 export interface Ptr<
   TAddr extends AddressSpace = AddressSpace,
-  TInner extends BaseData = BaseData, // can also be sampler or texture (╯'□')╯︵ ┻━┻
+  TInner extends StorableData = StorableData, // can also be sampler or texture (╯'□')╯︵ ┻━┻
```
reczkok: Since the type now actually represents samplers and textures we can (**_sadly_**) remove the comment

apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
aleksanderkatan: I swapped it with private

apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
aleksanderkatan: If I'm understanding this correctly, `workgroup`, `storage` and `uniform` are available by default, they just cannot be used as function parameters. I added a note about it.

packages/typegpu/src/data/wgslTypes.ts
aleksanderkatan: I removed it, along with the `TgpuTexture` and `TgpuSampler` from the `StorableData` type. I checked and we cannot create handle pointers in WGSL. I also removed `handle` from docs (though I left it in code).

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -150,6 +156,22 @@ JavaScript does not support operator overloading.
 This means that, while you can still use operators for numbers, 
 you have to use supplementary functions from `typegpu/std` (*add, mul, eq, lt, ge...*) for operations involving vectors and matrices.
 
+* **Pointers** --
+Since WGSL pointers can only point to a narrow set of items called Storable Types, 
+TGPU tries to dereference all pointers used in TGSL automatically.
```
reczkok: This sentence is also no necessarily true - we will try to dereference when we think it's appropriate (which just happens to be nearly all the time if pointers are used as intended)
```suggestion
TypeGPU tries to dereference all pointers used in TGSL automatically. 
```

apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
```diff
@@ -344,6 +343,40 @@ const array = ArrayPartialSchema(2)([1.2, 19.29]);
...
+## Atomics
+
+To create a schema corresponding to an atomic data type, wrap `d.i32` or `d.u32` with `d.atomic`.
+
+```ts twoslash
+import * as d from 'typegpu/data';
+// ---cut---
+const AtomicI32 = d.atomic(d.i32);
+//    ^?
+```
```
reczkok, line 355: Maybe we could mention as a tip that you can only use atomic values via `std` in TGSL (and WGSL too for that matter)? We already mention it (TGSL) inside this page and I think this could prevent some confusion.

apps/typegpu-docs/src/content/docs/fundamentals/data-schemas.mdx
```diff
@@ -344,6 +343,40 @@ const array = ArrayPartialSchema(2)([1.2, 19.29]);
...
+
+For address spaces `function`, `private`, `workgroup`, `storage` and `uniform`, use `d.ptrFn`, `d.ptrPrivate`, `d.ptrWorkgroup`, `d.ptrStorage` and `d.ptrUniform` respectively.
+
+```ts twoslash
+import * as d from 'typegpu/data';
+// ---cut---
+const VecPtrFn = d.ptrFn(d.vec3f);
+//    ^?
+
+const VecPtrPrivate = d.ptrPrivate(d.vec2f);
```
reczkok: I don't know which naming convention to apply here. On one hand those are `schemas` but they function more like `tgpu.var`'s so maybe we should go with camel case?

apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
```diff
@@ -18,7 +18,7 @@ For the TGSL functions to work, you need to use the dedicated build plugin -- [u
 ## Usage
 
 Instead of using a WGSL code string, you can pass TGSL to the tgpu function shell as an argument instead.
-Functions from the WGSL standard library (*distance, arrayLength, workgroupBarrier, [etc.](https://github.com/software-mansion/TypeGPU/blob/release/packages/typegpu/src/std/index.ts)*) are accessible through the `typegpu/std` endpoint. 
+Functions from the WGSL standard library (*distance, arrayLength, workgroupBarrier, [etc.](https://github.com/software-mansion/TypeGPU/blob/release/packages/typegpu/src/std/index.ts)*) are accessible through the `typegpu/std` endpoint.
```
iwoplaza:
```suggestion
Functions from the WGSL standard library (*distance, arrayLength, workgroupBarrier, [etc.](https://github.com/software-mansion/TypeGPU/blob/release/packages/typegpu/src/std/index.ts)*) are accessible through the `typegpu/std` entrypoint.
```

iwoplaza (approved):
Looks great! 👏

PR #1605, opened 2025-08-13
Feat: distributions from uniform [0, 1)

cieplypolar:


apps/typegpu-docs/src/content/examples/simple/probability/index.html
```diff
@@ -0,0 +1,38 @@
...
+  }
+</style>
+<div id="help">
+  <h3 class="controls-header">Controls</h3>
+  <p class="controls"><b>Left Mouse Button / Drag:</b> Rotate plot</p>
+  <p class="controls">
+    <b>Right Mouse Button / Two-Finger Drag:</b> Move plot
+  </p>
+  <p class="controls"><b>Scroll:</b> Zoom</p>
+</div>
```
aleksanderkatan, line 38: Personal preference, but I'd hide the help box after one click/tab, or alternatively, I'd make it take more time for the help box to reappear

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,136 @@
...
+        sampleBufferSlotTempAlias.$[i] = distributionSlotTempAlias.$();
+      }
+    });
+
+    this.#dataMoreWorkersFunc = tgpu['~unstable'].computeFn({
+      in: { gid: d.builtin.globalInvocationId, count: d.builtin.numWorkgroups },
+      workgroupSize: [1],
+    })((input) => {
+      const gid = input.gid;
+      randf.seed(d.f32(gid.x) / input.count.x);
```
aleksanderkatan: Do we want the distributions to be deterministic? Right now, every time I select a distribution, the result is the same. On the other hand, this makes selecting more points look like just adding them, not adding and reshuffling.

apps/typegpu-docs/src/content/docs/ecosystem/typegpu-noise.mdx
```diff
@@ -100,10 +101,43 @@ const main = tgpu['~unstable'].fragmentFn({
...
+$$
+
+### Continuous
+- `sampleExclusive` - works the same as `sample`, but returns a value strictly between 0 and 1
+- `exponential` - returns a number based on the exponential distribution with a given `rate`. The `rate` must be greater than 0. It is commonly used for modeling the time until an event occurs
+$$
+\mathrm{Exponential}(\lambda) = -\frac{1}{\lambda} \ln(1 - U)
+$$
+
+- `normal` - returns a number based on the normal (Gaussian) distribution with a given mean (`mu`) and standard deviation (`sigma`). `sigma` must be `> 0`. Following transformation is called Box-Muller transform
```
aleksanderkatan, line 126: Do we want to spell out `mu` and `sigma` instead of using greek letters? If so, then it should be `mi`, right?

apps/typegpu-docs/src/content/examples/simple/probability/constants.ts
```diff
@@ -0,0 +1,14 @@
+import { Distribution, ExecutionMode } from './types.ts';
+
+export const cameraPositionGeo = [0, 0, 0.5];
+export const cameraPositionHist = [0, 0, 0.2];
+export const initialCameraAngle = 15;
+export const initialNumSamples = 2000;
+export const numSamplesOptions = [100, 1000, 2000, 5000, 10000, 50000];
```
aleksanderkatan: nit
```suggestion
export const numSamplesOptions = [100, 1000, 2000, 5000, 10000, 50000];
export const initialNumSamples = numSamplesOptions[2];
```

apps/typegpu-docs/src/content/examples/simple/probability/constants.ts
```diff
@@ -0,0 +1,14 @@
...
+export const initialDistribution: Distribution = Distribution.NORMAL;
+export const distributions: Distribution[] = Object.values(Distribution);
+export const initialExecutionMode: ExecutionMode = ExecutionMode.SINGLE;
+export const executionModes: ExecutionMode[] = Object.values(ExecutionMode);
+export const minNumSamples = 100;
+export const maxNumSamples = 65100;
+export const step = 100;
```
aleksanderkatan: When a variable is exported, IDEs usually don't mention the fact that they are unused
```suggestion
```

packages/typegpu-noise/src/random.ts
```diff
@@ -48,24 +76,89 @@ export const randInUnitCircle: TgpuFn<() => d.Vec2f> = tgpu
...
   });
 
+const safeNormalize: TgpuFn<(v: d.Vec3f) => d.Vec3f> = tgpu
+  .fn([d.vec3f], d.vec3f)((v) => {
+    let vNonZero = v;
+    if (length(v) === 0) {
+      vNonZero = v.add(EPS);
+    }
+    return normalize(vNonZero);
+  });
```
aleksanderkatan, line 89: Not sure about this function. How about we return `d.vec3f(1, 0, 0)` or `std.normalize(d.vec3f(1))` if length is smaller than `EPS` instead?

packages/typegpu-noise/src/random.ts
```diff
@@ -48,24 +76,89 @@ export const randInUnitCircle: TgpuFn<() => d.Vec2f> = tgpu
...
     return mul(sign(alignment), value);
   });
+
+export const randUniformExclusive: TgpuFn<() => d.F32> = tgpu
+  .fn([], d.f32)(() => {
+    // Our generator currently operates on the range [0, 1),
+    // so clamping with 1 - EPS isn't necessary.
+    // However, let's keep it, in case we change the generator's domain in the future.
+    return clamp(randomGeneratorSlot.value.sample(), 0 + EPS, 1 - EPS);
+  });
```
aleksanderkatan, line 134: What is our policy on `EPS`? If we "discretisize" our (0,1) range to multiples of `EPS`, then `EPS` would be twice as likely in this generator as `2*EPS`. How about we randomize in a loop, or rescale?

apps/typegpu-docs/src/content/examples/simple/probability/types.ts
```diff
@@ -0,0 +1,78 @@
...
+  IN_HEMISPHERE: 'inHemisphere',
+  ON_HEMISPHERE: 'onHemisphere',
+
+  BERNOULLI: 'bernoulli',
+
+  SAMPLE: 'sample',
+  EXPONENTIAL: 'exponential',
+  NORMAL: 'normal',
+  CAUCHY: 'cauchy',
+} as const;
```
aleksanderkatan, line 56: Is there a simple way to have scrollbars in our select controls? Right now, someone might see this list and think that these are all available options.

image

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,136 @@
...
+  readonly #dataSingleWorkerFunc: TgpuComputeFn;
+  readonly #dataMoreWorkersFunc: TgpuComputeFn;
+  readonly #distributionSlot: TgpuSlot<TgpuFn<() => d.Vec3f>>;
+  readonly #sampleBufferSlot;
+
+  constructor(root: TgpuRoot, count: number) {
+    console.assert(
+      count > 0,
+      'Count cannot be 0. Cannot create buffer of size 0',
+    );
```
aleksanderkatan: Maybe an assert for max size here as well? Also, I'm not sure, but I think the buffer would itself throw an error, so this may be unnecessary.

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,136 @@
...
+        'storage',
+      );
+    this.#samples = [];
+  }
+
+  get count() {
+    return this.#count;
+  }
+
+  async executeSingleWorker(
```
aleksanderkatan: Do we need the single execution mode? With numbers like these, a single worker manages just fine so there is no visible advantage from switching to parallel for the user. Also, there is no animation when switching between single and parallel.

apps/typegpu-docs/src/content/examples/simple/probability/index.ts
```diff
@@ -0,0 +1,124 @@
...
+});
+canvas.addEventListener('touchend', () => {
+  helpInfo.style.opacity = '1';
+});
+
+export const controls = {
+  'Reset Camera': {
+    onButtonClick: () => {
+      plotter.resetView(getCameraPosition(currentDistribution));
+    },
```
aleksanderkatan, line 61: nit: we can simplify one-line arrow functions
```suggestion
    onButtonClick: () =>
      plotter.resetView(getCameraPosition(currentDistribution)),
```

apps/typegpu-docs/src/content/examples/simple/probability/plotter.ts
```diff
@@ -0,0 +1,436 @@
...
+    // if we want to animate the transition between the old and new plots,
+    // we need to have the same number of samples,
+    // that's why I clamp them instead of filtering
+    const samplesFiltered = samples.map((sample) => ({
+      ...sample,
+      x: sample.x < defaultMinValue
+        ? defaultMinValue
+        : sample.x > defaultMaxValue
+        ? defaultMaxValue
+        : sample.x,
```
aleksanderkatan:
```suggestion
      x: clamp(sample.x, defaultMinValue, defaultMaxValue),
```

Also, we usually import `* as std` instead of importing functions directly.

apps/typegpu-docs/src/content/examples/simple/probability/index.ts
```diff
@@ -0,0 +1,124 @@
...
+canvas.addEventListener('mouseover', () => {
+  helpInfo.style.opacity = '0';
+});
+canvas.addEventListener('mouseout', () => {
+  helpInfo.style.opacity = '1';
+});
+// handle mobile devices
+canvas.addEventListener('touchstart', () => {
+  helpInfo.style.opacity = '0';
+});
```
aleksanderkatan: This gets rid of a warn in the logs.
```suggestion
}, { passive: true });
```

aleksanderkatan (approved):
Great example! 🎲⚗️
I left some comments, mainly nits.

apps/typegpu-docs/src/content/examples/simple/probability/index.html
cieplypolar: I will add 5s cooldown.

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
cieplypolar: hmm, I will think about it, maybe I will get the seed from the CPU

apps/typegpu-docs/src/content/docs/ecosystem/typegpu-noise.mdx
cieplypolar: $\LaTeX$ uses \mu as $\mu$.  I wanted to make clear, that I am referring to parameters' names.

packages/typegpu-noise/src/random.ts
cieplypolar: I think this function works just fine, if we can tell, that the `length != 0`, then elements of the vector certainly are not all 0.

packages/typegpu-noise/src/random.ts
cieplypolar: You are correct. However, I really don't want to put a while loop inside. Also, I'm not sure if rescaling can solve this issue, I think either way we must clamp something :(.

apps/typegpu-docs/src/content/examples/simple/probability/types.ts
cieplypolar: Unfortunately, `overflow-auto` doesn't do anything

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
cieplypolar: You are right, I should focus on sampling from multiple threads.

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
cieplypolar: I will add assertion for max size.

apps/typegpu-docs/src/content/examples/simple/probability/types.ts
aleksanderkatan: Then let's not worry about

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  constructor(root: TgpuRoot, count: number) {
+    console.assert(
+      count > 0,
+      'Count cannot be 0. Cannot create buffer of size 0',
+    );
+    console.assert(
+      count <= 65536,
+      'Count cannot exceed max workgroup grid X dimension size',
+    );
```
reczkok: I would merge those - also what will happen if we are outside of the range? This will only log and I assume some time of runtime error will happen. So I would either throw or just remove this (since we control what will be passed here anyway).
```suggestion
    console.assert(
      count > 0 && count <= 65536,
      'Count must be in range (0, 65536]',
    );
```

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,105 @@
...
+      .$usage(
+        'storage',
+      );
+    this.#seedBuffer = this.#root.createBuffer(
+      d.arrayOf(d.f32, count),
+      Array.from({ length: count }, () => Math.random()),
+    )
+      .$usage(
+        'storage',
+      );
```
reczkok:
```suggestion
    this.#samplesBuffer = this.#root
      .createBuffer(d.arrayOf(d.vec3f, count))
      .$usage('storage');
    this.#seedBuffer = this.#root
      .createBuffer(
        d.arrayOf(d.f32, count),
        Array.from({ length: count }, () => Math.random()),
      ).$usage('storage');
```

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,105 @@
...
+      this.#seedBuffer.as('mutable'),
+    );
+
+    this.#sampleBufferSlot = sampleBufferSlotTempAlias;
+    this.#distributionSlot = distributionSlotTempAlias;
+    this.#seedBufferSlot = seedBufferSlotTempAlias;
+
+    this.#dataMoreWorkersFunc = tgpu['~unstable'].computeFn({
+      in: { gid: d.builtin.globalInvocationId, count: d.builtin.numWorkgroups },
+      workgroupSize: [1],
```
reczkok: We could bump this (and dispatch less workgroups) but the time it takes is so insignificant I cant even time it properly
```suggestion
      workgroupSize: [32],
```

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,105 @@
...
+      sampleBufferSlotTempAlias.$[gid.x] = distributionSlotTempAlias.$();
+    });
+  }
+
+  set count(value: number) {
+    this.#count = value;
+    this.#samplesBuffer = this.#root.createBuffer(d.arrayOf(d.vec3f, value))
+      .$usage(
+        'storage',
+      );
```
reczkok:
```suggestion
    this.#samplesBuffer = this.#root
      .createBuffer(d.arrayOf(d.vec3f, value))
      .$usage('storage');
```

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  async executeMoreWorkers(
+    distribution: TgpuFn<() => d.Vec3f>,
+  ): Promise<d.v3f[]> {
+    const pipeline = this.#root['~unstable']
+      .with(this.#sampleBufferSlot, this.#samplesBuffer.as('mutable'))
+      .with(this.#seedBufferSlot, this.#seedBuffer.as('mutable'))
+      .with(this.#distributionSlot, distribution)
+      .withCompute(this.#dataMoreWorkersFunc as TgpuComputeFn)
+      .createPipeline();
```
reczkok: I think recreating the pipeline on distribution change (+cache) + moving the fixed usages to a bind group that only changes when we change the count is the way to go

cieplypolar:
TODO (in executor.ts):
- [ ] replace buffer slots with bind group layouts
- [ ] cache pipelines

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
cieplypolar: I put it here for myself to remember, that I operate only in 1D. Since this is just an example, I will remove it.

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,134 @@
...
+      Array.from({ length: this.#count }, () => Math.random()),
+    );
+  }
+
+  set count(value: number) {
+    this.#count = value;
+    if (this.#bufferCache.has(value)) {
+      // biome-ignore lint/style/noNonNullAssertion: just checked it above
+      [this.#samplesBuffer, this.#seedBuffer] = this.#bufferCache.get(value)!;
+    } else {
```
reczkok: I would do the same for pipeline cache
```suggestion
    const cacheEntry = this.#bufferCache.get(value);
    if (cacheEntry) {
      [this.#samplesBuffer, this.#seedBuffer] = cacheEntry;
    } else {
```

apps/typegpu-docs/src/content/examples/simple/probability/executor.ts
```diff
@@ -0,0 +1,134 @@
...
+
+  cachedPipeline(distribution: TgpuFn<() => d.Vec3f>) {
+    if (!import.meta.env.DEV) {
+      throw new Error('Function only for testing purposes');
+    }
+
+    if (!this.#pipelineCache.has(distribution)) {
+      const pipeline = this.#root['~unstable']
+        .with(this.#distributionSlot, distribution)
+        .withCompute(this.#dataMoreWorkersFunc as TgpuComputeFn)
```
iwoplaza: Is this type assertion required?

apps/typegpu-docs/src/content/examples/simple/probability/index.ts
```diff
@@ -0,0 +1,111 @@
...
+  },
+  'Reseed': {
+    onButtonClick: async () => {
+      executor.reseed();
+      await replot(
+        currentDistribution,
+        true,
+      );
+      plotter.resetView(getCameraPosition(currentDistribution));
+    },
```
iwoplaza: If we have a block of statements, I would prefer we use non-arrow functions. Also, could you check if the replot call can be a one-liner like so?
```suggestion
    async onButtonClick() {
      executor.reseed();
      await replot(currentDistribution, true);
      plotter.resetView(getCameraPosition(currentDistribution));
    },
```

apps/typegpu-docs/src/examples/simple/probability/index.ts
```diff
@@ -0,0 +1,108 @@
...
+    helpInfo.style.opacity = '1';
+  }, c.popupCooldown));
+// handle mobile devices
+canvas.addEventListener('touchstart', () => {
+  helpInfo.style.opacity = '0';
+}, { passive: true });
+canvas.addEventListener('touchend', () =>
+  setTimeout(() => {
+    helpInfo.style.opacity = '1';
+  }, c.popupCooldown));
```
iwoplaza: Tiny nit, could be more readable

```suggestion
canvas.addEventListener('mouseout', () => {
  setTimeout(() => {
    helpInfo.style.opacity = '1';
  }, c.popupCooldown);
});

// handle mobile devices

canvas.addEventListener('touchstart', () => {
  helpInfo.style.opacity = '0';
}, { passive: true });

canvas.addEventListener('touchend', () => {
  setTimeout(() => {
    helpInfo.style.opacity = '1';
  }, c.popupCooldown);
});
```

iwoplaza (approved):

PR #1606, opened 2025-08-13
refactor: Generation helpers

iwoplaza:


packages/typegpu/src/data/numeric.ts
```diff
@@ -144,7 +144,14 @@ const f32Cast = createDualImpl(
...
   // GPU implementation
-  (v) => snip(stitch`f32(${v})`, f32),
+  (arg): Snippet => {
+    if (!arg) return snip(0, f32);
+    if (typeof arg.value === 'number' || typeof arg.value === 'boolean') {
+      // Precomputing...
+      return snip(f32Cast[$internal].jsImpl(arg.value), f32);
+    }
+    return snip(stitch`f32(${arg})`, f32);
+  },
```
iwoplaza: Just did this for `f32` for now, but hopefully with my planned changes, this will be automagic for all dual impls in a future rework 🤞

packages/typegpu/src/std/operators.ts
```diff
@@ -75,11 +58,16 @@ export const add = createDualImpl(
...
-    const resultType = isSnippetNumeric(convLhs)
-      ? convRhs.dataType
-      : convLhs.dataType;
+  (...args) => {
+    const ctx = getResolutionCtx() as ResolutionCtx;
+    const { converted, commonType } = convertToCommonType({
+      ctx,
+      values: args,
+    });
+    const [lhs, rhs] = converted;
```
reczkok: I think this appears enough times to justify the helper

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -222,7 +222,12 @@ function createFn<ImplSchema extends AnyFn>(
             schemaCallWrapper(shell.argTypes[index] as unknown as AnyData, arg)
           ) as InferArgs<Parameters<ImplSchema>>;
 
-          return implementation(...castAndCopiedArgs);
+          const result = implementation(...castAndCopiedArgs);
+          // Casting the result to the appropriate schema
+          if (shell.returnType.type === 'void') {
+            return result;
+          }
+          return schemaCallWrapper(shell.returnType, result);
```
reczkok: Maybe we could move the responsibility away from the function by changing the schemaCallWrapper. We could make it return identity in case the schema is not callable (we also wouldn't need to throw then):
```ts
export function schemaCallWrapper<T>(schema: AnyData, item?: T): T {
  const maybeType = (schema as { type: string })?.type;

  if (typeof schema !== 'function' && !(maybeType in formatToWGSLType)) {
    return item as T;
  }

  const callSchema =
    (maybeType in formatToWGSLType
      ? formatToWGSLType[maybeType as keyof typeof formatToWGSLType]
      : schema) as unknown as ((item?: T) => T);

  if (item === undefined) {
    return callSchema();
  }
  return callSchema(item);
}
```

packages/typegpu/src/data/numeric.ts
```diff
@@ -134,7 +134,7 @@ export const i32: I32 = Object.assign(i32Cast, {
 
 const f32Cast = createDualImpl(
   // CPU implementation
-  (v?: number | boolean) => {
+  (v?: number | boolean | undefined) => {
```
reczkok: Do we want this? I think explicitly passing undefined is weird
```suggestion
  (v?: number | boolean) => {
```

packages/typegpu/src/data/numeric.ts
```diff
@@ -144,7 +144,14 @@ const f32Cast = createDualImpl(
     return Math.fround(v);
   },
   // GPU implementation
-  (v) => snip(stitch`f32(${v})`, f32),
+  (arg): Snippet => {
+    if (!arg) return snip(0, f32);
```
reczkok: Just to be consistent with our convention (which i'm happy to move away from btw)
```suggestion
    if (!arg) {
      return snip(0, f32);
    }
```

packages/typegpu/src/std/operators.ts
```diff
@@ -93,7 +73,7 @@ export const add = createDualImpl(
       return snip(cpuAdd(lhs.value as never, rhs.value as never), resultType);
```
reczkok, line 73: If we added a more restrictive instance check:
```ts
export function isNumericVecInstance(
  value: unknown,
): value is AnyNumericVecInstance {
  return isVecInstance(value) && !value.kind.endsWith('<bool>');
}
``` 
We could get rid of those casts (same in tho other ops)

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+      ...(detail.action === 'cast' && {
+        targetType: detail.targetType as U32 | F32 | I32 | F16,
+      }),
+    }),
+  );
+
+  const hasCasts = actions.some((action) => action.action === 'cast');
+
+  return { targetType: bestType, actions, hasImplicitConversions: hasCasts };
+}
```
reczkok: Now that I look at it there is absolutely no reason for this to be a map based on usage. Here is a cleaned up version:
```suggestion
function findBestType(
  types: AnyData[],
  uniqueTypes: AnyData[],
  allowImplicit: boolean,
): ConversionResult | undefined {
  let bestResult:
    | { type: AnyData; details: ConversionRankInfo[]; sum: number }
    | undefined;

  for (const targetType of uniqueTypes) {
    const details: ConversionRankInfo[] = [];
    let sum = 0;

    for (const sourceType of types) {
      const conversion = getConversionRank(
        sourceType,
        targetType,
        allowImplicit,
      );
      if (conversion.rank === Number.POSITIVE_INFINITY) {
        sum = Number.POSITIVE_INFINITY;
        break;
      }
      sum += conversion.rank;
      details.push(conversion);
    }

    if (sum < (bestResult?.sum ?? Number.POSITIVE_INFINITY)) {
      bestResult = { type: targetType, details, sum };
    }
  }

  if (!bestResult) {
    return undefined;
  }

  const actions: ConversionResultAction[] = bestResult.details.map((
    detail,
    index,
  ) => ({
    sourceIndex: index,
    action: detail.action,
    ...(detail.action === 'cast' && {
      targetType: detail.targetType as U32 | F32 | I32 | F16,
    }),
  }));

  return {
    targetType: bestResult.type,
    actions,
    hasImplicitConversions: actions.some((action) => action.action === 'cast'),
  };
}
```

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+export function getBestConversion(
+  types: AnyData[],
+  targetTypes?: AnyData[],
+): ConversionResult | undefined {
+  if (types.length === 0) return undefined;
+
+  const uniqueTypes = [...new Set(types.map(undecorate))];
+  const uniqueTargetTypes = targetTypes
+    ? [...new Set(targetTypes.map(undecorate))]
+    : uniqueTypes;
```
reczkok: (I know that i'm nitpicking my own code but it hurts😆)
```suggestion
  const uniqueTargetTypes = [
    ...new Set((targetTypes || types).map(undecorate)),
  ];
```

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+  const explicitResult = findBestType(types, uniqueTargetTypes, false);
+  if (explicitResult) {
+    return explicitResult;
+  }
+
+  const implicitResult = findBestType(types, uniqueTargetTypes, true);
+  if (implicitResult) {
+    implicitResult.hasImplicitConversions = implicitResult.actions.some(
+      (action) => action.action === 'cast',
+    );
```
reczkok: `findBestType` already handles that so we can clean that up
```suggestion
```

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -0,0 +1,369 @@
...
+    }
+    return {
+      targetType: undecorate(targetType),
+      actions: [actionDetail],
+      hasImplicitConversions: conversion.action === 'cast',
+    };
+  }
+
+  return undefined;
+}
```
reczkok: This function is unused currently (we always use `convertToCommonType` in its place)

packages/typegpu/src/types.ts
```diff
@@ -229,7 +229,19 @@ export interface ResolutionCtx {
...
 
-  resolve(item: unknown, schema?: AnyData | UnknownData | undefined): string;
+  /**
+   * Returns the WGSL code representing `item`.
+   *
+   * @param item The value to resolve
+   * @param schema Additional information about the item's data type
+   * @param exact Should the inferred value of the resulting code be typed exactly as `schema` (true),
+   *              or is being assignable to `schema` enough (false). Default is false.
+   */
```
reczkok, line 239: Do we actually ensure compatibility if false? To me, it looks more like this forces the type to be the schema regardless of anything.

packages/typegpu/tests/tgsl/conversion.test.ts
```diff
@@ -0,0 +1,414 @@
...
+  pre: '',
+} as unknown as GenerationCtx;
+
+beforeEach(() => {
+  INTERNAL_setCtx(mockCtx);
+});
+
+afterEach(() => {
+  INTERNAL_setCtx(undefined);
+});
```
reczkok: I think it's effectively the same
```suggestion
beforeAll(() => {
  INTERNAL_setCtx(mockCtx);
});
```

reczkok (approved):
Great rework! Why only do f32 now, won’t the rest of the logic be analogous?

packages/typegpu/src/data/numeric.ts
iwoplaza: Yup, working automagically in https://github.com/software-mansion/TypeGPU/pull/1609

packages/typegpu/src/std/operators.ts
iwoplaza: I guess so! I am removing these manual precomputations in the next PR though, so these changes would be removed anyway.

packages/typegpu/src/types.ts
iwoplaza: > Do we actually ensure compatibility if false?

I believe so! If not, then we definitely should. Up until this point, when returning snippets from std functions, we were returning values that were guaranteed to embody their type _exactly_, but that's overkill for most cases. Apart from variable assignments and the division operator (probably a handful more), a value that is assignable to `type` is sufficient.

For example, after the changes in this PR, we can expect the following behavior:
```ts
ctx.resolve(123, d.f32,  /* exact */ false) // 123
ctx.resolve(123, d.f32,  /* exact */ true)  // 123f

ctx.resolve(0, d.i32,  /* exact */ false) // 0
ctx.resolve(0, d.i32,  /* exact */ true)  // 0i
```

The `exact` properly only really has an effect on numeric literals that have abstract types assignable to more concrete types, as all other WGSL types seem to be nominal.

reczkok: Yes, I agree that when `exact` is passed we make sure that the numerics are correctly suffixed. What I'm saying is that we do not ensure that the resolved value is assignable if we omit the flag. For example:
```ts
ctx.resolve(1.23, d.i32,  /* exact */ false) // 1.23
```
`1.23` is not _assignable_ to `i32` and I don't think we will do anything about it inside `resolve`.

iwoplaza: That's true, the property of assignability only applies to snippets, and during resolution, we trust that the value we pass is compatible.

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -677,21 +698,19 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
-          this.resolve(element, schema.elementType as AnyData)
-        )
+      return stitch`array<${elementTypeString}, ${schema.elementCount}>(${
+        item.map((element) => snip(element, schema.elementType as AnyData))
       })`;
     }
 
     if (Array.isArray(item)) {
-      return `array(${item.map((element) => this.resolve(element))})`;
+      return stitch`array(${item.map((element) => this.resolve(element))})`;
```
aleksanderkatan, line 707:
```suggestion
      return stitch`array(${item})`;
```
`stitch` resolves anyway, right? Can we make `stitch` signature more general so that we don't need to call resolve multiple times in other places?

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -594,19 +597,16 @@ export function generateStatement(
...
-    const returnValue = returnNode !== undefined
-      ? ctx.resolve(
-        generateTypedExpression(
-          ctx,
-          returnNode,
-          ctx.topFunctionReturnType,
-        ).value,
-      )
-      : undefined;
+    if (returnNode) {
```
reczkok: Since the return node can be false (bool) we need to be more specific
```suggestion
    if (returnNode !== undefined) {
```

packages/typegpu/src/resolutionCtx.ts
iwoplaza: We can definitely think about this. We can actually try to change things around and allow ONLY snippets to be resolved (so `resolve(snippet, exact)`), but that would require a more thorough refactor. Let's shelve this thought for later.

PR #1611, opened 2025-08-17
docs: Using rolldown in translator app

iwoplaza:


packages/unplugin-typegpu/src/rolldown-browser.ts
```diff
@@ -0,0 +1,8 @@
+import type { Options } from './common.ts';
+import { rollUpImpl } from './rollup-impl.ts';
+
+export default (options: Options) => {
+  // The unplugin API is based on the rollup/rolldown APIs, so it
+  // should just be a compatible rolldown plugin.
+  return rollUpImpl(options);
```
iwoplaza: Just return rollup Impl

packages/unplugin-typegpu/src/common.ts
aleksanderkatan, line 1: Why were these moved here specifically? From what I understand, in this file there are functions used both in rollup and in babel, and those are only used in rollup.

aleksanderkatan (commented):
LGTM!

packages/unplugin-typegpu/src/common.ts
iwoplaza: You're absolutely right, I'll move them where they belong

PR #1613, opened 2025-08-18
fix: Warn when external was omitted (postmortem)

aleksanderkatan:


packages/typegpu/tests/resolve.test.ts
```diff
@@ -39,11 +39,8 @@ describe('tgpu resolve', () => {
       },
       names: 'strict',
     });
-    expect(parse(resolved)).toBe(
-      parse(`
-          fn foo() { 
-            var g = 1000; 
-          }`),
+    expect(parse(resolved)).toMatchInlineSnapshot(
```
iwoplaza: If we omit `parse()`, then we'll get the actual result and snapshot that. In this case, it's going to be a matter of representing whitespace accurately.
```suggestion
    expect(resolved).toMatchInlineSnapshot(
```

packages/typegpu/tests/resolve.test.ts
```diff
@@ -597,20 +578,14 @@ describe('tgpu resolveWithContext', () => {
...
+      template: 'fn testFn() { return _EXT_.n + _EXT_.n; }',
       externals: { _EXT_: { n: 100 } },
-      names: 'strict',
     });
 
-    expect(parse(resolved.code)).toBe(parse(`
-      fn testFn() {
-        return 100 + 100;
-      }`));
+    expect(parse(resolved)).toMatchInlineSnapshot(
```
iwoplaza:
```suggestion
    expect(resolved).toMatchInlineSnapshot(
```

PR #1630, opened 2025-08-20
feat: Dispatch method

aleksanderkatan:
Changes:
- add `prepareDispatch` method that creates a compute pipeline and returns a dispatch function that runs it,
- add "Utilities" docs page describing the `dispatch` method,
- add an example testing whether dispatch runs correctly (it can be deleted if we don't deem it necessary),
- rewrite 3D Fish to use the dispatch for randomising fish positions.

### DIFF
--- a/apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
@@ -56,7 +56,7 @@ doubleUp(8);
 doubleUp(8);
 doubleUp(4);
 
-// no need to await `doubleUp()` because the command encoder
+// no need to call `onSubmittedWorkDone()` because the command encoder
 // will queue the read after `doubleUp` anyway
 console.log(await data.read()); // [0, 8, 16, 24, 16, 20, 24, 28]
 ```
--- a/apps/typegpu-docs/src/content/examples/tests/dispatch/index.ts
+++ b/apps/typegpu-docs/src/content/examples/tests/dispatch/index.ts
@@ -108,11 +108,7 @@ if (!table) {
   throw new Error('Nowhere to display the results');
 }
 runTests().then((result) => {
-  if (result) {
-    table.innerText = 'Tests succeeded!';
-  } else {
-    table.innerText = 'Tests failed.';
-  }
+  table.innerText = `Tests ${result ? 'succeeded' : 'failed'}.`;
 });
 
 // #region Example controls and cleanup
--- a/apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/index.ts
+++ b/apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/index.ts
@@ -39,7 +39,7 @@ if (!table) {
   throw new Error('Nowhere to display the results');
 }
 runTests().then((result) => {
-  table.innerText = result ? 'Tests succeeded!' : 'Tests failed.';
+  table.innerText = `Tests ${result ? 'succeeded' : 'failed'}.`;
 });
 
 // #region Example controls and cleanup


apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,72 @@
...
+
+It is highly recommended NOT to use `dispatch` for:
+
+- Kernel calls that occur multiple times, for example on every frame. Every
+  `dispatch` call recreates the pipeline, which may impact the performance of
+  your app significantly. Create and reuse a pipeline instead.
+
+- Avoidable small calls. For example, small buffers can be initialized with
+  `buffer.write()` method, or (when using TgpuBuffers) by passing the data to
+  the constructor.
```
reczkok: I would explain why that is (that for small data the shader creation and dispatch is more costly than serialization). Otherwise the user won't know what `avoidable` refers to.

apps/typegpu-docs/src/content/examples/rendering/3d-fish/index.ts
```diff
@@ -95,28 +96,32 @@ function enqueuePresetChanges() {
...
-    }),
-  );
-  fishDataBuffers[0].write(positions);
-  fishDataBuffers[1].write(positions);
+    });
+    buffer0mutable.$[x] = data;
+    buffer1mutable.$[x] = data;
+  });
   enqueuePresetChanges();
 };
```
reczkok: Unfortunately this is slower than the cpu counterpart but not very significantly so we could leave it as a showcase

packages/typegpu/src/core/root/dispatch.ts
```diff
@@ -0,0 +1,56 @@
...
+      return;
+    }
+    wrappedCallback(id.x, id.y, id.z);
+  });
+
+  root['~unstable']
+    .withCompute(mainCompute)
+    .createPipeline()
+    .dispatchWorkgroups(...workgroupCount);
+}
```
reczkok: We could do this all using vectors and std but I'm not sure if that's even cleaner
```suggestion
import { builtin } from '../../builtin.ts';
import { u32 } from '../../data/numeric.ts';
import { vec3f, vec3u } from '../../data/vector.ts';
import type { v3u } from '../../data/wgslTypes.ts';
import { any, ge } from '../../std/boolean.ts';
import { ceil } from '../../std/numeric.ts';
import { computeFn } from '../function/tgpuComputeFn.ts';
import { fn } from '../function/tgpuFn.ts';
import type { TgpuRoot } from './rootTypes.ts';

/**
 * Changes the given array to an array of 3 numbers, filling missing values with 1.
 */
function sanitizeArray(arr: readonly number[]): v3u {
  return vec3u(arr[0] ?? 1, arr[1] ?? 1, arr[2] ?? 1);
}

/**
 * Dispatch a single-shot compute pipeline.
 * @param root A TgpuRoot.
 * @param size A 3d (or shorter) array holding the total number of threads to run.
 * @param callback A function that is parsed to WGSL and run on GPU. Its arguments are the global invocation ids of the call.
 * @param workgroupSize (optional) A 3d (or shorter) array holding the sizes of the workgroups. [1, 1, 1] by default.
 */
export function dispatch(
  root: TgpuRoot,
  size: readonly number[],
  callback: (x: number, y: number, z: number) => void,
  workgroupSize?: readonly number[],
): void {
  const checkedSize = sanitizeArray(size);
  const checkedWorkgroupSize = sanitizeArray(workgroupSize ?? []);
  const workgroupCount = ceil(
    vec3f(checkedSize).div(vec3f(checkedWorkgroupSize)),
  );

  const wrappedCallback = fn([u32, u32, u32])(callback);

  const mainCompute = computeFn({
    workgroupSize: checkedWorkgroupSize,
    in: { id: builtin.globalInvocationId },
  })(({ id }) => {
    'kernel';
    if (any(ge(id, checkedSize))) {
      return;
    }
    wrappedCallback(id.x, id.y, id.z);
  });

  root['~unstable']
    .withCompute(mainCompute)
    .createPipeline()
    .dispatchWorkgroups(workgroupCount.x, workgroupCount.y, workgroupCount.z);
}
```

packages/typegpu/src/core/root/init.ts
```diff
@@ -651,6 +652,14 @@ class TgpuRootImpl extends WithBindingImpl
     this.device.queue.submit([this._commandEncoder.finish()]);
     this._commandEncoder = null;
   }
+
+  dispatch(
+    size: readonly number[],
+    callback: (x: number, y: number, z: number) => void,
+    workgroupSize?: readonly number[],
```
reczkok: We could consider making the parameter an object - I would like to define my workgroup size right next to the dispatch size. What do you think?

reczkok (approved):
Love the implementation! 🙇 Left some comments to consider (most of them are stylistic considerations)

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,72 @@
...
+  const boidData = Boid({ index: x, pos: d.vec3f() });
+  boidsMutable.$[x] = boidData;
+});
+
+// (optional) wait for the dispatch to finish
+await root.device.queue.onSubmittedWorkDone();
+```
+
+Buffer initialization commonly uses random number generators. 
+For that, you can use the `@typegpu/noise` library.
```
cieplypolar:
```suggestion
For that, you can use the [`@typegpu/noise`](TypeGPU/ecosystem/typegpu-noise) library.
```

apps/typegpu-docs/src/content/examples/rendering/3d-fish/index.ts
```diff
@@ -95,28 +96,32 @@ function enqueuePresetChanges() {
 }
 
 const randomizeFishPositions = () => {
-  const positions: d.Infer<typeof ModelData>[] = Array.from(
-    { length: p.fishAmount },
-    () => ({
+  const buffer0mutable = fishDataBuffers[0].as('mutable');
+  const buffer1mutable = fishDataBuffers[1].as('mutable');
+  const now = performance.now();
```
cieplypolar:
```suggestion
  const now = performance.now() % 10000;
```
Suppose that the user stared at the fish and after a long time clicked randomize. All of the fish will collapse to 1 point.

packages/typegpu/src/core/root/init.ts
cieplypolar: I think it would be nice to allow users to configure `dispatch` such that 1 thread will handle more than 1 index. Now you can think of `dispatch` as a cuboid built from $1\times1\times1$ cubes. I propose that we allow user to define size of these cubes. This can be done in another PR along with benchmarks if this is faster.

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,72 @@
...
+// buffer of 1024x1024 floats
+const waterLevelMutable = root.createMutable(
+  d.arrayOf(d.arrayOf(d.f32, 1024), 1024),
+);
+
+root['~unstable'].dispatch([1024, 1024], (x, y) => {
+  'kernel';
+  randf.seed((x * 1024 + y) / 1024 ** 2);
+  waterLevelMutable.$[x][y] = randf.sample();
+}, [16, 16]);
```
cieplypolar: I think we should write something more about the `workgroupSize` parameter, it was not clear to me what it is from the `dispatch` call above.

packages/typegpu/src/core/root/dispatch.ts
cieplypolar: Line with `ceil` looks better with vector operations to me.

packages/typegpu/src/core/root/init.ts
cieplypolar: I agree

cieplypolar (commented):
Really convenient feature! 🦖

packages/typegpu/src/core/root/init.ts
aleksanderkatan: I would argue that would be the case where an user should create their own pipeline, but we can think about it

apps/typegpu-docs/src/content/examples/rendering/3d-fish/index.ts
aleksanderkatan: I left it so that dispatch is used in at least one non-test example (as test examples are soon to be hidden from users).

packages/typegpu/src/core/root/init.ts
aleksanderkatan: Done

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
aleksanderkatan: I added a sentence more, would you add more? I guess we could add a link to some docs explaining workgroup size if it is still not enough

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,113 @@
...
+title: Utilities
+description: A list of various utilities provided by TypeGPU.
+---
+
+## Dispatch
+
+The `prepareDispatch` function simplifies running simple computations on the GPU.
+Behind the scenes, `prepareDispatch` wraps the callback into a `TgpuFn`, then it creates a compute pipeline and returns a dispatch function that runs the pipeline.
+
+This might help omit the serialization overhead for filling out buffers with initial data.
```
reczkok:
```suggestion
The `prepareDispatch` function simplifies running simple computations on the GPU.
Under the hood, it wraps the callback in a `TgpuFn`, creates a compute pipeline, and returns a dispatch function to execute it.

This can help reduce serialization overhead when initializing buffers with data.
```

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+dispatch().then(() => console.log("Dispatch finished!"));
+```
+
+:::note
+Remember to mark the callback with `'kernel'` directive to let TypeGPU know that this function is TGSL.
+:::
+
+The returned dispatch function can be called multiple times.
+The pipeline is reused, so there is no extra overhead of successive calls.
```
reczkok:
```suggestion
The returned dispatch function can be called multiple times.
Since the pipeline is reused, there’s no additional overhead for subsequent calls.
```

apps/typegpu-docs/src/content/examples/rendering/3d-fish/index.ts
```diff
@@ -94,29 +95,40 @@ function enqueuePresetChanges() {
...
       applySeaDesaturation: 1,
-    }),
-  );
-  fishDataBuffers[0].write(positions);
-  fishDataBuffers[1].write(positions);
+    });
+    buffer0mutable.$[x] = data;
+    buffer1mutable.$[x] = data;
+  },
+});
```
reczkok: Let's bump the workgroup size to something higher than `1`. I would recommend `32`.

packages/typegpu/src/prepareDispatch.ts
```diff
@@ -0,0 +1,84 @@
...
+import { u32 } from './data/numeric.ts';
+import { vec3f, vec3u } from './data/vector.ts';
+import type { v3u } from './data/wgslTypes.ts';
+import { any, ge } from './std/boolean.ts';
+import { ceil } from './std/numeric.ts';
+
+/**
+ * Changes the given array to a vec of 3 numbers, filling missing values with 1.
+ */
+function sanitizeArray(arr: readonly number[]): v3u {
```
reczkok: I don't think this name is very fitting since there is no much sanitization going on. Maybe something like:
```suggestion
function toVec3(arr: readonly number[]): v3u {
```
Also I just noticed we are not safeguarding against someone passing in `0`.

packages/typegpu/src/prepareDispatch.ts
```diff
@@ -0,0 +1,84 @@
...
+ * Call the returned dispatch function to run the GPU computations.
+ * The returned dispatch function can be called multiple times.
+ * The returned dispatch function returns a promise that resolves when the device queue finishes its jobs.
+ * @param options.root A TgpuRoot.
+ * @param options.callback A function that is parsed to WGSL and run on GPU. Its arguments are the global invocation ids of the call.
+ * @param options.size A 3d (or shorter) array holding the total number of threads to run.
+ * @param options.workgroupSize (optional) A 3d (or shorter) array holding the sizes of the groups of threads.
+ * [1, 1, 1] by default. Setting this to bigger values might speed up the computation due to better caching.
+ * The callback is not called when the gid would exceed option.size.
+ */
```
reczkok:
```suggestion
/**
 * Creates a dispatch function for a compute pipeline.
 * 
 * The returned function can be called multiple times to run GPU computations.
 * It returns a promise that resolves when the device queue completes its work.
 * 
 * @param options.root A TgpuRoot instance.
 * @param options.callback A function converted to WGSL and executed on the GPU. Its arguments correspond to the global invocation IDs.
 * @param options.size A 3D (or shorter) array specifying the total number of threads to run.
 * @param options.workgroupSize (optional) A 3D (or shorter) array specifying the workgroup dimensions. Defaults to [1, 1, 1].
 * 
 * The callback is not invoked for any invocation IDs that exceed `options.size`.
 */
```

reczkok (approved):
Looks great! Left mostly stylistic nits that are safe to ignore if you don't like them.

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
aleksanderkatan: Nice, thanks! I should have run those through gpt or something.

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,113 @@
...
+  size: [2048] as const,
+  callback: (x) => {
+    // TGSL function body to run 2048 times on the GPU
+    'kernel';
+    const boidData = Boid({ index: x, pos: d.vec3f() });
+    boidsMutable.$[x] = boidData;
+  },
+});
+
+dispatch().then(() => console.log("Dispatch finished!"));
```
iwoplaza: Since the amount of threads can change without having to recompile, I would consider splitting those two.

```ts
await dispatch(root, (x) => {
  'kernel';
  const boidData = Boid({ index: x, pos: d.vec3f() });
  boidsMutable.$[x] = boidData;
})(2048);

console.log("Dispatch finished!");
```

or:

```ts
const updateBoids = dispatch(root, (x) => {
  'kernel';
  const boidData = Boid({ index: x, pos: d.vec3f() });
  boidsMutable.$[x] = boidData;
});

await updateBoids(2048);
console.log("Dispatch finished!");
```

I guess we could determine the dimensionality based on the number of arguments in the callback. We would have to expose an API to retrieve that information though.
Thoughts @aleksanderkatan @cieplypolar @reczkok @lursz ?

aleksanderkatan: The api is now exactly like proposed, modulo the function name (`prepareDispatch`)

packages/typegpu/src/prepareDispatch.ts
```diff
@@ -0,0 +1,87 @@
...
+    const sanitizedSize = toVec3(size);
+    const workgroupCount = ceil(vec3f(sanitizedSize).div(vec3f(workgroupSize)));
+    sizeUniform.write(sanitizedSize);
+    pipeline.dispatchWorkgroups(
+      workgroupCount.x,
+      workgroupCount.y,
+      workgroupCount.z,
+    );
+    root['~unstable'].flush();
+    return root.device.queue.onSubmittedWorkDone();
```
reczkok: I think we can skip returning this promise. I see no feasible use case for awaiting the work to be done apart from timing (which the use can still do by calling this on his own). I think returning the promise will make users await it which is always detrimental.

aleksanderkatan: Done 👌

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -0,0 +1,111 @@
...
+const doubleUp = prepareDispatch(root, (x) => {
+  'kernel';
+  data.$[x] *= 2;
+});
+
+doubleUp(8);
+doubleUp(8);
+doubleUp(4);
+
+// no need to await `doubleUp()` because the command encoder
```
reczkok: Now even `no way to await`😄
```suggestion
// no need to call `onSubmittedWorkDone()` because the command encoder
```
we could also just remove this comment

apps/typegpu-docs/src/content/examples/tests/dispatch/index.ts
```diff
@@ -0,0 +1,124 @@
...
+const table = document.querySelector<HTMLDivElement>('.result');
+if (!table) {
+  throw new Error('Nowhere to display the results');
+}
+runTests().then((result) => {
+  if (result) {
+    table.innerText = 'Tests succeeded!';
+  } else {
+    table.innerText = 'Tests failed.';
+  }
```
reczkok: nit
```suggestion
  table.innerText = `Tests ${result ? 'succeeded' : 'failed'}.`;
```

apps/typegpu-docs/src/content/examples/tests/tgsl-parsing-test/index.ts
```diff
@@ -34,20 +34,17 @@ async function runTests() {
   return await result.read();
 }
 
+const table = document.querySelector<HTMLDivElement>('.result');
+if (!table) {
+  throw new Error('Nowhere to display the results');
+}
+runTests().then((result) => {
+  table.innerText = result ? 'Tests succeeded!' : 'Tests failed.';
```
reczkok: (just to be consistent - alternatively we can do `table.innerText = result ? 'Tests succeeded!' : 'Tests failed.'` for both)
```suggestion
  table.innerText = `Tests ${result ? 'succeeded' : 'failed'}.`;
```

iwoplaza (approved):
This is awesome, no notes! 👏

PR #1631, opened 2025-08-21
dev: Create tests for code produced by examples

### LINKED ISSUE #1627: dev: Create an automated tool that tracks shader code generated by our examples and allows us to easily diff it
This approach could help us spot potential generation improvements or regressions in more complex code than our existing tests, and highlight breaking changes that may require running the browser test suite. The simplest way forward is to mock WebGPU and run the examples with a spy on `createShaderModule`.

reczkok:
This PR adds tests for code passed to `createShaderModule` by our examples and creates inline snapshots for them to help us detect changes in our generated code. It also adds an API to hopefully make adding tests for new examples straightforward. 

closes #1627


iwoplaza (approved):
_EXTRA AWESOME_ 🪩

packages/typegpu/tests/examples/individual/3d-fish.test.ts
aleksanderkatan, line 1: Are we sure that we want to run these tests with our unit tests? They do not update when we modify the examples while in `test:unit:watch`, and they take a while to finish on the first run. Can we run them in a different script (and include that script in `nr test`)?

aleksanderkatan (commented):
Looks amazing!

packages/typegpu/tests/examples/individual/3d-fish.test.ts
reczkok: I will merge this and add an issue to move those tests to typegpu-docs where they should probably be and we can take care of this then. Right now this would require some uncomfortable vitest config changes and they don't take that long.

PR #1635, opened 2025-08-21
fix: Wider abstractInt range

reczkok:


packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -178,7 +178,8 @@ export function getTypeForIndexAccess(
 }
 
 export function numericLiteralToSnippet(value: number): Snippet {
-  if (Number.isInteger(value) && value >= MIN_INT32 && value <= MAX_INT32) {
+  // WGSL AbstractInt is a 64-bit two's complement integer
+  if (Number.isInteger(value) && value >= MIN_INT64 && value <= MAX_INT64) {
```
reczkok: If I understand [ECMAScript Language Specification](https://tc39.es/ecma262) correctly comparison between number and bigint should be fine.

Image

Image

Image

cieplypolar: I am not sure if this is correct. `Number.isInteger` return correct result up to Number.MAX_SAFE_INTEGER, which is around 2^53. What happens if we encounter `AbstractInt`, which can be larger than that?

cieplypolar: Okay, I checked larger integers will return true, but floats also. Other problem is, what if user types for example $2^{53} + 1$. We will incorrectly (or JS will) round it down to $2^{53}$. We cannot guarantee behavior of WGSL, which can handle it. I propose to use `Number.isSafeInteger` and throw warning that things get rounded. On the other hand, it is expected behavior of JS. Let me know what you think.

cieplypolar: Look at this code snippet
```ts
const b = root.createMutable(d.f32);
const f = tgpu['~unstable'].computeFn({ workgroupSize: [1] })(() => {
  b.$ = 1;
  const t = 0n;
  if (t > 9007199254740993n) {
    b.$ = 7;
  }
});
console.log(tgpu.resolve({ externals: { f } }));
```
It resolves to
```rust
@group(0) @binding(0) var<storage, read_write> b_1: f32;

@compute @workgroup_size(1) fn f_0() {
  b_1 = 1;
  var t = 0;
  if ((t > 9007199254740992)) {
    b_1 = 7;
  }
}
```
I'm afraid we cannot do much about it :(

cieplypolar (approved):
Great work 🦕

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -210,13 +210,30 @@ const Transpilers: Partial<
...
+      );
+    }
+    return [NODE.numericLiteral, String(Number(node.value)) ?? ''];
   },
 
   NumericLiteral(ctx, node) {
     return [NODE.numericLiteral, String(node.value) ?? ''];
   },
 
+  BigIntLiteral(ctx, node) {
```
cieplypolar, line 230: Looking for `Literal` and pattern that we have `node.bigint` and `BigIntLiteral`don't we need the `RegexLiteral` (apart from `node.regex`)?

reczkok: If no generator matches, we’ll end up throwing anyway, so creating a separate branch just for that feels unnecessary. The Acorn representation merges everything, so I handle it there since we always match the branch regardless. I could add it to the Babel representation as well, just for consistency.

reczkok: We will throw error like: `Unsupported JS functionality: RegexLiteral` which I think is good enough.

cieplypolar: Sounds good to me.

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -81,12 +81,12 @@ describe('wgslGenerator', () => {
...
         value: '12e10',
-        wgsl: '12e10',
-        dataType: abstractFloat,
+        wgsl: '120000000000',
+        dataType: abstractInt,
       },
       scientificNegativeExponentLiteral: {
         value: '12e-4',
-        wgsl: '12e-4',
+        wgsl: '0.0012',
```
iwoplaza: I am wondering why it chose decimal notation here, since the scientific notation is shorter 🤔. Any thoughts? @reczkok

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -103,10 +103,10 @@ describe('wgslGenerator', () => {
 
     expect(parsedBody).toStrictEqual([
       NODE.block,
-      Object.entries(literals).map(([key, { value }]) => [
+      Object.entries(literals).map(([key, { wgsl }]) => [
```
iwoplaza: Since we're testing `tinyest` here, we should expect it to equal `value`, not `wgsl`.
```suggestion
      Object.entries(literals).map(([key, { value }]) => [
```

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -103,10 +103,10 @@ describe('wgslGenerator', () => {
...
         NODE.const,
         key,
-        [NODE.numericLiteral, value],
+        [NODE.numericLiteral, wgsl],
```
iwoplaza:
```suggestion
        [NODE.numericLiteral, value],
```

iwoplaza (approved):
MERGE

PR #1642, opened 2025-08-23
fix: Better regex that includes potential query parameters

### LINKED ISSUE #1641: bug: Unplugin default regex does not take into account potential query params
Our current default regex for matching files for the plugin to transform is:
```regex
/\.m?[jt]sx?$/,
```
This works fine most of the time but sometimes bundlers use files with query parameters like:

Image

And since we expect the file to end with compatible extension it will not be transformed.

reczkok:
closes #1641


iwoplaza (approved):
Awesome change! 💜

PR #1643, opened 2025-08-25
docs: Add simple glass effect example

reczkok:


apps/typegpu-docs/src/content/examples/simple/liquid-glass/index.ts
```diff
@@ -0,0 +1,272 @@
...
+    );
+    const blue = std.textureSampleBias(
+      sampledView,
+      sampler,
+      uv.sub(dir.mul(offset)),
+      blur,
+    );
+    return d.vec3f(red.x, green.y, blue.z);
+  },
+);
```
iwoplaza: Does this work with Deno?
```suggestion
)((uv, offset, dir, blur) => {
  const red = std.textureSampleBias(
    sampledView,
    sampler,
    uv.add(dir.mul(offset)),
    blur,
  );
  const green = std.textureSampleBias(
    sampledView,
    sampler,
    uv,
    blur,
  );
  const blue = std.textureSampleBias(
    sampledView,
    sampler,
    uv.sub(dir.mul(offset)),
    blur,
  );
  return d.vec3f(red.x, green.y, blue.z);
});
```

apps/typegpu-docs/src/content/examples/simple/liquid-glass/index.ts
```diff
@@ -0,0 +1,272 @@
...
+
+let frameId: number;
+function render() {
+  frameId = requestAnimationFrame(render);
+  pipeline.withColorAttachment({
+    view: context.getCurrentTexture().createView(),
+    loadOp: 'clear',
+    storeOp: 'store',
+  }).draw(3);
+  root['~unstable'].flush();
```
iwoplaza: Should be unnecessary
```suggestion
```

iwoplaza (approved):
🧊🧊🧊

apps/typegpu-docs/src/content/examples/simple/liquid-glass/thumbnail.png
aleksanderkatan, line 1: Can you consider giving the glass a very slight tint so that it is always visible where the rectangle is?

packages/typegpu/src/std/texture.ts
```diff
@@ -131,6 +131,82 @@ export const textureSample: TextureSampleOverload = createDualImpl(
...
+    sampler: TgpuSampler,
+    coords: v3f,
+    bias: number,
+  ): v4f;
+  <T extends TgpuSampledTexture<'cube', F32>>(
+    texture: T,
+    sampler: TgpuSampler,
+    coords: v3f,
+    bias: number,
+  ): v4f;
```
aleksanderkatan: nit (this way we have the same number of entries that WGSL spec does)
```suggestion
  <T extends TgpuSampledTexture<'3d', F32> | TgpuSampledTexture<'cube', F32>>(
    texture: T,
    sampler: TgpuSampler,
    coords: v3f,
    bias: number,
  ): v4f;
```

packages/typegpu/src/std/texture.ts
```diff
@@ -131,6 +131,82 @@ export const textureSample: TextureSampleOverload = createDualImpl(
...
+  <T extends TgpuSampledTexture<'cube-array', F32>>(
+    texture: T,
+    sampler: TgpuSampler,
+    coords: v3f,
+    arrayIndex: number,
+    bias: number,
+  ): v4f;
+};
+
+export const textureSampleBias: TextureSampleBiasOverload = createDualImpl(
```
aleksanderkatan, line 185: Can we use the new `dualImpl` api here?

apps/typegpu-docs/src/content/examples/simple/liquid-glass/index.ts
```diff
@@ -0,0 +1,269 @@
...
+    blur,
+  );
+  return d.vec3f(red.x, green.y, blue.z);
+});
+
+const fragmentShader = tgpu['~unstable'].fragmentFn({
+  in: { uv: d.vec2f },
+  out: d.vec4f,
+})(({ uv }) => {
+  const mousePos = mousePosUniform.value;
```
aleksanderkatan: nit 
```suggestion
  const mousePos = mousePosUniform.$;
```

apps/typegpu-docs/src/content/examples/simple/liquid-glass/index.ts
```diff
@@ -0,0 +1,269 @@
...
+  }
+  const rect = canvas.getBoundingClientRect();
+  const x = (event.clientX - rect.left) / rect.width;
+  const y = (event.clientY - rect.top) / rect.height;
+  if (x < 0 || x > 1 || y < 0 || y > 1) {
+    return;
+  }
+  mousePosUniform.write(d.vec2f(x, y));
+}
+canvas.addEventListener('mousemove', updateMousePos);
```
aleksanderkatan: Can we make the rectangle move even when the mouse is not on the canvas?
```suggestion
  mousePosUniform.write(d.vec2f(x, y));
}
window.addEventListener('mousemove', updateMousePos);
```
(We'd also have to remove the event listener in cleanup)

packages/typegpu/src/std/texture.ts
reczkok: What would be the advantage of that?

apps/typegpu-docs/src/examples/simple/liquid-glass/index.ts
```diff
@@ -259,11 +296,30 @@ export const controls = {
...
 };
 
 export function onCleanup() {
   if (frameId) {
     cancelAnimationFrame(frameId);
   }
+  window.removeEventListener('mousemove', handleMouseMove);
+  canvas.removeEventListener('touchmove', handleTouchMove);
+  canvas.removeEventListener('click', handleClick);
+  canvas.removeEventListener('touchstart', handleTouchStart);
```
aleksanderkatan: I think we don't need to remove events from the canvas, only from the window since it persists between examples
```suggestion
```

PR #1644, opened 2025-08-25
feat: Add support for ** operator

aleksanderkatan:
JS `**` operator accepts numbers (of course), and WGSL `pow` only accepts floats.

Changes: 
- remove exception throw on `**` use,
- parse `**` to pow, with forced cast to f32/f16,
- add exception throw on `**=` use.

Let me know what you think about the `**=` operator. I think the following transpilation may be a little too drastic, and I don't think the operator is all that useful.
```ts
// TGSL
let a = d.u32(2);
a **= 4;
```
```wgsl
// WGSL
let a = 2u;
a = u32(pow(f32(a), 4f));
```

### DIFF
--- a/packages/typegpu/src/std/numeric.ts
+++ b/packages/typegpu/src/std/numeric.ts
@@ -818,12 +818,7 @@ function powCpu<T extends AnyFloatVecInstance | number>(
   if (typeof base === 'number' && typeof exponent === 'number') {
     return (base ** exponent) as T;
   }
-  if (
-    typeof base === 'object' &&
-    typeof exponent === 'object' &&
-    'kind' in base &&
-    'kind' in exponent
-  ) {
+  if (isVecInstance(base) && isVecInstance(exponent)) {
     return VectorOps.pow[base.kind](base, exponent) as T;
   }
   throw new Error('Invalid arguments to pow()');
--- a/packages/typegpu/src/std/numeric.ts
+++ b/packages/typegpu/src/std/numeric.ts
@@ -1,4 +1,5 @@
-import { snip } from '../data/snippet.ts';
+import { createDualImpl, dualImpl } from '../core/function/dualImpl.ts';
+import { stitch } from '../core/resolve/stitch.ts';
 import { smoothstepScalar } from '../data/numberOps.ts';
 import {
   abstractFloat,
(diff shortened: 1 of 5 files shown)


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -192,6 +192,11 @@ const opCodeToCodegen = {
   '-': sub[$internal].gpuImpl,
   '*': mul[$internal].gpuImpl,
   '/': div[$internal].gpuImpl,
+  '**': (lhsExpr: Snippet, rhsExpr: Snippet) => {
+    const [convLhs, convRhs] =
+      convertToCommonType([lhsExpr, rhsExpr], [f32, f16]) ?? [lhsExpr, rhsExpr];
+    return snip(stitch`pow(${convLhs}, ${convRhs})`, convLhs.dataType);
+  },
```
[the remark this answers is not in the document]
aleksanderkatan: it... already uses the `convLhs.dataType`

reczkok: We could move `pow` to the new `dualImpl` api like:
```ts
function powCpu(base: number, exponent: number): number;
function powCpu<T extends AnyFloatVecInstance>(
  base: T,
  exponent: T,
): T;
function powCpu<T extends AnyFloatVecInstance | number>(
  base: T,
  exponent: T,
): T {
  if (typeof base === 'number' && typeof exponent === 'number') {
    return (base ** exponent) as T;
  }
  if (
    typeof base === 'object' &&
    typeof exponent === 'object' &&
    'kind' in base &&
    'kind' in exponent
  ) {
    return VectorOps.pow[base.kind](base, exponent) as T;
  }
  throw new Error('Invalid arguments to pow()');
}

export const pow = dualImpl({
  name: 'pow',
  signature: (...args) => {
    const uargs = unify(args, [f32, f16, abstractFloat]) ?? args;
    return {
      argTypes: uargs,
      returnType: isNumericSchema(uargs[0]) ? uargs[1] : uargs[0],
    };
  },
  normalImpl: powCpu,
  codegenImpl: (lhs, rhs) => stitch`pow(${lhs}, ${rhs})`,
});
```
Then this simplifies to:
```suggestion
  '**': pow[$internal].gpuImpl,
```
And we get some free comptime powing

reczkok:
```diff
  "fn power() {
-   var n = pow(2, 4);
+   var n = 16.;
  }"
```

packages/typegpu/src/std/numeric.ts
```diff
@@ -804,31 +806,41 @@ export const normalize = createDualImpl(
...
+): T {
+  if (typeof base === 'number' && typeof exponent === 'number') {
+    return (base ** exponent) as T;
+  }
+  if (
+    typeof base === 'object' &&
+    typeof exponent === 'object' &&
+    'kind' in base &&
+    'kind' in exponent
+  ) {
```
iwoplaza: Tiny nit, more resilient to refactors
```suggestion
  if (isVecInstance(base) && isVecInstance(exponent)) {
```

iwoplaza (approved):
Yup yup yup! 🙏

PR #1652, opened 2025-08-27
feat: Texture schemas and complete texture api makeover

reczkok:


packages/typegpu/src/data/dataIO.ts
```diff
@@ -447,6 +447,58 @@ const dataWriters = {
...
+    throw new Error('Textures are not implemented yet!');
+  },
+  'texture_cube_array'() {
+    throw new Error('Textures are not implemented yet!');
+  },
+  'texture_depth_2d'() {
+    throw new Error('Textures are not implemented yet!');
+  },
+  'texture_depth_2d_array'() {
+    throw new Error('Textures are not implemented yet!');
```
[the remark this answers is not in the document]
reczkok: Yeah, will do

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -89,65 +152,25 @@ export interface TgpuTexture<TProps extends TextureProps = TextureProps>
...
-      TexelFormatToChannelType[
-        SampledFormatOptions<TProps> extends TFormat ? TProps['format']
-          : TFormat
-      ]
-    >;
-  }[TUsage];
+  createView(
+    ...args: this['usableAsSampled'] extends true ? []
+      : [ValidateTextureViewSchema<this, WgslTexture>]
+  ): TgpuTextureView<DefaultViewSchema<TProps>>;
```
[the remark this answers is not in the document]
reczkok: Well actually the schema is required for usage flag validation

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1536,6 +1558,11 @@ export type FlatInterpolatableData =
   | FlatInterpolatableAdditionalBaseType
   | Decorated<FlatInterpolatableAdditionalBaseType>;
 
+export type PrimitiveNumericData =
+  | F32
+  | I32
+  | U32;
```
aleksanderkatan: Nit: maybe we could find a different name? This one suggests to me that `F16` should be included.

apps/typegpu-docs/src/examples/tests/texture-test/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "title": "Texture Test",
+  "category": "tests",
+  "tags": ["experimental"]
```
aleksanderkatan:
```suggestion
  "tags": ["experimental"],
  "dev": true
```

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -145,7 +145,7 @@ export interface ColorAttachment {
    * A {@link GPUTextureView} describing the texture subresource that will be output to for this
    * color attachment.
    */
-  view: (TgpuTexture & Render) | GPUTextureView;
+  view: (TgpuTexture & RenderFlag) | GPUTextureView;
```
aleksanderkatan, line 152: It confused me when I saw a texture passed to a `view` prop. Could we consider renaming this or adding a different prop? If we want to avoid a breaking change, can we instead include more info about it in the JSDoc?

packages/typegpu/src/core/root/init.ts
```diff
@@ -422,13 +421,7 @@ class TgpuRootImpl extends WithBindingImpl
...
   unwrap(resource: TgpuBuffer<AnyData>): GPUBuffer;
   unwrap(resource: TgpuTexture): GPUTexture;
-  unwrap(
-    resource:
-      | TgpuReadonlyTexture
-      | TgpuWriteonlyTexture
-      | TgpuMutableTexture
-      | TgpuSampledTexture,
-  ): GPUTextureView;
+  unwrap(resource: TgpuTextureView): GPUTextureView;
```
aleksanderkatan, line 427: Do we not provide unwrap for storage textures?

packages/typegpu/src/core/texture/textureFormats.ts
```diff
@@ -162,75 +42,161 @@ export type StorageTextureTexelFormat =
...
+  | 'r16sint'
+  | 'r16float'
+  | 'rgb10a2unorm'
+  | 'rgb10a2uint'
+  | 'rg11b10ufloat';
 
-export type TexelFormatToDataType = typeof texelFormatToDataType;
-export type TexelFormatToDataTypeOrNever<T> = T extends
-  keyof TexelFormatToDataType ? TexelFormatToDataType[T] : never;
+// TODO: use this for texture view validation
```
aleksanderkatan, line 59: Is there a ticket created for this TODO?

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -1,76 +1,164 @@
...
+  | 'depth32float-stencil8';
+export type DepthFormats = 'depth16unorm' | 'depth24plus' | 'depth32float';
+export type StencilFormats = 'stencil8';
+
+export type AspectsForFormat<T extends GPUTextureFormat> =
+  GPUTextureFormat extends T ? TextureAspect
+    : T extends DepthStencilFormats ? 'depth' | 'stencil'
+    : never | T extends DepthFormats ? 'depth'
+    : never | T extends StencilFormats ? 'stencil'
+    : never | 'color';
```
aleksanderkatan: I don't get these `never`s, why are they here?
```suggestion
export type AspectsForFormat<T extends GPUTextureFormat> =
  GPUTextureFormat extends T ? TextureAspect
    : T extends DepthStencilFormats ? 'depth' | 'stencil'
    : T extends DepthFormats ? 'depth'
    : T extends StencilFormats ? 'stencil'
    : 'color';
```

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -240,86 +233,76 @@ export function isTexture<T extends TgpuTexture>(
...
+    this.aspects = (
+      format === 'depth24plus-stencil8' || format === 'depth32float-stencil8'
+        ? ['depth', 'stencil']
+        : format === 'depth16unorm' || format === 'depth24plus' ||
+            format === 'depth32float'
+        ? ['depth']
+        : format === 'stencil8'
+        ? ['stencil']
+        : ['color']
+    ) as AspectsForFormat<this['props']['format']>[];
```
aleksanderkatan: :(

packages/typegpu/src/core/texture/textureUtils.ts
```diff
@@ -0,0 +1,452 @@
...
+    [width, height, 1],
+  );
+
+  const renderTexture = device.createTexture({
+    size: [targetTexture.width, targetTexture.height, 1],
+    format: targetTexture.format,
+    usage: GPUTextureUsage.RENDER_ATTACHMENT | GPUTextureUsage.COPY_SRC,
+  });
+
+  const pipeline = device.createRenderPipeline({
```
aleksanderkatan, line 233: Could we use Tgpu pipelines here? Is there a reason not to do so?

aleksanderkatan (approved):
This one is huuge! GJ 〰️

packages/typegpu/src/core/root/init.ts
reczkok: Texture view can be both so I think we provide it

packages/typegpu/src/core/texture/textureFormats.ts
reczkok: I will create it

packages/typegpu/src/core/texture/texture.ts
reczkok: It's a remnant of the old type - will fix

packages/typegpu/src/core/texture/texture.ts
reczkok: :(

packages/typegpu/src/core/texture/textureUtils.ts
reczkok: Using our own API's inside the library can be problematic as it sometimes lead to impossible resolution (for example if texture needs pipeline and pipeline needs bind group and bind group needs texture we have a problem). This way we avoid any complicated relationships between files.

packages/typegpu/src/core/pipeline/renderPipeline.ts
reczkok: image

It's not just to avoid a breaking change - that's how it is in WebGPU API

apps/typegpu-docs/src/examples/image-processing/image-tuning/index.ts
```diff
@@ -24,9 +24,20 @@ const Adjustments = d.struct({
...
+  & TgpuTexture<
+    { size: [number, number, number]; format: 'rgba16float'; dimension: '3d' }
+  >
+  & SampledFlag;
+let imageTexture:
+  & TgpuTexture<{
+    size: [number, number];
+    format: 'rgba8unorm';
+  }>
+  & SampledFlag;
```
iwoplaza:
```suggestion
  & TgpuTexture<{ size: [number, number]; format: 'rgba8unorm' }>
  & SampledFlag;
```

packages/typegpu/src/core/root/configurableImpl.ts
```diff
@@ -8,13 +8,25 @@ import {
   type TgpuSlot,
 } from '../slot/slotTypes.ts';
 import type { Configurable } from './rootTypes.ts';
+import type { TgpuTextureView } from '../texture/texture.ts';
+import type { WgslStorageTexture, WgslTexture } from '../../data/index.ts';
```
iwoplaza: Let's not import from the barrel file

packages/typegpu/src/core/texture/externalTexture.ts
```diff
@@ -25,28 +32,59 @@ export function isExternalTexture<T extends TgpuExternalTexture>(
 
 export class TgpuExternalTextureImpl
   implements TgpuExternalTexture, SelfResolvable {
-  public readonly resourceType = 'external-texture';
-  public readonly [$wgslDataType]: AnyData;
+  readonly [$runtimeResource] = true;
+  readonly [$wgslDataType] = {} as WgslExternalTexture;
```
iwoplaza: Since these are not being used as runtime values anymore, these should be irrelevant
```suggestion
```

packages/typegpu/src/core/texture/externalTexture.ts
```diff
@@ -25,28 +32,59 @@ export function isExternalTexture<T extends TgpuExternalTexture>(
...
+  readonly [$repr] = undefined as unknown as WgslExternalTexture;
```
iwoplaza: Do we ever call `d.Infer` on these? If not, we can remove it
```suggestion
```

packages/typegpu/src/data/dataTypes.ts
```diff
@@ -19,6 +19,11 @@ import type {
 import { $internal } from '../shared/symbols.ts';
 import type { Prettify } from '../shared/utilityTypes.ts';
 import { vertexFormats } from '../shared/vertexFormat.ts';
+import type {
+  WgslExternalTexture,
+  WgslStorageTexture,
+  WgslTexture,
+} from './index.ts';
```
iwoplaza: Don't import from barrel files 🙏

packages/typegpu/src/data/texture.ts
```diff
@@ -0,0 +1,714 @@
...
+type StorageTextureLiteral = `texture_storage_${
+  | '1d'
+  | '2d'
+  | '2d_array'
+  | '3d'}`;
+
+export interface WgslTexture<
+  TProps extends Partial<WgslTextureProps> = WgslTextureProps,
+> extends BaseData {
+  readonly [$repr]: this;
```
iwoplaza: I'm not totally convinced that runtime texture types should be the same as schema types

packages/typegpu/src/shared/repr.ts
```diff
@@ -1,4 +1,6 @@
+import type { TgpuTexture } from '../core/texture/texture.ts';
 import type { Disarray, Undecorate } from '../data/dataTypes.ts';
+import type { WgslStorageTexture, WgslTexture } from '../data/index.ts';
```
iwoplaza: 👀👀👀

iwoplaza (changes requested):

packages/typegpu/src/data/texture.ts
reczkok: Yeah, neither am I but I don't have a better idea at the moment

reczkok: I was thinking as a part of a follow up PR we can move the current `repr` to `gpuRepr` and make `repr` be whatever we want the texture to be serializable as (possibly our own interface that makes it convenient to write to textures). Something like:
```suggestion
  readonly [$gpuRepr]: this;
  readonly [$repr]: TextureData<TProps>; // ImageBitmap | ArrayBuffer | ... (?)
```

iwoplaza: The problem with instance types having the same "type" as schemas, is that schemas are assignable to everywhere runtime types are:

image

Sampling a texture schema doesn't make sense, but it's fine on the type level

iwoplaza: I would propose we export near-empty instance types (there are no props on textures, so we're saved). We can call them the same as the schemas, and export them alongside the constructor functions.

image

Then we can enforce these types

image

reczkok: I like it - will do

packages/typegpu/src/data/texture.ts
```diff
@@ -476,11 +516,18 @@ export function texture2dArray<T extends PrimitiveNumericData>(sampleType?: T) {
     }));
 }
 
-export function textureCube<T extends PrimitiveNumericData>(
+export interface textureCube<
```
aleksanderkatan, line 519: Why are these types named in lowercase?

reczkok: This is our convention for instance types (so like `Vec3f`/`v3f` we have `WgslTexture2d`/`texture2d`)

packages/typegpu/src/core/root/configurableImpl.ts
```diff
@@ -8,13 +8,25 @@ import {
...
   with<T extends AnyWgslData>(
     slot: TgpuSlot<T> | TgpuAccessor<T>,
     value: T | TgpuFn<() => T> | TgpuBufferUsage<T> | Infer<T>,
+  ): Configurable;
+  with<T extends AnyWgslData>(
+    slot: TgpuSlot<T> | TgpuAccessor<T>,
+    value: T extends WgslTexture | WgslStorageTexture
+      ? TgpuTextureView<T> | TgpuFn<() => T> | Infer<T>
+      : T | TgpuFn<() => T> | TgpuBufferUsage<T> | Infer<T>,
   ): Configurable {
```
iwoplaza: I don't think this needs overloads, as it's just the implementation. The `Configurable` interface defines the public API

PR #1655, opened 2025-08-28
docs: Clouds example

lursz:
https://github.com/user-attachments/assets/11c4d43f-637a-41a7-a281-2cbc18488cde

### DIFF
--- a/apps/typegpu-docs/src/examples/rendering/clouds/consts.ts
+++ b/apps/typegpu-docs/src/examples/rendering/clouds/consts.ts
@@ -1,6 +1,5 @@
 import * as d from 'typegpu/data';
 
-export const MAX_DEPTH = 10;
 export const FOV_FACTOR = 1;
 
 export const SUN_DIRECTION = d.vec3f(1.0, 0.0, 0.0);


lursz:
Reverted perlin noise, due to performance issues (both impl with texture and without)

lursz:
Once approved, I will add example resolution test for it.

apps/typegpu-docs/src/examples/rendering/clouds/consts.ts
```diff
@@ -0,0 +1,28 @@
...
+export const WHITE = d.vec3f(1.0, 1.0, 1.0);
+export const DARK = d.vec3f(0.2, 0.2, 0.2);
+export const SKY = d.vec3f(0.6, 0.45, 0.75);
+export const SUN = d.vec3f(1.0, 0.7, 0.3);
+
+export const raymarchSlot = tgpu.slot<TgpuFn>();
+export const sampledViewSlot = tgpu.slot<TgpuSampledTexture<'2d', d.F32>>();
+export const samplerSlot = tgpu.slot<TgpuSampler>();
+export const timeAccess = tgpu['~unstable'].accessor(d.f32);
+export const resolutionAccess = tgpu['~unstable'].accessor(d.vec2f);
```
aleksanderkatan: Why not use these resources directly? What is the advantage of using these? Is this for abstraction?

apps/typegpu-docs/src/content/examples/rendering/clouds/utils.ts
```diff
@@ -0,0 +1,125 @@
...
+        albedo.z * lighting.z,
+      );
+      const premul = d.vec4f(
+        lit.x * density,
+        lit.y * density,
+        lit.z * density,
+        density,
+      );
+
+      res = std.add(res, std.mul(premul, LIGHT_ABSORPTION - res.w));
```
aleksanderkatan: nit
```suggestion
      const lit = albedo.mul(lighting);
      const premul = d.vec4f(lit, 1).mul(density);
      res = res.add(premul.mul(LIGHT_ABSORPTION - res.w));
```

apps/typegpu-docs/src/examples/rendering/clouds/index.ts
aleksanderkatan, line 1: Sometimes there are clearly visible vertical or horizontal divisions, do you know where they come from and if they are fixable?

image

aleksanderkatan (changes requested):
To unlock "approve", submit one "example resolution test" to the 🏧

apps/typegpu-docs/src/examples/rendering/clouds/consts.ts
lursz: Pretty much. I briefly tried to introduce cloud profiles into the example, but all my endeavours came out looking ugly. I left the abstraction for future extensibility and better file fragmentation

apps/typegpu-docs/src/examples/rendering/clouds/index.ts
lursz: This happens because clouds are generated from tiled 2d noise (`noise` function in `utils.ts`). We can change this, but it would hurt performance which isn't great anyways

lursz:
Merged `main` in order to access new texture api.

apps/typegpu-docs/src/examples/rendering/clouds/index.ts
aleksanderkatan: This isn't significant, so whatever

apps/typegpu-docs/src/examples/rendering/clouds/consts.ts
```diff
@@ -0,0 +1,28 @@
+import * as d from 'typegpu/data';
+
+export const MAX_DEPTH = 10;
```
aleksanderkatan: This variable is unused

apps/typegpu-docs/src/examples/rendering/clouds/index.ts
```diff
@@ -0,0 +1,176 @@
...
+
+root.device.queue.writeTexture(
+  { texture: root.unwrap(noiseTexture) },
+  noiseData,
+  { bytesPerRow: NOISE_TEXTURE_SIZE * 4 },
+  { width: NOISE_TEXTURE_SIZE, height: NOISE_TEXTURE_SIZE },
+);
+
+const bindGroup = root.createBindGroup(cloudsLayout, {
+  params: paramsUniform.buffer,
```
aleksanderkatan, line 72: I guess we can show in an example that we can still bind bindless resources 🤷

reczkok: I tend to prefer that to calling `.$usage` if I know my buffer will only be used one way

PR #1669, opened 2025-09-04
fix: Change visibility

### LINKED ISSUE #1654: fix: Better defaults for visibility in bind group layouts
Currently, when we try to write to a mutable buffer from fragment shader using TGLS code, we receive the following error:
> [!CAUTION]
> `Entry point's stage (ShaderStage::Fragment) is not in the binding visibility in the layout (ShaderStage::Compute).` 
This issue is related to our default setting, which limits the visibility of mutable buffers to `['compute']`. We could change this default to include both `['compute', 'fragment']`.

Reproduction:
```ts
const b1 = root.createBuffer(d.arrayOf(d.f32, 100 * 100)).$usage(
  'storage',
);
const b2 = b1.as('mutable');

const mainFragment = tgpu['~unstable'].fragmentFn({
  in: { uv: d.vec2f },
  out: d.vec4f,
})((input) => {
  const uv = input.uv.add(1).div(2);
  const grided = std.floor(uv.mul(100));

  b2.$[d.u32(grided.x * 100 + grided.y)] = grided.x * 100 +
    grided.y;

  return d.vec4f();
});
``` 

Temporary workaround:
```ts
  b: {
    storage: d.arrayOf(d.f32, 100 * 100),
    access: 'mutable',
    visibility: ['fragment'],
  },
});

const group = root.createBindGroup(layout, {
  b: b1,
});

const mainFragment = tgpu['~unstable'].fragmentFn({
  in: { uv: d.vec2f },
  out: d.vec4f,
})((input) => {
  const uv = input.uv.add(d.f32(1)).div(d.f32(2));
  const grided = std.floor(uv.mul(100));

  layout.bound.b.$[d.u32(grided.x * 100 + grided.y)] =
    grided.x * 100 +
    grided.y;

  return d.vec4f();
});
```

cieplypolar:
Changes:
- Mutable resources have now visibility [`compute`,  `fragment`]

Closes #1654.

### DIFF
--- a/packages/typegpu/src/tgpuBindGroupLayout.ts
+++ b/packages/typegpu/src/tgpuBindGroupLayout.ts
@@ -87,7 +87,7 @@ export type TgpuLayoutEntryBase = {
    * By default, each resource is visible to all shader stage types, but
    * depending on the underlying implementation, this may have performance implications.
    *
-   * @default ['compute'] for mutable resources
+   * @default ['compute', 'fragment] for mutable resources
    * @default ['compute','vertex','fragment'] for everything else
    */
   visibility?: TgpuShaderStage[];
@@ -416,7 +416,7 @@ export class MissingBindingError extends Error {
 // Implementation
 // --------------
 
-const DEFAULT_MUTABLE_VISIBILITY: TgpuShaderStage[] = ['compute'];
+const DEFAULT_MUTABLE_VISIBILITY: TgpuShaderStage[] = ['compute', 'fragment'];
 const DEFAULT_READONLY_VISIBILITY: TgpuShaderStage[] = [
   'compute',
   'vertex',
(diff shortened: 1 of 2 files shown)


packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -87,7 +87,7 @@ export type TgpuLayoutEntryBase = {
    * By default, each resource is visible to all shader stage types, but
    * depending on the underlying implementation, this may have performance implications.
    *
-   * @default ['compute'] for mutable resources
+   * @default ['compute', 'fragment] for mutable resources
```
iwoplaza:
```suggestion
   * @default ['compute', 'fragment'] for mutable resources
```

PR #1670, opened 2025-09-05
fix/impr: Fix nested compiled writers and enable them for initial data

reczkok:
Up until now we never actually used the compiled writers for initial data in buffers. While enabling it I discovered a major bug which makes nested loops in writers not work. This PR addresses both those issues.

### DIFF
--- a/packages/typegpu/src/data/numeric.ts
+++ b/packages/typegpu/src/data/numeric.ts
@@ -23,7 +23,7 @@ export const abstractFloat = {
 } as AbstractFloat;
 
 const boolCast = dualImpl({
-  name: 'boolCast',
+  name: 'bool',
   signature: (arg) => ({ argTypes: arg ? [arg] : [], returnType: bool }),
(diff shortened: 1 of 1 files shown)


packages/typegpu/src/data/compiledIO.ts
```diff
@@ -151,9 +151,12 @@ export function buildWriter(
   node: wgsl.BaseData,
   offsetExpr: string,
   valueExpr: string,
+  depth = 0,
```
cieplypolar, line 154:
```suggestion
  nestingLevel = 0,
```
I found this parameter name misleading. At first, I thought it referred to the total depth of nesting.

cieplypolar (approved):
Great work! 🦕

packages/typegpu/src/data/compiledIO.ts
reczkok: Is `currentDepth` or `currDepth` better?

PR #1680, opened 2025-09-09
docs: Add LCG generator to example

cieplypolar:
Changes:
- added LCG generator
- refactored executor in `probability`


apps/typegpu-docs/src/examples/algorithms/probability/index.ts
```diff
@@ -86,18 +107,24 @@ export const controls = {
...
+          c.generators.map((gen) =>
+            tgpu.resolve({
+              externals: {
+                p: executor.pipelineCacheGet(
+                  getPRNG(dist).prng,
+                  getGenerator(gen),
+                ),
+              },
+            })
+          )
```
aleksanderkatan: I think it would be cleaner as a nested for...of. Right now the result of the map is not used, so it would be more fitting for forEach anyway. 
Also, if we do not get rid of the default, this test has duplicate resolves for the default generator.

PR #1683, opened 2025-09-10
fix: Narrow types in std

aleksanderkatan:
Changes:
- rewrite most of numeric functions to use `dualImpl` instead of `createDualImpl`,
- remove the privateRemarks comments,
- when needed, extract normalImpls of std functions and write explicit overloads.

I have never used the links in private remarks and they were incomplete anyway so I decided to remove them. I can bring them back (and maybe make them public?) if you request so.

### DIFF
--- a/packages/typegpu/src/data/numberOps.ts
+++ b/packages/typegpu/src/data/numberOps.ts
@@ -19,3 +19,14 @@ export const divInteger = (lhs: number, rhs: number) => {
   }
   return Math.trunc(lhs / rhs);
 };
+
+export function bitcastU32toF32Impl(n: number): number {
+  const dataView = new DataView(new ArrayBuffer(4));
+  dataView.setUint32(0, n, true);
+  return dataView.getFloat32(0, true);
+}
+export function bitcastU32toI32Impl(n: number): number {
+  const dataView = new DataView(new ArrayBuffer(4));
+  dataView.setUint32(0, n, true);
+  return dataView.getInt32(0, true);
+}
--- a/packages/typegpu/src/data/vectorOps.ts
+++ b/packages/typegpu/src/data/vectorOps.ts
@@ -1,6 +1,12 @@
 import { $internal } from '../shared/symbols.ts';
 import { mat2x2f, mat3x3f, mat4x4f } from './matrix.ts';
-import { clamp, divInteger, smoothstepScalar } from './numberOps.ts';
+import {
+  bitcastU32toF32Impl,
+  bitcastU32toI32Impl,
+  clamp,
+  divInteger,
+  smoothstepScalar,
+} from './numberOps.ts';
 import * as vectorConstructors from './vector.ts';
 import type * as wgsl from './wgslTypes.ts';
 import type { VecKind } from './wgslTypes.ts';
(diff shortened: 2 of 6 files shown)


packages/typegpu/src/std/numeric.ts
```diff
@@ -353,160 +378,189 @@ export const dot = dualImpl({
...
-);
+  codegenImpl: (e1, e2) => stitch`dot4U8Packed(${e1}, ${e2})`,
+  name: 'dot4U8Packed',
+});
 
-export const dot4I8Packed = createDualImpl(
-  // CPU implementation
-  (e1: number, e2: number): number => {
+export const dot4I8Packed = dualImpl({
+  signature: (lhs, rhs) => ({ argTypes: [u32, u32], returnType: i32 }),
```
aleksanderkatan, line 399: The arg types were `[i32, i32]`, but this was an error

packages/typegpu/src/data/vector.ts
```diff
@@ -273,6 +273,24 @@ export const vecTypeToConstructor = {
...
+  vec3i: i32,
+  vec3u: u32,
+  'vec3<bool>': bool,
+  vec4f: f32,
+  vec4h: f16,
+  vec4i: i32,
+  vec4u: u32,
+  'vec4<bool>': bool,
+} as const;
+
```
reczkok: We actually don't need to add this back since we have the `primitive` field on vector schemas
```suggestion
```

packages/typegpu/src/std/numeric.ts
```diff
@@ -45,468 +51,521 @@ import { mul, sub } from './operators.ts';
...
-/**
- * @privateRemarks
- * https://www.w3.org/TR/WGSL/#dot-builtin
- */
 export const dot = dualImpl({
   name: 'dot',
-  signature: (...argTypes) => ({ argTypes, returnType: f32 }),
+  signature: (e1, e2) => ({
+    argTypes: [e1, e2],
+    returnType: vecTypeToElement[e1.type as keyof typeof vecTypeToElement],
```
reczkok:
```suggestion
    returnType: (e1 as VecData).primitive,
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -106,21 +107,7 @@ const kindToSchema = {
...
-  vec3h: f16,
-  vec3i: i32,
-  vec3u: u32,
-  'vec3<bool>': bool,
-  vec4f: f32,
-  vec4h: f16,
-  vec4i: i32,
-  vec4u: u32,
-  'vec4<bool>': bool,
+  ...vecTypeToElement,
```
reczkok:
```suggestion
```
And we can do:
```ts
export function getTypeForIndexAccess(
  dataType: AnyData,
): AnyData | UnknownData {
  // array
  if (isWgslArray(dataType) || isDisarray(dataType)) {
    return dataType.elementType as AnyData;
  }

  // matrix
  if (dataType.type in indexableTypeToResult) {
    return indexableTypeToResult[
      dataType.type as keyof typeof indexableTypeToResult
    ];
  }

  // vector
  if (isVec(dataType)) {
    return dataType.primitive;
  }

  return UnknownData;
}
```

PR #1692, opened 2025-09-11
feat: tgpu.namespace

iwoplaza:


packages/typegpu/src/core/resolve/namespace.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    throw new Error(`Unsupported event: ${event}`);
+  }
+}
+
+export interface NamespaceOptions {
+  names?: 'random' | 'strict' | undefined;
+}
+
+export function requestUniqueName(
```
reczkok: I'm not sure abut this name. When I see `request` my mind immediately goes to `asynchronous` and `possibly rejected` neither of which apply here. Maybe `getUniqueName`?

reczkok (approved):
Amazing! It fits perfectly into the existing API

packages/typegpu/src/core/resolve/namespace.ts
```diff
@@ -0,0 +1,99 @@
...
+  >;
+
+  listeners: {
+    [K in keyof NamespaceEventMap]: Set<(event: NamespaceEventMap[K]) => void>;
+  };
+}
+
+type NamespaceEventMap = {
+  'name': { target: object; name: string };
+};
```
aleksanderkatan, line 36: What other events do we plan on supporting?

packages/typegpu/src/core/resolve/namespace.ts
```diff
@@ -0,0 +1,99 @@
...
+
+type DetachListener = () => void;
+
+export interface Namespace {
+  readonly [$internal]: NamespaceInternal;
+
+  on<TEvent extends keyof NamespaceEventMap>(
+    event: TEvent,
+    listener: (event: NamespaceEventMap[TEvent]) => void,
+  ): DetachListener;
```
aleksanderkatan, line 46: This is currently only used in tests, in which cases do we plan on using this?

aleksanderkatan (approved):
✅✅✅

packages/typegpu/src/core/resolve/namespace.ts
iwoplaza: None so far, but this patterns allows us to add more "hooks" into the resolution process if they're needed. Hard to determine what they'd be, as the need for namespace surfaced only after working on @typegpu/three

packages/typegpu/src/core/resolve/namespace.ts
iwoplaza: In @typegpu/three for inspecting what name was given for a specific resource (private variable in that case).

PR #1694, opened 2025-09-12
feat(@typegpu/react): Implement useUniformValue hook

### LINKED ISSUE #1689: feat(@typegpu/react): useUniformValue
This hook should accept two parameters:
- `schema: TSchema`: The schema that will be used to allocate a GPU buffer, and serialize values to VRAM.
- `initialValue?: TValue`: The value that the buffer will hold initially. **Changes to this value after the initial component render are ignored.**

## Signature:
```ts
useUniformValue<TSchema extends d.AnyWgslData, TValue extends d.Infer<TSchema>>(schema: TSchema, initialValue?: TValue | undefined): UniformValue<TSchema, TValue>;
```

## Usage:

```ts
const Component = () => {
  const time = useUniformValue(d.f32, 0);
  
  // Runs every frame on the CPU
  useFrame(() => {
    // The .value property allows mutable access on the CPU    
    time.value = performance.now() / 1000;
  });

  const { ref } = useRender({
    fragment: () => {
      'kernel';
      // The .$ property allows readonly access on the GPU
      const t = time.$;
      return d.vec4f(sin(t) * 0.5 + 0.5, 0, 0, 1);
    },
  });

  return <canvas ref={ref}></canvas>;
};

```

piaccho:
## Changes

* Added an implementation of the `useUniformValue` hook
* Updated the triangle example

---

Closes #1689


packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,45 @@
...
+>(
+  schema: TSchema,
   initialValue?: TValue | undefined,
 ): UniformValue<TSchema, TValue> {
-  // TODO: Implement
+  const root = useRoot();
+
+  const uniformBuffer = useMemo(() => {
+    // biome-ignore lint/suspicious/noExplicitAny: Constraint validation issue with ValidateUniformSchema
+    return root.createUniform(schema as any, initialValue);
```
reczkok: #1706 will add the ValidateUniformSchema type to the public API so you can wrap `TSchema` in it - this will fix the type error and provide validation for the function

Image

piaccho: Excellent! This was a temporary workaround. I will provide proper modification when mentioned PR is merged.

packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,45 @@
...
+  }, [root, schema, initialValue]);
```
iwoplaza: RIght now, allocated buffers are not being cleanup up on unmount, and if the passed-in schema is a compound schema (like the result of calling d.arrayOf(), the buffer is reallocated on every render.

For now, I would just take the first schema and initial value, and ignore subsequent changes to these values.

Going by React suggestions, the result of `useMemo` "could" be thrown out if there is memory pressure, or if the view is inactive for some time, which is not what we want. Let's try to use `useState` for allocating the buffer.

```ts
const [uniformBuffer] = useState(() => root.createUniform(schema, initialValue));
```

packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,45 @@
...
+          uniformBuffer.write(newValue);
+        }
+      },
+      get $() {
+        return uniformBuffer.$;
+      },
+    };
+  }, [schema, uniformBuffer, initialValue]);
+
+  return uniformValue;
```
iwoplaza: Let's make this value stable (have the same object identity across renders). This means having the dependecy array blank: [].

packages/typegpu-react/src/use-uniform-value.ts
piaccho: This definitely makes sense. I will provide improved implementation based on it.

packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,55 @@
...
+        }
+      },
+      get $() {
+        return uniformBuffer.$;
+      },
+    };
+  }, []);
+
+  return uniformValue;
 }
```
reczkok: If we wrap the schema in the validation type and cast the return type instead of the buffer we get proper validation

Before:
Image

After:
Image

```suggestion
export function useUniformValue<
  TSchema extends d.AnyWgslData,
  TValue extends d.Infer<TSchema>,
>(
  schema: ValidateUniformSchema<TSchema>,
  initialValue?: TValue | undefined,
): UniformValue<TSchema, TValue> {
  const root = useRoot();

  const [uniformBuffer] = useState(() => {
    return root.createUniform(
      schema,
      initialValue,
    );
  });

  useEffect(() => {
    return () => {
      uniformBuffer.buffer.destroy();
    };
  }, [uniformBuffer]);

  // biome-ignore lint/correctness/useExhaustiveDependencies: This value needs to be stable
  const uniformValue = useMemo(() => {
    let currentValue = initialValue;
    return {
      schema,
      get value() {
        return currentValue;
      },
      set value(newValue: TValue | undefined) {
        currentValue = newValue;
        if (newValue !== undefined) {
          uniformBuffer.write(newValue);
        }
      },
      get $() {
        return uniformBuffer.$;
      },
    };
  }, []);

  return uniformValue as UniformValue<TSchema, TValue>;
}
```

piaccho: Good catch. I missed it, thanks!

apps/typegpu-docs/src/content/examples/react/triangle/index.tsx
```diff
@@ -28,7 +34,7 @@ const reactRoot = createRoot(
 reactRoot.render(<App />);
 
 export function onCleanup() {
-  reactRoot.unmount();
+  setTimeout(() => reactRoot.unmount(), 0);
```
iwoplaza, line 37: What error forces this timeout?

packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,55 @@
...
+      schema,
+      initialValue,
+    );
+  });
+
+  useEffect(() => {
+    return () => {
+      uniformBuffer.buffer.destroy();
+    };
+  }, [uniformBuffer]);
```
iwoplaza: I don't think this will work in strict mode.

```suggestion
  const ref = useRef<number>();
  useEffect(() => {
    if (ref.current) {
      // Cancel deallocation
      clearTimeout(ref.current);
    }
    
    return () => {
      // Schedule deallocation
      ref.current = setTimeout(() => {
        uniformBuffer.buffer.destroy();
      }, 200);
    };
  }, [uniformBuffer]);
```

packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,55 @@
...
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: This value needs to be stable
+  const uniformValue = useMemo(() => {
+    let currentValue = initialValue;
```
iwoplaza, line 43: Schemas should generally be callable, in which case they return an initial value that matches that schema.

```suggestion
    let currentValue = initialValue ?? schema();
```

packages/typegpu-react/src/use-uniform-value.ts
```diff
@@ -1,13 +1,55 @@
...
+      schema,
+      get value() {
+        return currentValue;
+      },
+      set value(newValue: TValue | undefined) {
+        currentValue = newValue;
+        if (newValue !== undefined) {
+          uniformBuffer.write(newValue);
+        }
+      },
```
iwoplaza, line 54: Because of the previous comment, we don't have to worry about undefined.
```suggestion
      get value() {
        return currentValue;
      },
      set value(newValue: TValue) {
        currentValue = newValue;
        uniformBuffer.write(newValue);
      },
```

apps/typegpu-docs/src/content/examples/react/triangle/index.tsx
piaccho: It was related to the following error thrown on example unmount:
```
Attempted to synchronously unmount a root while React was already rendering. React cannot finish unmounting the root until the current render has completed, which may lead to a race condition.
```
Initially I thought the cause was a new uniform's lifecycle, but now I think it's because of the conflict between synchronous unmount and animation loop in `useFrame`.

packages/typegpu-react/src/use-uniform-value.ts
piaccho: You're right, I missed it.

packages/typegpu-react/src/use-uniform-value.ts
piaccho: Thanks for pointing that out! However, with current implementation:

```ts
export function useUniformValue<
  TSchema extends d.AnyWgslData,
  TValue extends d.Infer<TSchema>,
>(
  schema: ValidateUniformSchema<TSchema>,
  initialValue?: TValue | undefined,
): UniformValue<TSchema, TValue> {
```
schema is not callable:

```
This expression is not callable.
  Not all constituents of type 'AnyWgslData | `(Error) ${string}`' are callable.
    Type 'U16' has no call signatures.ts(2349)
```
I'm wondering how can I alter this to simplify the logic.

PR #1695, opened 2025-09-12
fix: Accessor type inference

iwoplaza:


packages/typegpu/tests/examples/individual/oklab.test.ts
```diff
@@ -197,7 +197,7 @@ describe('oklab example', () => {
       }
 
       fn gamutClipAdaptiveL05_9(lab: vec3f) -> vec3f {
-        var alpha = 0.2;
+        var alpha = 0.2f;
```
iwoplaza, line 200: The accessor's type was not carrying over to this variable declaration before these changes 🎉

packages/typegpu/tests/accessor.test.ts
```diff
@@ -170,11 +170,26 @@ describe('tgpu.accessor', () => {
           var color2 = redUniform;
           var color3 = getColor();
 
-          var colorX = vec3f(1, 0, 0).x;
+          var colorX = 1;
```
iwoplaza, line 173: Accessors populate with a constant value are not known at comp-time, which makes them even more useful 🎉

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -327,13 +328,15 @@ class FnCall<ImplSchema extends AnyFn> implements SelfResolvable {
...
   }
 
-  '~resolve'(ctx: ResolutionCtx): string {
+  [$resolve](ctx: ResolutionCtx): string {
     // We need to reset the indentation level during function body resolution to ignore the indentation level of the function call
-    return ctx.withResetIndentLevel(() =>
-      stitch`${ctx.resolve(this.#fn)}(${this.#params})`
-    );
+    return ctx.withResetIndentLevel(() => {
+      // TODO: Resolve the params first, then the function (just for consistency)
```
iwoplaza, line 337: A separate PR, since it would update pretty much every example's snapshot

packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -140,20 +144,21 @@ export function isComparisonSampler(
...
-    // TODO: do not treat self-resolvable as wgsl data (when we have proper sampler schemas)
-    // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have sampler schemas
-    this[$wgslDataType] = this as any;
     this[$internal] = {};
     setName(this, _membership.key);
   }
 
-  '~resolve'(ctx: ResolutionCtx): string {
+  // TODO: do not treat self-resolvable as wgsl data (when we have proper texture schemas)
+  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have texture schemas
```
reczkok: I don't think texture schemas help us here
```suggestion
  // TODO: do not treat self-resolvable as wgsl data (when we have proper sampler schemas)
  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have sampler schemas
```

packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -170,20 +175,20 @@ export class TgpuLaidOutSamplerImpl implements TgpuSampler, SelfResolvable {
...
```
reczkok: Same here
```suggestion
  // TODO: do not treat self-resolvable as wgsl data (when we have proper sampler schemas)
  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have sampler schemas
```

packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -230,12 +232,11 @@ class TgpuFixedSamplerImpl implements TgpuFixedSampler, SelfResolvable {
       _props.mipmapFilter === 'linear';
   }
 
-  $name(label: string) {
-    setName(this, label);
-    return this;
-  }
+  // TODO: do not treat self-resolvable as wgsl data (when we have proper texture schemas)
+  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have texture schemas
```
reczkok: And here
```suggestion
  // TODO: do not treat self-resolvable as wgsl data (when we have proper sampler schemas)
  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have sampler schemas
```

packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -283,12 +285,16 @@ class TgpuFixedComparisonSamplerImpl
     };
   }
 
+  // TODO: do not treat self-resolvable as wgsl data (when we have proper texture schemas)
+  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have texture schemas
```
reczkok: Aaaand here
```suggestion
  // TODO: do not treat self-resolvable as wgsl data (when we have proper sampler schemas)
  // biome-ignore lint/suspicious/noExplicitAny: This is necessary until we have sampler schemas
```

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -58,6 +59,34 @@ export class TgpuAccessorImpl<T extends AnyWgslData>
...
+    const ctx = getResolutionCtx()!;
+    const value = getGpuValueRecursively(ctx.unwrap(this.slot));
+
+    if (isBufferUsage(value) || isBufferShorthand(value)) {
+      return snip(value, this.schema);
+    }
+
+    if (isTgpuFn(value)) {
+      return value[$internal].gpuImpl();
+    }
```
reczkok: Also could we not make the gpu value of tgpu fn its gpu impl? Then it would be unified
```suggestion
    const ctx = getResolutionCtx()!;
    const value = getGpuValueRecursively(ctx.unwrap(this.slot));

    if (isTgpuFn(value)) {
      return value[$internal].gpuImpl();
    }
```

packages/typegpu/src/std/numeric.ts
```diff
@@ -52,7 +52,7 @@ import { mul, sub } from './operators.ts';
 type NumVec = AnyNumericVecInstance;
 
 function cpuAbs(value: number): number;
-function cpuAbs<T extends NumVec>(value: T): T;
+function cpuAbs<T extends NumVec | number>(value: T): T;
```
reczkok, line 55: What does that do?

packages/typegpu/src/types.ts
```diff
@@ -263,19 +268,30 @@ export interface ResolutionCtx {
...
+  return !!(value as SelfResolvable)?.[$resolve];
+}
+
+export interface WithGPUValue<T> {
+  readonly [$gpuValueOf]: T;
+}
+
+export interface WithOwnSnippet {
+  readonly [$ownSnippet]: Snippet;
+}
```
reczkok, line 291: I think this interface is underutilized - for example constant and buffer usage don't use it.

apps/typegpu-docs/src/examples/tests/tgsl-parsing-test/index.html
```diff
@@ -1 +1 @@
-<div class='result'>Wait for the tests to finish running...</div>;
+<div class="result">Wait for the tests to finish running...</div>;
```
aleksanderkatan:
```suggestion
<div class="result">Wait for the tests to finish running...</div>
```
This semicolon is just a string that is displayed, idk how and when it got there.

image

packages/typegpu/src/core/sampler/sampler.ts
iwoplaza: Ah, I just copied the warnings between resources. Thanks 🙏

packages/typegpu/src/std/numeric.ts
iwoplaza: It allows unions to be passed into `abs`
image

packages/typegpu/src/core/slot/accessor.ts
iwoplaza: Thanks for the suggestion! Regarding making fn's [$gpuValueOf] be it's gpuImpl, it would break slot behavior. If we had a slot that houses a function:

```ts
const fooSlot = tgpu.slot(tgpu.fn(...)(...));

const foo = () => {
  'kernel';
  // This erroneously calls the gpuImpl, not the dualImpl
  fooSlot.$();
}
```

Slot access would stop working in "normal" mode.

packages/typegpu/src/types.ts
iwoplaza: They do though 👀

image

Accessors also use this symbol only as part of their gpu value.

packages/typegpu/src/std/numeric.ts
```diff
@@ -950,7 +950,10 @@ function cpuRadians<T extends AnyFloatVecInstance | number>(value: T): T {
 
 export const radians = dualImpl({
   name: 'radians',
-  signature: (arg) => ({ argTypes: [arg], returnType: arg }),
+  signature: (...args) => {
+    const uargs = unify(args, [f32, f16, abstractFloat]) ?? args;
```
aleksanderkatan, line 954: Does unify allow vectors here as well? It seems so in a quick test I wrote, but I want to be sure

packages/typegpu/src/std/numeric.ts
```diff
@@ -1146,7 +1149,10 @@ function cpuStep<T extends AnyFloatVecInstance | number>(edge: T, x: T): T {
 
 export const step = dualImpl({
   name: 'step',
-  signature: (edge, x) => ({ argTypes: [edge, x], returnType: edge }),
+  signature: (...args) => {
+    const uargs = unify(args) ?? args;
```
```suggestion
    const uargs = unify(args, [f32, f16, abstractFloat]) ?? args;
```

aleksanderkatan (approved):
LGTM! 🦾

packages/typegpu/src/std/numeric.ts
reczkok: Unfiy basically always returns undefined if a vector gets passed in so we will fall back to args

packages/typegpu/src/types.ts
reczkok: Oh I just noticed this:
```ts
export const valueProxyHandler: ProxyHandler<
  SelfResolvable & WithOwnSnippet
> = {
```
Now I get it

PR #1708, opened 2025-09-15
feat: Base types in `console.log`

aleksanderkatan:
Changes:
- add a `toString` for matrices,
- support `f32`, `f16`, `i32`, all vectors and all matrices in `console.log`.


apps/typegpu-docs/src/examples/tests/log-test/index.ts
```diff
@@ -4,7 +4,10 @@ import * as d from 'typegpu/data';
 const root = await tgpu.init({
   unstable_logOptions: {
     logCountLimit: 32,
-    logSizeLimit: 32,
+    logSizeLimit: 128,
+  },
+  device: {
+    optionalFeatures: ['shader-f16'],
```
reczkok, line 11: If we have this feature as optional we should also take advantage of feature based pruning to make this test not fail if this feature is not available

iwoplaza (approved):
Amazing!! This is huuuuge

apps/typegpu-docs/src/examples/tests/log-test/index.ts
```diff
@@ -32,23 +36,56 @@ export const controls = {
...
+        console.log();
+        if (hasF16) {
+          console.log('--- f16 ---');
+          console.log(d.f16(3.14));
+          console.log(d.vec2h(1.1, -2.2));
+          console.log(d.vec3h(10.1, -20.2, 30.3));
+          console.log(d.vec4h(100.1, -200.2, 300.3, -400.4));
+        } else {
+          console.log("The 'shader-f16' flag is not enabled.");
+        }
```
reczkok: We could also do this but in this case I guess it doesn't matter (I will just use any excuse to use this function)
```suggestion
        if (std.extensionEnabled('f16')) {
          console.log('--- f16 ---');
          console.log(d.f16(3.14));
          console.log(d.vec2h(1.1, -2.2));
          console.log(d.vec3h(10.1, -20.2, 30.3));
          console.log(d.vec4h(100.1, -200.2, 300.3, -400.4));
        } else {
          console.log("The 'shader-f16' flag is not enabled.");
        }
```

PR #1716, opened 2025-09-16
tests: disco-example-tests

lursz:
depends on #1656 being merged into `main`

### DIFF
--- a/packages/typegpu/tests/examples/individual/disco.test.ts
+++ b/packages/typegpu/tests/examples/individual/disco.test.ts
@@ -1,0 +1,88 @@
+/**
+ * @vitest-environment jsdom
+ */
+
+import { describe, expect } from 'vitest';
+import { it } from '../../utils/extendedIt.ts';
+import { runExampleTest, setupCommonMocks } from '../utils/baseTest.ts';
+
+describe('disco example', () => {
+  setupCommonMocks();
+
+  it('should produce valid code', async ({ device }) => {
+    const shaderCodes = await runExampleTest({
+      category: 'rendering',
+      name: 'disco',
+      expectedCalls: 1,
+    }, device);
+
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/examples/simple/disco/index.ts
aleksanderkatan, line 1: Is this file duplicated? Should it exist?

apps/typegpu-docs/src/examples/simple/disco/index.ts
```diff
@@ -0,0 +1,128 @@
...
+          externals: {
+            mainVertex,
+            fragment,
+            timeAccess,
+            resolutionAccess,
+          },
+        });
+        device.createShaderModule({ code });
+      }
+      render();
```
aleksanderkatan: Should we cancel the previous `requestAnimationFrame(render)` or is it alright?

apps/typegpu-docs/src/examples/simple/disco/index.ts
aleksanderkatan, line 1: This example throws an error when switching to a different one.

lursz: Yeah it's a duplication, I'll remove it. The duplication emerged with our examples structure rework.

apps/typegpu-docs/src/examples/simple/disco/index.ts
lursz: Both are necessary in this impl

apps/typegpu-docs/src/examples/simple/disco/index.ts
lursz: This should have solved itself with the removal of the duplicate.

PR #1718, opened 2025-09-16
feat: Shell-less functions

iwoplaza:


packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -1,13 +1,21 @@
...
+    jsImpl: (v?: S) => S;
+  };
+};
+
 // deno-fmt-ignore
 export abstract class VecBase<S> extends Array implements SelfResolvable {
   public readonly [$internal] = true;
   abstract get kind(): VecKind;
-  abstract castElement(): (v?: S) => S;
+  abstract schema(): VecSchema<S>;
```
reczkok: This makes more sense to me as a getter (or even just a field) since calling a schema for me means the instance type (even though I know that is not the case here)
```suggestion
  abstract get schema(): VecSchema<S>;
```
```suggestion
  abstract readonly schema: VecSchema<S>;
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -376,7 +377,15 @@ export class ResolutionCtxImpl implements ResolutionCtx {
   }
 
   get topFunctionReturnType() {
-    return this._itemStateStack.topFunctionReturnType;
+    const scope = this._itemStateStack.topFunctionScope;
+    if (!scope) {
+      throw new Error('Internal error, expected function scope to be present.');
+    }
```
reczkok: For internal errors we could use this:
```suggestion
    invariant(scope, 'Internal error, expected function scope to be present.');
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -405,6 +414,14 @@ export class ResolutionCtxImpl implements ResolutionCtx {
     return this._itemStateStack.defineBlockVariable(id, dataType);
   }
 
+  reportReturnType(dataType: AnyData) {
+    const scope = this._itemStateStack.topFunctionScope;
+    if (!scope) {
+      throw new Error('Internal error, expected function scope to be present.');
+    }
```
reczkok:
```suggestion
    invariant(scope, 'Internal error, expected function scope to be present.');
```

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -423,9 +442,37 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
+        const concreteTypes = [...uniqueReturnTypes].filter(isConcrete);
+        if (uniqueReturnTypes.length === 0) {
+          returnType = Void;
+        } else if (concreteTypes.length <= 1) {
+          // Trying to find a common type for the return values
+          returnType = unifyAuto(
+            uniqueReturnTypes,
+            concreteTypes.length > 0 ? concreteTypes : undefined,
+          )?.[0];
+        }
```
reczkok: I don't think we need `unifyAuto`. `getBestConversion` already does everything we need and a non implicit conversion with more than one concrete type is not possible.
```suggestion
      if (!returnType) {
        const returnTypes = scope.reportedReturnTypes;
        if (returnTypes.length === 0) {
          returnType = Void;
        } else {
          const conversion = getBestConversion(returnTypes);

          if (conversion && !conversion.hasImplicitConversions) {
            returnType = conversion.targetType;
          }
        }
```

packages/typegpu/src/data/vectorImpl.ts
iwoplaza: Unfortunately this doesn't work. I don't remember what error is shown though.

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -423,9 +433,35 @@ export class ResolutionCtxImpl implements ResolutionCtx {
 
     try {
       this.#shaderGenerator.initGenerator(this);
+      const body = this.#shaderGenerator.functionDefinition(options.body);
+
+      let returnType = options.returnType;
+      if (!returnType) {
+        const returnTypes = [...new Set(scope.reportedReturnTypes)];
```
reczkok: Unless we really care about no duplicates in the error `getBestConversion` will create a set under the hood

packages/typegpu/src/shared/meta.ts
```diff
@@ -73,6 +73,13 @@ export function setName(definition: object, name: string): void {
   setMetaData(definition, { name });
 }
 
+export function setNameForward(definition: object, name: string): void {
+  if (isForwarded(definition)) {
+    setNameForward(definition[$getNameForward] as object, name);
+  }
+  setMetaData(definition, { name });
+}
+
```
aleksanderkatan: Shouldn't this actually be the behavior of `setName`? Do we need another function?

packages/typegpu/src/data/vectorImpl.ts
aleksanderkatan: Maybe we could rename it to `getSchema` then? Or `getElementSchema` to be more precise

packages/typegpu/tests/tgsl/shellless.test.ts
aleksanderkatan, line 1: Have a test case
```ts
  it('handles nested shellless', () => {
    const fn1 = () => {
      'kernel';
      return 4.1;
    };

    const fn2 = () => {
      'kernel';
      return fn1();
    };

    const main = tgpu.fn([], d.f32)(() => {
      return fn2();
    });

    expect(asWgsl(main)).toMatchInlineSnapshot(`
      "fn fn1() -> f32 {
        return 4.1;
      }

      fn fn2() -> f32 {
        return fn1();
      }

      fn main() -> f32 {
        return fn2();
      }"
    `);
  });
```

aleksanderkatan (changes requested):
Great job! 🚫🐚

However, I think we really need some docs on this, not only for the users, but also for the team. I needed to create some tests to get a hold of what the expected behavior is like, e.g. that the expected type does not take part in selection of types for the function:
```ts
const f = () => {
  'kernel';
  return 1;
};

const main1 = tgpu.fn([], d.f32)(() => {
  return f();
});
```
(resolved `f` will not return a `f32` as I initially expected)

PR #1719, opened 2025-09-16
docs: Using shell-less functions in examples

iwoplaza:


aleksanderkatan (approved):
Nice! 📝

We should definitely add some publicly available docs about this feature before the next release

PR #1723, opened 2025-09-17
feat: Make std functions more generic

iwoplaza:


packages/typegpu/src/data/vectorOps.ts
aleksanderkatan, line 1: `wgsl.MatchingBoolInstance<T>` can also simplify `select` signature

packages/typegpu/src/std/boolean.ts
```diff
@@ -89,8 +106,20 @@ export const ne = dualImpl({
   codegenImpl: (lhs, rhs) => stitch`(${lhs} != ${rhs})`,
 });
 
-const cpuLt = <T extends AnyNumericVecInstance>(lhs: T, rhs: T) =>
-  VectorOps.lt[lhs.kind](lhs, rhs);
+function cpuLt(lhs: number, rhs: number): boolean;
```
aleksanderkatan, line 109: At some point we decided that if a plain js operator works, then we do not write overloads for std operators so as not to encourage users to complicate their code.
I assume we withdraw that decision to allow for overloaded functions, right? If so, the JSDocs would need updating

aleksanderkatan, line 109: I am not convinced about these new overloads. The following now becomes type safe, and it will generate invalid wgsl. It would be fine if we handled mixed comparisons though. Maybe we could make `unify` cast numbers to vectors by wrapping them?
```ts
function cmpLt(a: number | AnyNumericVecInstance, b: number | AnyNumericVecInstance) {
  "kernel";
  return lt(a, b);
}

cmpLt(3, vec2f());
```

packages/typegpu/src/std/boolean.ts
```diff
@@ -120,15 +152,30 @@ export const lt = dualImpl({
...
-    returnType: correspondingBooleanVectorSchema(argTypes[0]),
-  }),
+  signature: (...args) => {
+    const uargs = unify(args) ?? args;
+    return ({
+      argTypes: uargs,
+      returnType: correspondingBooleanVectorSchema(uargs[0]),
+    });
+  },
   normalImpl: <T extends AnyNumericVecInstance>(lhs: T, rhs: T) =>
```
aleksanderkatan, line 167: This function's interface is now inconsistent with other boolean functions

packages/typegpu/src/std/boolean.ts
```diff
@@ -158,19 +207,28 @@ export const gt = dualImpl({
...
```
aleksanderkatan, line 222: Here as well

aleksanderkatan (changes requested):

iwoplaza:
Too old

PR #1724, opened 2025-09-17
feat: Add a function to deep-compare schemas

### LINKED ISSUE #1550: feat: Add a function to deep-compare schemas

piaccho:
This PR introduces a new deep comparison utility for TypeGPU data schemas and adds tests to ensure correct behavior. 

---

Closes #1550

### DIFF
--- a/apps/typegpu-docs/src/examples/tests/tgsl-parsing-test/index.html
+++ b/apps/typegpu-docs/src/examples/tests/tgsl-parsing-test/index.html
@@ -1,1 +1,1 @@
-<div class='result'>Wait for the tests to finish running...</div>;
+<div class="result">Wait for the tests to finish running...</div>;
--- a/packages/typegpu/src/data/deepEqual.ts
+++ b/packages/typegpu/src/data/deepEqual.ts
@@ -1,0 +1,99 @@
+import type { AnyAttribute } from './attributes.ts';
+import { isLooseData, isUnstruct } from './dataTypes.ts';
+import type { AnyData } from './dataTypes.ts';
+import { isDecorated, isWgslArray, isWgslStruct } from './wgslTypes.ts';
+
+/**
+ * Performs a deep comparison of two TypeGPU data schemas.
+ *
+ * @param a The first data schema to compare.
+ * @param b The second data schema to compare.
+ * @returns `true` if the schemas are deeply equal, `false` otherwise.
+ *
+ * @example
+ * ```ts
+ * import { vec3f, struct, deepEqual } from 'typegpu/data';
+ *
+ * const schema1 = struct({ a: vec3f });
+ * const schema2 = struct({ a: vec3f });
+ * const schema3 = struct({ b: vec3f });
+ *
+ * console.log(deepEqual(schema1, schema2)); // true
+ * console.log(deepEqual(schema1, schema3)); // false
+ * ```
+ */
+export function deepEqual(a: AnyData, b: AnyData): boolean {
+  if (a === b) {
+    return true;
+  }
+
+  if (a.type !== b.type) {
+    return false;
+  }
+
+  if (isWgslStruct(a) && isWgslStruct(b)) {
+    const aProps = a.propTypes;
+    const bProps = b.propTypes;
+    const aKeys = Object.keys(aProps);
+    const bKeys = Object.keys(bProps);
+
+    if (aKeys.length !== bKeys.length) {
+      return false;
+    }
+
+    aKeys.sort();
+    bKeys.sort();
+
+    if (aKeys.join() !== bKeys.join()) {
+      return false;
+    }
+
+    for (const key of aKeys) {
+      if (!deepEqual(aProps[key], bProps[key])) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  if (isWgslArray(a) && isWgslArray(b)) {
+    return (
+      a.elementCount === b.elementCount &&
+      deepEqual(a.elementType as AnyData, b.elementType as AnyData)
+    );
+  }
+
+  if (isDecorated(a) && isDecorated(b)) {
+    if (!deepEqual(a.inner as AnyData, b.inner as AnyData)) {
+      return false;
+    }
+    if (a.attribs.length !== b.attribs.length) {
+      return false;
+    }
+    // TODO: A more robust comparison might be needed if attribute order is not guaranteed.
+    // For now, assuming attributes are ordered.
+    for (let i = 0; i < a.attribs.length; i++) {
+      const attrA = a.attribs[i] as AnyAttribute;
+      const attrB = b.attribs[i] as AnyAttribute;
+      if (attrA.type !== attrB.type) return false;
+      if (attrA.params?.length !== attrB.params?.length) return false;
+      if (attrA.params) {
+        for (let j = 0; j < attrA.params.length; j++) {
+          if (attrA.params[j] !== attrB.params[j]) return false;
+        }
+      }
+    }
+    return true;
+  }
+
+  if (isUnstruct(a) && isUnstruct(b)) {
+    return deepEqual(a, b);
+  }
+
+  if (isLooseData(a) && isLooseData(b)) {
+    // TODO: This is a simplified check. A a more detailed comparison might be necessary.
+    return JSON.stringify(a) === JSON.stringify(b);
+  }
+
+  return true;
+}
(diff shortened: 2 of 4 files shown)


packages/typegpu/src/data/deepEqual.ts
```diff
@@ -0,0 +1,155 @@
...
+    const bProps = b.propTypes;
+    const aKeys = Object.keys(aProps);
+    const bKeys = Object.keys(bProps);
+
+    if (aKeys.length !== bKeys.length) {
+      return false;
+    }
+
+    aKeys.sort();
+    bKeys.sort();
```
iwoplaza: I don't think we should sort keys when comparing for a couple of reasons:
1. It takes computing time
2. Order of keys matters when generating shaders and when serializing
3. The order of keys won't change randomly if the React component rerenders

```suggestion
```

packages/typegpu/src/data/deepEqual.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    if (aKeys.join() !== bKeys.join()) {
+      return false;
+    }
```
iwoplaza: The following for loop should already check if all keys in `a` have the same values in `b`. If the set of keys in `a` doesn't match the set of keys in `b`, we'll get `undefined` in some cases, so it should be fine.

packages/typegpu/src/data/deepEqual.ts
```diff
@@ -0,0 +1,155 @@
...
+
+    // Create comparable string representations for each attribute and sort them
+    // to handle cases where attributes are in a different order.
+    const getAttrKey = (attr: unknown): string => {
+      const anyAttr = attr as AnyAttribute;
+      return `${anyAttr.type}(${(anyAttr.params ?? []).join(',')})`;
+    };
+
+    const sortedAttrsA = a.attribs.map(getAttrKey).sort();
+    const sortedAttrsB = b.attribs.map(getAttrKey).sort();
```
iwoplaza:
```suggestion
    const attrsA = a.attribs.map(getAttrKey);
    const attrsB = b.attribs.map(getAttrKey);
```

packages/typegpu/src/data/deepEqual.ts
```diff
@@ -0,0 +1,155 @@
...
+    const bKeys = Object.keys(bProps);
+
+    if (aKeys.length !== bKeys.length) {
+      return false;
+    }
+
+    // For unstructs, order of properties matters.
+    if (aKeys.join() !== bKeys.join()) {
+      return false;
+    }
```
iwoplaza: I "think" the for below should check if they keys match as well

packages/typegpu/tests/data/deepEqual.test.ts
```diff
@@ -0,0 +1,169 @@
...
+  it('compares struct types', () => {
+    const struct1 = struct({ a: f32, b: vec2u });
+    const struct2 = struct({ a: f32, b: vec2u });
+    const struct3 = struct({ b: vec2u, a: f32 }); // different order
+    const struct4 = struct({ a: u32, b: vec2u }); // different prop type
+    const struct5 = struct({ a: f32, c: vec2u }); // different prop name
+    const struct6 = struct({ a: f32 }); // different number of props
+
+    expect(deepEqual(struct1, struct2)).toBe(true);
+    expect(deepEqual(struct1, struct3)).toBe(true); // property order shouldn't matter
```
iwoplaza: Property order does matter

iwoplaza (changes requested):

packages/typegpu/tests/data/deepEqual.test.ts
piaccho: In this case, two schemas - structures with the same properties but in a different order are considered unequal. If so, shouldn't we check whether the props key arrays of the two structs being compared are identical?

iwoplaza: We can rework the for loop to go through the keys in order, and comparing both the keys and the values.

packages/typegpu/src/data/deepEqual.ts
```diff
@@ -0,0 +1,126 @@
...
+  ) {
+    if (!deepEqual(a.inner as AnyData, b.inner as AnyData)) {
+      return false;
+    }
+    if (a.attribs.length !== b.attribs.length) {
+      return false;
+    }
+
+    // Create comparable string representations for each attribute and sort them
+    // to handle cases where attributes are in a different order.
```
aleksanderkatan: The following test does not pass, is this expected behavior?
```ts
it('ignores decorator order', () => {
  const decorated1 = size(8, align(16, u32));
  const decorated2 = align(16, size(8, u32));

  expect(deepEqual(decorated1, decorated2)).toBe(true);
});
```

piaccho: I think it is. I don't think we should treat them equally schemas because the order can have an impact on the memory layout with attributes like `align` and `size`. 

Analyzing this case:

- `size(8, align(16, u32))` - Takes a `u32`, aligns it to 16 bytes, and then ensures the total space it occupies is 8 bytes. But this is contradictory because the custom size (8) is smaller than the required alignment (16).
- `align(16, size(8, u32))` - Takes a `u32`, pad it to take up 8 bytes, and then align the start of this 8-byte block to a 16-byte boundary. This is a valid.

Please correct me if I’m wrong.

iwoplaza: Both cases are valid, for all attributes currently in WGSL (as far as I know), there is no difference depending on their order. However, it's hard to say whether an attribute will come about that does care about ordering. For that reason, and because our main use-case for this is React re-renders, let's take order into account when checking.

piaccho: Got it! Thanks for an explanation.

packages/typegpu/src/data/deepEqual.ts
```diff
@@ -0,0 +1,125 @@
...
+        return false;
+      }
+    }
+
+    return true;
+  }
+
+  if (isLooseData(a) && isLooseData(b)) {
+    // TODO: This is a simplified check. A a more detailed comparison might be necessary.
+    return JSON.stringify(a) === JSON.stringify(b);
```
aleksanderkatan: At this point, disarray, unstruct and loose-decorated have been handled, so only vertex formats remain.
```suggestion
    return a.type === b.type;
```

packages/typegpu/tests/data/deepEqual.test.ts
```diff
@@ -0,0 +1,173 @@
...
+    const disarray2 = disarrayOf(u32, 4);
+    const disarray3 = disarrayOf(u32, 5);
+
+    expect(deepEqual(unstruct1, unstruct2)).toBe(true);
+    expect(deepEqual(unstruct1, unstruct3)).toBe(false);
+    expect(deepEqual(disarray1, disarray2)).toBe(true);
+    expect(deepEqual(disarray1, disarray3)).toBe(false);
+  });
+
+  it('compares different kinds of types', () => {
```
aleksanderkatan, line 184: Have some free tests (imports still need to be updated)
```suggestion
  });

  it('compares vertex formats', () => {
    expect(deepEqual(uint16x2, uint8x4)).toBe(false);
    expect(deepEqual(uint16x2, uint32)).toBe(false);
    expect(deepEqual(uint16x2, uint16x2)).toBe(true);
  });

  it('compares unstructs with vertex formats', () => {
    const unstruct1 = unstruct({ a: uint16x2 });
    const unstruct2 = unstruct({ a: uint16x2 });
    const unstruct3 = unstruct({ a: uint32 });

    expect(deepEqual(unstruct1, unstruct2)).toBe(true);
    expect(deepEqual(unstruct1, unstruct3)).toBe(false);
  });

  it('compares different kinds of types', () => {
```

packages/typegpu/src/data/deepEqual.ts
aleksanderkatan: Actually, we already checked if the type is equal so we can just remove this entire if and fallback to returning true. Maybe we could include a comment explaining it though.

apps/typegpu-docs/src/examples/tests/tgsl-parsing-test/index.html
```diff
@@ -1 +1 @@
-<div class="result">Wait for the tests to finish running...</div>
+<div class="result">Wait for the tests to finish running...</div>;
```
iwoplaza: The formatter sometimes adds a semicolon at the end for some reason
```suggestion
<div class="result">Wait for the tests to finish running...</div>
```

iwoplaza (approved):
Awesome! Great job 👏

PR #1726, opened 2025-09-18
(WIP) chore: Tweaks to make Smelter work

iwoplaza:
- Explicit literals and naming random functions


packages/typegpu-noise/src/generator.ts
iwoplaza, line 1: Might not be necessary, since we should name object fields

packages/typegpu/src/shared/meta.ts
```diff
@@ -1,3 +1,5 @@
+// The version is inlined during build-time 🎉
```
reczkok, line 1: 🎉🎉

reczkok (approved):

PR #1732, opened 2025-09-19
feat: Support structs and arrays in `console.log`

aleksanderkatan:
1600 lines changed are just tests!

Changes:
- revamp serialization (WGSL side) so that it is no longer stateless but it is simpler to use (it still is O(n^2) due to arguments being copied),
- change `number[]` to `Uint32Array` in deserialization (JS side) so that the deserialization takes O(n) time,
- support structs and arrays in GPU `console.log`,
- slight refactor (`safeStringify.ts`->`stringify.ts`).


iwoplaza (approved):
UUuuuuugee! 👏

PR #1746, opened 2025-09-23
feat: Autoname functions in unplugin

aleksanderkatan:
Changes:
now the following works.
```ts
const myFun1 = () => {
  'kernel';
  return 0;
};
expect(getName(myFun1)).toBe('myFun1');

const myFun2 = function () {
  'kernel';
  return 0;
};
expect(getName(myFun2)).toBe('myFun2');

function myFun3() {
  'kernel';
  return 0;
}
expect(getName(myFun3)).toBe('myFun3');
```


packages/unplugin-typegpu/src/babel.ts
```diff
@@ -2,20 +2,21 @@ import * as Babel from '@babel/standalone';
 import type TemplateGenerator from '@babel/template';
 import type { TraverseOptions } from '@babel/traverse';
 import type * as babel from '@babel/types';
+import defu from 'defu';
```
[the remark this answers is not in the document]
aleksanderkatan: I have auto import optimization enabled. Though I guess it caused some merge conflicts in the past, I'll disable it.

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -139,32 +153,39 @@ function functionVisitor(ctx: Context): TraverseOptions {
...
+      const parent = path.parentPath.node;
+      if (containsKernelDirective(node)) {
+        path.replaceWith(transpileAndAutoname(ctx, node, parent));
         path.skip();
       }
     },
 
     FunctionDeclaration(path) {
       const node = path.node;
+      const parent = path.parentPath.node;
```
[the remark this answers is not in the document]
aleksanderkatan: It is used in line 178

packages/typegpu/src/shared/meta.ts
```diff
@@ -30,11 +30,18 @@ export type INTERNAL_GlobalExt = typeof globalThis & {
...
+      isNamable(exp) &&
+      (exp as unknown as { [$internal]: unknown })?.[$internal] &&
+      !getName(exp)
+    ) {
+      exp.$name(label);
+    } else if (typeof exp === 'function') {
+      setName(exp, label);
+    }
+    return exp;
+  },
```
iwoplaza, line 38: I would be tempted to call `setName` in all cases, as it now follows the chain of `$getNameForward`.
I would also move `isMarkedInternal` to `symbols.ts` just so that's it's more local to where the symbol is defined and to avoid circular deps, and use it here as well.

```suggestion
  '__TYPEGPU_AUTONAME__': <T>(exp: T, label: string): T => {
    if (isMarkedInternal(exp) && !getName(exp)) {
      setName(exp, label);
    }
    return exp;
  },
```

packages/unplugin-typegpu/test/auto-naming.test.ts
```diff
@@ -576,4 +632,58 @@ describe('[ROLLUP] auto naming', () => {
...
+
+    expect(await rollupTransform(code, { autoNamingEnabled: true }))
+      .toMatchInlineSnapshot(`
+        "const myFun1 = ((globalThis.__TYPEGPU_AUTONAME__ ?? (a => a))((($ => (globalThis.__TYPEGPU_META__ ??= new WeakMap()).set($.f = (() => {
+                'kernel';
+                return 0;
+              }), {
+                      v: 1,
+                      ast: {"params":[],"body":[0,[[10,[5,"0"]]]],"externalNames":[]},
+                      get externals() { return {}; },
```
iwoplaza, line 678: The __TYPEGPU_AUTONAME__ trick works because all but one case, we can depend on typegpu to be imported before we call it.

In the case of kernel functions, that's not always the case. We can have a kernel function that's in a file that does not depend on typegpu being imported prior. That's why I would suggest we take a slightly different approach in naming kernel functions (both shell-full and shell-less).

We can just add another property to the meta we associate with the kernel function:
```ts
{
  v: 1,
  name: "myFun1",
  ast: {"params":[],"body":[0,[[10,[5,"0"]]]],"externalNames":[]},
  get externals() { return {}; },
}
```

packages/typegpu/src/shared/meta.ts
aleksanderkatan: Will the functions passed in here be marked as internal though?

packages/unplugin-typegpu/test/auto-naming.test.ts
aleksanderkatan: Yeah this seems cleaner

packages/typegpu/src/shared/meta.ts
aleksanderkatan: I don't think we can do this, `$name` does more than `setName` in some cases.

image

packages/unplugin-typegpu/src/index.ts
```diff
@@ -198,6 +193,7 @@ const typegpu: UnpluginInstance<Options, false> = createUnplugin(
 
             const metadata = `{
               v: ${FORMAT_VERSION},
+              name: ${name ? `"${name}"` : 'undefined'},
```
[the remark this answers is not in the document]
aleksanderkatan: w/e

packages/typegpu/src/shared/meta.ts
iwoplaza: Okay, agreed!

iwoplaza (approved):
👏 !Awesome! 👏

PR #1748, opened 2025-09-24
feat(@typegpu/react): Add `useMirroredUniform` hook

### LINKED ISSUE #1714: feat(@typegpu/react): useMirroredUniform
The `useMirroredUniform` hook is used to "synchronise"/"mirror" a JS value from React onto the GPU.

| |useUniformValue|useMirroredUniform|
|-|-|-|
|mutable .value|✅|❌|
|reacts to change in parameters (schema, value)|❌|✅|

```ts
function Component(props: { color: d.v3f }) {
  // The uniform will be updated with the latest value of `props.color`
  // on every React render.
  const color = useMirroredUniform(d.vec3f, props.color);

  const { ref } = useRender({
    fragment: () => {
      'kernel';
      return d.vec4f(color.$, 1);
    },
  });

  return <canvas ref={ref}></canvas>
}
```

Doing the same with useUniformValue would require manually syncing the value either in a useEffect, or on every frame.

piaccho:
## To-do

- [x] Implement the `useMirroredUniform` hook
- [x] Add tests for the hook
- [ ] Add an example of usage of a new hook 

---

Closes #1714

### DIFF
--- a/packages/typegpu-react/src/use-mirrored-uniform.ts
+++ b/packages/typegpu-react/src/use-mirrored-uniform.ts
@@ -1,4 +1,4 @@
-import type * as d from 'typegpu/data';
+import * as d from 'typegpu/data';
 import { useRoot } from './root-context.tsx';
 import { useEffect, useMemo, useRef, useState } from 'react';
 import type { ValidateUniformSchema } from 'typegpu';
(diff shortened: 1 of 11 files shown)


packages/typegpu-react/src/use-mirrored-uniform.ts
```diff
@@ -0,0 +1,60 @@
...
+  useEffect(() => {
+    let currentBuffer = uniformBuffer;
+    if (!d.deepEqual(prevSchemaRef.current as d.AnyData, schema as d.AnyData)) {
+      currentBuffer.buffer.destroy();
+      currentBuffer = root.createUniform(schema, value);
+      setUniformBuffer(currentBuffer);
+      prevSchemaRef.current = schema;
+    }
+
+    currentBuffer.write(value);
```
iwoplaza: Tiny nit, we don't need to write to buffers that have just been created, as they're initialized with the new value

```suggestion
    if (!d.deepEqual(prevSchemaRef.current as d.AnyData, schema as d.AnyData)) {
      uniformBuffer.buffer.destroy();
      setUniformBuffer(root.createUniform(schema, value));
      prevSchemaRef.current = schema;
    } else {
      uniformBuffer.write(value);
    }
```

packages/typegpu-react/src/use-mirrored-uniform.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  const mirroredValue = useMemo(
+    () => ({
+      schema,
+      get $() {
+        return uniformBuffer.$;
+      },
+    }),
+    [schema, uniformBuffer],
+  );
```
iwoplaza, line 66: This value is not stable (depends on a possible unstable `schema` value). I would suggest to keep a `currentSchemaRef` and update it during rendering only if the deepEquals no longer passes

```ts
const currentSchemaRef = useRef(schema);
if (!d.deepEqual(currentSchemaRef.current, schema)) {
  currentSchemaRef.current = schema;
}
```

And then use that value in the dependency array:

```ts
[currentSchemaRef.current, uniformBuffer]
```

You can add comments explaining why you're writing that kind of code, since this seems like betraying rules of React, but you gotta do what you gotta do 😅

PR #1752, opened 2025-09-25
fix: `dualImpl` throws with missing `cpuImpl`

aleksanderkatan:
Changes:
- add `MissingCpuImplError`,
- allow strings as `cpuImpl` in `dualImpl`,
- fallback to `codegenImpl` when in codegen mode and cpuImpl is missing or throws `MissingCpuImplError`,
- rewrite some std functions to make use of the new utility.

### DIFF
--- a/packages/typegpu/package.json
+++ b/packages/typegpu/package.json
@@ -75,7 +75,6 @@
     "tsup": "catalog:build",
     "typescript": "catalog:types",
     "unplugin-typegpu": "workspace:*",
-    "wesl": "0.6.7",
     "wgpu-matrix": "catalog:example"
   },
   "packageManager": "pnpm@10.15.1+sha512.34e538c329b5553014ca8e8f4535997f96180a1d0f614339357449935350d924e22f8614682191264ec33d1462ac21561aff97f6bb18065351c162c7e8f6de67",
--- a/packages/typegpu/src/data/deepEqual.ts
+++ b/packages/typegpu/src/data/deepEqual.ts
@@ -1,0 +1,114 @@
+import type { AnyAttribute } from './attributes.ts';
+import { isDisarray, isLooseDecorated, isUnstruct } from './dataTypes.ts';
+import type { AnyData } from './dataTypes.ts';
+import {
+  isAtomic,
+  isDecorated,
+  isPtr,
+  isWgslArray,
+  isWgslStruct,
+} from './wgslTypes.ts';
+
+/**
+ * Performs a deep comparison of two TypeGPU data schemas.
+ *
+ * @param a The first data schema to compare.
+ * @param b The second data schema to compare.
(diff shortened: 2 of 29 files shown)


packages/typegpu/src/std/numeric.ts
```diff
@@ -556,6 +544,10 @@ export const fma = dualImpl({
 function cpuFract(value: number): number;
 function cpuFract<T extends AnyFloatVecInstance>(value: T): T;
 function cpuFract<T extends AnyFloatVecInstance | number>(value: T): T {
+  // biome-ignore lint/correctness/noConstantCondition lint/suspicious/noSelfCompare: this may potentially happen
+  if (true !== true) {
+    throw new Error('A random space ray has altered a bit on your PC. Sorry.');
+  }
```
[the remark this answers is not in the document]
aleksanderkatan: :c

packages/typegpu/src/std/numeric.ts
```diff
@@ -264,15 +268,14 @@ function cpuCountLeadingZeros<T extends AnyIntegerVecInstance>(value: T): T;
 function cpuCountLeadingZeros<T extends AnyIntegerVecInstance | number>(
   value: T,
 ): T {
-  throw new Error(
-    'CPU implementation for countLeadingZeros not implemented yet. Please submit an issue at https://github.com/software-mansion/TypeGPU/issues',
-  );
+  throw new MissingCpuImplError('countLeadingZeros');
```
iwoplaza: When I first looked at this, I thought that's the message was going to be shown to the user, but now I can see `cpuCountLeadingZeros` is only being used for types. Maybe we should make this more explicit to the devs seeing the code?

```suggestion
  throw new Error('Unreachable, only used as types');
```

packages/typegpu/src/core/function/dualImpl.ts
```diff
@@ -74,11 +81,22 @@ export function dualImpl<T extends (...args: never[]) => unknown>(
...
+    ) {
+      try {
+        return snip(
+          options.normalImpl(...converted.map((s) => s.value) as never[]),
+          returnType,
+        );
+      } catch (e) {
+        // if cpuImpl is not yet implemented, fallback to codegenImpl
+        // otherwise, rethrow error
+        if (!(e instanceof MissingCpuImplError)) {
```
reczkok, line 97: It took me embarrassingly long to figure out why we handle the case of a function CPU impl that throws a “missing CPU implementation” error. I later realized it is for partially implemented functions. I would leave a short comment here explaining it for future reference. If we ever get rid of partial implementations we can simplify this logic a bit and save some runtime work for each called dualimpl unless I am missing some other cases.

reczkok (approved):
Looks good!

PR #1755, opened 2025-09-26
feat: Ref/Value behavior tracking

iwoplaza:


aleksanderkatan:
Something weird happened to the vaporrave example, some of the effects seem to have disappeared.

apps/typegpu-docs/src/examples/rendering/3d-fish/compute.ts
```diff
@@ -130,5 +130,5 @@ export const computeShader = tgpu['~unstable'].computeFn({
     fishData.direction,
   );
   fishData.position = std.add(fishData.position, translation);
-  layout.$.nextFishData[fishIndex] = fishData;
+  layout.$.nextFishData[fishIndex] = ModelData(fishData);
```
aleksanderkatan: This probably triggers the weird chromium offset copy bug. Above, we explicitly copy props one by one to avoid that.

apps/typegpu-docs/src/examples/rendering/3d-fish/index.ts
```diff
@@ -118,8 +118,8 @@ const randomizeFishPositionsOnGPU = prepareDispatch(root, (x) => {
     applySeaFog: 1,
     applySeaDesaturation: 1,
   });
-  buffer0mutable.$[x] = data;
-  buffer1mutable.$[x] = data;
+  buffer0mutable.$[x] = ModelData(data);
+  buffer1mutable.$[x] = ModelData(data);
```
aleksanderkatan: Here as well. Although, we probably can just remove the workaround, because it seems that more and more devices are fixed over time. Also, the issue on chromium website has been stagnant for the past 2 months.

apps/typegpu-docs/src/examples/simulation/stable-fluid/simulation.ts
```diff
@@ -111,10 +111,10 @@ export const advectFn = tgpu['~unstable'].computeFn({
   const clampedPos = std.clamp(
     prevPos,
     d.vec2f(-0.5),
-    d.vec2f(std.sub(d.vec2f(texSize.xy), d.vec2f(0.5))),
+    d.vec2f(texSize.xy).sub(d.vec2f(0.5)),
```
aleksanderkatan: nit
```suggestion
    d.vec2f(texSize.xy).sub(0.5),
```

apps/typegpu-docs/src/examples/simulation/stable-fluid/simulation.ts
```diff
@@ -111,10 +111,10 @@ export const advectFn = tgpu['~unstable'].computeFn({
...
   );
   const normalizedPos = std.div(
-    std.add(clampedPos, d.vec2f(0.5)),
+    clampedPos.add(d.vec2f(0.5)),
```
aleksanderkatan: nit
```suggestion
    clampedPos.add(0.5),
```

packages/typegpu/tests/examples/individual/3d-fish.test.ts
aleksanderkatan, line 1: Did we do performance tests regarding many dereferences?

packages/typegpu/tests/examples/individual/3d-fish.test.ts
```diff
@@ -218,8 +218,8 @@ describe('3d fish example', () => {
 
       fn applySinWave_4(index: u32, vertex: PosAndNormal_3, time: f32) -> PosAndNormal_3 {
         var a = -60.1;
-        var b = 0.8;
-        var c = 6.1;
+        const b = 0.8;
+        const c = 6.1;
```
aleksanderkatan, line 232: Why was `a` not turned into a const? In js the code looks as follows:
```ts
const a = -60.1;
const b = 0.8;
const c = 6.1;
```

packages/typegpu/tests/examples/individual/caustics.test.ts
```diff
@@ -106,34 +106,41 @@ describe('caustics example', () => {
...
+        var noise = abs(sample_8(vec3f((uv2 * 5), time2)));
+        return pow(vec3f((1 - noise)), profile);
+      }
+
+      fn caustics_17(uv: vec2f, time2: f32, profile: vec3f) -> vec3f {
         var distortion = sample_8(vec3f((uv * 0.5), (time2 * 0.2)));
         var uv2 = (uv + distortion);
         var noise = abs(sample_8(vec3f((uv2 * 5), time2)));
         return pow(vec3f((1 - noise)), profile);
       }
```
aleksanderkatan: Are there any potential side effects to this kind of situation? If so, do we want to warn users when a situation like this occurs? This is the result of the following: 
```ts
// Transforming coordinates to simulate perspective squash
  const cuv = d.vec2f(
    uv.x * (std.pow(uv.y * 1.5, 3) + 0.1) * 5,
    std.pow((uv.y * 1.5 + 0.1) * 1.5, 3) * 1,
  );
  // Generating two layers of caustics (large scale, and small scale)
  const c1 = std.mul(
    caustics(cuv, time.$ * 0.2, /* profile */ d.vec3f(4, 4, 1)),
    // Tinting
    d.vec3f(0.4, 0.65, 1),
  );
  const c2 = std.mul(
    caustics(std.mul(cuv, 2), time.$ * 0.4, /* profile */ d.vec3f(16, 1, 4)),
    // Tinting
    d.vec3f(0.18, 0.3, 0.5),
  );
```

packages/typegpu/src/data/snippet.ts
```diff
@@ -3,13 +3,51 @@ import type { AnyData, UnknownData } from './dataTypes.ts';
 import { DEV } from '../shared/env.ts';
 import { isNumericSchema } from './wgslTypes.ts';
 
+export type RefSpace =
```
aleksanderkatan: I think there is an collision of terms, "space" means two different things - the set of possible reference types, and a place in GPU memory. I would rename this type, for example to "RefVariant"

packages/typegpu/src/data/array.ts
```diff
@@ -53,7 +53,7 @@ export const arrayOf = createDualImpl(
       // Marking so the WGSL generator lets this function through
       partial[$internal] = true;
 
-      return snip(partial, UnknownData);
+      return snip(partial, UnknownData, /* ref*/ 'runtime');
```
aleksanderkatan: Should we actually require `RefSpace` from `UnknownData`? Are there cases where this makes sense, or is the 'runtime' just so that something is there?

packages/typegpu/src/data/matrix.ts
```diff
@@ -553,7 +560,7 @@ export const identity2 = createDualImpl(
   // CPU implementation
   () => mat2x2f(1, 0, 0, 1),
   // CODEGEN implementation
-  () => snip('mat2x2f(1, 0, 0, 1)', mat2x2f),
+  () => snip('mat2x2f(1, 0, 0, 1)', mat2x2f, /* ref */ 'runtime'),
```
aleksanderkatan: This is a constant matrix, can it be 'constant'?

packages/typegpu/src/data/vectorImpl.ts
```diff
@@ -41,12 +41,12 @@ export abstract class VecBase<S> extends Array implements SelfResolvable {
   [$resolve](): ResolvedSnippet {
     const schema = this[$internal].elementSchema;
     if (this.every((e) => !e)) {
-      return snip(`${this.kind}()`, schema);
+      return snip(`${this.kind}()`, schema, /* ref */ 'runtime');
```
aleksanderkatan: Can this be 'constant'?

packages/typegpu/src/tgsl/consoleLog/logGenerator.ts
```diff
@@ -41,7 +41,7 @@ export class LogGeneratorNullImpl implements LogGenerator {
     console.warn(
       "'console.log' is currently only supported in compute pipelines.",
```
aleksanderkatan: This is a little outdated
```suggestion
      "'console.log' is only supported when resolving pipelines.",
```

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -313,19 +319,23 @@ export function tryConvertSnippet(
...
+      /* ref */ snippet.ref,
+    );
   }
 
   const converted = convertToCommonType([snippet], [targetDataType], verbose);
 
   if (!converted) {
     throw new WgslTypeError(
-      `Cannot convert value of type '${snippet.dataType.type}' to type '${targetDataType.type}'`,
+      `Cannot convert value of type '${snippet.dataType}' to type '${targetDataType.type}'`,
```
aleksanderkatan: Why change this? Now this will display stuff like this:
```ts
(...args) => {
    if (inCodegenMode()) {
      return gpuImpl(...args);
    }
    if (typeof options.normalImpl === "string") {
      throw new MissingCpuImplError(options.normalImpl);
    }
    ret…
```

packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -106,36 +118,132 @@ const kindToSchema = {
...
+        ? 'constant'
+        : 'runtime',
+    );
+  }
+
+  if (target.dataType.type === 'bool' || isNumericSchema(target.dataType)) {
     // No props to be accessed here
-    return UnknownData;
+    return undefined;
   }
```
aleksanderkatan: Why do we worry about this?

packages/typegpu/src/tgsl/shellless.ts
```diff
@@ -3,19 +3,30 @@ import {
...
 import { concretize } from './generationHelpers.ts';
 
-interface ShelllessVariant {
-  argTypes: AnyData[];
-  value: ShelllessImpl;
-}
-
 type AnyFn = (...args: never[]) => unknown;
 
+function shallowEqualSchemas(a: AnyData, b: AnyData): boolean {
```
aleksanderkatan, line 16: Different structs will be equal. I'd rename this, though I have no ideas right now

packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -92,7 +96,7 @@ class TgpuConstImpl<TDataType extends AnyWgslData>
       return this[$gpuValueOf];
     }
 
-    return this.#value;
+    return schemaCallWrapper(this.dataType, this.#value);
```
aleksanderkatan: I'm worried about the performance of this. Sometimes these constants are not insignificant in size, e.g. in cube-marching, I use a `tgpu.const(d.arrayOf(d.u32, 256), ...)`. I don't use it on JS side, but still, if we allow that, it should not be unnecessarily slow.

packages/typegpu/src/core/function/dualImpl.ts
```diff
@@ -52,6 +48,12 @@ interface DualImplOptions<T extends (...args: never[]) => unknown> {
     | ((
       ...inArgTypes: MapValueToDataType<Parameters<T>>
     ) => { argTypes: AnyData[]; returnType: AnyData });
+  /**
+   * Whether the function should skip trying to execute the "normal" implementation if
+   * all arguments are known at compile time.
+   * @default false
+   */
+  readonly noComptime?: boolean | undefined;
```
aleksanderkatan, line 56: We already kind of have this mechanism, if the "normal" impl is a string, or if the "normal" impl throws a "MissingCpuImplError". Is it necessary to add a second way to do this? If so, can we get rid of the old way?

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -135,7 +136,7 @@ export function createFnCore(
         }
 
         ctx.addDeclaration(`${fnAttribute}fn ${id}${header}${body}`);
-        return snip(id, returnType);
+        return snip(id, returnType, /* ref */ 'runtime');
```
aleksanderkatan: I'm a little hesitant about calling everything 'runtime' bu default, maybe we could find a more fitting name, like 'unknown'?

packages/typegpu/src/data/dataTypes.ts
```diff
@@ -134,6 +134,21 @@ export function undecorate(data: AnyData): AnyData {
...
 }
 
+export function unptr(data: AnyData): AnyData {
+  if (data.type === 'ptr') {
+    return data.inner as AnyData;
+  }
+  return data;
+}
+
+export function toStorable(schema: AnyData): AnyData {
```
aleksanderkatan: Could you add some docs to this? At a glance, I don't know what it does e.g. to atomic types.

packages/typegpu/tests/examples/individual/3d-fish.test.ts
iwoplaza: We didn't, but since pointers are very restrictive in WGSL, I can't imagine they don't do any optimization on it.

If they do incur a performance hit, we can always implement usage-tracking (if a variable is used read-only in a function, we don't have to pass it as a pointer), but maybe it won't come to that

packages/typegpu/src/data/vectorImpl.ts
iwoplaza: I think so 🤔. I also think that we can improve constant tracking in operators, but that would have to be another PR.

packages/typegpu/src/tgsl/generationHelpers.ts
iwoplaza: Good point, it should fall through to `return undefined;` at the end of the function

packages/typegpu/src/core/function/dualImpl.ts
iwoplaza: We still want the normal impl to be called outside of codegen mode, that's the difference.

aleksanderkatan: Makes sense

apps/typegpu-docs/src/examples/rendering/3d-fish/index.ts
```diff
@@ -98,30 +98,31 @@ function enqueuePresetChanges() {
...
-      variant: randf.sample(),
-      applySinWave: 1,
-      applySeaFog: 1,
-      applySeaDesaturation: 1,
-    });
-    buffer0mutable.$[x] = data;
-    buffer1mutable.$[x] = data;
+const randomizeFishPositionsPipeline = root[
+  '~unstable'
+].createGuardedComputePipeline((x) => {
```
reczkok: I would revert the formatting here

apps/typegpu-docs/src/examples/rendering/3d-fish/index.ts
```diff
@@ -199,8 +200,9 @@ let depthTexture = root.device.createTexture({
   usage: GPUTextureUsage.RENDER_ATTACHMENT,
 });
 
-const simulatePipeline = root['~unstable']
-  .createGuardedComputePipeline(simulate);
+const simulatePipeline = root['~unstable'].createGuardedComputePipeline(
+  simulate,
+);
```
reczkok: Formatting also seems worse here

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
```diff
@@ -0,0 +1,199 @@
...
+  return Boid({ pos: vec3(), vel: vec3(0, 1, 0) });
+};
+
+const boid = createBoid();
+//    ^?
+```
+
+:::caution
+We could do everything at runtime, transforming the code a bit so that the TypeGPU shader generator can have access to
+a function's JS source code, along with references to values referenced from the outer scope.
```
aleksanderkatan, line 44: I don't understand this sentence. Did you mean that we could parse the JS to AST at runtime? If so, is this warn necessary if it is already mentioned below?

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
aleksanderkatan, line 1: Great docs!

Do we plan on publishing this or is this just internal?

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
```diff
@@ -0,0 +1,199 @@
...
+  get externals() { return { Boid, vec3 }; },
+});
+```
+
+## Snippets
+
+Snippets are the basis for TypeGPU shader code generation. They are immutable objects that hold three values:
+- *value*: A piece of WGSL code, or something "resolvable" to a piece of WGSL code
+- *dataType*: The inferred WGSL type of `value` [(more here)](#data-types)
+- *origin*: An enumerable of where the value came from (if it's a reference to an existing value, or ephemeral)
```
cieplypolar:
```suggestion
- *origin*: An enumerable of where the `value` came from (if it's a reference to an existing value, or ephemeral)
```

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
```diff
@@ -0,0 +1,199 @@
...
+    const newBoid = Boid();
+
+    const boidPos = newBoid.pos;
+  };
+  ```
+  Generates:
+  ```wgsl
+  fn foo() {
+    let firstBoid = (&boids[0]); // typed as ptr<storage, Boid, read_write>
+    var newBoid = Boid(); // typed as Boid
```
aleksanderkatan, line 155: Could you also show what `Boid(firstBoid)` resolves to?

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
```diff
@@ -0,0 +1,199 @@
...
+  fn advance(pos: vec3f) {
+    pos.x += 1;
+  }
+
+  fn main() {
+    let pos = vec3f(0, 0, 0);
+    advance(pos);
+    // pos.x === 0 in WGSL
+  }
+  ```
```
aleksanderkatan:
```
- <root>
- fn*:main
- fn*:main()
- fn*:advance(vec3f): 'pos.x += 1f' is invalid, because non-pointer arguments cannot be mutated.
```

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
```diff
@@ -0,0 +1,199 @@
...
+
+There are essentially three types of origins:
+- **Ephemeral Origins**: These origins represent values that are created or derived from other values. They are typically used for creating new instances or
+  performing operations that produce new values. Examples include creating a new `Boid` instance or calculating a new position based on an existing one. These
+  include `'runtime'` and `'constant'`.
+- **Referential Origins**: These origins represent references to existing values. They are typically used for accessing or modifying existing data. Examples
+  include accessing the position of an existing `Boid` instance or modifying the position of an existing `Boid` instance. These include `'uniform'`, `'mutable'`, `'readonly'`, `'workgroup'`, `'private'`, `'function'`, `'handle'` and `'constant-ref'`.
+- **Argument Origins**: This group is dedicated to exactly one origin: 'argument'. It represents values that are passed as arguments to functions.
```
aleksanderkatan, line 205: Could you elaborate in one sentence each about `constant` and `constant-ref`? Others I think are clear from the name.

apps/typegpu-docs/src/examples/rendering/3d-fish/compute.ts
```diff
@@ -1,22 +1,12 @@
...
-    position: layout.$.currentFishData[fishIndex].position,
-    direction: layout.$.currentFishData[fishIndex].direction,
-    scale: layout.$.currentFishData[fishIndex].scale,
-    variant: layout.$.currentFishData[fishIndex].variant,
-    applySeaDesaturation:
-      layout.$.currentFishData[fishIndex].applySeaDesaturation,
-    applySeaFog: layout.$.currentFishData[fishIndex].applySeaFog,
-    applySinWave: layout.$.currentFishData[fishIndex].applySinWave,
-  });
+  const fishData = layout.$.currentFishData[fishIndex];
```
aleksanderkatan, line 9: There is still one of these left in the `render.ts` file.

packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -43,16 +52,35 @@ export function constant<TDataType extends AnyWgslData>(
 // Implementation
 // --------------
 
+function deepFreeze<T extends object>(object: T): T {
+  // Retrieve the property names defined on object
+  const propNames = Reflect.ownKeys(object);
```
aleksanderkatan, line 57: We don't want to freeze props under symbols, right?

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -75,7 +88,16 @@ export class TgpuAccessorImpl<T extends AnyWgslData>
...
+    const ownSnippet = getOwnSnippet(value);
+    if (ownSnippet) {
+      return ownSnippet;
+    }
+
+    // Doing a deep copy each time so that we don't have to deal with refs
+    return schemaCallWrapper(
+      this.schema,
+      snip(value, this.schema, /* ref */ 'constant'),
+    );
```
aleksanderkatan, line 100: Does this work? `schemaCallWrapper` calls `schema(item)`, and I don't think that schema constructors handle snippets well...

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -725,18 +814,117 @@ ${this.ctx.pre}else ${alternate}`;
...
+- Try 'const ${rawId} = ${rhsStr}' if you won't reassign '${rawId}' later.
+-----`,
+            );
+          }
+          varType = 'let';
+        }
+
+        if (stmtType === NODE.const) {
+          if (eq.origin === 'argument') {
+            // Arguments cannot be mutated, so we 'let' them be (kill me)
```
aleksanderkatan, line 905: 💀

packages/typegpu/src/tgsl/shellless.ts
```diff
@@ -31,23 +49,67 @@ export class ShelllessRepository {
...
+        const ctx = getResolutionCtx()!;
+        throw new Error(
+          `Cannot pass constant references as function arguments. Explicitly copy them by wrapping them in a schema: '${
+            ctx.resolve(type).value
+          }(...)'`,
+        );
+      }
+
+      if (isPtr(type) && type.implicit) {
+        // If the pointer was made implicitly (e.g. by assigning a reference to a const variable),// then we dereference the pointer before passing it to the function. The main reason for this,
```
aleksanderkatan:
```suggestion
        // If the pointer was made implicitly (e.g. by assigning a reference to a const variable),
        // then we dereference the pointer before passing it to the function. The main reason for this,
```

packages/typegpu/src/data/ref.ts
```diff
@@ -0,0 +1,183 @@
...
+  Object.defineProperty(impl, $internal, {
+    value: {
+      jsImpl,
+      gpuImpl,
+      strictSignature: undefined,
+      argConversionHint: 'keep',
+    },
+  });
+
+  return impl as unknown as DualFn<<T>(value: T) => ref<T>>;
```
aleksanderkatan, line 101: Can we use `createDualImpl`?

apps/typegpu-docs/src/examples/simulation/gravity/compute.ts
```diff
@@ -9,18 +9,17 @@ const { none, bounce, merge } = collisionBehaviors;
 
 // tiebreaker function for merges and bounces
 const isSmaller = tgpu.fn([d.u32, d.u32], d.bool)((currentId, otherId) => {
-  if (
-    computeLayout.$.inState[currentId].mass <
-      computeLayout.$.inState[otherId].mass
-  ) {
+  const current = computeLayout.$.inState[currentId];
```
aleksanderkatan, line 12: It still is weird to me that this translates to a pointer. It makes sense in JS though. This may just be me coming from C, `t[i]` being equivalent to `*(t+i)`.

packages/typegpu/tests/examples/individual/log-test.test.ts
```diff
@@ -798,14 +798,14 @@ describe('console log example', () => {
...
         log4_17(3000000000u);
         log5_20(true);
         log6_23();
         log7_25();
-        log8_27(vec2f(1.1f, -2.2));
-        log9_30(vec3f(10.1f, -20.2, 30.3f));
-        log10_33(vec4f(100.1f, -200.2, 300.3f, -400.4));
+        log8_27(vec2f(1.100000023841858, -2.200000047683716));
+        log9_30(vec3f(10.100000381469727, -20.200000762939453, 30.299999237060547));
+        log10_33(vec4f(100.0999984741211, -200.1999969482422, 300.29998779296875, -400.3999938964844));
```
aleksanderkatan, line 808: I didn't catch it, what change causes these to go through casting again?

packages/typegpu/tests/tgsl/argumentOrigin.test.ts
```diff
@@ -0,0 +1,148 @@
...
+
+    const main = () => {
+      'use gpu';
+      foo(1);
+    };
+
+    expect(asWgsl(main)).toMatchInlineSnapshot(`
+      "fn foo(a: i32) {
+        a += 1i;
+      }
```
aleksanderkatan: This is not a valid WGSL.

```
Error while parsing WGSL: :4:9 error: cannot assign to parameter 'a'
        a += 1i;
        ^

:4:9 note: parameters are immutable
        a += 1i;
        ^

:3:8 note: parameter 'a' declared here
fn foo(a: i32) {
       ^
```

packages/typegpu/tests/tgsl/argumentOrigin.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    };
+
+    expect(asWgsl(main)).toMatchInlineSnapshot(`
+      "struct Foo {
+        a: f32,
+      }
+
+      fn foo(_arg_0: Foo) {
+        _arg_0.a += 1f;
+      }
```
aleksanderkatan: Here as well. Immutable is immutable.
```
Error while parsing WGSL: :7:9 error: cannot assign to value of type 'f32'
        _arg_0.a += 1f;
        ^^^^^^^^

:7:9 note: parameters are immutable
        _arg_0.a += 1f;
        ^^^^^^

:6:14 note: parameter '_arg_0' declared here
      fn foo(_arg_0: Foo) {
```

packages/typegpu/src/core/constant/tgpuConstant.ts
cieplypolar: I believe we don't need to worry. I'm almost certain that all symbolic properties are references.

packages/typegpu/src/core/function/dualImpl.ts
```diff
@@ -68,27 +70,36 @@ export function dualImpl<T extends (...args: never[]) => unknown>(
...
+      !options.noComptime &&
+      converted.every((s) => isKnownAtComptime(s)) &&
       typeof options.normalImpl === 'function'
     ) {
       try {
         return snip(
           options.normalImpl(...converted.map((s) => s.value) as never[]),
           returnType,
+          // Functions give up ownership of their return value
+          /* ref */ 'constant',
```
cieplypolar: origin
```suggestion
          /* origin */ 'constant',
```

packages/typegpu/src/core/function/dualImpl.ts
```diff
@@ -100,7 +111,12 @@ export function dualImpl<T extends (...args: never[]) => unknown>(
       }
     }
 
-    return snip(options.codegenImpl(...converted), returnType);
+    return snip(
+      options.codegenImpl(...converted),
+      returnType,
+      // Functions give up ownership of their return value
+      /* ref */ 'runtime',
```
cieplypolar:
```suggestion
      /* origin */ 'runtime',
```

packages/typegpu/tests/examples/individual/3d-fish.test.ts
iwoplaza: It's now a constant. It was caused by the `-` unary operator resolving too early, returning a 'runtime' origin snippet

packages/typegpu/tests/examples/individual/caustics.test.ts
iwoplaza: No longer happens thanks to explicit refs and derefs 🫡

packages/typegpu/src/data/array.ts
iwoplaza: It's just arbitrary, though it would make sense for it to be `constant`, as it doesn't change at shader runtime...

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -135,7 +137,7 @@ export function createFnCore(
...
```
cieplypolar:
```suggestion
        return snip(id, returnType, /* origin */ 'runtime');
```

packages/typegpu/src/core/function/fnCore.ts
cieplypolar: For me, `runtime` is fine

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
iwoplaza: Just internal for now, might take stuff out from it and plug it into the main docs, or create a more "Advanced" guide

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
iwoplaza: Yes, we could do the parsing at runtime, but we choose not to.

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
iwoplaza: This is not a real example, it's to showcase how the 'argument' origin guards us from misusing passing referential values to functions.

packages/typegpu/src/core/slot/accessor.ts
iwoplaza: It's called in GPU mode, and since they're dualImpl, they expect to be called with snippets

packages/typegpu/tests/examples/individual/log-test.test.ts
iwoplaza: I think it's because now it's a `dualImpl`, not a `createDualImpl`

packages/typegpu/src/data/ref.ts
iwoplaza: I tried, but the API of `dualImpl` is a tiny bit too abstracted away for `ref`

packages/typegpu/src/core/variable/tgpuVariable.ts
```diff
@@ -117,11 +122,12 @@ class TgpuVarImpl<TScope extends VariableScope, TDataType extends AnyData>
 
   get [$gpuValueOf](): InferGPU<TDataType> {
     const dataType = this.#dataType;
+    const ref = isNaturallyEphemeral(dataType) ? 'runtime' : this.#scope;
```
cieplypolar:
```suggestion
    const origin = isNaturallyEphemeral(dataType) ? 'runtime' : this.#scope;
```

packages/typegpu/src/data/disarray.ts
```diff
@@ -59,7 +59,7 @@ export const disarrayOf = createDualImpl(
       // Marking so the WGSL generator lets this function through
       partial[$internal] = true;
 
-      return snip(partial, UnknownData);
+      return snip(partial, UnknownData, /* origin */ 'runtime');
```
cieplypolar:
```suggestion
      return snip(partial, UnknownData, /* origin */ 'constant');
```
Based on the array implementation. I'm not sure if this is correct, certainly makes sense, because `partial` function is compTime constant.

packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -43,16 +52,35 @@ export function constant<TDataType extends AnyWgslData>(
...
+  readonly #value: DeepReadonly<InferGPU<TDataType>>;
 
   constructor(
     public readonly dataType: TDataType,
     value: InferGPU<TDataType>,
   ) {
-    this.#value = value;
+    this.#value = value && typeof value === 'object'
+      ? deepFreeze(value) as DeepReadonly<InferGPU<TDataType>>
+      : value as DeepReadonly<InferGPU<TDataType>>;
```
reczkok, line 83: Why exactly do we care about freezing the value that was passed in?

packages/typegpu/src/data/dataTypes.ts
```diff
@@ -135,6 +135,20 @@ export function undecorate(data: AnyData): AnyData {
...
+  if (data.type === 'ptr') {
+    return data.inner as AnyData;
+  }
+  return data;
+}
+
+/**
+ * Dereferences `schema` if it's a pointer schema. Leaves it be otherwise.
+ */
+export function toStorable(schema: AnyData): AnyData {
```
reczkok: The name is a little misleading since it does not guarantee that we get a storable type from it. We would need to concretize too. Also we do have a type for StorableData (but it's incomplete since it does not include textures and samplers)

apps/typegpu-docs/src/content/docs/reference/shader-generation.mdx
aleksanderkatan: Oh okay, now I get it

packages/typegpu/src/core/slot/accessor.ts
aleksanderkatan: Okay, makes sense

packages/typegpu/tests/tgsl/argumentOrigin.test.ts
cieplypolar, line 1: You can never have too many tests
```ts
it('should fail on returning an argument reference', () => {
  const foo = (a: d.v3f) => {
    'use gpu';
    return a;
  };

  const main = () => {
    'use gpu';
    foo(d.vec3f(7));
  };

  expect(() => asWgsl(main)).toThrowErrorMatchingInlineSnapshot(`
    - <root>
    - fn*:main
    - fn*:main()
    - fn*:foo(vec3f): Cannot return references to arguments, returning 'a'. Copy the argument before returning it.]
  `);
});
```

packages/typegpu/src/core/constant/tgpuConstant.ts
iwoplaza: Because we want devs to be aware that mutating this value cannot be reflected in the generated WGSL, and we want to have less restrictions on using these values in TypeGPU functions.

PR #1756, opened 2025-09-27
chore: Simplify tests using inline snapshots

iwoplaza:
Why?
- Easier to update expected output when a part of how shaders are generated changes
- We can track the visual quality of the produced WGSL code


reczkok (approved):
Looks great! Much more convenient

PR #1758, opened 2025-09-29
fix: Big int handling

aleksanderkatan:
Now the following does not warn, and the cast is unnecessary.
`const FLT_MAX = f32(3.40282346e38);`

### DIFF
--- a/packages/typegpu/src/tgsl/generationHelpers.ts
+++ b/packages/typegpu/src/tgsl/generationHelpers.ts
@@ -168,6 +168,9 @@ export function getTypeForIndexAccess(
 }
 
 export function numericLiteralToSnippet(value: number): Snippet {
+  if (Math.abs(value) >= 2 ** 63) {
+    return snip(value, abstractFloat);
+  }
   // WGSL AbstractInt uses 64-bit precision, but JS numbers are only safe up to 2^53 - 1.
   // Warn when values exceed this range to prevent precision loss.
   if (Number.isInteger(value)) {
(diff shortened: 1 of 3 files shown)


packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -168,6 +168,9 @@ export function getTypeForIndexAccess(
 }
 
 export function numericLiteralToSnippet(value: number): Snippet {
+  if (Math.abs(value) > 2 ** 63) {
```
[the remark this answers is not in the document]
aleksanderkatan: Literally not true, you can check in console that `Math.abs(2 ** 64) > 2 ** 63` is true.

cieplypolar: are you sure about $2^{63}$ imo limit should be  $2^{53}$.

cieplypolar (changes requested):

packages/typegpu/src/tgsl/generationHelpers.ts
cieplypolar: Sorry, my bad $2^{63}$ is correct, but we should check for $\geq$

cieplypolar:
According to your question, I believe if we take `bigint` as the argument we can handle the whole range of `abstractInt`. So imo we shouldn't cast numbers between $2^{53}$ and $2^{63} - 1$ as `floats`.

packages/typegpu/src/tgsl/generationHelpers.ts
aleksanderkatan: How about `value >= 2 ** 63 || value < -(2 ** 63)`?

cieplypolar: oh yea, forgot that we take `Math.abs`

PR #1765, opened 2025-09-30
fix: Destroyed buffer reference in Disco

aleksanderkatan:


apps/typegpu-docs/src/examples/rendering/disco/index.ts
```diff
@@ -104,7 +104,8 @@ export const controls = {
       }[value];
       if (patternIndex !== undefined) {
         currentPipeline = pipelines[patternIndex];
-        render();
+        cancelAnimationFrame(frameId);
+        frameId = requestAnimationFrame(render);
```
iwoplaza: I just removed these two lines, and it works fine
```suggestion
```

PR #1769, opened 2025-09-30
fix: Don't track buffers and textures just to dispose of them

iwoplaza:
Early on, we decided to track buffers and textures that are created using the root. This made sense for creating roots out of already existing devices (tgpu.initFromDevice), as by destroying the root, you would destroy every resource created only with that root. This however increases the risk of leaking memory if the user doesn't properly dispose of allocated resources.

I think we should remove this "feature", in favor of decreasing the risk of introducing memory leaks.

### DIFF
--- a/apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/tgsl.mdx
@@ -8,7 +8,7 @@ This feature is under heavy development and is yet to reach stability.
 :::
 
 **TGSL (TypeGPU Shading Language)** is a subset of JavaScript used to define functions that run on the GPU via TypeGPU.
-It works by transpiling JavaScript into a compact AST format, called [tinyest](https://www.npmjs.com/package/tinyest), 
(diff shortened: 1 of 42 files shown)


apps/typegpu-docs/src/content/docs/fundamentals/roots.mdx
```diff
@@ -65,8 +65,8 @@ To unwrap a `TgpuVertexLayout` make sure to explicitly mark each of its attribut
 
 ## Destroying resources
 
-Calling `root.destroy()` will destroy all resources created with it.
-It will also destroy the underlying WebGPU device, if it wasn't originally passed in via the `initFromDevice` function.
+Calling `root.destroy()` will destroy all resources created with it. However, if an existing device is passed in via `tgpu.initFromDevice()`, this method
+is a no-op.
```
cieplypolar: I don't think putting the 'no-op' shorthand in the documentation is a good idea. Personally, I had to google it.

aleksanderkatan: I think we could just list exactly what it does, to avoid any ambiguity. "will destroy all resources created with it" is not actually true.

PR #1771, opened 2025-10-01
fix: Console.log in drawindexed

aleksanderkatan:


apps/typegpu-docs/src/examples/tests/log-test/index.ts
```diff
@@ -207,18 +237,6 @@ export const controls = {
...
-      try {
-        prepareDispatch(root, () => {
-          'kernel';
-          console.log(d.mat4x4f(), d.mat4x4f(), 1);
-        }).dispatch();
-      } catch (err) {
-        console.log(err);
-      }
-    },
-  },
```
cieplypolar, line 221: What's the reason for deleting this test?

aleksanderkatan: It does not need to use the GPU, so no need to keep it in an example. The same test already appears in unit tests.

PR #1776, opened 2025-10-02
docs: 2d slime mold simulation

reczkok:


apps/typegpu-docs/src/examples/simulation/slime-mold/index.ts
```diff
@@ -0,0 +1,355 @@
...
+    max: 100,
+    step: 1,
+    onSliderChange: (newValue: number) => {
+      params.writePartial({ moveSpeed: newValue });
+    },
+  },
+  'Sensor Angle': {
+    initial: 0.5,
+    min: 0,
+    max: Math.PI,
```
aleksanderkatan:
```suggestion
    max: 3.14,
```

image

apps/typegpu-docs/src/examples/simulation/slime-mold/index.ts
```diff
@@ -0,0 +1,355 @@
...
+  root.createBindGroup(renderLayout, {
+    state: textures[i],
+  })
+);
+
+let lastTime = performance.now();
+let currentTexture = 0;
+
+function frame() {
+  const now = performance.now();
```
aleksanderkatan: `frame` can accept timestamp, it might be better to use that instead of `performance.now()`

apps/typegpu-docs/src/examples/simulation/slime-mold/index.ts
```diff
@@ -0,0 +1,355 @@
...
+}).dispatch(NUM_AGENTS);
+
+const params = root.createUniform(Params, {
+  deltaTime: 0.016,
+  moveSpeed: 30.0,
+  sensorAngle: 0.5,
+  sensorDistance: 9.0,
+  turnSpeed: 2.0,
+  evaporationRate: 0.05,
+});
```
aleksanderkatan: Could you put the options in a constant and use it in the example controls init values?

aleksanderkatan (approved):
Great example! 🇲🇩

PR #1780, opened 2025-10-06
docs: Fixed buffers

aleksanderkatan:


apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
```diff
@@ -332,3 +332,82 @@ const indexBuffer = root
...
+
+const bindGroup = root
+  .createBindGroup(bindGroupLayout, { points: pointsBuffer });
+
+const mainCompute = tgpu['~unstable'].computeFn({
+  in: { gid: d.builtin.globalInvocationId },
+  workgroupSize: [1],
+})((x) => {
+  // Access and modify the bound buffer via the layout
+  bindGroupLayout.$.points[0] = d.vec2i();
```
reczkok: I don't think this code would work since we didn't specify proper `access`

PR #1793, opened 2025-10-08
docs: JS Kernels

iwoplaza:


apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
 Their signatures are fully visible to TypeScript, enabling tooling and static checks.
 Dependencies, including GPU resources or other functions, are resolved automatically, with no duplication or name clashes.
 This also supports distributing shader logic across multiple modules or packages.
 Imported functions from external sources are automatically resolved and embedded into the final shader when referenced.
 
 ## Defining a function
 
-In order to construct a TypeGPU function, you need to start by defining its shell, an object holding only the input and output types.
+:::note[WGSL enthusiasts!]
+Don't let the JavaScript scare you! TypeGPU functions can be implemented using either WGSL or JS, both being able to call one another.
```
reczkok, line 20: Maybe `discourage` over `scare`? Scare sounds a little condescending to me.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+const neighborhood = (a: number, r: number) => {
+  'use gpu';
+  return d.vec2f(a - r, a + r);
+};
+
+// ---cut---
+const main = () => {
+  'use gpu';
+  // Call from another function
+  return neighborhood(1.1, 0.5);
```
reczkok, line 57: Those numbers are quite carefully chosen I see 😄 I think we should either wrap them in d.f32() or at least explain the dangers of using plain literals (eg: showing the 1.0 case)

EDIT: I just now reached a section where you explain it in depth - maybe we should do a short warning with a reference to that in case people get trigger happy and do not reach that?

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+- Using operators like `+`, `-`, `*`, `/`, etc. is perfectly valid on numbers.
+- TS types are properly inferred, feel free to hover over the variables to see their types.
+- The generated code closely matches your source code.
+
+### Code transformation
+
+To make this all work, we perform a small transformation to functions marked with `'use gpu'`, leaving all other code untouched. Every project's setup is different, and we want to be as non-invasive as possible. The [unplugin-typegpu](/TypeGPU/tooling/unplugin-typegpu) package hooks into existing bundlers and build tools, extracts ASTs from TypeGPU functions and compacts them into our custom format called [tinyest](https://www.npmjs.com/package/tinyest). This metadata is injected into the final JS bundle, then used to efficiently generate equivalent WGSL at runtime.
+
+:::tip
+If you wish for all WGSL generation to happen at build time, combine [unplugin-macros](https://github.com/unplugin/unplugin-macros) and our [resolve API](/TypeGPU/fundamentals/resolve).
```
reczkok, line 99: `all WGSL generation at build time` is not possible most of the time so we shouldn't sell it like that - maybe all build time known WGSL generation?

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+:::
+
+* **Calling other functions** --
+Only functions marked with `'use gpu'` can be called from within a shader. An exception to that rule is [`console.log`](/TypeGPU/fundamentals/utils#consolelog), which allows for tracking runtime behavior
+of shaders in a familiar way.
+
+* **Operators** --
+JavaScript does not support operator overloading.
+This means that, while you can still use operators for numbers,
+you have to use supplementary functions from `typegpu/std` (*add, mul, eq, lt, ge...*) for operations involving vectors and matrices, or use a fluent interface (*abc.mul(xyz), ...*)
```
reczkok, line 309: We could mention the //ts-ignore (no @ since of course there is a user with that username) trick for people who like cursed code 😅

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+
+* **Math.\*** --
+Utility functions on the `Math` object can't automatically run on the GPU, but can usually be swapped with functions exported from `typegpu/std`.
```
reczkok, line 312: We could mention that constants are fine

apps/typegpu-docs/package.json
```diff
@@ -50,10 +50,10 @@
     "remeda": "^2.21.2",
     "sharp": "^0.34.2",
     "starlight-blog": "^0.23.2",
-    "starlight-typedoc": "^0.21.4",
+    "starlight-typedoc": "^0.19.0",
     "tinybench": "^3.1.0",
-    "typedoc": "^0.28.13",
-    "typedoc-plugin-markdown": "^4.3.0",
+    "typedoc": "^0.27.9",
```
reczkok, line 55: Is there a particular reason for this change?

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
-- <a href="https://webgpufundamentals.org/webgpu/lessons/webgpu-wgsl.html" target="_blank" rel="noopener noreferrer">WebGPU Shading Language</a>
+:::caution[May require unplugin-typegpu]
+To write TypeGPU functions in JavaScript/TypeScript, you need to install and configure [unplugin-typegpu](/TypeGPU/tooling/unplugin-typegpu).
+If you're planning on only using WGSL, you can skip installing it.
 :::
 
-TypeGPU functions let you define shader logic in a modular and type-safe way.
+**TypeGPU functions** let you define shader logic in a modular and type-safe way.
 Their signatures are fully visible to TypeScript, enabling tooling and static checks.
 Dependencies, including GPU resources or other functions, are resolved automatically, with no duplication or name clashes.
```
aleksanderkatan, line 13: The dependencies are resolved automatically, but the externals are collected automatically only with unplugin & TGSL, do we want to mention it here?

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+  // Call from another function
+  return neighborhood(1.1, 0.5);
+};
+
+// #1) Can be called in JS
+const range = main();
+//    ^?
+
+// #2) Used to generate WGSL
+const wgsl = tgpu.resolve({ externals: { main } });
```
aleksanderkatan, line 65: I'd be a little more explicit here, since this is close to the very beginning of the docs.
```suggestion
const wgsl = tgpu.resolve({ template: "", externals: { main } });
```
I really think we should add a simplified resolve api btw

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+
+const neighborhood = (a: number, r: number) => {
+  'use gpu';
+  return d.vec2f(a - r, a + r);
+};
+
+// ---cut---
+const main = () => {
+  'use gpu';
+  // Call from another function
```
aleksanderkatan, line 56:
```suggestion
```
I read it as 
docs: "There are three main ways to use TypeGPU functions." 
me: ok, what are they?
docs: "Call from another function"

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+````
+
+You can already notice a few things about TypeGPU functions:
+- Using operators like `+`, `-`, `*`, `/`, etc. is perfectly valid on numbers.
+- TS types are properly inferred, feel free to hover over the variables to see their types.
+- The generated code closely matches your source code.
+
+### Code transformation
+
+To make this all work, we perform a small transformation to functions marked with `'use gpu'`, leaving all other code untouched. Every project's setup is different, and we want to be as non-invasive as possible. The [unplugin-typegpu](/TypeGPU/tooling/unplugin-typegpu) package hooks into existing bundlers and build tools, extracts ASTs from TypeGPU functions and compacts them into our custom format called [tinyest](https://www.npmjs.com/package/tinyest). This metadata is injected into the final JS bundle, then used to efficiently generate equivalent WGSL at runtime.
```
aleksanderkatan, line 96: We do touch other code, we autoname stuff

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+// ---cut---
+const main = () => {
+  'use gpu';
+  // A very easy case, just floating point literals, so f32 by default
+  return neighborhood(1.1, 0.5);
+};
+```
+
+TypeGPU then propagates those types into the function body and analyses the types returned by the function.
+If it cannot unify them into a single type, it will throw an error.
```
aleksanderkatan, line 141: `them` may be ambiguous here
```suggestion
If it cannot unify the types found in different return statements into a single type, it will throw an error.
```

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+import * as d from 'typegpu/data';
+import * as std from 'typegpu/std';
+// ---cut---
+const double = <T extends d.v2f | d.v3f | d.v4f>(a: T): T => {
+  'use gpu';
+  return std.mul(a, a);
+};
+```
+
+### Standard library
```
aleksanderkatan, line 196: I think that up to this point, it wasn't mendioned that we cannot use e.g. "Math.abs" instead, so this may be confusing, I'd put this section a little lower

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+fn getColor(t: f32) -> vec3f {
+  if (t > 0.5) {
+    return vec3f(0.5, 0.5, 0);
+  }
+  return mix(vec3f(1, 0, 0), vec3f(0, 1, 0), t);
+}
+```
+
+Notice how `from` and `to` are inlined, and how `std.mix(from, to, constantMix)` was precomputed. TypeGPU leverages the
+fact that these values are known at shader compilation time, and can be optimized away. All other instructions are kept as is,
```
aleksanderkatan: I'd bold this, or even put it in a warning, I think this is really important

aleksanderkatan (commented):
Great docs!
I leave some nits

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
```diff
@@ -3,99 +3,445 @@ title: Functions
...
+
+You can generally assume that all JavaScript syntax is supported, and in the occasion that it is not, we'll throw a
+descriptive error either at build time or at runtime (when compiling the shader).
+
+:::note
+Our aim with TypeGPU functions is not to allow arbitrary JavaScript to be supported in the context of shaders, **rather to allow for shaders to be written in JavaScript**. This distinction means we won't support every JavaScript feature, only those that make sense in the context of graphics programming.
+:::
+
+* **Calling other functions** --
+Only functions marked with `'use gpu'` can be called from within a shader. An exception to that rule is [`console.log`](/TypeGPU/fundamentals/utils#consolelog), which allows for tracking runtime behavior
```
cieplypolar, line 303: `Only functions marked with 'use gpu'` - what about shelled functions? They don't have to be marked with `'use gpu'`, however we can call them from within a shader.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
cieplypolar, line 586: I think we should include an example using the TypeGPU function in addition to raw WGSL. When using TypeGPU functions, we need to explicitly name the `in` parameter, for example: `(input) => { ... }`.

apps/typegpu-docs/src/content/docs/fundamentals/variables.mdx
cieplypolar, line 8:
```suggestion
When using TypeGPU for resolution, there is no need to ever define either `var<uniform>` or `var<storage[, access_mode]>`, as such definitions are added automatically.
```

apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
cieplypolar, line 383: I will add a comment about `as('...')`, because we can make buffer a fixed resource.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
iwoplaza: I wouldn't want people to abuse that and feel like we're endorsing it, as it introduces more problems down the line (result inferred as `any`, not working properly on the CPU)

apps/typegpu-docs/package.json
iwoplaza: There is, the newer versions of these packages didn't generate any API Reference docs, but they also didn't fail the `nr build` command.

apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
iwoplaza: I am not sure I understand what you mean 👀

PR #1794, opened 2025-10-08
feat: Move prepareDispatch to root, and enable slot bindings

iwoplaza:


apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -3,20 +3,20 @@ title: Utilities
 description: A list of various utilities provided by TypeGPU.
 ---
 
-## *prepareDispatch*
+## prepareDispatch
```
aleksanderkatan: Why? Other functions in docs are in cursive, like *withCompute*, *resolveWithContext*
```suggestion
## *prepareDispatch*
```

packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -181,6 +196,41 @@ export interface WithBinding {
...
+   * action.dispatch();
+   * ```
+   *
+   * @example
+   * One parameter means n-threads will be executed in parallel.
+   *
+   * ```ts
+   * const action = root.prepareDispatch((x) => {
+   *   'kernel';
+   *   console.log('I am the ', x, ' thread');
```
aleksanderkatan:
```suggestion
   *   console.log('I am the', x, 'thread');
```

packages/typegpu/src/core/root/init.ts
```diff
@@ -93,13 +99,123 @@ import type {
...
+      this.#root,
+      this.#pipeline.with(bindGroup),
+    );
+  }
+
+  dispatch(): void {
+    this.#pipeline.dispatchWorkgroups(1);
+    this.#root.flush();
+  }
+}
```
aleksanderkatan: Why do we need it? In all cases we just fill the missing number of threads with 1 until we have a vector if 3, it does not seem to be any special case...

packages/typegpu/src/core/root/init.ts
```diff
@@ -123,6 +239,54 @@ class WithBindingImpl implements WithBinding {
...
+      // because we do not run unplugin before shipping typegpu package
+      const mainCompute = computeFn({
+        workgroupSize: [1],
+      })(callback as (() => undefined));
+
+      return new PreparedZeroDispatchImpl(
+        root,
+        this.withCompute(mainCompute).createPipeline(),
+      );
+    }
```
aleksanderkatan: The tests succeed if I just remove this

```suggestion
```

apps/typegpu-docs/src/examples/tests/dispatch/index.ts
aleksanderkatan, line 1: Could you add a test with slots here?

packages/typegpu/src/core/root/init.ts
```diff
@@ -93,13 +99,123 @@ import type {
...
+   * this method takes in the number of threads to run in each dimension.
+   */
+  dispatch(...threads: TArgs): void {
+    const sanitizedSize = toVec3(threads);
+    const workgroupCount = ceil(
+      vec3f(sanitizedSize).div(vec3f(this.#workgroupSize)),
+    );
+    if (!allEq(sanitizedSize, this.#lastSize)) {
+      // Only updating the size if it has changed from the last
+      // invocation. This removes the need for flushing.
```
aleksanderkatan, line 187: Does it? What if dispatch uses another buffers that have been updated in meantime?

packages/typegpu/src/core/root/init.ts
iwoplaza: This does change the generated code, as we can simplify two things:
- No need to send a uniform with the thread count
- The callback can become the entry function, so we generate one function instead of two

packages/typegpu/src/core/root/init.ts
iwoplaza: I meant this won't flush in this is part of a sequence of dispatches. If somebody writes in between that sequence anyway, then it's going to flush

packages/typegpu/src/core/root/init.ts
aleksanderkatan: Do we really need to complicate the code to optimize a dispatch of a single thread? This seems to me like a mostly debug option

iwoplaza: Reverted the optimization for the sake of simplicity

PR #1801, opened 2025-10-10
feat: Better constant handling for ref/value tracking

iwoplaza:


packages/typegpu/src/tgsl/generationHelpers.ts
```diff
@@ -200,7 +200,7 @@ export function accessProp(
       propType,
       /* ref */ isRef(target) && isNaturallyRef(propType)
         ? target.ref
-        : target.ref === 'constant'
+        : target.ref === 'constant' || target.ref === 'constant-ref'
         ? 'constant'
```
aleksanderkatan, line 204: If we access a prop of a const struct that turns out to be struct as well, wouldn't that result in 'constant-ref' again?
Or does 'constant-ref' just mean a reference to a WGSL const?

packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -18,11 +17,17 @@ import { valueProxyHandler } from '../valueProxyUtils.ts';
 // Public API
 // ----------
 
+type DeepReadonly<T> = T extends { [$internal]: unknown } ? T
+  : T extends unknown[] ? ReadonlyArray<DeepReadonly<T[number]>>
```
aleksanderkatan, line 21: Why do we leave internal stuff as is?

packages/typegpu/src/tgsl/generationHelpers.ts
iwoplaza: `constant-ref` is specifically a reference to a tgpu.const's value, and non-primitives inside that value are also `constant-ref` (see the check for isRef and isNaturallyRef just above this ternary)

```ts
const foo = tgpu.const(Boid, { pos: d.vec3f(), vel: d.vec3f() });

const bar = () => {
  'use gpu';
  foo.$; // constant-ref
  foo.$.pos; // constant-ref
  foo.$.pos.x; // constant
};
```

packages/typegpu/src/core/constant/tgpuConstant.ts
iwoplaza: Because things like v2f, v3f, v4f, ... would match the later checks, and the types could get unreadable. We could create readonly variants of these types, but that would double the amount of types we export. We can think about it.

PR #1819, opened 2025-10-14
feat: Sampler schemas

reczkok:


packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -1,134 +1,88 @@
...
+    wgslComparisonSampler(),
+    props,
+    branch,
+  ) as TgpuFixedComparisonSampler;
 }
 
-export function isSampler(resource: unknown): resource is TgpuSampler {
+export function isSampler(
+  resource: unknown,
+): resource is TgpuSampler {
```
iwoplaza:
```suggestion
export function isSampler(resource: unknown): resource is TgpuSampler {
```

iwoplaza (approved):

packages/typegpu/src/data/sampler.ts
```diff
@@ -0,0 +1,151 @@
...
+  minFilter?: GPUFilterMode;
+  /**
+   * Specifies behavior for sampling between mipmap levels.
+   */
+  mipmapFilter?: GPUMipmapFilterMode;
+  lodMinClamp?: number;
+  /**
+   * Specifies the minimum and maximum levels of detail, respectively, used internally when
+   * sampling a texture.
+   */
```
aleksanderkatan, line 29: I don't think any jsdocs appear for `lodMinClamp`

packages/typegpu/src/data/sampler.ts
```diff
@@ -0,0 +1,151 @@
...
+   * angles. Higher {@link GPUSamplerDescriptor.maxAnisotropy} values indicate the maximum ratio of
+   * anisotropy supported when filtering.
+   *
+   * Most implementations support {@link GPUSamplerDescriptor.maxAnisotropy} values in range
+   * between 1 and 16, inclusive. The used value of {@link GPUSamplerDescriptor.maxAnisotropy}
+   * will be clamped to the maximum value that the platform supports.
+   * The precise filtering behavior is implementation-dependent.
+   */
+  maxAnisotropy?: number;
+}
```
aleksanderkatan, line 87: Can we use this approach?

```suggestion
export interface WgslComparisonSamplerProps extends WgslSamplerProps {
  compare: GPUCompareFunction;
}
```

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -573,15 +577,13 @@ class TgpuBindGroupLayoutImpl<
...
+        // biome-ignore lint/suspicious/noExplicitAny: <no need for type magic>
+        (this.bound[key] as any) = new TgpuLaidOutSamplerImpl(
+          entry.sampler === 'comparison'
+            ? wgslComparisonSampler()
+            : wgslSampler(),
+          membership,
+        );
       }
 
       if ('sampler' in entry) {
```
aleksanderkatan: Are there two ifs with the same condition here? Should there be a `comparison-sampler` instead?

packages/typegpu/tests/bindGroupLayout.test.ts
```diff
@@ -570,12 +569,12 @@ describe('TgpuBindGroup', () => {
 
       root.createBindGroup(layout, {
         // @ts-expect-error
-        foo: sampler({ minFilter: 'linear' }),
+        foo: root.createSampler({ minFilter: 'linear' }),
       });
```
aleksanderkatan: nit: I think we shoud wrap all type errors in `() => ...` since the behavior is undefined

aleksanderkatan (approved):
🦣🦣🦣

packages/typegpu/src/data/sampler.ts
reczkok: We could - I just kept it the way it was

packages/typegpu/src/tgpuBindGroupLayout.ts
reczkok: This is a great catch! The second if should just be removed since the first handles both sampler types. This was a remnant of the old implementation not having a schema

packages/typegpu/src/data/sampler.ts
reczkok: There could be an issue if someone has a different version of WebGPUTypes tho

PR #1822, opened 2025-10-15
feat: Always flushing

iwoplaza:


packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -424,8 +413,8 @@ class TgpuBufferImpl<TData extends AnyData> implements TgpuBuffer<TData> {
       sizeOf(this.dataType),
     );
 
-    device.queue.submit([commandEncoder.finish()]);
-    await device.queue.onSubmittedWorkDone();
+    this.#device.queue.submit([commandEncoder.finish()]);
+    await this.#device.queue.onSubmittedWorkDone();
```
cieplypolar:
```suggestion
```
You don't have to wait for all tasks on the queue to finish. `await mapAsync` is enough because it automatically waits for the buffer to become available.

packages/typegpu/src/core/querySet/querySet.ts
```diff
@@ -122,35 +125,34 @@ class TgpuQuerySetImpl<T extends GPUQueryType> implements TgpuQuerySet<T> {
...
         this[$internal].resolveBuffer,
         0,
         this[$internal].readBuffer,
         0,
         this.count * BigUint64Array.BYTES_PER_ELEMENT,
       );
-      this._group.device.queue.submit([commandEncoder.finish()]);
-      await this._group.device.queue.onSubmittedWorkDone();
+      this.#device.queue.submit([commandEncoder.finish()]);
+      await this.#device.queue.onSubmittedWorkDone();
```
cieplypolar: This is also unnecessary.
```suggestion
```

PR #1827, opened 2025-10-16
feat: Initial @typegpu/glsl POC

iwoplaza:
This PR makes alterations to the shader generator system to allow an alternate shader language to be generated (GLSL in this case). It's definitely not a complete solution, but I think it would be best to work on it in parts, as this can get hard to rebase very fast.


iwoplaza:
Done as part other PRs

PR #1828, opened 2025-10-16
chore: Rename 'dispatch' to 'dispatchThreads'

iwoplaza:


apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
aleksanderkatan, line 1: "...returns an object with dispatch method that executes..."

aleksanderkatan (approved):

PR #1831, opened 2025-10-16
feat: Common `fullScreenTriangle` vertex function

iwoplaza:


packages/typegpu/src/common/fullScreenTriangle.ts
```diff
@@ -0,0 +1,13 @@
+import { builtin } from '../builtin.ts';
+import { vertexFn } from '../core/function/tgpuVertexFn.ts';
+import { vec2f } from '../data/vector.ts';
+
+export const fullScreenTriangle = vertexFn({
```
aleksanderkatan, line 21: I'd add some docs with an example usage, what do you think?

PR #1835, opened 2025-10-18
feat: Shell-less entry functions, createComputePipeline, createRenderPipeline

iwoplaza:
https://github.com/user-attachments/assets/b7b8f849-a7cf-4f75-ac85-6a0e5598f533

### DIFF
--- a/apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
@@ -176,7 +176,7 @@ The underlying WebGPU resources for TypeGPU pipelines are created lazily, just b
 :::
 
 
-### *createGuardedComputePipeline*
+### createGuardedComputePipeline
 
 The `createGuardedComputePipeline` method streamlines running simple computations on the GPU. 
 Instead of dispatching workgroups, the guarded pipeline allows calling an exact number of GPU threads. Think of it as a parallelized `for` loop.
--- a/packages/typegpu/src/core/function/tgpuFragmentFn.ts
+++ b/packages/typegpu/src/core/function/tgpuFragmentFn.ts
@@ -75,6 +75,7 @@ type TgpuFragmentFnShellHeader<
 
 type CleanIO<T> = T extends Record<string, BaseData>
   ? Prettify<UndecorateRecord<OmitBuiltins<T>>>
+  // a trick to use a non-record type in place of a record parameter
   : Prettify<UndecorateRecord<OmitBuiltins<{ a: T }>>> extends
     { a: infer Result } ? Result
   : never;
--- a/packages/typegpu/src/data/autoStruct.ts
+++ b/packages/typegpu/src/data/autoStruct.ts
@@ -96,7 +96,7 @@ export class AutoStruct implements BaseData, SelfResolvable {
     if (!this.#cachedStruct) {
       this.#cachedStruct = createIoSchema(
         Object.fromEntries(
-          Object.entries(this.#allocated).map(([key, alloc]) => {
+          Object.values(this.#allocated).map((alloc) => {
             return [alloc.prop, alloc.type];
           }),
         ),
--- a/packages/typegpu/src/data/instanceToSchema.ts
+++ b/packages/typegpu/src/data/instanceToSchema.ts
@@ -9,14 +9,17 @@ import type {
   Mat4x4f,
   U16,
   U32,
+  Vec2b,
   Vec2f,
   Vec2h,
   Vec2i,
   Vec2u,
+  Vec3b,
   Vec3f,
   Vec3h,
   Vec3i,
   Vec3u,
+  Vec4b,
   Vec4f,
   Vec4h,
   Vec4i,
@@ -37,6 +40,9 @@ type KindToSchemaMap = {
   vec2u: Vec2u;
   vec3u: Vec3u;
   vec4u: Vec4u;
+  'vec2<bool>': Vec2b;
+  'vec3<bool>': Vec3b;
+  'vec4<bool>': Vec4b;
   mat2x2f: Mat2x2f;
   mat3x3f: Mat3x3f;
   mat4x4f: Mat4x4f;


packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -230,9 +258,70 @@ export interface WithBinding {
...
+      VertexOut,
+      FragmentOut
+    >,
+  ): TgpuRenderPipeline<FragmentOut>;
+
   /**
    * Creates a compute pipeline that executes the given callback in an exact number of threads.
-   * This is different from `withCompute(...).createPipeline()` in that it does a bounds check on the
+   * This is different from `createComputePipeline` in that it does a bounds check on the
    * thread id, where as regular pipelines do not and work in units of workgroups.
```
aleksanderkatan, line 431:
```suggestion
   * thread id, and it works in units of workgroups.
```

apps/typegpu-docs/src/examples/rendering/smoky-triangle/index.ts
```diff
@@ -0,0 +1,168 @@
...
+const root = await tgpu.init();
+
+const fromColor = root.createUniform(d.vec3f);
+const polarCoords = root.createUniform(d.u32);
+const squashed = root.createUniform(d.u32);
+const toColor = root.createUniform(d.vec3f);
+const sharpness = root.createUniform(d.f32);
+const distortion = root.createUniform(d.f32);
+const time = root.createUniform(d.f32);
+const grainSeed = root.createUniform(d.f32);
```
aleksanderkatan: Even an example like this can get kind of close to the uniform limit per bind group (12). Have we considered possible approaches to amend this?
```
The number of uniform buffers (13) in the Vertex stage exceeds the maximum per-stage limit (12).
 - While validating binding counts
 - While validating [BindGroupLayoutDescriptor ""<unnamed> - Automatic Bind Group & Layout""]
```

- The error is readable, so users can just figure out that they can wrap these values into a single struct to only use 1 binding (though this may be non-trivial for new users).
- We could check the limits ourselves and throw more descriptive errors.
- We could group primitive uniforms into a struct ourselves.
- We could "bleed" the elements over the limit to another catch-all bind groups.

packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -230,9 +258,70 @@ export interface WithBinding {
...
+      FragmentIn,
+      FragmentOut
+    >,
+  ): TgpuRenderPipeline<FragmentOut>;
+  createRenderPipeline<
+    VertexIn extends VertexInConstrained,
+    VertexOut extends VertexOutConstrained,
+    FragmentOut extends FragmentOutInferred,
+  >(
+    descriptor: TgpuRenderPipelineDescriptor__ShelllessFrag<
```
aleksanderkatan: Is shellless vertex + shelled frag not allowed?

aleksanderkatan (approved):
Docs (including pipeline docs) still use the old API, is this meant for another ticket?

packages/typegpu/src/core/root/rootTypes.ts
iwoplaza: It becomes a false statement

packages/typegpu/src/core/root/rootTypes.ts
iwoplaza: Currently yes, we can add it if somebody really wants or needs it.

apps/typegpu-docs/src/examples/rendering/smoky-triangle/index.ts
iwoplaza: The 4th "bleed" solution was my head canon for a while, though I would prefer if we intervened as little as possible. With that in mind, the 3rd would be even more invasive.

I think I'd prefer the 2nd solution for now.

iwoplaza: Created an issue for this https://github.com/software-mansion/TypeGPU/issues/2059

packages/typegpu/src/core/root/rootTypes.ts
aleksanderkatan: Ah, i swapped those in my mind, sorry

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -82,147 +98,203 @@ const mainVertex = tgpu['~unstable'].vertexFn({
...
+console.log(await buffer2.read()); // [4, 8, 16, 32];
 ```
 
-### *withCompute*
+It is recommended NOT to use guarded compute pipelines for:
 
-Creating a compute pipeline is even easier -- the `withCompute` method accepts just a `TgpuComputeFn` with no additional parameters.
-Please note that compute pipelines are separate identities from render pipelines. You cannot combine `withVertex` and `withFragment` methods with `withCompute` in a singular pipeline.
+- More complex compute shaders.
+When using guarded compute pipelines, it is impossible to change workgroup sizes, or effectively utilize workgroup shared memory.
```
reczkok: I wouldn't say that it's impossible to `effectively utilize workgroup shared memory` now that I read it, why? I would say it's much more important to mention that you cannot use builtin values

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -82,147 +98,203 @@ const mainVertex = tgpu['~unstable'].vertexFn({
...
+  'use gpu';
+  layout.$.values[x] *= 2;
+});
+
+doubleUpPipeline.with(bindGroup1).dispatchThreads(3);
+doubleUpPipeline.with(bindGroup2).dispatchThreads(4);
+
+console.log(await buffer1.read()); // [2, 4, 6];
+console.log(await buffer2.read()); // [4, 8, 16, 32];
 ```
```
aleksanderkatan: If I'm not mistaken, this applies to all pipelines, so we could mention it outside of guarded section

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -82,147 +98,203 @@ const mainVertex = tgpu['~unstable'].vertexFn({
...
-In general, when using vertex and fragment functions with TypeGPU pipelines, it is not necessary to set locations on the IO struct properties.
-The library automatically matches up the corresponding members (by their names) and assigns common locations to them.
-When a custom location is provided by the user (via the `d.location` attribute function) it is respected by the automatic assignment procedure,
-as long as there is no conflict between vertex and fragment location value.
+
+### *createGuardedComputePipeline*
+
+The `createGuardedComputePipeline` method streamlines running simple computations on the GPU.
+Under the hood, it creates a compute pipeline that calls the provided callback only if the current thread ID is within the requested range, and returns an object with a `dispatchThreads` method that executes the pipeline.
+Since the pipeline is reused, there’s no additional overhead for subsequent calls.
```
aleksanderkatan: Since regular computePipeline is now explained directly above, I'd be more explicit about the differences between these two in this paragraph

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -82,147 +98,203 @@ const mainVertex = tgpu['~unstable'].vertexFn({
...
-Creating a compute pipeline is even easier -- the `withCompute` method accepts just a `TgpuComputeFn` with no additional parameters.
-Please note that compute pipelines are separate identities from render pipelines. You cannot combine `withVertex` and `withFragment` methods with `withCompute` in a singular pipeline.
+- More complex compute shaders.
+When using guarded compute pipelines, it is impossible to change workgroup sizes, or effectively utilize workgroup shared memory.
+For such cases, a manually created pipeline would be more suitable.
 
-### *createPipeline*
+- Small calls.
+Usually, for small data the shader creation and dispatch is more costly than serialization.
+Small buffers can be more efficiently initialized with the `buffer.write()` method.
```
aleksanderkatan: I'd remove these bullet points from here. I think the first one would be better located in a note after the first paragraph noting the differences between regular and guarded pipelines. The second one could be made into a warning, but we could also just get rid of it.

packages/typegpu/src/core/function/autoIO.ts
```diff
@@ -0,0 +1,153 @@
...
+  InferRecord,
+} from '../../shared/repr.ts';
+import { $internal, $resolve } from '../../shared/symbols.ts';
+import type { ResolutionCtx, SelfResolvable } from '../../types.ts';
+import { createFnCore, type FnCore } from './fnCore.ts';
+import type { BaseIOData } from './fnTypes.ts';
+
+export type AnyAutoCustoms = Record<
+  string,
+  number | InferGPU<BaseIOData> | number[]
```
reczkok: I think this covers all possibilities? The only possibility to have an array here is the clipDistances builtin which will be handled by the other types
```suggestion
  InferGPU<BaseIOData>
```

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -236,7 +308,11 @@ renderPipeline
   })
   .draw(3);
 
+// Assuming computePipeline is created with createComputePipeline
 computePipeline.dispatchWorkgroups(16);
+
+// Assuming computePipeline2 is created with createGuardedComputePipeline
+computePipeline2.dispatchThreads(4);
```
aleksanderkatan: How about this?

```suggestion
computePipeline.dispatchWorkgroups(16);

guardedComputePipeline.dispatchThreads(4);
```

packages/typegpu/src/core/function/autoIO.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Only used internally
+ */
+export class AutoVertexFn implements SelfResolvable {
+  // Prototype properties
+  declare [$internal]: true;
+  declare resourceType: 'auto-vertex-fn';
+
+  impl: AutoVertexFnImpl;
```
reczkok: Why do we need to expose the `impl` (or for that matter store it at all) if it's SelfResolvable? (The same for Fragment)

apps/typegpu-docs/src/examples/image-processing/background-segmentation/index.ts
```diff
@@ -168,9 +167,53 @@ const blurPipelines = [false, true].map((flip) =>
...
-  .createPipeline();
+  .createRenderPipeline({
+    vertex: common.fullScreenTriangle,
+    fragment: ({ uv }) => {
+      'use gpu';
+      const originalColor = std.textureSampleBaseClampToEdge(
+        drawWithMaskLayout.$.inputTexture,
+        drawWithMaskLayout.$.sampler,
+        uv,
+      );
```
aleksanderkatan: nit: 
I think inlining shaders like this makes the index.ts less concise, especially when there is a dedicated `shaders.ts` file in the example.

apps/typegpu-docs/src/examples/simple/triangle-next/index.ts
```diff
@@ -0,0 +1,62 @@
...
+      d.vec2f(0.5, 1.0),
+      d.vec2f(0.0, 0.0),
+      d.vec2f(1.0, 0.0),
+    ];
+
+    return {
+      $position: d.vec4f(pos[$vertexIndex], 0, 1),
+      uv: uv[$vertexIndex],
+    };
+  },
```
aleksanderkatan, line 35: How about this?
```suggestion
const pos = tgpu.const(d.arrayOf(d.vec2f, 3), [
  d.vec2f(0.0, 0.5),
  d.vec2f(-0.5, -0.5),
  d.vec2f(0.5, -0.5),
]);

const uv = tgpu.const(d.arrayOf(d.vec2f, 3), [
  d.vec2f(0.5, 1.0),
  d.vec2f(0.0, 0.0),
  d.vec2f(1.0, 0.0),
]);

const pipeline = root['~unstable'].createRenderPipeline({
  vertex: ({ $vertexIndex }) => {
    'use gpu';
    return {
      $position: d.vec4f(pos.$[$vertexIndex], 0, 1),
      uv: uv.$[$vertexIndex],
    };
  },
```

packages/typegpu/src/core/function/autoIO.ts
```diff
@@ -0,0 +1,153 @@
...
+} from '../../shared/repr.ts';
+import { $internal, $resolve } from '../../shared/symbols.ts';
+import type { ResolutionCtx, SelfResolvable } from '../../types.ts';
+import { createFnCore, type FnCore } from './fnCore.ts';
+import type { BaseIOData } from './fnTypes.ts';
+
+export type AnyAutoCustoms = Record<
+  string,
+  number | InferGPU<BaseIOData> | number[]
+>;
```
aleksanderkatan: Why isn't this type just this?
```suggestion
export type AnyAutoCustoms = Record<string, InferGPU<BaseIOData>>;
```
number is already a part of InferGPU, and I don't think a number array can be a vertex attribute nor an interstage variable

packages/typegpu/src/core/function/tgpuVertexFn.ts
```diff
@@ -46,48 +48,64 @@ export type VertexOutConstrained = IORecord<
...
  */
-export type TgpuVertexFnShell<
-  VertexIn extends VertexInConstrained,
-  VertexOut extends VertexOutConstrained,
-> =
-  & TgpuVertexFnShellHeader<VertexIn, VertexOut>
-  & ((
+export interface TgpuVertexFnShell<
+  // We force the variance to be covariant, since shells are just containers of
+  // schemas that conincidentally can be called to create a fragment function.
```
aleksanderkatan:
```suggestion
  // schemas that coincidentally can be called to create a vertex function.
```
Isn't this shell's main purpose though? Can you give an example use case where out/out is move convenient than in/out?

packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -207,10 +233,10 @@ function createFragmentFn(
     [$resolve](ctx: ResolutionCtx): ResolvedSnippet {
       const inputWithLocation = shell.in
         ? createIoSchema(shell.in, ctx.varyingLocations)
-          .$name(`${getName(this) ?? ''}_Input`)
         : undefined;
 
       if (inputWithLocation) {
+        setName(inputWithLocation, `${getName(this) ?? ''}_Input`);
```
aleksanderkatan: nit: I'd prioritise using `.$name` wherever possible, since it sometimes does more than just the `setName`
```suggestion
        inputWithLocation.$name(`${getName(this) ?? ''}_Input`);
```

packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -225,11 +254,177 @@ export interface Configurable extends Withable<Configurable> {
...
+          | ((
+            input: AutoFragmentIn<
+              Assume<
+                InferGPURecord<FragmentInFromVertexOut<TVertexOut>>,
+                AnyAutoCustoms
+              >
+            >,
+          ) => AutoFragmentOut<Assume<TFragmentOut, AnyAutoCustoms | v4f>>);
+        targets: FragmentOutToTargets<NoInfer<TFragmentOut>>;
+      }),
```
aleksanderkatan, line 315: Can we deduplicate type code here? It's difficult to parse. Are the descriptor types the same?

packages/typegpu/src/index.ts
```diff
@@ -128,8 +131,18 @@ export type {
...
+} from './core/function/autoIO.ts';
 export type { TgpuDeclare } from './core/declare/tgpuDeclare.ts';
 export type { Namespace } from './core/resolve/namespace.ts';
 // Exported for being able to track use of these global extensions easier,
 // and to establish a solid contract between tooling using them.
 export type { INTERNAL_GlobalExt } from './shared/meta.ts';
+
+/** @deprecated Use TgpuRenderPipeline.Descriptor */
+export type TgpuRenderPipelineDescriptor =
+  import('./core/pipeline/renderPipeline.ts').TgpuRenderPipeline.Descriptor;
```
aleksanderkatan, line 152: Why is this a dynamic import? Does it matter?

packages/typegpu/src/data/autoStruct.ts
```diff
@@ -0,0 +1,119 @@
...
+    key: string,
+    dataType: BaseData,
+  ): { prop: string; type: BaseData } {
+    let alloc = this.#allocated[key];
+    if (!alloc) {
+      let wgslKey = key;
+      // Builtins always start with '$'
+      if (wgslKey.includes('$')) {
+        // Finding a unique key
+        wgslKey = key.replace('$', ''); // Starting with the original key without '$' (identity for non-builtins)
```
aleksanderkatan: User defined prop name can include multiple dollar signs, as well as be invalid in many other ways
```suggestion
        wgslKey = key.replaceAll('$', ''); // Starting with the original key without '$' (identity for non-builtins)
```

Can we use nameRegistry here somehow to clean up the key and make it unique? Even if the tradeoff would be polluting our set of used names, I think it may be worth it for clarity and less of similar code in our repo

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
aleksanderkatan: It is already explained in `Resource bindings` section, so this can be removed.

packages/typegpu/src/index.ts
aleksanderkatan: I guess it is so that it is easier to delete when we eventually do this

apps/typegpu-docs/src/examples/simple/triangle-next/index.ts
aleksanderkatan: there is no test for this example

packages/typegpu/src/core/root/rootTypes.ts
aleksanderkatan: I'm sorry, these types require so much parametrisation and are just different enough from each other that I was not able simplify this monstrosity.

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
aleksanderkatan: The docs were rewritten

packages/typegpu/src/core/function/autoIO.ts
aleksanderkatan: I removed this

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -18,7 +18,10 @@ It abstracts away the standard WebGPU procedures to offer a convenient, type-saf
 
 ## Creating pipelines
 
-A pipeline definition starts with the [root](/TypeGPU/fundamentals/roots) object and follows a builder pattern.
+A pipeline can be defined with one of the following methods on the [root](/TypeGPU/fundamentals/roots) object:
+- [createRenderPipeline](#createrenderpipeline)
+- [createComputePipeline](#createcomputepipeline)
+- [createGuardedComputePipeline](#createguardedcomputepipeline)
```
cieplypolar, line 24: what about builder pattern?

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
```diff
@@ -183,46 +142,112 @@ const fragment = tgpu['~unstable'].fragmentFn({
...
+});
+```
 
 :::caution
-The underlying WebGPU resource is created lazily, that is just before the first execution or as part of a `root.unwrap`, not immediately after the `createPipeline` invocation.
+The underlying WebGPU resources for TypeGPU pipelines are created lazily, just before the first execution or as part of a `root.unwrap`, not immediately after the `createRenderPipeline` or `createComputePipeline` invocation.
+:::
+
+
+### *createGuardedComputePipeline*
```
cieplypolar:
```suggestion
### createGuardedComputePipeline
```
for consistency

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -3,116 +3,82 @@ title: Utilities
...
+Usually `'runtime'` (the default) is a safe bet, but if you're sure that the expression or
+computation is constant (either a reference to a constant, a numeric literal,
+or an operation on constants), then pass `'constant'` as it might lead to better
+optimizations.
 
-The callback is not called if the global invocation id of a thread would exceed the size in any dimension.
-:::
+If what the expression is a direct reference to an existing value (e.g. a uniform, a
+storage binding, ...), then choose from `'uniform'`, `'mutable'`, `'readonly'`, `'workgroup'`,
+`'private'` or `'handle'` depending on the address space of the referred value.
```
cieplypolar, line 81: what is a `handle`?

packages/typegpu/src/data/autoStruct.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+
+  provideProp(
+    key: string,
+    dataType: BaseData,
+  ): { prop: string; type: BaseData } {
+    let alloc = this.#allocated[key];
+    if (!alloc) {
+      let wgslKey = key.replaceAll('$', '');
+      let idx = 1;
```
cieplypolar:
```suggestion
      let idx = 0;
```
Now, we start indexing from 2.

packages/typegpu/src/data/autoStruct.ts
```diff
@@ -0,0 +1,116 @@
...
+      while (wgslKey in this.#validProps && this.#usedWgslKeys.has(wgslKey)) {
```
cieplypolar:
```suggestion
      while (this.#usedWgslKeys.has(wgslKey)) {
```
do we need the left condition?

packages/typegpu/src/data/autoStruct.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    return alloc;
+  }
+
+  get completeStruct() {
+    if (!this.#cachedStruct) {
+      this.#cachedStruct = createIoSchema(
+        Object.fromEntries(
+          Object.entries(this.#allocated).map(([key, alloc]) => {
```
cieplypolar:
```suggestion
          Object.values(this.#allocated).map((alloc) => {
```

packages/typegpu/src/data/instanceToSchema.ts
```diff
@@ -0,0 +1,56 @@
...
+  Vec3i,
+  Vec3u,
+  Vec4f,
+  Vec4h,
+  Vec4i,
+  Vec4u,
+  Void,
+} from './wgslTypes.ts';
+
+type KindToSchemaMap = {
```
cieplypolar, line 30: what about boolean vectors

packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -56,49 +64,62 @@ export type FragmentOutConstrained = IOLayout<
...
 > = {
   readonly in: FragmentIn | undefined;
   readonly out: FragmentOut;
   readonly returnType: IOLayoutToSchema<FragmentOut>;
   readonly entryPoint: 'fragment';
 };
 
+type CleanIO<T> = T extends Record<string, BaseData>
+  ? Prettify<UndecorateRecord<OmitBuiltins<T>>>
+  : Prettify<UndecorateRecord<OmitBuiltins<{ a: T }>>> extends
```
cieplypolar, line 79: add comment that explains necessity of this, please :)

apps/typegpu-docs/src/content/docs/fundamentals/pipelines.mdx
aleksanderkatan: The old API is already deprecated, we plan on removing it in the future so let's not encourage anyone to use it

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
aleksanderkatan: It's a WGSL address space used for variable declaration and pointer types, that is never stated explicitly. It's reserved for textures and samplers. See more here: https://www.w3.org/TR/WGSL/#address-space

packages/typegpu/src/data/autoStruct.ts
aleksanderkatan: There were zero tests that enter this if branch, even including examples...

iwoplaza: Would be beneficial to have at least one test for this collision avoidance, I agree

aleksanderkatan: It doesn't seem to work, and it didn't work before my changes to autostruct either, I'll try to fix this.

image

aleksanderkatan:
I suspect that shaderStageSlot is not filled when using these, I'll check this out

packages/typegpu/src/data/autoStruct.ts
aleksanderkatan: Using `||` instead of `&&` seems to kind of do the trick, though the renamed name is not passed to fragment so the user has to manually add `_2`, which of course is unacceptable.

image

If we don't pass any renaming map, then this collision resolution relies on the order of items in the struct. We could sort them, but then again, they don't need to have the same props...

I will just throw an error when I detect a situation like this, users can handle renaming a badly named variable.

packages/typegpu/src/data/instanceToSchema.ts
aleksanderkatan: Fair point

packages/typegpu/src/data/autoStruct.ts
aleksanderkatan: Now an error is thrown, and only on the second condition as you suggested

PR #1838, opened 2025-10-20
chore: Rename `prepareDispatch` to `createGuardedComputePipeline`

iwoplaza:


lursz (approved):
Good job 👍🏿 All examples works flawlessly

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -3,10 +3,10 @@ title: Utilities
 description: A list of various utilities provided by TypeGPU.
 ---
 
-## *prepareDispatch*
+## *root.createGuardedComputePipeline*
 
-The `prepareDispatch` function streamlines running simple computations on the GPU.
-Under the hood, it wraps the callback in a `TgpuFn`, creates a compute pipeline, and returns an object with a `dispatchThreads` method that executes the pipeline.
+The `root.createGuardedComputePipeline` method streamlines running simple computations on the GPU.
+Under the hood, it creates a compute pipeline that calls the provided callback only if the current thread ID is within the requested range, and returns an object with a `dispatchThreads` method that executes the pipeline.
```
aleksanderkatan, line 9: nit: I guess this may be what people who are getting into compute shaders stumble upon, and the "if the current thread ID is within the requested range" might be confusing to somebody who does not know about workgroups.

iwoplaza: I can see that. It might be beneficial to write a blog post explaining what workgroups are, and how compute shaders are being dispatched. Then we can link to that explanation here, and maybe include a visual representation of this difference.

iwoplaza: I made an issue for it:
https://github.com/software-mansion/TypeGPU/issues/1849

PR #1853, opened 2025-10-23
fix: Remove redundant .with slot calls & chore: Cleanup ItemStateStack

### LINKED ISSUE #1854: impr: Optimize slot resolution
`resolutionCtx.ts`, the `withSlots` method.

After the patch from #1853, the slots are resolved correctly, but inefficiently (one by one, instead of all at once).

aleksanderkatan:
Changes:
- fix #1854,
- clean up stack layers code.

### DIFF
--- a/.github/workflows/treeshake-test.yml
+++ b/.github/workflows/treeshake-test.yml
@@ -106,13 +106,13 @@ jobs:
                 await github.rest.issues.updateComment({
                   ...context.repo,
                   comment_id: existingComment.id,
-                  body: botCommentIdentifier + comparison,
+                  body: botCommentIdentifier + '\n' + comparison,
                 });
               } else {
                 await github.rest.issues.createComment({
                   ...context.repo,
                   issue_number: issueNumber,
-                  body: botCommentIdentifier + comparison,
+                  body: botCommentIdentifier + '\n' + comparison,
                 });
               }
             }
@@ -124,6 +124,5 @@ jobs:
               await createOrUpdateComment(issueNumber);
             }
             await core.summary
-              .addHeading('Tree-shake Test Results')
               .addRaw(comparison)
               .write();


aleksanderkatan:
I'm pretty sure I found a small bug in the resolution code. Take a look at these two snippets:

`resolutionCtx.ts`
```ts
  unwrap<T>(eventual: Eventual<T>): T {
    if (isProviding(eventual)) {
      return this.withSlots(
        eventual[$providing].pairs,
        () => this.unwrap(eventual[$providing].inner) as T,
      );
    }
```

`tgpuFn.ts`
```ts
    with(
      slot: TgpuSlot<unknown> | TgpuAccessor,
      value: unknown,
    ): TgpuFn<ImplSchema> {
      return createBoundFunction(fn, [
        ...pairs,
        [isAccessor(slot) ? slot.slot : slot, value],
      ]);
    },
```

I think that if we call `.with` multiple times, like:
```ts
    const getSize = tgpu.fn([], d.f32)`...`
      .$uses({ ... })
      .with(sizeSlot, 1)
      .with(colorSlot, RED)
      .with(shapeSlot, 2);
```

then `withSlots` is called with `[sizeSlot, colorSlot, shapeSlot]`, then with `[sizeSlot, colorSlot]`, then with `[sizeSlot]`.

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -141,11 +142,13 @@ class ItemStateStackImpl implements ItemStateStack {
...
       bindingMap: new WeakMap(pairs),
+      usedSet: new WeakSet(),
     });
   }
 
-  popSlotBindings() {
-    this.pop('slotBinding');
+  popSlotBindings(): WeakSet<TgpuSlot<unknown>> {
+    const slotBindings = this.pop('slotBinding') as SlotBindingLayer;
+    return slotBindings.usedSet;
```
reczkok: We could do a little cleanup with the pop method since it can and should return a narrowed type:
```ts
type StackLayer =
  | ItemLayer
  | SlotBindingLayer
  | FunctionScopeLayer
  | BlockScopeLayer;

class ItemStateStackImpl implements ItemStateStack {
  private _stack: StackLayer[] = [];
  ...
  pop<T extends StackLayer['type']>(type: T): Extract<StackLayer, { type: T }>;
  pop(type?: StackLayer['type']): StackLayer | undefined;
  pop(type?: StackLayer['type']) {
    ...
  }
```
```suggestion
    return this.pop('slotBinding').usedSet;
```

reczkok (approved):
Great work!

packages/typegpu/src/resolutionCtx.ts
aleksanderkatan: I changed the type to the following:
```ts
pop<T extends StackLayer['type']>(type: T): Extract<StackLayer, { type: T }>;
pop(): StackLayer | undefined;
```
And I left the push methods unchanged. Let me know what you think

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -290,7 +290,6 @@ function createBoundFunction<ImplSchema extends AnyFn>(
       value: unknown,
     ): TgpuFn<ImplSchema> {
       return createBoundFunction(fn, [
```
iwoplaza: I am wondering if passing in "innerFn" instead of "fn" and keeping the ... spread will lead to less wrapping (less layers to resolve)

aleksanderkatan: I thought it would break some things like toStrings, this is why I created another issue for that, but it turns out toStrings are broken right now:
```ts
  it('includes slot bindings in toString', () => {
    const firstSlot = tgpu.slot<number>();
    const secondSlot = tgpu.slot<number>();
    const thirdSlot = tgpu.slot<number>();

    const getSize = tgpu.fn([], d.f32)(() =>
      firstSlot.$ + secondSlot.$ + thirdSlot.$
    )
      .with(firstSlot, 1)
      .with(secondSlot, 2)
      .with(thirdSlot, 3);

    expect(getSize.toString()).toMatchInlineSnapshot(`"fn:getSize[thirdSlot=3]"`);
  });
```
And they work after the change:
```ts
  it('includes slot bindings in toString', () => {
    const firstSlot = tgpu.slot<number>();
    const secondSlot = tgpu.slot<number>();
    const thirdSlot = tgpu.slot<number>();

    const getSize = tgpu.fn([], d.f32)(() =>
      firstSlot.$ + secondSlot.$ + thirdSlot.$
    )
      .with(firstSlot, 1)
      .with(secondSlot, 2)
      .with(thirdSlot, 3);

    expect(getSize.toString()).toMatchInlineSnapshot(
      `"fn:getSize[firstSlot=1, secondSlot=2, thirdSlot=3]"`,
    );
  });
```
So yes, I guess I will go with that!

iwoplaza:
I am unsure if I want unused slots to always be warned about. I can imagine a library that accepts a callback and provides it with a slot, and the callback can choose to use the slot or not at all. I am not yet sure how to both allow devs to have this introspection, and to not flood the console with false positives. Any thoughts?

iwoplaza (approved):

PR #1856, opened 2025-10-24
impr: Error messages for invalid array and `tgpu.const` usage

aleksanderkatan:
Changes:
- better error when indexing an external array with a runtime index,
- better error when using a const inside TGSL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,11 +8,41 @@
 
 </div>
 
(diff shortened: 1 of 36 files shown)


iwoplaza (approved):
Awesome changes! 👏

iwoplaza:
Could you change the name of the PR slightly to describe in more detail what errors were improved? This will make it easier to generate change notes

PR #1858, opened 2025-10-24
feat: new homepage

lursz:
New landing page for TypeGPU, with interactive islands running TypeGPU examples

After running the dev script the new landing page is at:
http://localhost:4321/TypeGPU/new


apps/typegpu-docs/src/components/TgpuExamples.astro
```diff
@@ -0,0 +1,97 @@
...
+];
+---
+
+{/* Interactive gallery */}
+<div
+	class="sm:hidden mb-4 px-4 py-3 border dark:text-white border-accent-500 text-sm text-center"
+>
+	Hint: tap the example with two fingers to preview
+</div>
+<div class="gap-6 grid sm:grid-cols-2 xl:grid-cols-2">
```
sharabai:
```suggestion
<div class="gap-6 grid sm:grid-cols-2">
```

apps/typegpu-docs/src/components/TgpuExamples.astro
```diff
@@ -0,0 +1,97 @@
...
+	{
+		galleryItems.map((item) => (
+			<div
+				class="group relative aspect-square overflow-hidden"
+				data-example-card
+				style="touch-action: pan-x pan-y;"
+			>
+				<div class="bottom-1/40 inset-x-2 px-3 z-10 absolute flex items-center justify-between gap-3 bg-white h-16 text-blue-900">
```
sharabai: image
On designs you have 24px for desktop and 16px on mobile. Not what you have.

apps/typegpu-docs/astro.config.mjs
iwoplaza, line 1: The changes in this file are forcing code snippets in the docs to be dark mode. Also, I think the code snippets on the landing page should respect the light/dark mode settings, or it's at least something we should explore.

iwoplaza, line 1: The bg of the code editor doesn't fit the bg of the page. I think we'd need to create our own custom theme.

image

apps/typegpu-docs/src/components/ThemeSwitcher.astro
iwoplaza, line 1: The color transitions on these buttons are causing an unpleasant moment of mismatch between the color themes.
Also, the GitHub button doesn't react to the light/dark theme.

https://github.com/user-attachments/assets/ec94238f-8018-493f-83c9-55bd9188b974

apps/typegpu-docs/src/pages/new/index.astro
```diff
@@ -0,0 +1,302 @@
...
+
+    <!-- Hero Section -->
+    <section
+      class="flex flex-col items-center mx-auto lg:px-72 md:max-w-214 lg:max-w-[1024px]"
+    >
+      <div class="flex flex-col gap-4 md:gap-12 px-6 lg:px-0">
+        <div
+          class="font-bold text-[#001A72] dark:text-gray-100 text-2xl md:text-6xl"
+        >
+          Type<GradientTitle title="GPU" />
```
iwoplaza, line 88: The gradient goes in the opposite direction

Implementation:
image

Design:
image

apps/typegpu-docs/src/layouts/PageLayout.astro
```diff
@@ -73,6 +73,31 @@ const { title, theme = 'light' } = Astro.props;
...
+          storedTheme = null;
+        }
+
+        const systemPrefersDark =
+          window.matchMedia && window.matchMedia('(prefers-color-scheme: dark)').matches;
+
+        const themeToApply = storedTheme || doc.dataset.theme || (systemPrefersDark ? 'dark' : 'light');
+
+        applyTheme(themeToApply);
+      })();
```
iwoplaza: Why wrapped in an IIFE?

apps/typegpu-docs/src/components/TgpuExamples.astro
iwoplaza, line 1: I see there's sometimes both "Loading..." text and/or control widgets that pop-up when activating an example tile. I think for the sake of simplifying this component and `HoverExampleIsland`, we should duplicate those examples for the purpose of embedding them here, and making them more simplified and easier to interact with.

apps/typegpu-docs/src/layouts/PageLayout.astro
lursz: It's a leftover, I intended to keep some variables out of global scope. Will get removed

apps/typegpu-docs/astro.config.mjs
lursz: Yeah I've just noticed, it replaced the theme for all code snippets. Would we like to have the exact same theme everywhere (homepage + docs) or keep them separated?

apps/typegpu-docs/src/components/ThemeSwitcher.astro
lursz: added `transition-colors duration-300 ease-in-out` across whole page, but I'm not in favour of it

apps/typegpu-docs/src/pages/new/index.astro
lursz: Nice eye for a detail. Fixed

apps/typegpu-docs/astro.config.mjs
aleksanderkatan, line 1: At some zoom levels the indents become inconsistent
image

aleksanderkatan, line 1: Sometimes, when I hover over an example, it stays blurred (I don't hover over the link). Does it mean it fails to load?

image

aleksanderkatan, line 1: I can see the compression here 

image

aleksanderkatan, line 1: Can we get rid of this loading text?

https://github.com/user-attachments/assets/a2a2e248-7ce0-493c-a74b-399fcc740e10

aleksanderkatan, line 1: This example sometimes flickers when hovering over it, any ideas on what may be the cause?

https://github.com/user-attachments/assets/1b214805-0bfc-454b-b5d8-83ad492e3ca1

aleksanderkatan, line 1: The popup here overlaps with the link. There was a suggestion by @reczkok that might help us eliminate these popups, so this may not be an issue.

image

aleksanderkatan, line 1: Not sure about this, but I think clicking on this symbol usually opens a new page

image

aleksanderkatan, line 1: The jump in the blurred background is quite jarring. When capturing this, you could make the example run in a loop by instead of `t`, passing `std.sin(t)` to the time buffer. Though I'm not sure if I like this blurred background, couldn't we zoom in the example to also take in the entire canvas?

https://github.com/user-attachments/assets/5dc37f30-e252-48cc-842f-9d31048aa5d1

aleksanderkatan, line 1: Should hovering over example link buttons blur the examples?

aleksanderkatan, line 1: Why is youtube so mean? (I couldn't get the miniatures to load)

image

apps/typegpu-docs/src/components/Backlog.astro
```diff
@@ -0,0 +1,148 @@
+---
+import { Image } from "astro:assets";
+import as_a_foundation from "../assets/hero/as_a_foundation.svg";
+import as_a_puzzle_piece from "../assets/hero/as_a_puzzle_piece.svg";
+import for_libraries from "../assets/hero/for_libraries.svg";
+---
+
+<div class="gap-8 grid grid-cols-1 md:grid-cols-3">
```
aleksanderkatan, line 8: We'll change these values in another PR, right?

apps/typegpu-docs/astro.config.mjs
aleksanderkatan, line 1: The horizontal line here is missing

image

apps/typegpu-docs/src/pages/new/index.astro
```diff
@@ -0,0 +1,298 @@
...
+    </section>
+
+    <!-- Call to Action -->
+    <section
+      class="flex flex-col items-center gap-16 mx-auto px-6 lg:px-72 md:max-w-214 lg:max-w-[1024px]"
+    >
+      <h2
+        class="font-bold text-blue-950 dark:text-gray-100 text-3xl lg:text-5xl text-center leading-tight"
+      >
+        Ready to try out <GradientTitle title="TypeGPU" />?
```
aleksanderkatan: There is a space after "TypeGPU", and the question mark breaks to a new line, can you fix this?

image

apps/typegpu-docs/astro.config.mjs
aleksanderkatan, line 1: I think these three could probably autoselect when swiping on mobile. Could we use code from the https://swmansion.com/ carrousel?

image

aleksanderkatan (commented):
Nice page!

I know nothing about frontend, so I just leave some nits that came to me when looking at the page

apps/typegpu-docs/astro.config.mjs
lursz: Yeah, it basically means that the example failed to load. Should happen pretty seldom tho

lursz: This is an animated webp that quickly grows in size. I agree there are pixels visible here or there, but if we want to keep the page snappy and quick, that's something we have deal with

lursz: It's a feature not a bug! xd
(seriously tho, on weaker phones loading some of our examples takes a while)

lursz: It quickly switches between the live example and the picture of it, and those two aren't pixel perfect unfortunately. Still, that's such a minute thing I'd just leave it

lursz: I don't think I can remove the link license-wise, but if I can I'll just remove it from the photo

lursz: I think the current meta is to leave the choice to the user - if you want to open in the current tab use LMB, if in new use MMB

lursz: Yeah, there is sth wrong with the transition, but otherwise I'd leave it as is - I quite like the blurred look

lursz: This behaviour wasn't defined on figma so it's up to us I guess

lursz: This is a CORS issue, should be working out of the box in deployment build

apps/typegpu-docs/src/components/Backlog.astro
lursz: yeah, we can always fine-tune it later

apps/typegpu-docs/astro.config.mjs
lursz: Nice catch. I just assumed this line is some figma overlay, my mistake

lursz: Could you please provide the device sizes for which it happens? I tested for everything until Ipad mini (8 inches) and stopped caring for PC version, because anything under that is a mobile anyway

apps/typegpu-docs/src/pages/new/index.astro
lursz: Fixed!

apps/typegpu-docs/astro.config.mjs
lursz: Not bad idea, I'll leave the decision to the designer

aleksanderkatan: I don't understand the last sentence. This is a pc version (chrome), this also happens when stretching the page (100% zoom).

https://github.com/user-attachments/assets/4aa284b3-b21e-463b-8854-06d5147e8455

aleksanderkatan: Well, it happened to me at least 3 times while testing, one of which was on my first ever hover after a refresh. I don't think we can ignore this

aleksanderkatan: Can we at least center it? Right now it definitely does not look intended

aleksanderkatan: This is not what I mean, I understand the pixel offsets. I mean the frames where the example looks like this:

image

image

aleksanderkatan: You're right, I checked some pages and they do that, and this meta makes sense.

aleksanderkatan: I wouldn't do it, I kept hovering over it on accident which resets the example for no reason

lursz: Is this only Jelly Slider thing? Cause I can't think of any example-independent bug that may cause this
Edit: confirmed, this is a bounding box of a slider and only happens in this particular example

lursz: Oh I didn't know this existed on large screen formats. Nice catch, fixed

lursz: It's really difficult to reproduce, this might be a race condition, will look into it

iwoplaza:
Superseded by https://github.com/software-mansion/TypeGPU/pull/2331

PR #1874, opened 2025-10-30
feat: Resolvable guarded compute pipelines

aleksanderkatan:
We decided not to make the guarded pipelines resolvable, but to instead expose the underlying pipeline.

### DIFF
--- a/packages/typegpu/package.json
+++ b/packages/typegpu/package.json
@@ -73,7 +73,7 @@
   },
   "homepage": "https://typegpu.com",
   "devDependencies": {
-    "@ark/attest": "^0.49.0",
+    "@ark/attest": "^0.53.0",
     "@typegpu/tgpu-dev-cli": "workspace:*",
     "@webgpu/types": "catalog:types",
     "arktype": "catalog:",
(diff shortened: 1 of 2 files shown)


packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -94,6 +95,18 @@ export interface TgpuGuardedComputePipeline<TArgs extends number[] = number[]> {
...
+
+  /**
+   * The underlying pipeline used during `dispatchThreads`.
+   */
+  pipeline: TgpuComputePipeline;
+
+  /**
+   * The buffer used to automatically pass the thread count to the underlying pipeline during `dispatchThreads`.
+   * In case of guarded pipelines that dispatch in less than 3 dimensions,
+   * the remaining values should be filled with 1.
```
iwoplaza:
```suggestion
   * The buffer used to automatically pass the thread count to the underlying pipeline during `dispatchThreads`.
   * For pipelines with a dimension count lower than 3, the remaining coordinates are expected to be 1.
```

iwoplaza (approved):
Yeshh! ✨

PR #1880, opened 2025-11-02
fix: Always resolve with strict types

iwoplaza:


packages/typegpu/tests/examples/individual/vaporrave.test.ts
aleksanderkatan, line 1: Some of the examples started to report implicit casts. Could you investigate whether those are caused by intended behavior?

packages/typegpu/tests/examples/individual/oklab.test.ts
```diff
@@ -61,65 +61,65 @@ describe('oklab example', () => {
...
-            k2 = 0.08285427;
-            k3 = 0.1254107;
-            k4 = 0.14503204;
+            k2 = 0.08285427f;
+            k3 = 0.1254107f;
+            k4 = 0.14503204f;
             wl = -1.2684380046;
-            wm = 2.6097574011;
+            wm = 2.6097574011f;
             ws = -0.3413193965;
```
aleksanderkatan, line 83: Why do only positive literals have "f"?

aleksanderkatan (approved):
Looks good! 📸

packages/typegpu/tests/examples/individual/oklab.test.ts
iwoplaza: It's complicated

PR #1881, opened 2025-11-02
docs: Soft shadows & bounce lighting in the "Jelly Slider" example

iwoplaza:


reczkok:
obraz

The only slight artifact I could find is the shadow "expanding" towards the top but that's barely noticeable

PR #1886, opened 2025-11-04
feat: Resolve benchmark github action

cieplypolar:


apps/typegpu-docs/src/pages/resolve/index.astro
aleksanderkatan, line 1: I see that the page does not respond to the light theme, but /benchmark acts the same so we may ignore it (at least until the landing page rework)

apps/typegpu-docs/astro.config.mjs
```diff
@@ -36,6 +36,9 @@ export default defineConfig({
       'Cross-Origin-Opener-Policy': 'same-origin',
     },
   },
+  image: {
+    domains: ['raw.githubusercontent.com'],
+  },
```
aleksanderkatan: Why do we need that?

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
```diff
@@ -0,0 +1,128 @@
...
+const buttonUtilityClasses =
+  '-translate-y-1/2 absolute top-1/2 rounded-full border border-gray-700 bg-gray-800 p-4 text-gray-150 transition-all duration-300 ease-in-out hover:bg-gray-700 hover:text-white active:bg-gray-500 active:text-white z-1';
+const chevronUtilityClasses = 'w-4 h-4 sm:w-8 sm:h-8';
+
+export default function PlotGallery() {
+  // this is for infinite effect
+  const extendedPlots = useMemo(
+    () => [plots[plots.length - 1], ...plots, plots[0]],
+    [],
+  );
```
reczkok: I think we don't really need this - creating logic for wrapping in callbacks is trivial and we don't care about the dependency (plots never changes).

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const [currentIndex, setCurrentIndex] = useState(1);
+  const [isTransitioning, setIsTransitioning] = useState(false);
```
reczkok, line 32:
```suggestion
  const isTransitioningRef = useRef(false);
```
I would go for this - isTransitioning does not need to cause re-renders and it will simplify transition handling

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const nextSlide = useCallback((isTransitioning: boolean) => {
+    if (isTransitioning) return;
+    setIsTransitioning(true);
+    setCurrentIndex((prev) => prev + 1); // to avoid deps
+  }, []);
```
reczkok, line 38:
```diff lang=ts
+ const slideCount = plots.length;
```
```suggestion
  const nextSlide = useCallback(() => {
    if (isTransitioningRef.current) return;
    isTransitioningRef.current = true;

    setCurrentIndex((prev) => prev === slideCount - 1 ? 0 : prev + 1);
  }, [slideCount]);
```

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+  const handleTransitionEnd = useCallback((index: number) => {
+    setIsTransitioning(false);
+
+    if (index === 0) {
+      setCurrentIndex(plots.length);
+    } else if (index === extendedPlots.length - 1) {
+      setCurrentIndex(1);
+    }
+  }, [extendedPlots]);
```
reczkok: This will also be much simpler if we just do transitioning as a ref and the logic for wrapping in callbacks
```suggestion
  const handleTransitionEnd = useCallback(() => {
    isTransitioningRef.current = false;
  }, []);
```

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
reczkok, line 1: This is a little over-engineered but since it's for dev I don't really mind it (all the comments are just general feedback for learning purposes).
- We treat plots like they are not constant but they are so we should use it (for example there is no need to do useMemo for the plots - if we wanted to do extended plots we could just do that out of the component)
- There is no expensive calculation going on here so the optimizations are a little overkill but good practice
- Rest of the feedback is in the other comments

apps/typegpu-docs/astro.config.mjs
cieplypolar: Actually, we don't, my bad

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
cieplypolar: I don't see a simple solution inside the callbacks, if I just use modulo length, then the carousel will go through all the plots

apps/typegpu-docs/src/components/resolve/PlotGallery.tsx
cieplypolar: I appreciate the feedback. I deleted `useMemo`, but I can't change `isTransitioning` to `useRef`. `isTransitioning` state plays a crucial role in smooth carousel wrap when we transition from the last plot -> to the first plot (and first -> last).
1. Because plots are extended with 'guard' plots, the animation appears to move continuously to the right (left).
2. Then, I handle the transform property. I set `isTransitioning` to false, and change the index to real plot (not the guarding one)
3. The CSS (this one containing animation logic) rerenders and we don't see the carousel going through all the plots.

PR #1895, opened 2025-11-06
feat: Inspect vector type in shader function

iwoplaza:


packages/tinyest/src/nodes.ts
```diff
@@ -155,7 +160,9 @@ export type AssignmentOperator =
   | '&='
   | '**='
   | '||='
-  | '&&=';
+  | '&&='
```
aleksanderkatan, line 163: These don't generate valid WGSL

cieplypolar: Ooops, you're right. They probably should be added to `OP_MAP` with appropriate error.

PR #1900, opened 2025-11-07
docs: Optimize 3d slime mold example

reczkok:
- reduce steps and introduce jitter to make up for the visual quality
- rewrite blur to do 6 filtered samples instead of 3x3x3 loads


aleksanderkatan:
Yeah, I don't think I can find a way to optimize it any more at this point

PR #1902, opened 2025-11-09
docs: "Jelly Switch" example

iwoplaza:


apps/typegpu-docs/src/examples/rendering/jelly-switch/index.ts
```diff
@@ -0,0 +1,792 @@
...
+
+canvas.addEventListener('mouseup', (event) => {
+  switchBehavior.pressed = false;
+  switchBehavior.toggled = !switchBehavior.toggled;
+  event.stopPropagation();
+});
+
+window.addEventListener('mouseup', (event) => {
+  switchBehavior.pressed = false;
+});
```
reczkok: we usually do those at the end

reczkok (approved):
More Jelly!

PR #1903, opened 2025-11-10
feat: "create-typegpu" CLI package

iwoplaza:
Executing `npm create typegpu` has the same behavior as `npx typegpu@latest`. It's important to note that it always targets the latest released version of `@typegpu/cli`


packages/create-typegpu/bin.mjs
```diff
@@ -0,0 +1,44 @@
...
+    process.exit(1);
+  };
+}
+
+(async () => {
+  const windows = process.platform === 'win32';
+  const npxCommand = windows ? 'npx.cmd' : 'npx';
+
+  const code = await asyncSpawn(npxCommand, ['@typegpu/cli@latest', ...process.argv.slice(2)], {
+    stdio: 'inherit',
```
cieplypolar, line 40:
```suggestion
    stdio: 'inherit',
    shell: windows, // needs to be ran through the shell on Windows
```

aleksanderkatan (approved):

PR #1911, opened 2025-11-13
Geometry package: lines & circles

deluksic:
This package is a heavy WIP, but we decided to merge it as is, in order to have easier time iterating the design. For now, we add two shapes:
- lines
- circles

The idea with this package is to allow easy setup of vertex shaders for rendering different shapes.

The new examples are marked `dev` so they don't appear in docs for now.

image

image

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ const main = () => {
   return neighborhood(1.1, 0.5);
 };
 
-const wgsl = tgpu.resolve({ externals: { main } });
+const wgsl = tgpu.resolve([main]);
 //    ^? string
 
 //
--- a/apps/bun-example/index.ts
+++ b/apps/bun-example/index.ts
@@ -10,8 +10,6 @@ const createRandomBoid = tgpu.fn([], Boid)(() => {
   return { pos: randf.inUnitCube() };
 });
 
-const shaderCode = tgpu.resolve({
-  externals: { createRandomBoid },
-});
+const shaderCode = tgpu.resolve([createRandomBoid]);
 
 console.log(shaderCode);
--- a/apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
@@ -61,7 +61,7 @@ const range = main();
 //    ^?
 
 // #2) Used to generate WGSL
-const wgsl = tgpu.resolve({ externals: { main } });
+const wgsl = tgpu.resolve([main]);
 //    ^?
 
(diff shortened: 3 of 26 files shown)


apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -256,15 +254,16 @@ function useResizableCanvas(exampleHtmlRef: RefObject<HTMLDivElement | null>) {
...
 
-      const onResize = () => {
-        newCanvas.width = frame.clientWidth * window.devicePixelRatio;
-        newCanvas.height = frame.clientHeight * window.devicePixelRatio;
+      const onResize: ResizeObserverCallback = (entries) => {
+        const size = entries[0]?.devicePixelContentBoxSize[0];
+        if (size) {
+          newCanvas.width = size.inlineSize;
+          newCanvas.height = size.blockSize;
+        }
```
iwoplaza, line 262: Interesting! What does this address?

apps/typegpu-docs/tsconfig.json
```diff
@@ -5,6 +5,10 @@
     "jsxImportSource": "react",
     "types": ["@webgpu/types"]
   },
-  "include": [".astro/types.d.ts", "**/*"],
+  "include": [
+    ".astro/types.d.ts",
+    "**/*",
+    "../../packages/typegpu-geometry/src/lines/indices.ts"
+  ],
```
iwoplaza: This sometimes happens by mistake, gets automatically added by the IDE.

```suggestion
  "include": [".astro/types.d.ts", "**/*"],
```

packages/typegpu-geometry/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "sideEffects": false,
+  "scripts": {
+    "build": "unbuild",
+    "test:types": "pnpm tsc --p ./tsconfig.json --noEmit",
+    "prepublishOnly": "tgpu-dev-cli prepack"
+  },
+  "keywords": [],
+  "license": "MIT",
+  "peerDependencies": {
+    "typegpu": "^0.5.9"
```
iwoplaza:
```suggestion
    "typegpu": "^0.8.0"
```

apps/typegpu-docs/src/components/ExampleView.tsx
deluksic: This is the only way to get pixel-perfect canvas resolution that I know of! Any other method which scales css pixels is bound to round differently than what the browser engine decides.

https://web.dev/articles/device-pixel-content-box

PR #1917, opened 2025-11-15
feat: tgpu.comptime, tgpu.rawCodeSnippet and `this` allowed in TypeGPU shader functions

iwoplaza:
Changes:
- `externals` is now an arrow function instead of a getter, so that it can capture the `this` value from the outer scope.
- `this` is treated like any other identifier, except for when generating the metadata object in unplugin. We cannot use shorthand syntax for `this`
- `tgpu['~unstable'].comptime` is a way to create functions that are callable from within shader functions, but only allowed to execute at compile-time. They accept only compile-time known arguments, and throw otherwise.
- `tgpu['~unstable'].rawCodeSnippet` lets a JS-implemented function refer to a WGSL expression that TypeGPU doesn't have control over (like in the case of our Three.js integration)

### DIFF
--- a/apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
+++ b/apps/typegpu-docs/src/content/docs/fundamentals/functions/index.mdx
@@ -323,7 +323,7 @@ function manhattanDistance(a: d.v3f, b: d.v3f) {
   const dy = std.abs(a.y - b.y);
   const dz = std.abs(a.z - b.z);
 
-  return std.max(dx, std.max(dy, dz));
+  return std.max(dx, dy, dz);
 }
 ```
 
--- a/apps/typegpu-docs/src/examples/rendering/disco/index.ts
+++ b/apps/typegpu-docs/src/examples/rendering/disco/index.ts
@@ -1,9 +1,8 @@
 import tgpu from 'typegpu';
 import * as d from 'typegpu/data';
-import { mainVertex } from './shaders/vertex.ts';
 import { resolutionAccess, timeAccess } from './consts.ts';
 import {
-  mainFragment,
+  mainFragment1,
   mainFragment2,
   mainFragment3,
   mainFragment4,
@@ -11,6 +10,7 @@ import {
   mainFragment6,
   mainFragment7,
 } from './shaders/fragment.ts';
+import { mainVertex } from './shaders/vertex.ts';
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
 const context = canvas.getContext('webgpu') as GPUCanvasContext;
@@ -32,7 +32,7 @@ const resolutionUniform = root.createUniform(
 );
 
 const fragmentShaders = [
-  mainFragment,
+  mainFragment1,
   mainFragment2,
   mainFragment3,
   mainFragment4,
@@ -105,6 +105,16 @@ export const controls = {
       if (patternIndex !== undefined) {
         currentPipeline = pipelines[patternIndex];
       }
+    },
+  },
+  'Test Resolution': import.meta.env.DEV && {
+    onButtonClick() {
+      const namespace = tgpu['~unstable'].namespace();
+      Array.from({ length: 6 }).map((_, i) =>
+        root.device.createShaderModule({
+          code: tgpu.resolve([pipelines[i + 1]], { names: namespace }),
+        })
+      );
     },
   },
 };
--- a/apps/typegpu-docs/src/examples/rendering/disco/shaders/fragment.ts
+++ b/apps/typegpu-docs/src/examples/rendering/disco/shaders/fragment.ts
@@ -21,33 +21,32 @@ const accumulate = tgpu.fn(
   d.vec3f,
 )((acc, col, weight) => acc.add(col.mul(weight)));
 
-export const mainFragment = tgpu['~unstable'].fragmentFn({
+// Variation1
+export const mainFragment1 = tgpu['~unstable'].fragmentFn({
   in: { uv: d.vec2f },
   out: d.vec4f,
 })(({ uv }) => {
-  {
-    const originalUv = aspectCorrected(uv);
+  const originalUv = aspectCorrected(uv);
 
-    let aspectUv = d.vec2f(originalUv);
-    let accumulatedColor = d.vec3f();
-    for (let iteration = 0.0; iteration < 5.0; iteration++) {
-      aspectUv = std.fract(aspectUv.mul(1.3 * std.sin(timeAccess.$))).sub(0.5);
-      let radialLength = std.length(aspectUv) *
-        std.exp(-std.length(originalUv) * 2);
-      radialLength = std.sin(radialLength * 8 + timeAccess.$) / 8;
-      radialLength = std.abs(radialLength);
-      radialLength = std.smoothstep(0.0, 0.1, radialLength);
-      radialLength = 0.06 / radialLength;
+  let aspectUv = d.vec2f(originalUv);
+  let accumulatedColor = d.vec3f();
+  for (let iteration = 0.0; iteration < 5.0; iteration++) {
+    aspectUv = std.fract(aspectUv.mul(1.3 * std.sin(timeAccess.$))).sub(0.5);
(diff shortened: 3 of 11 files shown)


packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
```diff
@@ -0,0 +1,166 @@
...
+ * //   return EXISTING_GLOBAL * 2;
+ * // }
+ * ```
+ */
+export function rawCodeSnippet<TDataType extends AnyData>(
+  expression: string,
+  type: TDataType,
+  origin: RawCodeSnippetOrigin | undefined = 'runtime',
+): TgpuRawCodeSnippet<TDataType> {
+  return new TgpuRawCodeSnippetImpl(expression, type, 'runtime');
```
aleksanderkatan:
```suggestion
  return new TgpuRawCodeSnippetImpl(expression, type, origin);
```

packages/unplugin-typegpu/src/rollup-impl.ts
```diff
@@ -198,7 +198,11 @@ export const rollUpImpl = (rawOptions: Options) => {
               v: ${FORMAT_VERSION},
               name: ${name ? `"${name}"` : 'undefined'},
               ast: ${embedJSON({ params, body, externalNames })},
-              get externals() { return {${externalNames.join(', ')}}; },
+              externals: () => ({${
+            externalNames.map((e) => e === 'this' ? '"this": this' : e).join(
```
aleksanderkatan, line 202: Can you add some tests for this?

packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
```diff
@@ -0,0 +1,166 @@
...
+}
+
+// --------------
+// Implementation
+// --------------
+
+class TgpuRawCodeSnippetImpl<TDataType extends AnyData>
```
aleksanderkatan, line 94: Please add some resolve tests for the raw code snippet ($uses, constant inlining etc.)

packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
```diff
@@ -0,0 +1,166 @@
...
+    if (!inCodegenMode()) {
+      throw new Error('Raw code snippets can only be used on the GPU.');
+    }
+
+    return this[$gpuValueOf];
+  }
+
+  get value(): InferGPU<TDataType> {
+    return this.$;
+  }
```
aleksanderkatan, line 165: Do we still plan to deprecate and remove the `value` in favor of `$`? If so, maybe we can skip adding the `value` getter in new resources

packages/typegpu/src/core/function/comptime.ts
```diff
@@ -0,0 +1,76 @@
...
+ *   .comptime(() => Math.random());
+ *
+ * const getColor = (diffuse: d.v3f) => {
+ *   'use gpu';
+ *   const albedo = hsvToRgb(injectRand01(), 1, 0.5);
+ *   return albedo.mul(diffuse);
+ * };
+ * ```
+ */
+export function comptime<T extends (...args: never[]) => unknown>(
```
aleksanderkatan, line 35: Please add tests for this...

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -142,9 +142,13 @@ export function createFnCore(
       // get data generated by the plugin
       const pluginData = getMetaData(implementation);
```
aleksanderkatan, line 144: smol
```suggestion

    // Passing a record happens prior to version 0.9.0
    // TODO: Support for this can be removed down the line
```

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -143,6 +143,14 @@ const Transpilers: Partial<
     return node.name;
   },
 
+  ThisExpression(ctx) {
+    if (ctx.ignoreExternalDepth === 0) {
+      ctx.externalNames.add('this');
+    }
+
+    return 'this';
```
aleksanderkatan: I'm not sure what exactly happens here, and whether this is relevant, but `this` is a reserved word and cannot be used to name variables, struct props, structs or functions

iwoplaza: That's true! Could be simplified to return just "this"

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -142,9 +142,15 @@ export function createFnCore(
...
+      // Passing a record happens prior to version 0.9.0
+      // TODO: Support for this can be removed down the line
+      const pluginExternals = typeof pluginData?.externals === 'function'
+        ? pluginData.externals()
+        : pluginData?.externals;
+
+      if (pluginExternals) {
         const missing = Object.fromEntries(
-          Object.entries(pluginData.externals).filter(
+          Object.entries(pluginExternals).filter(
```
reczkok, line 153: Do we need to support both? If we have mismatched versions of the library and the plugin there will be issues anyway and if they match there shouldn't be an issue

packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
```diff
@@ -0,0 +1,166 @@
...
+  replaceExternalsInWgsl,
+} from '../resolve/externals.ts';
+import { valueProxyHandler } from '../valueProxyUtils.ts';
+
+// ----------
+// Public API
+// ----------
+
+/**
+ * Extra declaration that shall be included in final WGSL code,
```
reczkok: My lord 🏰🤴
```suggestion
 * Extra declaration that will be included in final WGSL code
```

packages/typegpu/tests/tgsl/codeGen.test.ts
```diff
@@ -34,4 +34,23 @@ describe('codeGen', () => {
       `);
     });
   });
+
+  it('should properly resolve this', ({ root }) => {
```
reczkok:
```suggestion
  it('should properly resolve the "this" keyword', ({ root }) => {
```

packages/typegpu/tests/tgsl/codeGen.test.ts
```diff
@@ -34,4 +34,23 @@ describe('codeGen', () => {
...
+      myBuffer = root.createUniform(d.u32);
+      myFn = tgpu.fn([], d.u32)(() => {
+        return this.myBuffer.$;
+      });
+    }
+
+    const myController = new MyController();
+
+    expect(tgpu.resolve([myController.myFn])).toMatchInlineSnapshot(`
+      "@group(0) @binding(0) var<uniform> item_1: u32;
```
reczkok, line 49: No auto naming? 🤔

packages/typegpu/tests/tgsl/extensionEnabled.test.ts
reczkok, line 1: What caused this? 🤔

packages/typegpu/src/core/function/fnCore.ts
aleksanderkatan: When somebody publishes a typegpu library, the code is already transformed, and provides the externals as a record. Therefore, you would have to bump all such libraries at once (provided that authors actually updated them already).

packages/typegpu/tests/tgsl/codeGen.test.ts
aleksanderkatan: I created a detailed issue for this: #1940

packages/typegpu/tests/tgsl/extensionEnabled.test.ts
aleksanderkatan: I'm pretty sure that `.toMatchInlineSnapshot` tries to ignore indent and this was a manual stylistic change.
The test still passes if I indent everything, and logging the result shows no unnecessary indentation.

PR #1924, opened 2025-11-20
feat: Accept multiple arguments in min and max

aleksanderkatan:


reczkok (approved):
Free test case  

```ts
  it('does comptime reduction', () => {
    const myMax = tgpu.fn([], d.u32)(() => {
      'use gpu';
      return std.max(12, 33, 444, 12333);
    });

    expect(myMax())
      .toStrictEqual(d.u32(12333));
    expect(tgpu.resolve([myMax])).toMatchInlineSnapshot(`
      "fn myMax() -> u32 {
        return 12333u;
      }"
    `);
  });
```

PR #1928, opened 2025-11-21
fix: Added missing 'Test Resolution' button in disco example

### LINKED ISSUE #1926: fix: Disco example has broken test
Waits for 7 calls but receives only 1 :(

cieplypolar:
Closes #1926.


aleksanderkatan (approved):
LGTM!

PR #1932, opened 2025-11-21
WIP fix: Align @geometry with main

### LINKED ISSUE #1930: fix: Fix 'Lines Combinations' and 'Wind Map' examples

cieplypolar:
- [x] wgsl resolution test for lines combinations example (tricky, because there are a lot of options)

Closes #1930.

### DIFF
--- a/packages/typegpu-geometry/src/circle.ts
+++ b/packages/typegpu-geometry/src/circle.ts
@@ -13,8 +13,8 @@ const SubdivLevelResult = struct({
 
 const getSubdivLevel = tgpu.fn([u32], SubdivLevelResult)(
   (vertexIndex) => {
-    let totalVertexCount = u32();
-    for (let level = u32(); level < 8; level += 1) {
+    let totalVertexCount = u32(0);
+    for (let level = u32(0); level < 8; level += 1) {
       const pointCount = u32(3) * (u32(1) << level);
       const triangleCount = select(
         u32(1),
@@ -69,7 +69,7 @@ export const circle = tgpu.fn([u32], vec2f)(
 
 export function circleVertexCount(subdivLevel: number) {
   let totalVertexCount = 3;
-  for (let level = u32(); level < subdivLevel; level += 1) {
+  for (let level = u32(0); level < subdivLevel; level += 1) {
     totalVertexCount += 9 * (1 << level);
   }
   return totalVertexCount;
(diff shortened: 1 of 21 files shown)


packages/typegpu-geometry/src/circle.ts
```diff
@@ -13,8 +13,8 @@ const SubdivLevelResult = struct({
 
 const getSubdivLevel = tgpu.fn([u32], SubdivLevelResult)(
   (vertexIndex) => {
-    let totalVertexCount = u32(0);
-    for (let level = u32(0); level < 8; level += 1) {
+    let totalVertexCount = u32();
+    for (let level = u32(); level < 8; level += 1) {
```
deluksic: is there a reason to change these two lines? I kinda preferred seeing the explicit 0 in there

cieplypolar: Force of habit, 0s restored ;)

packages/typegpu-geometry/src/lines/caps/index.ts
aleksanderkatan, line 1: This does not look intended, did it look like this before the example stopped working?

image

Also, the last two test cases disappear when the segment becomes too short.

aleksanderkatan (approved):
Nice!

Do we plan on keeping these examples as dev only?

iwoplaza:
> Do we plan on keeping these examples as dev only?

Yes, until we publish the @typegpu/geometry package.

packages/typegpu-geometry/src/lines/caps/index.ts
deluksic: Yeah, I know about these issues. I have this new PR open which rewrites the implementation, then I'll think about what to do in these specific cases. It is not completely obvious to me what strategy is visually best yet.

deluksic: oh wait, this exact issue I have not seen before. But maybe look at the new implementation PR and see if you can rebase your changes. I have already aligned (all?) the implementations with recent `main`.

deluksic: I had another look, and indeed this happens when you try using "arrow" caps. They are kinda fiddly in terms of cases where they work nicely. When there is such a short and thick segment, the arrow is way bigger than it, causing the issue you are seeing. So in this case, yes it is expected.

deluksic:
@cieplypolar  Please let me know if there are more things to align in #1935, otherwise we can close this one.

PR #1934, opened 2025-11-21
fix: Jelly switch WGSL resolution test

### LINKED ISSUE #1933: fix: Jelly switch doesn't have a WGSL resolution test

cieplypolar:
Apart from the test, I added a simple mock for audio loading.

Closes #1933.


packages/typegpu/tests/examples/utils/commonMocks.ts
```diff
@@ -111,6 +112,109 @@ export function mockMnistWeights() {
   });
 }
 
+export function mockAudioLoading() {
+  // Mock fetch for audio files
+  const originalFetch = global.fetch;
+  vi.stubGlobal('fetch', async (url: string) => {
```
reczkok: This will work until we run this mock alongside image loading mock - then one will override the other. We need a better system for this. Maybe our mocks can share a list of callbacks for a given identifier and enabling a mock simply appends the callback. Then right before running the example we stub global and call each of the callbacks before the original implementation.

reczkok (changes requested):
Great work! I

packages/typegpu/tests/examples/utils/commonMocks.ts
cieplypolar: Yeah, you are right, we should rewrite it

cieplypolar: Done

PR #1935, opened 2025-11-22
@typegpu/geometry Simplify variable width lines implementation

deluksic:
This is a complete rewrite of the `lineSegmentVariableWidth` utility, switching from 2 triangles per line and max-area triangle joins to 4 triangles per line and fan-out joins. The new approach produces more overdraw, but it is a MUCH simpler approach. We should have a purpose-built line rendering for optimizing triangle counts and sizes.

New abilities:
- color based on "distance" to centerline
- render only one half, for example for outlines which don't cover inside or outside

| image | image |
|-|-|
| image | image |

### DIFF
--- a/apps/typegpu-docs/tests/individual-example-tests/circles.test.ts
+++ b/apps/typegpu-docs/tests/individual-example-tests/circles.test.ts
@@ -71,16 +71,11 @@ describe('circles example', () => {
         @location(1) @interpolate(flat) instanceIndex: u32,
       }
 
-      struct mainVertexMaxArea_Input {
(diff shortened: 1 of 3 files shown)


apps/typegpu-docs/src/examples/geometry/lines-combinations/index.ts
reczkok: Here is a nice little pattern you might like
Image

deluksic: awesome, done

PR #1937, opened 2025-11-25
docs/fix: Add point light shadow example

reczkok:
- adds a new example that does dynamic point light shadows with instancing
- fixes a couple of bugs
- unifies mobile camera controls and adds pinch to zoom


apps/typegpu-docs/src/examples/rendering/cubemap-reflection/index.ts
cieplypolar, line 1:
```ts
const touchEndEventListener = (e: TouchEvent) => {
  if (e.touches.length === 1) {
    isDragging = true;
    prevX = e.touches[0].clientX;
    prevY = e.touches[0].clientY;
  } else {
    isDragging = false;
  }
};
```
By doing so, we enable users to look around once they lift one finger after zooming.

apps/typegpu-docs/src/examples/rendering/phong-reflection/setup-orbit-camera.ts
cieplypolar, line 1: The Same change as in `cubemap reflection` could be applied.

apps/typegpu-docs/src/examples/rendering/point-light-shadow/box-geometry.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  static get indexCount() {
+    return BoxGeometry.#indexCount;
+  }
+
+  static clearBuffers() {
+    BoxGeometry.#vertexBuffer = null;
+    BoxGeometry.#indexBuffer = null;
+  }
```
cieplypolar, line 142:
```suggestion
  static clearBuffers() {
    BoxGeometry.#vertexBuffer = null;
    BoxGeometry.#indexBuffer = null;
    BoxGeometry.#indexCount = 0;
  }
```

apps/typegpu-docs/src/examples/rendering/point-light-shadow/camera.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  #computeData() {
+    const view = m.mat4.lookAt(
+      d.vec3f(-this.#position.x, this.#position.y, this.#position.z),
+      d.vec3f(-this.#target.x, this.#target.y, this.#target.z),
+      d.vec3f(-this.#up.x, this.#up.y, this.#up.z),
+      d.mat4x4f(),
+    );
+    m.mat4.scale(view, d.vec3f(-1, 1, 1), view);
```
cieplypolar:
```suggestion
  #computeData() {
    const view = m.mat4.lookAt(
      d.vec3f(-this.#position.x, this.#position.y, this.#position.z),
      d.vec3f(-this.#target.x, this.#target.y, this.#target.z),
      d.vec3f(-this.#up.x, this.#up.y, this.#up.z),
      d.mat4x4f(),
    );
    m.mat4.scale(view, d.vec3f(-1, 1, 1), view);
```
This mirroring with -1 * is counterintuitive for me. We can easily get rid of it by changing how we update orbit

apps/typegpu-docs/src/examples/rendering/point-light-shadow/point-light.ts
```diff
@@ -0,0 +1,124 @@
...
+  TgpuRoot,
+} from 'typegpu';
+import * as d from 'typegpu/data';
+import { BoxGeometry } from './box-geometry.ts';
+import { Camera } from './camera.ts';
+import type { Scene } from './scene.ts';
+import { instanceLayout, vertexLayout } from './types.ts';
+
+const FACE_CONFIGS = [
+  { name: 'right', dir: d.vec3f(1, 0, 0), up: d.vec3f(0, 1, 0) },
```
cieplypolar:
```suggestion
  { name: 'right', offsetDir: d.vec3f(1, 0, 0), up: d.vec3f(0, 1, 0) },
```

apps/typegpu-docs/src/examples/rendering/point-light-shadow/index.ts
```diff
@@ -0,0 +1,613 @@
...
+  );
+  orbitingCube.scale = d.vec3f(0.5, 0.5, 0.5);
+  orbitingCubes.push(orbitingCube);
+  scene.add(orbitingCube);
+}
+
+const floorCube = new BoxGeometry(root);
+floorCube.scale = d.vec3f(10, 0.1, 10);
+floorCube.position = d.vec3f(0, -0.5, 0);
+scene.add(floorCube);
```
cieplypolar: I know performance-wise it won't make a difference, but can you add a method to the scene that accepts an array of objects instead of creating a new buffer for each object?

apps/typegpu-docs/src/examples/rendering/point-light-shadow/index.ts
cieplypolar, line 1: While drawing  the depth preview, we can easily add borders for each camera view
```ts
  const ratio = 2 / 3;
  if (std.abs(2 * uv.y - 1.0) <= 0.003 * ratio) {
    return d.vec4f(0, 1, 0, 1);
  }
  if (std.abs(3 * uv.x - 1.0) <= 0.003 || std.abs(3 * uv.x - 2.0) <= 0.003) {
    return d.vec4f(0, 1, 0, 1);
  }
```
Green color is arbitrary, probably should be changed ;)

apps/typegpu-docs/src/examples/rendering/two-boxes/index.ts
```diff
@@ -398,94 +404,85 @@ canvas.addEventListener('touchend', () => {
...
+  } else if (e.button === 2) {
     isRightDragging = true;
   }
-  prevX = event.clientX;
-  prevY = event.clientY;
+  prevX = e.clientX;
+  prevY = e.clientY;
 });
 
+canvas.addEventListener('touchstart', (e) => {
```
cieplypolar, line 422: Two finger drag should rotate the cubes (according to the info note).
Either change the info note or the behavior of two finger drag.

packages/typegpu/tests/examples/individual/point-light-shadow.test.ts
```diff
@@ -0,0 +1,175 @@
+/**
+ * @vitest-environment jsdom
+ */
+
+import { describe, expect } from 'vitest';
+import { it } from '../../utils/extendedIt.ts';
+import { runExampleTest, setupCommonMocks } from '../utils/baseTest.ts';
+import { mockResizeObserver } from '../utils/commonMocks.ts';
+
+describe('perlin noise example', () => {
```
cieplypolar:
```suggestion
describe('point light shadow example', () => {
```

apps/typegpu-docs/src/examples/rendering/point-light-shadow/index.ts
```diff
@@ -0,0 +1,613 @@
...
+    lastPinchDist = Math.sqrt(dx * dx + dy * dy);
+  }
+}, { passive: false });
+
+const mouseUpEventListener = () => {
+  isDragging = false;
+};
+window.addEventListener('mouseup', mouseUpEventListener);
+
+const touchEndEventListener = () => {
```
cieplypolar: same as above (handle if 1 finger still touches the window)

apps/typegpu-docs/src/examples/rendering/point-light-shadow/box-geometry.ts
reczkok: Is there any particular reason for that? The index count is constant

apps/typegpu-docs/src/examples/rendering/cubemap-reflection/index.ts
reczkok: Great idea!

apps/typegpu-docs/src/examples/rendering/point-light-shadow/point-light.ts
reczkok: I think dir is fine since this is the vector that tells us which direction the camera is looking. Offset dir would refer to offsetting the lookAt position which is less intuitive to me

apps/typegpu-docs/src/examples/rendering/point-light-shadow/camera.ts
reczkok: It actually will break all the shadows

reczkok: Fixed it

cieplypolar (approved):

PR #1941, opened 2025-11-25
fix: Deleted unnecessary braces in disco example

cieplypolar:


apps/typegpu-docs/src/examples/rendering/disco/index.ts
```diff
@@ -109,9 +109,10 @@ export const controls = {
   },
   'Test Resolution': import.meta.env.DEV && {
     onButtonClick() {
+      const namespace = tgpu['~unstable'].namespace();
```
iwoplaza, line 112: Was this meant to reduce duplication of common definitions between variants?

Also, this PR switched from "strict" back into "random" naming. Was that intentional?

PR #1945, opened 2025-12-02
impr: three.js and its types bump to r181

cieplypolar:
```ts
TSL.ShaderNodeObject -> THREE.TSL.NodeObject
```


cieplypolar:
Done

PR #1948, opened 2025-12-02
feat: ThreeJS 'Compute Geometry' with tgpu

cieplypolar:


apps/typegpu-docs/src/examples/threejs/compute-geometry/index.html
```diff
@@ -0,0 +1 @@
+<canvas></canvas>
```
aleksanderkatan:
```suggestion
<canvas data-fit-to-container></canvas>
```

apps/typegpu-docs/public/assets/threejs/compute-geometry/LeePerrySmith_License.txt
aleksanderkatan, line 1: image

Do we care that the licence is not listed here?

apps/typegpu-docs/src/examples/threejs/compute-geometry/index.ts
aleksanderkatan, line 1: I know the same happens on the three.js site, but can we disable torso rotation on mobile? Right now the jiggle cannot be achieved without rotating the model

apps/typegpu-docs/public/assets/threejs/compute-geometry/LeePerrySmith_License.txt
cieplypolar: I will add it, just in case

apps/typegpu-docs/src/examples/threejs/compute-geometry/index.ts
cieplypolar: sure, let me know if it works on your device

aleksanderkatan: Works just fine, I can also scale and move with two fingers

PR #1950, opened 2025-12-02
impr: Fix errors and types around textures

reczkok:
- support for omitting empty objects in `withVertex` and `withFragment` calls
- type support for non float fragment outputs
- reworked texture format data (parsed instead of stored)
- added support for render views
- looser types for both color attachments and depth attachments  
- reworked the texture test to include more formats (including non filterable ones)
- reworked the mipmap and resampling generation

### DIFF
--- a/packages/typegpu/src/core/texture/textureUtils.ts
+++ b/packages/typegpu/src/core/texture/textureUtils.ts
@@ -73,62 +73,86 @@ type BlitResources = {
   sampler: GPUSampler;
 };
 
-const blitCache = new WeakMap<GPUDevice, Map<string, BlitResources>>();
+type DeviceCache = {
+  vertexModule: GPUShaderModule;
+  filterableResources: Map<
+    boolean,
+    { fragmentModule: GPUShaderModule; sampler: GPUSampler }
+  >;
+  layoutResources: Map<
+    string,
+    { bindGroupLayout: GPUBindGroupLayout; pipelineLayout: GPUPipelineLayout }
+  >;
+};
+
+const blitCache = new WeakMap<GPUDevice, DeviceCache>();
+
+function getOrCreateDeviceCache(device: GPUDevice): DeviceCache {
+  let cache = blitCache.get(device);
+  if (!cache) {
+    cache = {
+      vertexModule: device.createShaderModule({
+        code: FULLSCREEN_VERTEX_SHADER,
+      }),
+      filterableResources: new Map(),
+      layoutResources: new Map(),
+    };
+    blitCache.set(device, cache);
+  }
+  return cache;
+}
 
 function getBlitResources(
   device: GPUDevice,
   filterable: boolean,
   sampleType: GPUTextureSampleType,
 ): BlitResources {
-  let cache = blitCache.get(device);
-  if (!cache) {
-    cache = new Map();
-    blitCache.set(device, cache);
+  const cache = getOrCreateDeviceCache(device);
+
+  let filterableRes = cache.filterableResources.get(filterable);
+  if (!filterableRes) {
+    filterableRes = {
+      fragmentModule: device.createShaderModule({
+        code: filterable ? SAMPLE_FRAGMENT_SHADER : GATHER_FRAGMENT_SHADER,
+      }),
+      sampler: device.createSampler(
+        filterable ? { magFilter: 'linear', minFilter: 'linear' } : {},
+      ),
+    };
+    cache.filterableResources.set(filterable, filterableRes);
   }
 
-  const key = `${filterable}:${sampleType}`;
-  let resources = cache.get(key);
-  if (resources) {
-    return resources;
+  const layoutKey = `${filterable}:${sampleType}`;
+  let layoutRes = cache.layoutResources.get(layoutKey);
+  if (!layoutRes) {
+    const bindGroupLayout = device.createBindGroupLayout({
+      entries: [
+        {
+          binding: 0,
+          visibility: GPUShaderStage.FRAGMENT,
+          texture: { sampleType },
+        },
+        {
+          binding: 1,
+          visibility: GPUShaderStage.FRAGMENT,
+          sampler: { type: filterable ? 'filtering' : 'non-filtering' },
+        },
+      ],
+    });
+    layoutRes = {
+      bindGroupLayout,
+      pipelineLayout: device.createPipelineLayout({
+        bindGroupLayouts: [bindGroupLayout],
+      }),
+    };
+    cache.layoutResources.set(layoutKey, layoutRes);
   }
 
-  const vertexModule = device.createShaderModule({
-    code: FULLSCREEN_VERTEX_SHADER,
-  });
-  const fragmentModule = device.createShaderModule({
-    code: filterable ? SAMPLE_FRAGMENT_SHADER : GATHER_FRAGMENT_SHADER,
-  });
-
-  const bindGroupLayout = device.createBindGroupLayout({
-    entries: [
-      {
-        binding: 0,
-        visibility: GPUShaderStage.FRAGMENT,
-        texture: { sampleType },
-      },
-      {
-        binding: 1,
-        visibility: GPUShaderStage.FRAGMENT,
-        sampler: { type: filterable ? 'filtering' : 'non-filtering' },
-      },
-    ],
-  });
-  const pipelineLayout = device.createPipelineLayout({
-    bindGroupLayouts: [bindGroupLayout],
-  });
-  const sampler = device.createSampler(
-    filterable ? { magFilter: 'linear', minFilter: 'linear' } : {},
-  );
-
-  resources = {
-    vertexModule,
-    fragmentModule,
-    bindGroupLayout,
-    pipelineLayout,
-    sampler,
+  return {
+    vertexModule: cache.vertexModule,
+    ...filterableRes,
+    ...layoutRes,
   };
-  cache.set(key, resources);
-  return resources;
 }
 
 type BlitOptions = {
@@ -182,7 +206,7 @@ function blit(options: BlitOptions): void {
 }
 
 export function clearTextureUtilsCache(device: GPUDevice): void {
-  blitCache.get(device)?.clear();
+  blitCache.delete(device);
 }
 
 function validateBlitFormat(
--- a/packages/typegpu/tests/texture.test.ts
+++ b/packages/typegpu/tests/texture.test.ts
@@ -453,6 +453,55 @@ Overload 3 of 4, '(schema: "(Error) Texture not usable as storage, call $usage('
         expect(device.mock.queue.submit).toHaveBeenCalled();
       });
 
+      it('caches blit resources appropriately per level', ({ root, device }) => {
+        const createTex = (format: GPUTextureFormat) =>
+          root.createTexture({ size: [32, 32], format, mipLevelCount: 2 })
+            .$usage('render');
+
+        const getCalls = () => ({
+          shaderModule: device.mock.createShaderModule.mock.calls.length,
+          sampler: device.mock.createSampler.mock.calls.length,
+          bindGroupLayout: device.mock.createBindGroupLayout.mock.calls.length,
+          pipelineLayout: device.mock.createPipelineLayout.mock.calls.length,
+        });
+
+        // First filterable texture
+        createTex('rgba8unorm').generateMipmaps();
+        expect(getCalls()).toEqual({
+          shaderModule: 2,
+          sampler: 1,
+          bindGroupLayout: 1,
+          pipelineLayout: 1,
+        });
+
+        // Same format - all cached
+        createTex('rgba8unorm').generateMipmaps();
+        expect(getCalls()).toEqual({
+          shaderModule: 2,
+          sampler: 1,
+          bindGroupLayout: 1,
+          pipelineLayout: 1,
+        });
+
+        // Different filterable format - still uses same cache (both filterable floats)
+        createTex('rgba16float').generateMipmaps();
+        expect(getCalls()).toEqual({
+          shaderModule: 2,
+          sampler: 1,
+          bindGroupLayout: 1,
+          pipelineLayout: 1,
+        });
+
+        // Unfilterable format - new fragment shader, sampler, and layouts (vertex module reused)
+        createTex('r32float').generateMipmaps();
+        expect(getCalls()).toEqual({
+          shaderModule: 3,
+          sampler: 2,
+          bindGroupLayout: 2,
+          pipelineLayout: 2,
+        });
+      });
+
       it('calls queue.writeTexture when write is called with buffer data', ({ root, device }) => {
         const texture = root.createTexture({
           size: [4, 4],


packages/typegpu/tests/examples/individual/gravity.test.ts
```diff
@@ -23,11 +23,33 @@ describe('gravity example', () => {
         mock3DModelLoading();
         mockResizeObserver();
       },
-      expectedCalls: 4,
+      expectedCalls: 6,
```
reczkok, line 28: In the actual example we don't need to resample since all the dimensions match but our mock doesn't - that's why we see the resampling code here

packages/typegpu/tests/textureFormats.test.ts
```diff
@@ -0,0 +1,307 @@
...
+      [
+        ['r32uint', 4, u32, vec4u, ['uint'], true],
+        ['r32sint', 4, i32, vec4i, ['sint'], true],
+        ['r32float', 4, f32, vec4f, ['float', 'unfilterable-float'], true],
+        ['rg32uint', 8, u32, vec4u, ['uint'], true],
+        ['rg32sint', 8, i32, vec4i, ['sint'], true],
+        ['rg32float', 8, f32, vec4f, ['float', 'unfilterable-float'], true],
+        ['rgba32uint', 16, u32, vec4u, ['uint'], true],
+        ['rgba32sint', 16, i32, vec4i, ['sint'], true],
+        ['rgba32float', 16, f32, vec4f, ['float', 'unfilterable-float'], true],
```
aleksanderkatan, line 110: 32 bits float textures only support float sample type if the appropriate feature is present. If I'm not mistaken, textures could keep a reference to the root, and `getTextureFormatInfo` is used mainly inside textures. Do we plan on checking that in runtime in the future?

packages/typegpu/tests/textureFormats.test.ts
```diff
@@ -0,0 +1,307 @@
...
+        ['rgba8unorm', 4, f32, vec4f, ['float', 'unfilterable-float'], true],
+        [
+          'rgba8unorm-srgb',
+          4,
+          f32,
+          vec4f,
+          ['float', 'unfilterable-float'],
+          true,
+        ],
+        ['rgba8snorm', 4, f32, vec4f, ['float', 'unfilterable-float'], false],
```
aleksanderkatan: Same question about ["texture-formats-tier1"](https://www.w3.org/TR/webgpu/#texture-formats-tier1). Do we maybe want to say `true` by default because this is possible *sometimes*?

There are more such ifs in the table.

packages/typegpu/tests/textureFormats.test.ts
```diff
@@ -0,0 +1,307 @@
...
+    );
+  });
+
+  describe('packed 32-bit formats', () => {
+    it.each(
+      [
+        ['rgb10a2uint', 4, u32, vec4u, ['uint'], true],
+        ['rgb10a2unorm', 4, f32, vec4f, ['float', 'unfilterable-float'], true],
+        ['rg11b10ufloat', 4, f32, vec4f, ['float', 'unfilterable-float'], true],
+        ['rgb9e5ufloat', 4, f32, vec4f, ['float', 'unfilterable-float'], true],
```
aleksanderkatan: I don't see this one in the table, is this really renderable?

image

packages/typegpu/tests/textureFormats.test.ts
```diff
@@ -0,0 +1,307 @@
...
+          f32,
+          vec4f,
+          ['depth', 'unfilterable-float'],
+          true,
+        ],
+        ['depth32float', 4, f32, vec4f, ['depth', 'unfilterable-float'], true],
+        ['depth32float-stencil8', 8, f32, vec4f, [
+          'depth',
+          'unfilterable-float',
+        ], true],
```
aleksanderkatan: I don't really understand this table, but in specs I see some uints, shouldn't they be here as well?

image

packages/typegpu/src/core/texture/textureFormats.ts
```diff
@@ -56,246 +60,126 @@ export type StorageTextureFormats =
...
+    format.startsWith('bc') ||
+    format.startsWith('etc2') ||
+    format.startsWith('eac') ||
+    format.startsWith('astc')
+  ) {
+    return false;
   }
+  if (format.includes('snorm')) return false;
+  return true;
+}
```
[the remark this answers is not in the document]
reczkok: Thar's just false

packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -112,15 +112,26 @@ export interface WithCompute {
   createPipeline(): TgpuComputePipeline;
 }
 
+type IsEmptyRecord<T> = keyof T extends never ? true
+  : T extends Record<string, never> ? true
+  : false;
```
aleksanderkatan:
```suggestion
type IsEmptyRecord<T> = T extends Record<string, never> ? true : false;
```
Can this be simplified or is there some case I don't see?

packages/typegpu/src/core/texture/textureUtils.ts
```diff
@@ -27,289 +22,294 @@ export function getImageSourceDimensions(
...
 
-type CachedResources = {
-  vertexShader: GPUShaderModule;
-  fragmentShader: GPUShaderModule;
+@vertex
+fn vs_main(@builtin(vertex_index) i: u32) -> VertexOutput {
+  const pos = array(vec2f(-1, -1), vec2f(3, -1), vec2f(-1, 3));
+  const uv = array(vec2f(0, 1), vec2f(2, 1), vec2f(0, -1));
+  return VertexOutput(vec4f(pos[i], 0, 1), uv[i]);
+}`;
```
aleksanderkatan, line 44: Can we use `fullScreenTriangle` here?

aleksanderkatan (approved):
Great changes! 🛫📗

packages/typegpu/tests/textureFormats.test.ts
reczkok: There is a runtime check:
image

packages/typegpu/src/core/texture/textureUtils.ts
reczkok: I'd rather not

reczkok: It's very simple and won't ever change and imports are trouble

packages/typegpu/src/core/root/rootTypes.ts
reczkok: The `keyof T` check is safer if we ever get something that is not an object passed here - we should be fine with the simpler one tho

apps/typegpu-docs/src/examples/tests/texture-test/index.ts
```diff
@@ -16,56 +16,127 @@ const imageBitmap = await createImageBitmap(
...
+  const mipLevels = calculateMipLevels(size[0], size[1]);
+  console.log(
+    `Creating texture: ${format}, size: ${size[0]}x${
+      size[1]
+    }, mips: ${mipLevels}, filterable: ${isFilterable(format)}`,
+  );
+
+  return root['~unstable']
+    .createTexture({
+      size: [size[0], size[1]],
```
cieplypolar:
```suggestion
      size,
```

apps/typegpu-docs/src/examples/tests/texture-test/index.ts
```diff
@@ -16,56 +16,127 @@ const imageBitmap = await createImageBitmap(
...
+    bias: biasUniform,
+    channel: channelUniform,
+  });
 
-const pipeline = root['~unstable']
-  .withVertex(fullScreenTriangle, {})
-  .withFragment(fragmentFunction, { format: presentationFormat })
-  .createPipeline();
+  const pipeline = root['~unstable']
+    .withVertex(fullScreenTriangle, {})
```
cieplypolar:
```suggestion
    .withVertex(fullScreenTriangle)
```

apps/typegpu-docs/src/examples/tests/texture-test/index.ts
```diff
@@ -16,56 +16,127 @@ const imageBitmap = await createImageBitmap(
...
+    if (channel == 3) { return vec4f(color.bbb, 1.0); }
+    if (channel == 4) { return vec4f(color.aaa, 1.0); }
+
+    return color;
+  }`.$uses({
+    texture: layout.bound.myTexture,
+    sampler,
+    bias: biasUniform,
+    channel: channelUniform,
+  });
```
cieplypolar, line 116:
```suggestion
  const textureSampleBiasTGSL = tgpu.fn([
    d.texture2d(d.f32),
    d.sampler(),
    d.vec2f,
    d.f32,
  ], d.vec4f)`(texture, sampler, uv, bias) -> vec4f {
    return textureSampleBias(texture, sampler, uv, bias);
  }`;

  const textureNumLevelsTGSL = tgpu.fn([
    d.texture2d(d.f32),
  ], d.u32)`(texture) -> u32 {
    return textureNumLevels(texture);
  }`;

  const caclulateMipLevel = tgpu.fn([
    d.texture2d(d.f32),
    d.vec2f,
  ], d.u32)`(texture, uv) -> u32 {
    let size = vec2f(textureDimensions(texture));
    let dx = dpdx(uv * size);
    let dy = dpdy(uv * size);

    let max_delta = max(length(dx), length(dy));

    return u32(log2(max_delta));
  }`;

  const texture = layout.bound.myTexture;
  const fragmentFunction = tgpu['~unstable'].fragmentFn({
    in: { uv: d.vec2f },
    out: d.vec4f,
  })((input) => {
    const color = textureSampleBiasTGSL(
      texture.$,
      sampler.$,
      input.uv,
      biasUniform.$,
    );

    const numLevels = textureNumLevelsTGSL(texture.$);
    const mipLevel = caclulateMipLevel(texture.$, input.uv);
    // don't judge, it works
    if (d.u32(input.uv.x * 1000) === 0 && d.u32(input.uv.y * 1000) === 0) {
      console.log('Mip level count:', numLevels);
      console.log('Mip level:', mipLevel);
    }

    if (channelUniform.$ === 1) return d.vec4f(color.xxx, 1.0);
    if (channelUniform.$ === 2) return d.vec4f(color.yyy, 1.0);
    if (channelUniform.$ === 3) return d.vec4f(color.zzz, 1.0);
    if (channelUniform.$ === 4) return d.vec4f(color.www, 1.0);

    return color;
  });
```
I rewrote it to TGSL to see if negative bias does something (I don't see any effect). It doesn't because we are always on level 0.

Note that the `caclulateMipLevel` is vibecoded.

packages/typegpu/src/core/texture/textureFormats.ts
```diff
@@ -56,246 +60,153 @@ export type StorageTextureFormats =
...
-      'r8unorm', 'r8uint', 'r8sint',
-      'r16uint', 'r16sint', 'r16float',
-      'rgba8unorm', 'rgba8uint', 'rgba8sint',
-      'rgba16uint', 'rgba16sint', 'rgba16float',
-      'r32uint', 'r32sint', 'r32float',
-      'rgba32uint', 'rgba32sint', 'rgba32float',
-    ];
+  // Depth/stencil
+  if (format === 'stencil8') return 1;
+  if (format === 'depth32float-stencil8') return 8;
```
cieplypolar: Are you sure its 8 and not 5?

packages/typegpu/src/core/texture/textureUtils.ts
```diff
@@ -27,289 +22,294 @@ export function getImageSourceDimensions(
...
+        sampler: { type: filterable ? 'filtering' : 'non-filtering' },
+      },
+    ],
+  });
+  const pipelineLayout = device.createPipelineLayout({
+    bindGroupLayouts: [bindGroupLayout],
+  });
+  const sampler = device.createSampler(
+    filterable ? { magFilter: 'linear', minFilter: 'linear' } : {},
+  );
```
cieplypolar: Can we cache `vertexModule`, `fragmentModule` and `sampler`? That's only 5 variables and we would avoid unnecessary calls to the device.

packages/typegpu/src/core/texture/textureUtils.ts
```diff
@@ -27,289 +22,294 @@ export function getImageSourceDimensions(
...
-    cached = {
-      vertexShader,
-      fragmentShader,
-      bindGroupLayout,
-      pipelineLayout,
-      sampler,
-    };
-    deviceCache.set(cacheKey, cached);
+): void {
+  if (targetTexture.dimension === '3d') {
```
cieplypolar:
```suggestion
  if (targetTexture.dimension !== '2d') {
```

cieplypolar (commented):
Great work! Code is much cleaner. Love the use of `textureGather`. Left some comments where I wasn't sure about the correctness.

packages/typegpu/src/core/texture/textureFormats.ts
reczkok: Sadly there is no guarantee so I assume the worst:
image

apps/typegpu-docs/src/examples/tests/texture-test/index.ts
reczkok: If TGSL then why not go all the way? Quite a lot of code

packages/typegpu/src/core/texture/textureUtils.ts
reczkok: Aren't we?

packages/typegpu/src/core/texture/textureFormats.ts
reczkok: I just noticed that I am inconsistent with the assumptions. I will change all to be worst case I guess?

cieplypolar: good idea

packages/typegpu/src/core/texture/textureUtils.ts
cieplypolar: I think we create shader modules for each resource, but we could only per device.

reczkok: Oh - yeah I guess we could

reczkok: There aren't a lot of combinations (for most people exactly two)

apps/typegpu-docs/src/examples/tests/texture-test/index.ts
cieplypolar: We didn't implement `textureNumLevels` in std

```ts
const textureNumLevelsTGSL = tgpu.fn([
    d.texture2d(d.f32),
  ], d.u32)`(texture) -> u32 {
    return textureNumLevels(texture);
  }`;
  const texture = layout.bound.myTexture;
  const fragmentFunction = tgpu['~unstable'].fragmentFn({
    in: { uv: d.vec2f },
    out: d.vec4f,
  })((input) => {
    const color = std.textureSampleBias(
      texture.$,
      sampler.$,
      input.uv,
      biasUniform.$,
    );

    const size = d.vec2f(std.textureDimensions(texture.$));
    const dx = std.dpdx(input.uv.mul(size));
    const dy = std.dpdy(input.uv.mul(size));
    const max_delta = std.max(std.length(dx), std.length(dy));
    const numLevels = textureNumLevelsTGSL(texture.$);
    const mipLevel = d.u32(std.log2(max_delta));
    // don't judge, it works
    if (d.u32(input.uv.x * 1000) === 0 && d.u32(input.uv.y * 1000) === 0) {
      console.log('Mip level count:', numLevels);
      console.log('Mip level:', mipLevel);
    }
    if (channelUniform.$ === 1) return d.vec4f(color.xxx, 1.0);
    if (channelUniform.$ === 2) return d.vec4f(color.yyy, 1.0);
    if (channelUniform.$ === 3) return d.vec4f(color.zzz, 1.0);
    if (channelUniform.$ === 4) return d.vec4f(color.www, 1.0);
    return color;
  });
```

cieplypolar: You don't have to accept this change, but I think it could be beneficial for debugging if we could see the mip level

reczkok: Is the manual mip level consistent with the chosen one considering bias? I don't think so

PR #1951, opened 2025-12-03
docs: Background segmentation

lursz:
Introducing variety of in-house developed models and pulling them directly form hugging face.


apps/typegpu-docs/src/examples/image-processing/background-segmentation/schemas.ts
```diff
@@ -15,6 +15,8 @@ export const prepareModelInputLayout = tgpu.bindGroupLayout({
   sampler: { sampler: 'filtering' },
 });
 
+export const cropBoundsSlot = tgpu.slot<TgpuUniform<d.Vec4f>>();
```
iwoplaza: I'd recommend an accessor for this usecase instead. We don't really care that it's a uniform, rather that it's a vec4f

```suggestion
export const cropBoundsAccess = tgpu['~unstable'].accessor(d.vec4f);
```

reczkok (approved):

apps/typegpu-docs/src/examples/image-processing/background-segmentation/model.ts
```diff
@@ -1,30 +1,151 @@
 import * as ort from 'onnxruntime-web/webgpu';
+import type { ModelConfig } from './schemas.ts';
 
 export const MODEL_WIDTH = 320;
 export const MODEL_HEIGHT = 320;
 
+const modelCache = new Map<string, ArrayBuffer>();
```
reczkok, line 7: Do we clear the cache in example cleanup?

lursz: We do not. Nice catch

aleksanderkatan: Do we need to? Won't this just get GCed?
Edit: oh, import caching

PR #1952, opened 2025-12-03
docs: Jump flood algorithm example

reczkok:
It's pretty - next step is to use it to create an outline around a 3d object in a scene 😉

thumbnail

### DIFF
--- a/apps/typegpu-docs/src/examples/algorithms/jump-flood/index.ts
+++ b/apps/typegpu-docs/src/examples/algorithms/jump-flood/index.ts
@@ -424,10 +424,35 @@ const onMouseUp = () => {
   lastDrawPos = null;
 };
 
+const onTouchStart = (e: TouchEvent) => {
+  isDrawing = true;
+  lastDrawPos = null;
+  const rect = canvas.getBoundingClientRect();
+  const touch = e.touches[0];
+  interpolateAndDraw(touch.clientX - rect.left, touch.clientY - rect.top);
+};
+
+const onTouchMove = (e: TouchEvent) => {
+  if (!isDrawing) return;
+  e.preventDefault();
+  const rect = canvas.getBoundingClientRect();
+  const touch = e.touches[0];
+  interpolateAndDraw(touch.clientX - rect.left, touch.clientY - rect.top);
+};
+
+const onTouchEnd = () => {
+  isDrawing = false;
+  lastDrawPos = null;
+};
+
 canvas.addEventListener('mousedown', onMouseDown);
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/examples/algorithms/jump-flood/index.ts
```diff
@@ -0,0 +1,330 @@
...
+
+export const controls = {
+  Reset: {
+    onButtonClick: reset,
+  },
+  'Seed density (%)': {
+    initial: 0.1,
+    min: 0.001,
+    max: 5,
+    step: 0.001,
```
aleksanderkatan: Maybe we could do this a select? It's really difficult to select values smaller than 0.1, and those are interesting as well

apps/typegpu-docs/src/examples/algorithms/jump-flood/index.ts
```diff
@@ -0,0 +1,330 @@
...
+let floodBindGroups = createFloodBindGroups(textures);
+let initBindGroup = root.createBindGroup(initLayout, {
+  writeView: textures[0],
+});
+
+const initializeRandom = root['~unstable'].createGuardedComputePipeline(
+  (x, y) => {
+    'use gpu';
+    const size = std.textureDimensions(initLayout.$.writeView);
+    randf.seed2(d.vec2f(x, y).div(d.vec2f(size)));
```
aleksanderkatan: How about we pass time in a uniform to the seed so that the same % don't always have the same outcome?

aleksanderkatan (approved):
Great example! 🥐🍦

apps/typegpu-docs/src/examples/algorithms/jump-flood/index.ts
reczkok: But it will be just as hard if we just do 2 of them - how many would you like?

reczkok: I will add a non-linear mapping so we get more precision at the lower ends
image

apps/typegpu-docs/src/examples/algorithms/jump-flood/index.ts
```diff
@@ -0,0 +1,506 @@
...
+  ));
+  brushSizeUniform.write(brushSize);
+  drawSeed.with(resources.pingPongBindGroups[sourceIdx]).dispatchThreads(
+    canvas.width,
+    canvas.height,
+  );
+  swap();
+  render();
+}
+
```
lursz, line 324: nit: on mobile devices you can't really drag the brush

Issue #1940, opened 2025-11-25
feat: More autonaming


aleksanderkatan:
Those are the cases when we perform autonaming as of right now:
``` ts
    node.type === 'VariableDeclarator' &&
    node.id.type === 'Identifier' &&
```
``` ts
    node.type === 'AssignmentExpression' &&
    node.left.type === 'Identifier' &&
```
``` ts
    (node.type === 'Property' || node.type === 'ObjectProperty') &&
    node.key.type === 'Identifier' &&
```

Take a look at the following code:
``` ts
import tgpu, { TgpuUniform } from 'typegpu';
import * as d from 'typegpu/data';
const root = await tgpu.init();

class MyController {
  myBuffer1 = root.createUniform(d.u32);
  myBuffer2: TgpuUniform<d.U32>;

  constructor(myBuffer3: TgpuUniform<d.U32>) {
    this.myBuffer2 = root.createUniform(d.u32);
  }
}
```
- `myBuffer1` won't be autonamed (ClassProperty),
- `myBuffer2` won't be autonamed (AssignmentExpression), because the left side is an assignment expression, not an identifier,
- `myBuffer3` won't be auto(re)named.

We could autoname `myBuffer1` and `myBuffer2`, however I see no feasible way how we could do do it with `myBuffer3`.

PR #1953, opened 2025-12-04
feat: More autonaming

aleksanderkatan:
See #1940 description

### DIFF
--- a/apps/typegpu-docs/src/examples/rendering/3d-fish/load-model.ts
+++ b/apps/typegpu-docs/src/examples/rendering/3d-fish/load-model.ts
@@ -50,11 +50,7 @@ export async function loadModel(
     .$usage('sampled', 'render')
     .$name(`texture from ${texturePath}`);
 
-  root.device.queue.copyExternalImageToTexture(
-    { source: imageBitmap },
-    { texture: root.unwrap(texture) },
-    [imageBitmap.width, imageBitmap.height],
-  );
+  texture.write(imageBitmap);
 
   return {
     vertexBuffer,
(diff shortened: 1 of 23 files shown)


packages/unplugin-typegpu/src/common.ts
```diff
@@ -250,16 +250,45 @@ export function performExpressionNaming<T extends acorn.AnyNode | babel.Node>(
...
+ * tryFindIdentifier('this.myBuffer'); // 'myBuffer'
+ * tryFindIdentifier('[a, b]'); // undefined
+ */
+function tryFindIdentifier(
+  node: acorn.AnyNode | babel.Node,
+): string | undefined {
+  if (node.type === 'Identifier') {
+    return node.name;
+  }
+  if (node.type === 'MemberExpression') {
```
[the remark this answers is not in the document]
aleksanderkatan: The should be no false positives with the current approach.

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -63,7 +63,8 @@ export class TgpuAccessorImpl<T extends AnyWgslData>
       | Infer<T>
       | undefined = undefined,
   ) {
-    this.slot = slot(defaultValue);
+    // NOTE: unplugin runs on package typegpu, so we have to avoid auto-naming triggering here
```
reczkok: I thought it didn't 🤔

reczkok (approved):
Very cool

packages/typegpu/src/core/slot/accessor.ts
aleksanderkatan: So did I, but at least when running CI locally, it does

iwoplaza: It runs during typegpu-docs CI, which imports typegpu and uses unplugin-typegpu. This means apps that use typegpu and unplugin-typegpu, with the appropriate settings, can encounter that same behavior. I think we should adjust the comment to specify that this code CAN go through unplugin-typegpu in certain setups.

PR #1961, opened 2025-12-08
chore: Remove 'createDualImpl'

iwoplaza:


packages/typegpu/src/core/function/comptime.ts
```diff
@@ -52,8 +52,14 @@ export function comptime<T extends (...args: never[]) => unknown>(
...
-      return gpuImpl(...args as MapValueToSnippet<Parameters<T>>);
+    const ctx = getResolutionCtx();
+    if (ctx && ctx.mode.type === 'codegen') {
+      ctx.pushMode(new NormalState());
+      try {
+        return gpuImpl(...args as MapValueToSnippet<Parameters<T>>);
+      } finally {
+        ctx.popMode('normal');
+      }
     }
```
iwoplaza, line 63: This was also fixed in #1957, but had to fix it here as well to make tests pass

packages/typegpu/src/core/function/dualImpl.ts
```diff
@@ -135,9 +110,11 @@ export function dualImpl<T extends (...args: never[]) => unknown>(
...
       jsImpl: options.normalImpl,
       gpuImpl,
-      strictSignature: typeof options.signature !== 'function'
-        ? options.signature
-        : undefined,
+      get strictSignature() {
+        return typeof options.signature !== 'function'
+          ? options.signature
+          : undefined;
+      },
```
iwoplaza, line 117: Some operations on matrices have circular dependencies at runtime, this defers accessing options.signature to a later point.

packages/typegpu/src/data/array.ts
```diff
@@ -37,39 +35,14 @@ interface WgslArrayConstructor {
  * @param elementType The type of elements in the array.
  * @param elementCount The number of elements in the array.
  */
-export const arrayOf = createDualImpl(
```
iwoplaza, line 40: I love comptime 💕

packages/typegpu/src/data/vector.ts
```diff
@@ -332,5 +332,9 @@ function makeVecSchema<TValue, S extends number | boolean>(
       [$repr]: undefined as TValue,
     });
 
+  // it's a workaround for circular dependencies caused by us using schemas in the shader generator
+  // biome-ignore lint/suspicious/noExplicitAny: explained above
+  (VecImpl.prototype as any).schema = schema;
```
reczkok, line 338: Maybe a TODO here to try and resolve this without workarounds would be a good idea so it does not get lost

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1927,3 +1927,12 @@ export function isNaturallyEphemeral(schema: unknown): boolean {
...
 }
+
+export function WORKAROUND_getSchema<T extends AnyVecInstance | AnyMatInstance>(
+  vec: T,
+): VecData | MatData {
+  // it's a workaround for circular dependencies caused by us using schemas in the shader generator
+  // these schema properties are assigned on the prototype of vector and matrix instances
+  // biome-ignore lint/suspicious/noExplicitAny: explained above
+  return (vec as any).schema;
+}
```
reczkok, line 1939: Same here

> Maybe a TODO here to try and resolve this without workarounds would be a good idea so it does not get lost

packages/typegpu/src/std/atomic.ts
reczkok, line 1: We should unify how we handle functions that have no `normalImpl`. In this file we have two different approaches (here shown on the same example for a better side to side):
```ts
export const workgroupBarrier = dualImpl({
  // ...
  normalImpl: () =>
    console.warn('workgroupBarrier is a no-op outside of CODEGEN mode.'),
  noComptime: true,
});
```
```ts
export const workgroupBarrier = dualImpl<() => void>({
  // ...
  normalImpl: 'workgroupBarrier is a no-op outside of CODEGEN mode.',
  noComptime: true, // actually not needed in this case
});
```
In the `dualImpl` the condition we have to satisfy to run the function directly is:
```ts
if (
  !options.noComptime &&
  converted.every((s) => isKnownAtComptime(s)) &&
  typeof options.normalImpl === 'function'
) 
```
I think both of the approaches have their own drawbacks but we should decide on one of them

reczkok (approved):

packages/typegpu/tests/array.test.ts
```diff
@@ -254,7 +254,7 @@ describe('array', () => {
       [Error: Resolution of the following tree failed:
       - <root>
       - fn:foo
-      - fn:arrayOf: Cannot create array schema with count unknown at compile-time: 'count']
+      - fn:arrayOf: Called comptime function with runtime-known values: 'count']
```
cieplypolar, line 257: While I was reading the new implementation of the array, I was worried about loosing informative errors and unfortunately it did happen. I don't see an easy fix for that, also don't want to go back to `createDualImpl`...

packages/typegpu/src/std/atomic.ts
```diff
@@ -1,258 +1,133 @@
...
-export const textureBarrier = createDualImpl(
-  // CPU implementation
-  () => console.warn('textureBarrier is a no-op outside of CODEGEN mode.'),
-  // CODEGEN implementation
-  () => snip('textureBarrier()', Void, /* origin */ 'runtime'),
-  'textureBarrier',
-);
+export const textureBarrier = dualImpl({
+  name: 'textureBarrier',
+  normalImpl: () => 'textureBarrier is a no-op outside of CODEGEN mode.',
```
cieplypolar:
```suggestion
  normalImpl: 'textureBarrier is a no-op outside of CODEGEN mode.',
```

packages/typegpu/src/std/atomic.ts
```diff
@@ -1,258 +1,133 @@
...
-  (a) => {
-    if (isWgslData(a.dataType) && a.dataType.type === 'atomic') {
-      return snip(
-        stitch`atomicLoad(&${a})`,
-        a.dataType.inner,
-        /* origin */ 'runtime',
-      );
+export const atomicLoad = dualImpl<<T extends AnyAtomic>(a: T) => number>({
+  name: 'atomicLoad',
+  normalImpl: 'Atomic operations are not supported outside of CODEGEN mode.',
```
cieplypolar: 'Atomic operations are not supported outside of CODEGEN mode.' 
This string could be stored be in a const variable to avoid repetition.

packages/typegpu/src/std/derivative.ts
```diff
@@ -1,133 +1,93 @@
...
-export const fwidthFine = createDualImpl(
-  cpuFwidthFine,
-  (value) =>
-    snip(stitch`fwidthFine(${value})`, value.dataType, /* origin */ 'runtime'),
-  'fwidthFine',
-);
+export const dpdx = dualImpl<
+  & ((value: number) => number)
+  & (<T extends AnyFloat32VecInstance>(value: T) => T)
+>({
```
cieplypolar: Since we got rid of overloads and we have types now, we could create a type for that
```ts
type DpdxSignature = 
  & ((value: number) => number)
  & (<T extends AnyFloat32VecInstance>(value: T) => T);
```

packages/typegpu/src/std/derivative.ts
```diff
@@ -1,133 +1,93 @@
...
+  name: 'dpdx',
+  normalImpl: 'Derivative builtins are not allowed on the CPU',
```
cieplypolar: 'Derivative builtins are not allowed on the CPU'
This string could be stored in a const variable.

packages/typegpu/src/std/matrix.ts
```diff
@@ -35,91 +36,80 @@ const gpuRotationZ4 = rotationZ4[$internal].gpuImpl;
...
-      matrix.dataType,
-      /* origin */ 'runtime',
-    ),
-  'rotateX4',
-);
+export const rotateX4 = dualImpl({
+  name: 'rotateX4',
+  normalImpl: (matrix: m4x4f, angle: number) =>
+    cpuMul(cpuRotationX4(angle), matrix),
+  signature: (_matrix, angle) => ({
```
cieplypolar: Am I correct, that _matrix and angle are snippets and that's why we don't do anything with them?

packages/typegpu/tests/array.test.ts
iwoplaza: I know, it's definitely a regression. I'll try to hack on a solution.

iwoplaza: On the other hand... it's kind of readable and explanatory.

PR #1970, opened 2025-12-11
impr: Improve the clouds example

reczkok:
This PR aims to address performance issues in the clouds example while improving readability of the code. It also improves the visual result (subjective).

https://github.com/user-attachments/assets/d12c6982-706f-404e-b9e1-816ecc6a8f1e


lursz (approved):
Great work!

PR #1985, opened 2025-12-16
impr: Ignore console.logs in vertex shaders

aleksanderkatan:
Before:
image

After:
image

Changes:
- remove `isEntry` from all functions shells (it was unused),
- add `entryPoint` to make entry functions differentiable,
- add type guards for entry functions (now they are no longer used in this pr, but since the code is already here, let's keep them),
- add an internal slot that holds the info about current stage,
- based on the slot, warn and fallback to a comment when a console.log is used on vertex stage (an overload is created if the function is used on both stages).

### DIFF
--- a/apps/typegpu-docs/src/examples/tests/log-test/index.ts
+++ b/apps/typegpu-docs/src/examples/tests/log-test/index.ts
@@ -35,13 +35,7 @@ const mainFragment = tgpu['~unstable'].fragmentFn({
   return d.vec4f(0.769, 0.392, 1.0, 1);
 });
 
-const context = canvas.getContext('webgpu') as GPUCanvasContext;
-
-context.configure({
-  device: root.device,
-  format: presentationFormat,
-  alphaMode: 'premultiplied',
-});
+const context = root.configureContext({ canvas, alphaMode: 'premultiplied' });
 
 // #region Example controls and cleanup
 
@@ -213,52 +207,9 @@ export const controls = {
   },
   'Render pipeline': {
     onButtonClick: () => {
-      const pipeline = root['~unstable']
-        .withVertex(mainVertex)
-        .withFragment(mainFragment, { format: presentationFormat })
-        .createPipeline();
-
-      pipeline
-        .withColorAttachment({
-          view: context.getCurrentTexture().createView(),
-          clearValue: [0, 0, 0, 0],
-          loadOp: 'clear',
-          storeOp: 'store',
-        })
-        .draw(3);
-    },
-  },
-  'Render pipeline helper': {
-    onButtonClick: () => {
-      const helper = tgpu.fn([])(() => {
-        'use gpu';
-        console.log('A log made from a helper used during both stages.');
-      });
-
-      const mainVertex = tgpu['~unstable'].vertexFn({
-        in: { vertexIndex: d.builtin.vertexIndex },
-        out: { pos: d.builtin.position },
-      })((input) => {
-        console.log('A log made from vertex stage.');
-        helper();
-
-        const positions = [
-          d.vec2f(0, 0.5),
-          d.vec2f(-0.5, -0.5),
-          d.vec2f(0.5, -0.5),
-        ];
-
-        return { pos: d.vec4f(positions[input.vertexIndex], 0, 1) };
-      });
-
-      const mainFragment = tgpu['~unstable'].fragmentFn({
-        in: { pos: d.builtin.position },
-        out: d.vec4f,
-      })(() => {
-        console.log('A log made from fragment stage.');
-        helper();
-
-        return d.vec4f(0.769, 0.392, 1.0, 1);
+      const context = root.configureContext({
+        canvas,
+        alphaMode: 'premultiplied',
       });
 
       const pipeline = root['~unstable']
--- a/apps/typegpu-docs/src/components/ExampleLayout.tsx
+++ b/apps/typegpu-docs/src/components/ExampleLayout.tsx
@@ -1,4 +1,4 @@
-import { useAtom, useSetAtom } from 'jotai';
+import { useSetAtom } from 'jotai';
 import type { ReactNode } from 'react';
 import { useId, useRef } from 'react';
 import CrossSvg from '../assets/cross.svg';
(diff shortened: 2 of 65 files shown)


packages/typegpu/src/core/function/tgpuComputeFn.ts
```diff
@@ -135,9 +120,13 @@ export function computeFn<
       stripTemplate(arg, ...values),
     );
 
-  return Object.assign(Object.assign(call, shell), {
-    does: call,
-  }) as TgpuComputeFnShell<ComputeIn>;
+  return Object.assign(call, shell);
```
cieplypolar, line 124: Is there a reason, why we don't have to cast this to `TgpuComputeFnShell<T>` like in fragment and vertex functions?

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -732,11 +747,24 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
+      // if we're resolving an entrypoint function, we want to update this._currentStage
+      const stage = getItemStage(item);
+      const resolutionAction = () => {
+        if (stage) {
+          this._currentStage = stage;
+        }
+        const result = this._getOrInstantiate(item);
+        if (stage) {
+          this._currentStage = undefined;
+        }
```
cieplypolar:
```suggestion
        this._currentStage = undefined;
```
We can skip this if

packages/typegpu/tests/tgsl/consoleLog.test.ts
cieplypolar:
```suggestion
      namespace: namespace(),
```

cieplypolar:
```suggestion
  it('Parses console.log in a stray function to a comment and warns', () => {
```

packages/typegpu/tests/tgsl/consoleLog.test.ts
```diff
@@ -37,6 +37,74 @@ describe('wgslGenerator with console.log', () => {
     expect(consoleWarnSpy).toHaveBeenCalledTimes(1);
   });
 
+  it('Throws appropriate error when in a vertex shader', ({ root }) => {
```
cieplypolar:
```suggestion
  it('Throws appropriate error when in a vertex shader', () => {
```

packages/typegpu/tests/tgsl/consoleLog.test.ts
```diff
@@ -37,6 +37,74 @@ describe('wgslGenerator with console.log', () => {
...
+        myLog(5);
+        return { pos: d.vec4f() };
+      },
+    );
+    const fs = tgpu['~unstable'].fragmentFn({ out: d.vec4f })(() => {
+      return d.vec4f();
+    });
+
+    const pipeline = root['~unstable']
+      .withVertex(vs, {})
```
cieplypolar:
```suggestion
      .withVertex(vs)
```

packages/typegpu/tests/tgsl/consoleLog.test.ts
cieplypolar:
```suggestion
      .withVertex(vs)
```

packages/typegpu/src/core/function/tgpuComputeFn.ts
aleksanderkatan: ~~We don't have to. In other cases this is necessary.~~
Damn my reading comprehension sucks today.
I don't know, I didn't investigate it.

packages/typegpu/src/resolutionCtx.ts
aleksanderkatan: ~~In the case of `console.log`, we probably could. However it might matter e.g. in `tgpu.resolve([fragmentFn, otherFn])`, if we ever use `ctx._currentStage` for anything else.~~ <- that's why we cannot remove the if entirely

We cannot skip the check, because this will reset the `_currentStage` after any function finishes the resolution. The following test fails after applying the suggested change.
```ts
  it('Does not reset the state after visiting one function', () => {
    const fn1 = (n: number) => {
      'use gpu';
    };

    const fn2 = (n: number) => {
      'use gpu';
      console.log(n);
    };

    const vs = tgpu['~unstable'].vertexFn({ out: { pos: d.builtin.position } })(
      () => {
        fn1(1);
        fn2(2);
        return { pos: d.vec4f() };
      },
    );
    expect(() => tgpu.resolve([vs])).toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - vertexFn:vs
      - fn*:fn2(i32)
      - fn:consoleLog: 'console.log' is not supported during vertex shader stage.]
    `);
  });
```

packages/typegpu/src/core/function/tgpuComputeFn.ts
aleksanderkatan: From what I see, it is an issue regarding `createVertexFn` not being generic, as well as the `OmitBuiltins` not being consistent. I think it's the simplest solution to just cast.

packages/typegpu/src/resolutionCtx.ts
cieplypolar: you are correct, completely forgot that `stage` is shared

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -334,6 +338,12 @@ export class ResolutionCtxImpl implements ResolutionCtx {
   readonly #namespace: NamespaceInternal;
   readonly #shaderGenerator: ShaderGenerator;
 
+  /**
+   * Holds info about the currently resolved shader stage (if there is any).
+   * Note that if a function is used both in vertex and fragment stage,
+   * then it will only go through the process during the vertex stage.
+   */
+  private _currentStage: ShaderStage;
```
iwoplaza: This will only allow shader code to ask which stage they're being resolved in **first**. If we instead create an internal slot that provides the value of the stage currently being resolved, then branches that actually use this value will be able to have branching logic based on which stage they're in.

I can definitely see someone wanting to log a value inside of a utility function that happens to execute in a compute shader and a fragment shader, and not being able to do so because of this error. If we instead emit a no-op and warn the user that a console log has been skipped because it's in a vertex shader, then it would still allow both shader stages to compile, and the other console log to run.

aleksanderkatan: It is so clean and it just works, great idea

apps/typegpu-docs/src/examples/tests/log-test/index.ts
```diff
@@ -215,16 +215,56 @@ export const controls = {
   },
   'Render pipeline': {
     onButtonClick: () => {
-      const context = canvas.getContext('webgpu') as GPUCanvasContext;
+      const pipeline = root['~unstable']
+        .withVertex(mainVertex, {})
```
cieplypolar:
```suggestion
        .withVertex(mainVertex)
```

PR #1988, opened 2025-12-17
feat: Inform user about type mismatches

aleksanderkatan:
I did not take into account the t3 shorthands, since I don't know how to do it without losing the ability to declare `uvec2` etc.

The example should probably be made into a unit test.

### DIFF
--- a/apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
+++ b/apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
@@ -333,7 +333,7 @@ async function render() {
     await verletSim.update(renderer);
   }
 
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
```diff
@@ -333,7 +333,7 @@ async function render() {
     await verletSim.update(renderer);
   }
 
-  await renderer.renderAsync(scene, camera);
+  await renderer.render(scene, camera);
```
iwoplaza: This `await` seems to be redundant
```suggestion
  renderer.render(scene, camera);
```

PR #1995, opened 2025-12-22
docs: Common code between examples

aleksanderkatan:
Changes:
- add `common` folder to examples,
- import all `common` files when opening in StackBlitz,
- list used `common` files in our examples,
- rewrite `setup-orbit-camera.ts` to be common (only where it was already used).

### DIFF
--- a/apps/typegpu-docs/src/components/ExampleView.tsx
+++ b/apps/typegpu-docs/src/components/ExampleView.tsx
@@ -310,7 +310,7 @@ function filterRelevantTsFiles(
   ];
 
   for (const common of commonFiles) {
-    for (const src of tsFiles) {
+    for (const src of srcFiles) {
       if (src.content.includes(`common/${common.path}`)) {
         tsFiles.push(common);
         break;
--- a/apps/typegpu-docs/src/examples/exampleContent.ts
+++ b/apps/typegpu-docs/src/examples/exampleContent.ts
@@ -148,16 +148,11 @@ export const common = R.pipe(
     eager: true,
     import: 'default',
   }) as Record<string, string>,
-  R.mapValues((content: string, key: string): ExampleCommonFile => {
-    const pathRelToCommon = pathe.relative('./', key);
-    const filePath = pathRelToCommon.split('/')[1];
-
-    return {
-      common: true,
-      path: filePath,
-      content,
-    };
-  }),
+  R.mapValues((content: string, key: string): ExampleCommonFile => ({
+    common: true,
+    path: pathe.basename(key),
+    content,
+  })),
   R.values(),
 );
 


apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -274,3 +280,27 @@ function useResizableCanvas(exampleHtmlRef: RefObject<HTMLDivElement | null>) {
...
+    ...srcFiles,
+  ];
+
+  for (const common of commonFiles) {
+    for (const src of tsFiles) {
+      if (src.content.includes(`common/${common.path}`)) {
+        tsFiles.push(common);
+        break;
+      }
+    }
```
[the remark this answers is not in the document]
aleksanderkatan: It should have been srcFiles instead of tsFiles

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -274,3 +280,27 @@ function useResizableCanvas(exampleHtmlRef: RefObject<HTMLDivElement | null>) {
...
+  srcFiles: ExampleSrcFile[],
+  commonFiles: ExampleCommonFile[],
+) {
+  const tsFiles: (ExampleSrcFile | ExampleCommonFile)[] = [
+    ...srcFiles,
+  ];
+
+  for (const common of commonFiles) {
+    for (const src of tsFiles) {
+      if (src.content.includes(`common/${common.path}`)) {
```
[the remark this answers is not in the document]
aleksanderkatan: w/e

apps/typegpu-docs/src/components/ExampleView.tsx
```diff
@@ -290,3 +296,27 @@ function useResizableCanvas(exampleHtmlRef: RefObject<HTMLDivElement | null>) {
...
+function filterRelevantTsFiles(
+  srcFiles: ExampleSrcFile[],
+  commonFiles: ExampleCommonFile[],
+) {
+  const tsFiles: (ExampleSrcFile | ExampleCommonFile)[] = [
+    ...srcFiles,
+  ];
+
+  for (const common of commonFiles) {
+    for (const src of srcFiles) {
```
iwoplaza, line 313: Would this change allow common file to reference each other, and show up in the tabs?
```suggestion
    for (const src of [...srcFiles, ...commonFiles]) {
```

aleksanderkatan: Not really. Or, I guess, it would, but with a side effect. If common1.ts references common2.ts, then common2.ts would be included in every example, regardless of whether the example uses common1.ts or not. We would need to do a BFS or something.

iwoplaza: I see. Could you create an issue for this so we can do it in another PR?

aleksanderkatan: #2038

PR #2000, opened 2025-12-22
fix: Giving local declarations unique names if they clash with global declarations, and vice-versa + 'strict' the default naming scheme in pipelines

iwoplaza:


packages/typegpu/src/nameRegistry.ts
```diff
@@ -229,45 +235,78 @@ export function isValidIdentifier(ident: string): boolean {
...
+  constructor() {
+    this.#usedNames = new Set<string>(bannedTokens);
+    this.#usedFunctionScopeNamesStack = [];
+  }
+
+  get usedFunctionScopeNames(): Set<string> | undefined {
+    return this
+      .#usedFunctionScopeNamesStack[
+        this.#usedFunctionScopeNamesStack.length - 1
+      ];
```
aleksanderkatan, line 252: nit
```suggestion
    return this.#usedFunctionScopeNamesStack.at(-1);
```

packages/typegpu/src/nameRegistry.ts
```diff
@@ -229,45 +235,78 @@ export function isValidIdentifier(ident: string): boolean {
...
+  }
+
+  makeUnique(primer: string | undefined, global: boolean): string {
+    const sanitizedPrimer = sanitizePrimer(primer);
+    const name = this.getUniqueVariant(sanitizedPrimer);
+
+    if (global) {
+      this.#usedNames.add(name);
+    } else {
+      this.usedFunctionScopeNames?.add(name);
```
cieplypolar, line 262: I don't like the fact that the function scope stack can be empty. I’d prefer to add an invariant (it is never empty) or a comment explaining why this state is possible

packages/typegpu/src/nameRegistry.ts
aleksanderkatan, line 1: We should probably add all of std to the keywords list

image

packages/typegpu/src/nameRegistry.ts
```diff
@@ -229,45 +235,78 @@ export function isValidIdentifier(ident: string): boolean {
...
+    }
+
+    return name;
+  }
 
   makeValid(primer: string): string {
-    if (isValidIdentifier(primer)) {
+    if (isValidIdentifier(primer) && !this.#usedNames.has(primer)) {
+      this.usedFunctionScopeNames?.add(primer);
```
cieplypolar, line 270: same as above

cieplypolar (commented):
Looks good 🎄

Issue #2001, opened 2025-12-23
impr: Clean up nameRegistry.ts


aleksanderkatan:
- [ ] remove random name schema,
- [ ] merge `makeUnique` and `makeValid` into one function

reczkok:
Done/Oboslete

PR #2002, opened 2025-12-23
fix: Collisions with builtins

aleksanderkatan:
Changes:
- move struct prop validation from resolution to struct creation,
- add builtins to `makeValid` checks (not used for prop naming).

I also created an issue for cleaning up the `nameRegistry.ts` (#2001).


cieplypolar (approved):
Looks good 🎄

PR #2006, opened 2026-01-07
fix/feat: Fix type in `textureSampleCompare` and add `textureSampleCompareLevel`

reczkok:


packages/typegpu/src/std/texture.ts
```diff
@@ -567,6 +568,70 @@ export const textureSampleCompare = dualImpl({
...
+): number;
+function textureSampleCompareLevelCpu(
+  _texture: WgslTexture,
+  _sampler: comparisonSampler,
+  _coords: v2f | v3f,
+  _depthRefOrArrayIndex: number,
+  _depthRefOrOffset?: number | v2i,
+  _maybeOffset?: v2i,
+): number {
+  throw new Error(
```
iwoplaza:
```suggestion
  throw new MissingCpuImplError(
```

PR #2010, opened 2026-01-07
feat: More capable accessors

iwoplaza:
Accessors could always provide:
- Comptime-known values
- `tgpu.fn` functions that return a value matching the schema
- Buffer usages and buffer shorthands that hold the schema

Now, accessors can provide:
- tgpu.const references
- tgpu.*Var references
- Shell-less functions
- Deep references to nested structures (with regular callbacks)
- comptime functions that are ran when requesting the value

There's also a new resource: `tgpu.mutableAccessor`. It's currently functionally the same as `tgpu.accessor` but with a more restricted API, and with future PRs, values accessed from regular accessors will be immutable.

Things to do in subsequent PRs:
- https://github.com/software-mansion/TypeGPU/issues/2015
- https://github.com/software-mansion/TypeGPU/issues/2013


packages/typegpu/src/core/root/init.ts
```diff
@@ -203,17 +205,12 @@ class WithBindingImpl implements WithBinding {
   ) {}
 
   with<T extends AnyWgslData>(
```
cieplypolar: Why not `AnyData`? I've seen changes from `AnyWgslData` to `AnyData` before.
```suggestion
  with<T extends AnyData>(
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -92,7 +95,11 @@ interface TgpuFnBase<ImplSchema extends AnyFn> extends TgpuNamable {
   with<T>(slot: TgpuSlot<T>, value: Eventual<T>): TgpuFn<ImplSchema>;
   with<T extends AnyData>(
     accessor: TgpuAccessor<T>,
-    value: TgpuFn<() => T> | TgpuBufferUsage<T> | Infer<T>,
+    value: AccessorIn<NoInfer<T>>,
```
reczkok, line 98: Is there any particular reason we use `NoInfer` here and not in the other `.with` signatures (on root for example)?

reczkok: Nevermind, they are there in the public APIs

packages/typegpu/src/core/slot/slotTypes.ts
```diff
@@ -43,26 +46,61 @@ export interface TgpuDerived<T> {
   '~compute'(): T;
 }
 
+export type AccessorIn<TSchema extends AnyData> = TSchema extends
+  WgslTexture | WgslStorageTexture ? (
+    | (() => Infer<TSchema> | TgpuTextureView<TSchema>)
+    | Infer<TSchema>
```
reczkok, line 52: I'm doubting that this is actually the correct signature. I don't think there is a situation where the user is able to pass `Infer<WgslTexture | WgslStorageTexture>` that would produce valid code. User-defined functions cannot return texures (since WGSL requires return types to be [constructible](https://www.w3.org/TR/WGSL/#constructible-types)). I think a TextureView may be the only valid input.

iwoplaza: I agree, `Infer<WgslTexture | WgslStorageTexture>` seems impossible to retrieve at comp-time. I'll remove that 🙌 

> User-defined functions cannot return texures (since WGSL requires return types to be [constructible](https://www.w3.org/TR/WGSL/#constructible-types)).

It's not possible to distinguish 'use gpu' from non-'use gpu' marked functions on the type-level, but this is meant to accept only non-'use gpu' functions that can access a nested resource, for example:

```ts
const fooSchema = d.textureStorage2d('rgba16float', 'write-only');

  foo: { storageTexture: fooSchema },
});

const fooAccess = tgpu.mutableAccessor(fooSchema, () => layout.$.foo);
```

PR #2018, opened 2026-01-08
docs: ducks

lursz:
New example, ported from threejs


lursz:
Comments in the code resemble comments in the og code, so it's easier to understand where something came from.

apps/typegpu-docs/src/examples/threejs/ducklings/gpuHelpers.ts
```diff
@@ -0,0 +1,75 @@
...
+    const bottomY = std.max(0, std.sub(y, 1));
+    const topY = std.min(std.add(y, 1), std.sub(d.i32(width), 1));
+
+    const westIndex = d.u32(std.add(std.mul(y, d.i32(width)), leftX));
+    const eastIndex = d.u32(std.add(std.mul(y, d.i32(width)), rightX));
+    const southIndex = d.u32(std.add(std.mul(bottomY, d.i32(width)), x));
+    const northIndex = d.u32(std.add(std.mul(topY, d.i32(width)), x));
+
+    return NeighborIndices({ northIndex, southIndex, eastIndex, westIndex });
+  };
```
reczkok, line 29: 🤔
```suggestion
  const getNeighborIndices = (index: number) => {
    'use gpu';
    const width = d.u32(WIDTH);
    const x = index % WIDTH;
    const y = d.u32(index / WIDTH);

    const leftX = std.max(0, x - 1);
    const rightX = std.min(x + 1, width - 1);
    const bottomY = std.max(0, y - 1);
    const topY = std.min(y + 1, width - 1);

    const westIndex = y * width + leftX;
    const eastIndex = y * width + rightX;
    const southIndex = bottomY * width + x;
    const northIndex = topY * width + x;

    return NeighborIndices({ northIndex, southIndex, eastIndex, westIndex });
  };
```

reczkok (changes requested):
I have a couple of problems with this PR:
- The TGSL functions wouldn’t want to make me use TypeGPU - please move to normal operators and go through the logic to eliminate useless casts, etc.
- Take advantage of TypeGPU mechanics (like slots) to make the logic nicer. computeHeightBtoA and computeHeightAtoB are basically the same, so you could halve the code using slots (I’m pretty sure).
- Separating gpuHelpers into a separate file makes no sense to me if you need to create them using a function. I would either figure out a way to make the helpers top‑level (which I’m pretty sure is possible using accessors/slots) or not do it at all.

PR #2023, opened 2026-01-08
chore: `pnpm run spec-report` - querying the current spec for missing std functions

iwoplaza:


iwoplaza:
Not really the approach we want to go with down the line

PR #2029, opened 2026-01-09
feat: tgpu.fn(callback) for providing slots and accessors

lursz:
A new overload for `tgpu.fn` that accepts a shell-less callback and returns a callable that has a `.with()` method. This allows shell-less functions to also be provided with slot and accessor overrides.

## Example
```ts
const countAccess = tgpu.accessor(d.f32);

const getDouble = () => {
  'use gpu';
  return countAccess.$ * 2;
};
// ^? () => number

const getDouble4 = tgpu.fn(getDouble).with(countAccess, 4);
//    ^? TgpuGenericFn<() => number>

const main = () => {
  'use gpu';
  return getDouble4(); // should return 8
};
```

## Requirements
- If a shell-less function `foo` is used in code, and `tgpu.fn(foo)` is also used in code, it should only generate one definition.

### DIFF
--- a/packages/typegpu/src/core/function/tgpuFn.ts
+++ b/packages/typegpu/src/core/function/tgpuFn.ts
@@ -181,7 +181,7 @@ function createFn<ImplSchema extends AnyFn>(
     Parameters<ImplSchema>,
     Extract<ReturnType<ImplSchema>, BaseData>
   >,
-  implementation: Implementation<ImplSchema>,
+  _implementation: Implementation<ImplSchema>,
 ): TgpuFn<ImplSchema> {
   type This = TgpuFnBase<ImplSchema> & SelfResolvable & {
     [$getNameForward]: FnCore;
@@ -189,11 +189,12 @@ function createFn<ImplSchema extends AnyFn>(
 
   let pairs: SlotValuePair[] = [];
   // Unwrapping generic functions
-  if (isGenericFn(implementation)) {
-    pairs = implementation[$providing]?.pairs ?? [];
-    implementation = implementation[$internal].inner as Implementation<
-      ImplSchema
-    >;
+  let implementation: Implementation<ImplSchema>;
+  if (isGenericFn(_implementation)) {
+    pairs = _implementation[$providing]?.pairs ?? [];
+    implementation = _implementation[$internal].inner as typeof implementation;
+  } else {
+    implementation = _implementation;
   }
 
   const core = createFnCore(implementation as Implementation, '');
@@ -350,8 +351,6 @@ function createGenericFn<T extends AnyFn>(
   inner: T,
   pairs: SlotValuePair[],
 ): TgpuGenericFn<T> {
-  type This = TgpuGenericFn<T>;
-
   const fnBase = {
     [$internal]: { inner },
     resourceType: 'generic-function' as const,
@@ -379,7 +378,7 @@ function createGenericFn<T extends AnyFn>(
     return inner(...args) as ReturnType<T>;
   };
 
-  const genericFn = Object.assign(call, fnBase) as unknown as This;
+  const genericFn = Object.assign(call, fnBase) as unknown as TgpuGenericFn<T>;
 
   // Inheriting name from `inner`, if it exists
   if (getName(inner)) {


packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -108,6 +109,28 @@ export type TgpuFn<ImplSchema extends AnyFn = (...args: any[]) => any> =
       };
   };
 
+/**
+ * A function wrapper that allows providing slot and accessor overrides for shellless functions
+ */
+export interface TgpuGenericFn<Callback extends AnyFn> {
```
```suggestion
export interface TgpuGenericFn<T extends AnyFn> {
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -108,6 +109,28 @@ export type TgpuFn<ImplSchema extends AnyFn = (...args: any[]) => any> =
...
+  readonly [$internal]: true;
+  readonly [$providing]?: Providing | undefined;
+  readonly resourceType: 'generic-function';
+  readonly callback: Callback;
```
iwoplaza: Let's not make `callback` part of the public API
```suggestion
  readonly [$internal]: {
    inner: T;
  };
  readonly [$providing]?: Providing | undefined;
  readonly resourceType: 'generic-function';
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -117,10 +140,22 @@ export function fn<
   Return extends AnyData,
 >(argTypes: Args, returnType: Return): TgpuFnShell<Args, Return>;
 
+export function fn<Callback extends AnyFn>(
+  callback: Callback,
+): TgpuGenericFn<Callback>;
```
```suggestion
export function fn<T extends AnyFn>(inner: T): TgpuGenericFn<T>;
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -328,3 +370,50 @@ function createBoundFunction<ImplSchema extends AnyFn>(
...
 }
+
+function createGenericFn<Callback extends AnyFn>(
+  callback: Callback,
+  pairs: SlotValuePair[],
+): TgpuGenericFn<Callback> {
+  type This = TgpuGenericFn<Callback>;
+
+  const fnBase = {
+    [$internal]: true as const,
```
```suggestion
    [$internal]: { inner },
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -328,3 +370,50 @@ function createBoundFunction<ImplSchema extends AnyFn>(
 
   return fn;
 }
+
+function createGenericFn<Callback extends AnyFn>(
+  callback: Callback,
+  pairs: SlotValuePair[],
+): TgpuGenericFn<Callback> {
```
```suggestion
function createGenericFn<T extends AnyFn>(
  inner: T,
  pairs: SlotValuePair[],
): TgpuGenericFn<T> {
```

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -328,3 +370,50 @@ function createBoundFunction<ImplSchema extends AnyFn>(
...
+  const fnBase = {
+    [$internal]: true as const,
+    resourceType: 'generic-function' as const,
+    callback,
+    [$providing]: pairs.length > 0
+      ? {
+        inner: callback,
+        pairs,
+      }
+      : undefined,
```
```suggestion
    [$providing]: pairs.length > 0
      ? { inner, pairs }
      : undefined,
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -549,26 +551,44 @@ ${this.ctx.pre}}`;
...
-        const args = argNodes.map((arg) => this.expression(arg));
-        const shellless = this.ctx.shelllessRepo.get(
-          callee.value as (...args: never[]) => unknown,
-          args,
+      if (!isMarkedInternal(callee.value) || isGenericFn(callee.value)) {
+        const slotPairs: SlotValuePair[] = isGenericFn(callee.value)
+          ? (callee.value[$providing]?.pairs ?? [])
+          : [];
+        const callback = isGenericFn(callee.value)
+          ? callee.value.callback
```
iwoplaza:
```suggestion
          ? callee.value[$internal].inner
```

packages/typegpu/tests/tgpuGenericFn.test.ts
iwoplaza, line 1: Could you add more tests to see if the `.with` API works on these generic functions?

aleksanderkatan, line 1: I think this api could use a paragraph in either functions or slots docs

aleksanderkatan (approved):
Nice! 🐚

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -111,6 +113,29 @@ export type TgpuFn<ImplSchema extends AnyFn = (...args: any[]) => any> =
...
+ */
+export interface TgpuGenericFn<T extends AnyFn> {
+  readonly [$internal]: {
+    inner: T;
+  };
+  readonly [$providing]?: Providing | undefined;
+  readonly resourceType: 'generic-function';
+
+  with<S>(slot: TgpuSlot<S>, value: Eventual<S>): TgpuGenericFn<T>;
+  with<S extends AnyData>(
```
cieplypolar: You should provide `with` overload for `accessor` and `mutableAccessor`. Also we have dedicated types for accessors input, I believe after merging main it would be `TgpuAccessor.In<T>` and `TgpuMutableAccessor.In<T>`

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -120,10 +145,20 @@ export function fn<
   Return extends AnyData,
 >(argTypes: Args, returnType: Return): TgpuFnShell<Args, Return>;
 
+export function fn<T extends AnyFn>(inner: T): TgpuGenericFn<T>;
```
cieplypolar, line 125:
```suggestion
export function fn<T extends AnyFn>(callback: T): TgpuGenericFn<T>;
```
For consistency

packages/typegpu/tests/tgpuGenericFn.test.ts
```diff
@@ -0,0 +1,199 @@
...
+        return 2f;
+      }
+
+      fn main() -> f32 {
+        return getValue();
+      }"
+    `);
+  });
+
+  it('generates one fucntion even when .with is called multiple times with the same arguments', () => {
```
cieplypolar:
```suggestion
  it('generates one function even when .with is called multiple times with the same arguments', () => {
```

packages/typegpu/tests/tgpuGenericFn.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    };
+
+    const getDouble4 = tgpu.fn(getDouble);
+
+    const main = () => {
+      'use gpu';
+      return getDouble4();
+    };
+
+    const wgsl = tgpu.resolve([main]);
```
cieplypolar:
```suggestion
```

packages/typegpu/tests/tgpuGenericFn.test.ts
```diff
@@ -0,0 +1,199 @@
...
+        return 0f;
+      }
+
+      fn main() -> f32 {
+        return getDouble();
+      }"
+    `);
+  });
+
+  it('does not duplicate the same function', () => {
```
cieplypolar, line 74: I think, we've seen that test before

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -315,3 +357,44 @@ function createBoundFunction<ImplSchema extends AnyFn>(
...
+): TgpuGenericFn<T> {
+  type This = TgpuGenericFn<T>;
+
+  const fnBase = {
+    [$internal]: { inner },
+    resourceType: 'generic-function' as const,
+    [$providing]: pairs.length > 0 ? { inner, pairs } : undefined,
+
+    with(
+      slot: TgpuSlot<unknown> | TgpuAccessor,
```
cieplypolar: Mutable accessor is missing. Also, is there a particular reason, why there isn't `comptime` wrapper like in `createBoundFunction`?

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -315,3 +357,44 @@ function createBoundFunction<ImplSchema extends AnyFn>(
...
 }
+
+function createGenericFn<T extends AnyFn>(
+  inner: T,
+  pairs: SlotValuePair[],
+): TgpuGenericFn<T> {
+  type This = TgpuGenericFn<T>;
+
+  const fnBase = {
+    [$internal]: { inner },
```
cieplypolar, line 355: Can we add `$name` property?

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -315,3 +357,44 @@ function createBoundFunction<ImplSchema extends AnyFn>(
...
+      return createGenericFn(inner, [
+        ...pairs,
+        [isAccessor(slot) ? slot.slot : slot, value],
+      ]);
+    },
+  };
+
+  const call = ((...args: Parameters<T>): ReturnType<T> => {
+    return inner(...args) as ReturnType<T>;
+  }) as T;
```
cieplypolar:
```suggestion
  });
```
Do we need this cast?

packages/typegpu/tests/tgpuGenericFn.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      return valueAccess.$;
+    }).with(valueAccess, 2);
+
+    const main = () => {
+      'use gpu';
+      return getValueGeneric();
+    };
+
+    expect(tgpu.resolve([main])).toMatchInlineSnapshot(`
+      "fn item() -> f32 {
```
cieplypolar: IMO it should get `getValueGeneric` name

packages/typegpu/tests/tgpuGenericFn.test.ts
```diff
@@ -0,0 +1,199 @@
+import { describe, expect } from 'vitest';
+import * as d from '../src/data/index.ts';
+import tgpu from '../src/index.ts';
+import { it } from './utils/extendedIt.ts';
+
+describe('TgpuGenericFn - shellless callback wrapper', () => {
```
cieplypolar, line 6:
```ts
it('can be called in js', () => {
  const getValue = () => {
    'use gpu';
    return 2;
  };

  const getValueGeneric = tgpu.fn(getValue);

  expect(getValueGeneric()).toBe(2);
});
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -591,31 +598,49 @@ ${this.ctx.pre}}`;
...
-        if (shellless) {
-          const converted = args.map((s, idx) => {
-            const argType = shellless.argTypes[idx] as AnyData;
-            return tryConvertSnippet(
-              this.ctx,
-              s,
-              argType,
-              /* verbose */ false,
+      if (!isMarkedInternal(callee.value) || isGenericFn(callee.value)) {
+        const slotPairs: SlotValuePair[] = isGenericFn(callee.value)
```
```suggestion
        const slotPairs = isGenericFn(callee.value)
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -591,31 +598,49 @@ ${this.ctx.pre}}`;
...
-        );
-        if (shellless) {
-          const converted = args.map((s, idx) => {
-            const argType = shellless.argTypes[idx] as AnyData;
-            return tryConvertSnippet(
-              this.ctx,
-              s,
-              argType,
-              /* verbose */ false,
+      if (!isMarkedInternal(callee.value) || isGenericFn(callee.value)) {
```
cieplypolar: Can you save the result of `isGenericFn` to some variable? You are calling it 3 times, which means there will be 3 jumps (unless JS caches the calls)

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -591,31 +598,49 @@ ${this.ctx.pre}}`;
...
+        const slotPairs: SlotValuePair[] = isGenericFn(callee.value)
+          ? (callee.value[$providing]?.pairs ?? [])
+          : [];
+        const callback = isGenericFn(callee.value)
+          ? callee.value[$internal].inner
+          : (callee.value as (...args: never[]) => unknown);
```
```suggestion
          : (callee.value as AnyFn);
```

packages/typegpu/tests/tgpuGenericFn.test.ts
```diff
@@ -0,0 +1,199 @@
...
+      "fn item() -> f32 {
+        return 2f;
+      }
+
+      fn main() -> f32 {
+        return item();
+      }"
+    `);
+  });
+});
```
cieplypolar, line 366: Is this intended behavior?
```ts
it('complex', () => {
  const valueAccess = tgpu['~unstable'].accessor(d.f32);
  const slot = tgpu.slot(5);

  const getValue = () => {
    'use gpu';
    return valueAccess.$ * slot.$;
  };

  const getValueGeneric = tgpu.fn(getValue).with(valueAccess, 2).with(
    valueAccess,
    4,
  ).with(slot, 7).with(slot, 5);
  const getValueGenericAgain = tgpu.fn(getValue).with(valueAccess, 4);
  const getValueShell = tgpu.fn([], d.f32)(getValue).with(
    valueAccess,
    4,
  );
  const getValueShell2 = tgpu.fn([], d.f32)(getValue).with(
    valueAccess,
    4,
  );

  const main = () => {
    'use gpu';
    return getValueShell() + getValueGeneric() + getValueGenericAgain() +
      getValueShell2();
  };

    "fn getValue() -> f32 {
      return 20f;
    }

    fn getValue_1() -> f32 {
      return 20f;
    }

    fn getValue_2() -> f32 {
      return 20f;
    }

    fn main() -> f32 {
      return (((getValue() + getValue_1()) + getValue_1()) + getValue_2());
    }"
  `);
});
```

packages/typegpu/tests/tgpuGenericFn.test.ts
aleksanderkatan: This would match our new approach to naming slotted functions. Could we also include a test with `getName`?

packages/typegpu/src/core/function/tgpuFn.ts
lursz: Yeah I agree, but based on previous PR comments we ought to do it the other way around (https://github.com/software-mansion/TypeGPU/pull/2029#discussion_r2687845114) or not at all

packages/typegpu/tests/tgpuGenericFn.test.ts
lursz: few tests perform testing that overlaps, but there are none that are equivalent

packages/typegpu/src/tgsl/wgslGenerator.ts
lursz: Nice catch

packages/typegpu/src/core/function/tgpuFn.ts
lursz: Agreed, it should be there as it's also going to be used in both contexts

packages/typegpu/tests/tgpuGenericFn.test.ts
aleksanderkatan: I'm not sure what exactly you're trying to test here.
Here's a test for overriding

```ts
  it('complex', () => {
    const valueAccess = tgpu['~unstable'].accessor(d.f32);
    const slot = tgpu.slot<number>();

    const getValue = tgpu.fn(() => {
      'use gpu';
      return valueAccess.$ * slot.$;
    })
      .with(valueAccess, 2)
      .with(valueAccess, 4)
      .with(slot, 7)
      .with(slot, 5);

    const getValueShelled = tgpu.fn([], d.f32)(() => {
      'use gpu';
      return valueAccess.$ * slot.$;
    })
      .with(valueAccess, 2)
      .with(valueAccess, 4)
      .with(slot, 7)
      .with(slot, 5);

    const main = () => {
      'use gpu';
      return getValue() + getValueShelled();
    };

      "fn item() -> f32 {
        return 20f;
      }

      fn getValueShelled() -> f32 {
        return 20f;
      }

      fn main() -> f32 {
        return (item() + getValueShelled());
      }"
    `);
  });
```

packages/typegpu/src/core/function/tgpuFn.ts
lursz: Done

packages/typegpu/tests/tgpuGenericFn.test.ts
lursz: Done

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -302,3 +345,52 @@ function createBoundFunction<ImplSchema extends AnyFn>(
...
+
+function createGenericFn<T extends AnyFn>(
+  inner: T,
+  pairs: SlotValuePair[],
+): TgpuGenericFn<T> {
+  type This = TgpuGenericFn<T>;
+
+  const fnBase = {
+    [$internal]: { inner },
+    [$getNameForward]: inner,
```
aleksanderkatan: This makes it so that inner function is renamed instead. This would be fair, since the inner function cannot be used in WGSL anyway, since it is slotted (if the user wraps a regular shellless in tgpu.fn, then skill issue), however this makes it so every slotted variant is named the same.

image

I'd copy instead the code from bound fn, where we inherit the name from the inner, but we never rename the inner fn.

packages/typegpu/src/shared/meta.ts
```diff
@@ -84,6 +84,7 @@ export function getName(definition: unknown): string | undefined {
 export function setName(definition: object, name: string | undefined): void {
   if (isForwarded(definition)) {
     setName(definition[$getNameForward] as object, name);
+    return;
```
iwoplaza, line 86: Has the same behavior, but we save memory by not storing unnecessary metadata

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -156,6 +187,15 @@ function createFn<ImplSchema extends AnyFn>(
     [$getNameForward]: FnCore;
   };
 
+  let pairs: SlotValuePair[] = [];
+  // Unwrapping generic functions
+  if (isGenericFn(implementation)) {
+    pairs = implementation[$providing]?.pairs ?? [];
+    implementation = implementation[$internal].inner as Implementation<
```
cieplypolar: I would rather create a new variable for the `implementation` than reassign function parameter.

packages/typegpu/src/core/function/tgpuFn.ts
```diff
@@ -302,3 +345,55 @@ function createBoundFunction<ImplSchema extends AnyFn>(
 
   return fn;
 }
+
+function createGenericFn<T extends AnyFn>(
+  inner: T,
+  pairs: SlotValuePair[],
+): TgpuGenericFn<T> {
+  type This = TgpuGenericFn<T>;
```
cieplypolar: We declare this type and use it only once.
`$name` and `with` methods could return it.

PR #2040, opened 2026-01-12
feat: Deprecate .value and .bound

iwoplaza:


packages/typegpu/src/core/resolve/tgpuResolve.ts
```diff
@@ -240,8 +241,16 @@ function resolveFromArray(
...
     [$resolve](ctx): ResolvedSnippet {
-      // biome-ignore lint/suspicious/useIterableCallbackReturn: <we just resolve>
-      items.forEach((item) => ctx.resolve(item));
+      for (const item of items) {
+        // Support for: tgpu.resolve([layout])
+        if (isBindGroupLayout(item)) {
+          for (const binding of Object.values(item[$internal].bound)) {
+            ctx.resolve(binding);
+          }
+        } else {
```
aleksanderkatan, line 250: Could we make layout implement the `$resolve` symbol instead?

iwoplaza: Because I believe it would then pass this check, and not allow `layout.$.foo` access in WGSL templates
image

PR #2042, opened 2026-01-12
chore: Analyse the typegpu entry points

aleksanderkatan:
Changes:
- added `generate-tests.ts` script in the `treeshake-test` directory, that iterates over exports from `typegpu`, `typegpu/data` and `typegpu/std`, and generates an import test for each,
- modified the CI to run the `generate-tests.ts`,
- disabled `webpack` by default, so now only `tsdown` runs on CI (because `tsdown` takes 10 seconds, and `webpack` takes 6 minutes),
- made `treeshake-test` runnable from the Actions menu (generates a summary instead of a comment, and runs `webpack` as well),
- modified the results message to include 3 tables: "interesting" static tests results,  "interesting" dynamic (generated) tests results, and all test results (a result is "interesting" if it undefined on target branch, or differs by at least 0.5% between branches).

The .yaml must be on main branch for the "Run workflow" button to become visible, I tested it on a repo fork.

### DIFF
--- a/apps/typegpu-docs/src/examples/rendering/jelly-switch/index.ts
+++ b/apps/typegpu-docs/src/examples/rendering/jelly-switch/index.ts
@@ -584,6 +584,7 @@ function createBindGroups() {
 
 let bindGroups = createBindGroups();
 
+let animationFrameHandle: number;
 function render(timestamp: number) {
   frameCount++;
   camera.jitter();
@@ -621,7 +622,7 @@ function render(timestamp: number) {
     .with(bindGroups.render[currentFrame])
     .draw(3);
 
-  requestAnimationFrame(render);
+  animationFrameHandle = requestAnimationFrame(render);
 }
 
 function handleResize() {
@@ -643,7 +644,7 @@ const resizeObserver = new ResizeObserver(() => {
 });
 resizeObserver.observe(canvas);
 
-requestAnimationFrame(render);
+animationFrameHandle = requestAnimationFrame(render);
 
 // #region Example controls and cleanup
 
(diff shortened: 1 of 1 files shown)


iwoplaza:
It should be possible for files to be virtual, not committed to the file system.

Also, in case you haven't yet thought about displaying the data concisely, maybe we should default to only showing outstanding differences in values (above a certain threshold)? That way we won't be desensitized to these messages.

apps/treeshake-test/package.json
```diff
@@ -5,7 +5,8 @@
   "description": "Treeshake testing app for TypeGPU",
   "type": "module",
   "scripts": {
-    "test": "node index.ts"
+    "generate-tests": "tsx generateTests.ts",
```
iwoplaza, line 8: What was failing when using plain `node`?

aleksanderkatan: `generateTests.ts` needs workspace typegpu import to know what exports to generate tests for. And in typegpu, we use typescript syntax unsupported by node 

```
SyntaxError [ERR_UNSUPPORTED_TYPESCRIPT_SYNTAX]:   x TypeScript parameter property is not supported in strip-only mode
    ,-[49:1]
 46 |     'Vertex formats are not host-shareable, use concrete types instead';
 47 |   // ---
 48 | 
 49 |   constructor(public readonly type: T) {}
    :                               ^^^^^^^
 50 | }
 51 | 
 52 | export const formatToWGSLType = {
```

iwoplaza: Aah, I see. It's okay then.

PR #2046, opened 2026-01-13
feat: Better snippet origin tracking

iwoplaza:
Changes:
- Not all arguments are automatically considered ephemeral, meaning unnecessary branching is no longer required.
- `blockVariable` is only responsible for generating the variable definition code (easily swappable with a different shader generator)


iwoplaza:
Already done as part of other PRs

PR #2047, opened 2026-01-14
fix: Add manual chunks with Rollup for prod

### LINKED ISSUE #2041: docs: Don't cache example files
Currently, all example files except for `index.ts` are cached. This makes it impossible to create a root in a file other than `index.ts`, because it will be cleaned up on example change, and on another example load it will attempt to create resources from a destroyed root.

Investigate whether this is necessary.

Potential benefits of this change:
- cleaner example code with less slots/wrappers for resource creation,
- easier point of entry for potential external example creators (it has already happened twice that somebody attempted delegating root to another file, and needed to refactor),
- less memory usage when opening multiple examples in a single session.

sharabai:
Fixes #2041 for examples in production, dev server still has those issues.

### DIFF
--- a/apps/typegpu-docs/astro.config.mjs
+++ b/apps/typegpu-docs/astro.config.mjs
@@ -13,6 +13,7 @@ import wasm from 'vite-plugin-wasm';
 import basicSsl from '@vitejs/plugin-basic-ssl';
 import rehypeMathJax from 'rehype-mathjax';
 import remarkMath from 'remark-math';
+import { relative, resolve } from 'pathe';
 
 /**
  * @template T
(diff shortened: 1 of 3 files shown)


iwoplaza:
Since it only fixes things in prod, I would hold off on this change until we migrate to Vite 8 (with rolldown), then try a similar trick again.

The other thing we could do to fix this is write an unplugin that swaps import calls with async calls to a user-provided function, where we could do this cache invalidation ourselves on a deeper level than we do currently.

Lastly, the use case that this change supports is in my opinion not a great way to architect an app/example to begin with. A root is something ephemeral, something that can get lost and needs to be recreated. Splitting the example into separate files that all import a root, or top-level buffers that   import a root, can lead to unclear coupling that is hard to decouple when we need to say instantiate the effect a few times with different parameters, or publish it as a library. There are ways to abstract and split an example that doesn't involve this tight coupling with an ephemeral resource.

iwoplaza:
We're going in the direction of not having side-effects when importing examples entirely.

PR #2048, opened 2026-01-14
docs: Rotating Triangle Tiles example

sharabai:
Triangle tiles that rotate.


apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/bezier.ts
```diff
@@ -0,0 +1,49 @@
+// Generate by AI
```
aleksanderkatan: Could you find some human made implementation and reference the author instead? Or, if it's just some wikipedia formulas, just remove this line

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/buffers.ts
```diff
@@ -0,0 +1,68 @@
...
+  getGridParams,
+  INITIAL_MIDDLE_SQUARE_SCALE,
+  INITIAL_STEP_ROTATION,
+} from './params.ts';
+
+const animationProgressUniform = root.createUniform(d.f32);
+
+const shiftedColorsBuffer = root.createReadonly(d.arrayOf(d.vec4f, 3), [
+  ...colors,
+]);
```
aleksanderkatan:
```suggestion
const shiftedColorsBuffer = root.createReadonly(d.arrayOf(d.vec4f, 3), colors);
```

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/meta.json
```diff
@@ -0,0 +1,6 @@
+{
+  "title": "Rotating Triangle Tiles",
+  "category": "simple",
+  "tags": ["experimental", "fragment shader", "instancing"],
+  "dev": false
```
aleksanderkatan:
```suggestion
  "tags": ["experimental", "fragment shader", "instancing"]
```

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/params.ts
```diff
@@ -0,0 +1,147 @@
...
+const cubicBezierControlPoints: [number, number, number, number] = [
+  0.18,
+  0.7,
+  0.68,
+  1.03,
+];
+
+function getCubicBezierControlPoints() {
+  return cubicBezierControlPoints;
+}
```
aleksanderkatan, line 21: Can this function be inlined?

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/params.ts
```diff
@@ -0,0 +1,147 @@
...
+    scale ?? DEFAULT_ROTATION_TO_MIDDLE_SQUARE_SCALE_ARRAY,
+  );
+}
+
+const GridParams = d.struct({
+  tileDensity: d.f32,
+  userScale: d.f32,
+  trianglesPerRow: d.u32,
+  triangleCount: d.u32,
+});
```
aleksanderkatan: I think this struct is unused

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/buffers.ts
```diff
@@ -0,0 +1,88 @@
...
+const aspectRatioAccess = tgpu['~unstable'].accessor(d.f32);
+
+const instanceInfoLayout = tgpu.bindGroupLayout({
+  instanceInfo: { storage: InstanceInfoArray },
+});
+
+function createBuffers(root: TgpuRoot) {
+  const animationProgressUniform = root.createUniform(d.f32);
+
+  const shiftedColorsBuffer = root.createReadonly(d.arrayOf(d.vec4f, 3), [
```
aleksanderkatan: nit:
typically we name buffers based on whether they were created using 
- `createBuffer`, 
- `createUniform` (equivalent to `.createBuffer(...).$usage('uniform').as('uniform')`), 
- `createReadonly` (`.createBuffer(...).$usage('storage').as('readonly')`), 
- `createMutable` (`.createBuffer(...).$usage('storage').as('mutable')`), 

so this one should be named `shiftedColorsReadonly`. Also, this could be a uniform, as 3 vectors is only 12 bytes (readonlys are best for large arrays).

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/transformations.ts
```diff
@@ -0,0 +1,97 @@
...
+    inputValue: number,
+    inputLowerEndpoint: number,
+    inputUpperEndpoint: number,
+    outputLowerEndpoint: number,
+    outputUpperEndpoint: number,
+  ) => {
+    const inputProgress = inputValue - inputLowerEndpoint;
+    const inputInterval = inputUpperEndpoint - inputLowerEndpoint;
+    const progressPercentage = inputProgress / inputInterval;
+    const outputInterval = std.sub(outputUpperEndpoint, outputLowerEndpoint);
```
aleksanderkatan:
```suggestion
    const outputInterval = outputUpperEndpoint - outputLowerEndpoint;
```
Also, my personal preference is to make functions with many arguments shellless, though its alright if you want to be sure that number types are inferred correctly. The `: number` annotations are redundant in shelled functions.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/transformations.ts
```diff
@@ -0,0 +1,97 @@
...
+
+const interpolateBezier = tgpu.fn(
+  [d.f32, d.f32, d.f32],
+  d.f32,
+)(
+  (
+    inputValue: number,
+    outputLowerEndpoint: number,
+    outputUpperEndpoint: number,
+  ) => {
```
aleksanderkatan:
```suggestion
const interpolateBezier = tgpu.fn([d.f32, d.f32, d.f32], d.f32)(
  (inputValue, outputLowerEndpoint, outputUpperEndpoint) => {
```
nit: this could be formatted like this

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
```diff
@@ -0,0 +1,215 @@
...
+    ).draw(6);
+
+  midgroundPipeline
+    .withColorAttachment({
+      view,
+      loadOp: 'load',
+      storeOp: 'store',
+    })
+    .with(getInstanceInfoBindGroup())
+    .draw(3, getGridParams().triangleCount);
```
aleksanderkatan:
```suggestion
  const shiftedColors = getShiftedColors(timestamp);
  shiftedColorsBuffer.write(shiftedColors);
  animationProgressUniform.write(
    ease((timestamp % getAnimationDuration()) / getAnimationDuration()),
  );

  const view = context.getCurrentTexture().createView();

  midgroundPipeline
    .withColorAttachment({
      view,
      loadOp: 'clear',
      storeOp: 'store',
      clearValue: shiftedColors[0],
    })
    .with(getInstanceInfoBindGroup())
    .draw(3, getGridParams().triangleCount);
```
`backgoundPipeline` only renders the entire canvas in one color, it can be replaced by `clearValue` in `midgroundPipeline`

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
```diff
@@ -0,0 +1,215 @@
...
+  stepRotationBuffer,
+} = createBuffers(root);
+
+updateAspectRatio(canvas.width, canvas.height, aspectRatioBuffer);
+
+context.configure({
+  device: root.device,
+  format: presentationFormat,
+  alphaMode: 'premultiplied',
+});
```
aleksanderkatan: In the meantime we implemented `root.configureContext()` API that does this for you, see other examples on main branch

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
aleksanderkatan, line 1: These changes are currently on the main branch, thus delaying the example showing up until the next release. You can rebase onto release if you wish to see it sooner!

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
```diff
@@ -0,0 +1,215 @@
...
+const foregroundPipeline = root['~unstable']
+  .with(drawOverNeighborsAccess, drawOverNeighborsBuffer)
+  .with(animationProgressAccess, animationProgressUniform)
+  .with(stepRotationAccess, stepRotationBuffer)
+  .with(shiftedColorsAccess, shiftedColorsBuffer)
+  .with(scaleAccess, scaleBuffer)
+  .with(aspectRatioAccess, aspectRatioBuffer)
+  .withVertex(foregroundVertex)
+  .withFragment(foregroundFragment, { format: presentationFormat })
+  .createPipeline();
```
aleksanderkatan: Unused slots are not included in pipelines, so you could do this if you wanted to:
```suggestion
const pipelineBase = root['~unstable']
  .with(animationProgressAccess, animationProgressUniform)
  .with(stepRotationAccess, stepRotationBuffer)
  .with(drawOverNeighborsAccess, drawOverNeighborsBuffer)
  .with(shiftedColorsAccess, shiftedColorsBuffer)
  .with(middleSquareScaleAccess, middleSquareScaleBuffer)
  .with(scaleAccess, scaleBuffer)
  .with(aspectRatioAccess, aspectRatioBuffer);

const midgroundPipeline = pipelineBase
  .withVertex(midgroundVertex)
  .withFragment(midgroundFragment, { format: presentationFormat })
  .createPipeline();

const foregroundPipeline = pipelineBase
  .withVertex(foregroundVertex)
  .withFragment(foregroundFragment, { format: presentationFormat })
  .createPipeline();
```

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
```diff
@@ -0,0 +1,215 @@
...
+const resizeObserver = new ResizeObserver(() => {
+  updateAspectRatio(canvas.width, canvas.height, aspectRatioBuffer);
+  updateGridParams(scaleBuffer, updateInstanceInfoBufferAndBindGroup);
+  updateInstanceInfoBufferAndBindGroup();
+});
+
+resizeObserver.observe(canvas);
+
+export function onCleanup() {
+  isRunning = false;
```
aleksanderkatan: `isRunning` is used in some old examples that weren't updated for some time.
This leaves one frame enqueued, and this sometimes causes some small issues. It would probably be better to save the frameId and call cancelAnimationFrame here instead

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/shaderModules.ts
```diff
@@ -0,0 +1,197 @@
...
+    stepRotationAccess.$ % SMALLEST_LOOPING_ROTATION_ANGLE,
+    stepRotationAccess.$ +
+      stepRotationAccess.$ % SMALLEST_LOOPING_ROTATION_ANGLE,
+  );
+
+  const scaleFactor = interpolateBezier(
+    animationProgressAccess.$,
+    0.5,
+    middleSquareScaleAccess.$,
+  );
```
aleksanderkatan: If I'm not mistaken, these two values are uniform among all the threads of a single pass, and right now, every thread recalculates it. It would probably be better to pass this in an uniform buffer.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/shaderModules.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  let calculatedPosition = rotate(vertexPosition, angle);
+  calculatedPosition = std.mul(calculatedPosition, scaleFactor);
```
aleksanderkatan: nit: we support infix operators for +-*/
```suggestion
  const calculatedPosition = rotate(vertexPosition, angle).mul(scaleFactor);
```

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/shaderModules.ts
```diff
@@ -0,0 +1,197 @@
...
+  in: MidgroundVertexOutput,
+  out: d.vec4f,
+})(({ maskP0, maskP1, maskP2, vertexClipPos }) => {
+  const e0 = edgeFunction(maskP0, maskP1, vertexClipPos);
+  const e1 = edgeFunction(maskP1, maskP2, vertexClipPos);
+  const e2 = edgeFunction(maskP2, maskP0, vertexClipPos);
+
+  if ((e0 > 0 || e1 > 0 || e2 > 0) && drawOverNeighborsAccess.$ === 0) {
+    std.discard();
+  }
```
aleksanderkatan: This `drawOverNeighbors` switch has made this shader a lot more complicated. Without this, we could probably simplify everything to only one straightforward pipeline. 

I tried to come up with something to avoid passing so much flat interpolated data and without discard. We could theoretically create a static stencil like the following:

image

And use it for "odd" triangles, use a negative stencil for "even" triangles, and no stencil for the front layer. Maybe @reczkok can come up with something better.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/buffers.ts
```diff
@@ -0,0 +1,88 @@
...
+  INITIAL_STEP_ROTATION,
+} from './params.ts';
+
+const stepRotationAccess = tgpu['~unstable'].accessor(d.f32);
+const shiftedColorsAccess = tgpu['~unstable'].accessor(d.arrayOf(d.vec4f, 3));
+const animationProgressAccess = tgpu['~unstable'].accessor(d.f32);
+const middleSquareScaleAccess = tgpu['~unstable'].accessor(d.f32);
+const drawOverNeighborsAccess = tgpu['~unstable'].accessor(d.u32);
+const scaleAccess = tgpu['~unstable'].accessor(d.f32);
+const aspectRatioAccess = tgpu['~unstable'].accessor(d.f32);
```
aleksanderkatan: tip: There is a pretty small limit of buffers that can be used without bind groups, usually it's only 12 (because this is the minimum limit for a single bind group, and those are bound under the hood). A simple way to avoid it is to group those into structs. In this case, it is not necessary, though it might clean up the code a little bit.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/params.ts
```diff
@@ -0,0 +1,147 @@
...
+function updateAspectRatio(
+  width: number,
+  height: number,
+  aspectRatioBuffer: TgpuUniform<d.F32>,
+) {
+  aspectRatio = width / height;
+  aspectRatioBuffer.write(aspectRatio);
+}
+
+let animationDuration = 1000;
```
aleksanderkatan, line 54: This variable is kept between example switches (triangles -> sth else -> triangles). It's just a number so this probably does not matter.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
```diff
@@ -0,0 +1,215 @@
...
+
+  resizeObserver.disconnect();
+  root.destroy();
+}
+
+// Example controls
+
+export const controls = {
+  'Tile density': {
```
aleksanderkatan, line 135: In examples we usually start controls' names with lowercase letters (exception: buttons)

aleksanderkatan (commented):
Great first example! 🦋

I tried to be educational in the comments, so many of them are not necessarily issues but stylistic preferences, nits or good practices.
If you have no time to address these comments, we can take over, no problem

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/buffers.ts
sharabai: The original idea for this example was that with the current implementation it's trivial to add more colors. The example requires three colors at minimum, but it's really easy to add more. It was more of a problem how to do that from the control panel that is available to the user. I would leave it as it is now and expand to more colors later.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/bezier.ts
sharabai: @aleksanderkatan maybe it was better to leave this as a comment in github PR. 
Just wanted to tell people not to ask me about the math here. 
I'm pretty sure this is generic, foundational math.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/meta.json
sharabai: @aleksanderkatan makes sense since it's a default value

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/params.ts
sharabai: @aleksanderkatan could be, but I just didn't finish one thing it the example. Right now, if you click on `Edit Cubic Bezier Points` it would open up with those hardcoded values. But now, after adding updating logic, this `getCubicBezierControsPoints` function makes sense because it gives different outputs.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/buffers.ts
sharabai: @aleksanderkatan, looking at it again, I was wrong. Your point stands. Even with more than 3 colors to choose from UI, the color buffer only needs 3 colors at the time. So you're completely right!

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/transformations.ts
sharabai: @aleksanderkatan thnx! I also prefer to have it without explicitly defined types. 
The problem was that I was getting some errors without it about implicit conversions. 
image
But now that I checked it again it only happens when `interpolateBezier` is not wrapped in `tgpu.fn`. This one is never used directly, so it can be not wrapped in `tgpu.fn`

sharabai: I think even the `std.sub` was part of my initial attempts to combat that warning.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/transformations.ts
sharabai: @aleksanderkatan thx, looks a lot better!

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
sharabai: @aleksanderkatan THAT'S HUGE!!! That's like one of the biggest things I didn't like in general. For some reason I was sure that I could only pass static values there. Don't ask me why. I think I got confused at some point and though that `withColorAttachment` was provided during pipeline definition, not it's `draw` method.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
sharabai: @aleksanderkatan nice, less boilerplate

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
sharabai: @aleksanderkatan  thx for letting me know. Did just that.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
sharabai: @aleksanderkatan Any change that reduces boilerplate code is a good change.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
sharabai: Ok, changed it so that it follows the example of the latest added docs examples with cancelAnimationFrame.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/shaderModules.ts
sharabai: @aleksanderkatan 
I agree. At this moment, it is recalculated for each triangle. 
But I'd hate to see it changed ﹘ I was thinking of adding a grayscale image that would act as a kind of map for each triangle where values from 0 to 255 have influence on animation of each triangle. Those values would represent the delay for each triangle to start rotation.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/shaderModules.ts
sharabai: @aleksanderkatan 
I had working stencils at [Add instances](https://github.com/software-mansion/TypeGPU/pull/2048/changes/9d58007151b3a278b41566f25571dbbc9df7c550) commit.

https://github.com/user-attachments/assets/9389e2b3-3167-44a6-9fbc-1a9bfd13d740

The problem was that with instances always drew over each other.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/params.ts
sharabai: Same for cubicBezierControlPoints, although I don't see any harm.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/index.ts
sharabai: @aleksanderkatan  
I don't know about that. I've seen many where capitalized names are the norm. Like jump-flood-voronoi, jump-flood-distance, jelly-slider to name a few. 
Imo, it's better to handle it at the level of Controls themselves at this point. Just make the names there capitalized or not if consistency in control names is important. At this point, it's a mix of both. Hard to say which one is a standard.

apps/typegpu-docs/src/examples/simple/rotating-triangle-tiles/buffers.ts
sharabai: @aleksanderkatan  got it, I'd keep in mind.

PR #2050, opened 2026-01-14
docs: Small @typegpu/three tutorial

iwoplaza:


apps/typegpu-docs/src/content/docs/ecosystem/typegpu-three/index.mdx
```diff
@@ -0,0 +1,253 @@
...
+
+```diff lang="ts"
+import * as THREE from 'three/webgpu';
++import * as TSL from 'three/tsl';
+import * as t3 from '@typegpu/three';
+
+const material = new THREE.MeshBasicNodeMaterial();
+
++const albedo = TSL.color('magenta').mul(0.5);
++const albedoAccess = t3.fromTSL(albedo, d.vec3f);
```
aleksanderkatan, line 90: `fromTSL` does not work insinde TGSL, right? Is the error descriptive? Maybe we should mention it around here

apps/typegpu-docs/src/content/docs/ecosystem/typegpu-three/index.mdx
```diff
@@ -0,0 +1,253 @@
...
+```
+
+TypeGPU:
+```ts
+for (let i = ptrStart; i < ptrEnd; i++) {
+  const springId = springListBuffer.$[i];
+  const springForce = springForceBuffer.$[springId];
+  const springVertexIds = springVertexIdBuffer.$[springId];
+  const factor = std.select(-1, 1, springVertexIds.x === idx);
+  force = force.add(springForce.mul(d.f32(factor)));
```
aleksanderkatan, line 251: When rewriting three.js code to TGSL, I accidentally rewrote:
```ts
const position = positionsBuffer.element(i);
position.addAssign(1);
```
to
```ts
const position = positionsBuffer.$[i];
position += 1;
```
I think some new users may come across this. Maybe we should include a warning about this caveat in the docs?

apps/typegpu-docs/src/content/docs/ecosystem/typegpu-three/index.mdx
iwoplaza: From what I remember, it works inside TGSL just fine 👀

cieplypolar: Yes, it does work
```ts
material.positionNode = t3.toTSL(() => {
  'use gpu';
  return t3.fromTSL(TSL.positionLocal, d.vec3f).$
    .mul(scaleBuffer.$[t3.instanceIndex.$])
    .add(posBuffer.$[t3.instanceIndex.$]);
});
```

PR #2051, opened 2026-01-14
feat: Stabilize tgpu.comptime

iwoplaza:
The API is pretty simple, we found uses for it internally and in our examples, seems like it's not going anywhere or changing in the near future.


aleksanderkatan (approved):
Nice, let people use it!

I'm still not convinced on the name `comptime`. I don't know what would fit better. Something with `resolve`, `inline`, `embed`, `inject`?

PR #2058, opened 2026-01-14
fix: Merge mobile and non-mobile state in our examples page

reczkok:
- Merge mobile an non-mobile atoms into one
- Clean up some React patterns
- Move the adaptive styles from relying on state to rely on media queries 
- It also seems to fix the bug with canvas sizing on dev server

## Known bug 
If the left menu toggle is the first interaction post reload its state is not reflected in the toggle position

https://github.com/user-attachments/assets/044c1a21-f219-4182-b85a-c99abbb49909

### DIFF
--- a/apps/typegpu-docs/src/pages/examples/index.astro
+++ b/apps/typegpu-docs/src/pages/examples/index.astro
@@ -1,11 +1,11 @@
 ---
-import { ExampleLayout } from "../../components/ExampleLayout";
-import ExamplePage from "../../components/ExamplePage";
-import PageLayout from "../../layouts/PageLayout.astro";
+import { ExampleLayout } from '../../components/ExampleLayout';
+import ExamplePage from '../../components/ExamplePage';
+import PageLayout from '../../layouts/PageLayout.astro';
 ---
 
 <PageLayout title="Examples | TypeGPU">
-    <ExampleLayout client:only="react">
-        <ExamplePage client:only="react" />
-    </ExampleLayout>
+  <ExampleLayout client:load>
+    <ExamplePage client:only="react" />
+  </ExampleLayout>
 </PageLayout>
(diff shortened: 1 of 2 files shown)


biome.json
```diff
@@ -11,6 +11,11 @@
       "!**/*.astro"
     ]
   },
+  "vcs": {
+    "enabled": true,
+    "clientKind": "git",
+    "useIgnoreFile": true
+  },
```
reczkok, line 18: I added this since biome was complaining about my `.zed/settings.json` file - this makes biome ignore all the files that are ignored by git

apps/typegpu-docs/src/pages/examples/index.astro
```diff
@@ -1,11 +1,11 @@
...
+import { ExampleLayout } from "../../components/ExampleLayout";
+import ExamplePage from "../../components/ExamplePage";
+import PageLayout from "../../layouts/PageLayout.astro";
 ---
 
 <PageLayout title="Examples | TypeGPU">
-  <ExampleLayout client:load>
-    <ExamplePage client:only="react" />
-  </ExampleLayout>
+    <ExampleLayout client:only="react">
```
reczkok: This seems to fix hydration

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -344,16 +345,14 @@ export function ControlPanel() {
...
-              {exampleControlParams.map((param) => paramToControlRow(param))}
-            </div>
-          </>
-        )
-        : null}
+      {isGPUSupported && (
+        <>
+          <h2 className='m-0 font-medium text-xl'>Example controls</h2>
+          <div className='grid grid-cols-2 items-center gap-4 overflow-auto p-1 pb-2'>
+            {exampleControlParams.map((param) => paramToControlRow(param))}
```
aleksanderkatan: nit
```suggestion
            {exampleControlParams.map(paramToControlRow)}
```

PR #2065, opened 2026-01-16
fix: Improve perlin noise internals and API with accessors

iwoplaza:
Apart from changing the perlin noise internals and API, there were a few changes to accessors:
- The `AccessorIn` and `MutableAccessorIn` types now live as `TgpuAccessor.In` and `TgpuMutableAccessor.In` respectively. This can help reduce unnecessary top-level export polution.
- Accessors can accept recursively indirect access functions (e.g. `() => () => () => layout.$.foo`). This is helpful when writing a library, and we're unsure whether we can safely access a property in "normal" mode.

### DIFF
--- a/packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
+++ b/packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
@@ -1,10 +1,10 @@
 import tgpu, {
   type Configurable,
   type StorageFlag,
+  type TgpuAccessor,
   type TgpuBuffer,
   type TgpuFn,
   type TgpuRoot,
-  type TgpuSlot,
   type UniformFlag,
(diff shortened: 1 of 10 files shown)


packages/typegpu-noise/src/perlin-2d/dynamic-cache.ts
```diff
@@ -239,13 +227,13 @@ export function dynamicCacheConfig<Prefix extends string>(
       [`${prefix}size`]: { uniform: d.vec2u },
       [`${prefix}memory`]: { storage: MemorySchema, access: 'readonly' },
     } as Layout<Prefix>,
-    valuesSlot,
```
iwoplaza, line 242: It's a slight breaking change, we can note that in the release notes

PR #2069, opened 2026-01-16
feat: Ternary operator

aleksanderkatan:
Changes:
- support for `undefined` in TGSL (previously parsed to `void0`, now parses to empty string),
- support for ternary operator (only with comptime-known condition),
- docs for branch pruning.

### DIFF
--- a/packages/typegpu/tests/tgsl/ternaryOperator.test.ts
+++ b/packages/typegpu/tests/tgsl/ternaryOperator.test.ts
@@ -18,11 +18,11 @@ describe('ternary operator', () => {
       ]),
     )
       .toMatchInlineSnapshot(`
-        "fn falseFn() -> u32 {
+        "fn trueFn() -> u32 {
           return 10u;
         }
 
-        fn falseFn_1() -> u32 {
+        fn falseFn() -> u32 {
           return 20u;
         }"
       `);
@@ -77,19 +77,19 @@ describe('ternary operator', () => {
       ]),
     )
       .toMatchInlineSnapshot(`
-        "fn threeFn() -> u32 {
+        "fn myFn() -> u32 {
           return -1u;
         }
 
-        fn threeFn_1() -> u32 {
+        fn oneFn() -> u32 {
           return 10u;
         }
 
-        fn threeFn_2() -> u32 {
+        fn twoFn() -> u32 {
           return 20u;
         }
 
-        fn threeFn_3() -> u32 {
+        fn threeFn() -> u32 {
           return 30u;
         }"
       `);


packages/typegpu/tests/tgsl/ternaryOperator.test.ts
```diff
@@ -0,0 +1,151 @@
...
+  it('should handle undefined', ({ root }) => {
+    const counter = root.createMutable(d.u32);
+
+    const myFunction = tgpu.fn([])(() => {
+      // biome-ignore lint/correctness/noConstantCondition: it's a test
+      false ? counter.$++ : undefined;
+    });
+    expect(tgpu.resolve([myFunction])).toMatchInlineSnapshot(`
+      "fn myFunction() {
+        ;
```
cieplypolar: simple check at the end of `statement` method will handle this single `;`.

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -150,6 +150,14 @@ const Transpilers: Partial<
     return [NODE.postUpdate, operator, argument];
   },
 
+  ConditionalExpression(ctx, node) {
+    const test = transpile(ctx, node.test) as tinyest.Expression;
+    const consequent = transpile(ctx, node.consequent) as tinyest.Expression;
+    const alternative = transpile(ctx, node.alternate) as tinyest.Expression;
+
+    return [NODE.conditionalExpr, test, consequent, alternative];
```
iwoplaza, line 158: IMO we should call it "conditional" if we want to follow the pattern

```suggestion
    return [NODE.conditional, test, consequent, alternative];
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -791,9 +807,8 @@ ${this.ctx.pre}}`;
     statement: tinyest.Statement,
   ): string {
     if (typeof statement === 'string') {
-      return `${this.ctx.pre}${
-        this.ctx.resolve(this.identifier(statement).value).value
-      };`;
+      const resolved = this.ctx.resolve(this.expression(statement).value).value;
```
iwoplaza: What does changing it from "identifier" to "expression" do? Was it intentional?

packages/tinyest-for-wgsl/src/parsers.ts
cieplypolar: But it is called `ConditionalExpression` in the AST we receive.

aleksanderkatan: Many expressions have the `Expr` suffix, why should we skip this one?

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: Not intentional, I'll fix this

aleksanderkatan: We get to this branch only when using `tgpu.declare`, which I didn't know existed until now (or when otherwise using identifiers as statements, but I see no other use at this moment). Is the `declare` api still maintained and do we consider it potentially useful?

PR #2073, opened 2026-01-19
feat: Variance annotation overrides to achieve better type behavior

iwoplaza:


packages/typegpu/src/core/buffer/bufferShorthand.ts
```diff
@@ -59,7 +59,7 @@ export interface TgpuReadonly<TData extends BaseData>
   // ---
 }
 
-export interface TgpuUniform<TData extends BaseData>
+export interface TgpuUniform<out TData extends BaseData>
```
aleksanderkatan, line 62: We could now do this
```suggestion
export interface TgpuUniform<out TData extends BaseData = BaseData>
```
This would allow code like:
```ts
let uniform: TgpuUniform;
uniform = root.createUniform(d.u32)
```
What do you think?

packages/typegpu/tests/struct.test.ts
```diff
@@ -406,6 +407,37 @@ describe('struct', () => {
...
+    foo(d.struct({ a: d.f32 }));
+  });
+
+  it('accepts structs with more properties', () => {
+    const foo = (_aStruct: d.WgslStruct<{ a: d.F32 }>) => {
+      // Does something with the struct...
+    };
+
+    // @ts-expect-error: It doesn't have the 'a' property
+    foo(d.struct({}));
```
aleksanderkatan: nit (I don't like it when erroneous code is executed in tests, in many cases it leads to undefined behavior)
```suggestion
    (() => foo(d.struct({})));
```

iwoplaza:
> Did you use any particular tool for detecting useless 'as' casts?

@aleksanderkatan Not really, I looked for them manually

packages/typegpu/src/core/buffer/bufferShorthand.ts
iwoplaza: I would rather the types be explicit anyway in that case:

```ts
let uniform: TgpuUniform<d.U32>;
uniform = root.createUniform(d.u32)
```

reczkok (approved):
Amazing!

PR #2074, opened 2026-01-19
feat: Better buffer validation

iwoplaza:


iwoplaza:
Already done in another PR

PR #2083, opened 2026-01-20
feat: Rename `derived` to `lazy` and stabilize

iwoplaza:


packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -208,6 +208,18 @@ export interface WithFragment<
   createPipeline(): TgpuRenderPipeline<Output>;
 }
 
+export interface Withable<TSelf> {
```
aleksanderkatan, line 211: Withable ❤️  
Do we plan on making TgpuFn and pipelines withable as well?

iwoplaza: Done!

reczkok, line 211: Let's goo! We all love Withable ❤️

PR #2085, opened 2026-01-20
feat: More predictable dual-impl behavior

iwoplaza:
- Provide `ctx` to dualImpl codegen functions directly, in preparation for changes required for future `@typegpu/glsl` features
- Calls no longer check whether they're in codegen more or normal mode, simplifying use of dual functions internally.


packages/typegpu/tests/tgslFn.test.ts
```diff
@@ -1008,6 +1008,35 @@ describe('tgsl fn when using plugin', () => {
...
+    const multiplierSlot = tgpu.slot(1);
+    const scale = tgpu.fn([d.f32], d.f32)((v) => {
+      'use gpu';
+      return v * multiplierSlot.$;
+    });
+
+    const main = () => {
+      'use gpu';
+      scale(2);
+      scale.with(multiplierSlot, 2)(2);
```
aleksanderkatan, line 1020: 🔥 🔥 🔥

packages/typegpu/src/shared/symbols.ts
```diff
@@ -25,6 +25,15 @@ export const $ownSnippet = Symbol(`typegpu:${version}:$ownSnippet`);
 
 export const $resolve = Symbol(`typegpu:${version}:$resolve`);
 
+/**
+ * Is a 'casting' schema, but it's meant to be called by the end user (e.g. vertex formats)
+ */
```
aleksanderkatan, line 31: At this point, why aren't vertex formats just callable? 
If that's not achievable, can we change this description to say that some schemas provide a cast via $cast, and others are just callable? Or is this not true?

iwoplaza: You're right, the description is not clear enough. Should be better now 🙌

PR #2086, opened 2026-01-21
impr: Remove unnecessary nested blocks in seed functions

cieplypolar:
- removed unnecessary nested blocks in seed functions
- better `undefined` handling in wgsl generator
- filtering out empty statements while creating a `block`

### DIFF
--- a/apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
+++ b/apps/typegpu-docs/src/examples/threejs/compute-cloth/index.ts
@@ -371,7 +371,11 @@ export const controls = defineControls({
     },
   },
   'Sheen Color': {
-    initial: d.vec3f(API.sheenColor),
+    initial: d.vec3f(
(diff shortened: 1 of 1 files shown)


packages/typegpu-noise/src/random.ts
iwoplaza, line 1: I think we could avoid the `// @ts-expect-error` comments if we did this instead:

```ts
const warnIfNotProvided = tgpu.comptime(
  (seedFnName: keyof typeof randomGeneratorSlot.$) => {
    if (!randomGeneratorSlot.$[seedFnName]) {
      console.warn(`Called \`randf.${seedFnName}\`, but it wasn't provided`);
    }
    
    return undefined;
  },
);

export const randSeed = tgpu.fn([d.f32])((seed) => {
  warnIfNotProvided('seed');
  randomGeneratorSlot.$.seed ? randomGeneratorSlot.$.seed(seed) : undefined;
});
```

reczkok (approved):

PR #2088, opened 2026-01-21
fix: Investigate why some examples work differently

aleksanderkatan:
Mnist Inference - It actually worked just like before, I compared it with an old commit.
Stable Fluids - -||-
3D Fish - The repulsion ray actually was weaker. I made it stronger, and on hover instead of on RMB.
Function Visualizer - I was unable to reproduce the texture size mismatch, I cleaned up unnecessary complications in texture size calculation.

### DIFF
--- a/apps/typegpu-docs/src/examples/rendering/3d-fish/index.ts
+++ b/apps/typegpu-docs/src/examples/rendering/3d-fish/index.ts
@@ -147,13 +147,8 @@ const camera = {
 };
 
 const cameraBuffer = root.createBuffer(Camera, camera).$usage('uniform');
-
-const mouseRayBuffer = root
-  .createBuffer(Line3, Line3({ origin: d.vec3f(), dir: d.vec3f() }))
-  .$usage('uniform');
-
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/examples/rendering/3d-fish/index.ts
```diff
@@ -150,10 +149,7 @@ const camera = {
 const cameraBuffer = root.createBuffer(Camera, camera).$usage('uniform');
 
 const mouseRayBuffer = root
-  .createBuffer(MouseRay, {
-    activated: 0,
-    line: Line3({ origin: d.vec3f(), dir: d.vec3f() }),
-  })
+  .createBuffer(Line3, Line3({ origin: d.vec3f(), dir: d.vec3f() }))
   .$usage('uniform');
```
iwoplaza: I think we could try and do the following, since the initial value is the default one:

```ts
const mouseRayBuffer = root.createBuffer(Line3).$usage('uniform');
```

PR #2092, opened 2026-01-21
refactor: Double down on BaseData

iwoplaza:
Using `AnyData` and `AnyWgslData` internally complicates our types, causing the TypeScript server to chug in generic use-cases, and occasionally crash with "Type instantiation was excessively deep". This PR makes more use of `BaseData` internally, which is a subtype of all schemas.

### DIFF
--- a/packages/typegpu/src/data/dataTypes.ts
+++ b/packages/typegpu/src/data/dataTypes.ts
@@ -30,6 +30,7 @@ import type { PackedData } from './vertexFormatData.ts';
 import * as wgsl from './wgslTypes.ts';
 import type { WgslComparisonSampler, WgslSampler } from './sampler.ts';
 import type { ResolutionCtx } from '../types.ts';
+import type { BaseData } from './wgslTypes.ts';
 
 /**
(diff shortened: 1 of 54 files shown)


packages/typegpu/src/data/dataTypes.ts
```diff
@@ -240,16 +247,8 @@ export type AnyConcreteData = Exclude<
...
-  readonly type: 'unknown';
-}
-
-export const UnknownData = {
-  type: 'unknown' as const,
-  toString() {
-    return 'unknown';
-  },
-} as UnknownData;
+export const UnknownData = Symbol('UNKNOWN');
```
aleksanderkatan, line 250: Does this truly need to be a symbol? Other schemas aren't symbols

iwoplaza: It benefits from being a symbol because after the switch to BaseData, we couldn't distinguish between `UnknownData` and other BaseData on the type level (since `type` was a string, not a union of literals)

PR #2096, opened 2026-01-22
feat: FPS counter

lursz:
Introducing FPS counter across all examples
image


apps/typegpu-docs/src/components/FpsCounter.tsx
iwoplaza, line 1: Since this is a component, why not put it in `typegpu-docs/src/components/` and call it `FpsCounter.tsx`?

lursz: Fair, fixed

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -315,6 +316,9 @@ export function ControlPanel() {
         'box-border flex max-h-[50%] flex-col gap-4 overflow-auto rounded-xl bg-grayscale-0 p-6 md:max-h-full',
       )}
     >
+      <FPSCounter
+        className={'flex items-center gap-3 rounded-xl bg-grayscale-20/80 px-4 py-2 font-mono text-grayscale-80 text-xs'}
+      />
```
iwoplaza: I feel like these styles make more sense inside of the component itself. They require knowledge of how elements are laid out.

lursz: aye aye done

apps/typegpu-docs/src/components/FpsCounter.tsx
```diff
@@ -0,0 +1,78 @@
...
+        metrics.frames.length > 0
+      ) {
+        const avg = metrics.frames.reduce((a, b) => a + b, 0) /
+          metrics.frames.length;
+
+        if (fpsRef.current) {
+          fpsRef.current.innerText = Math.round(1000 / avg).toString();
+        }
+        if (frameTimeRef.current) {
+          frameTimeRef.current.innerText = avg.toFixed(2);
```
aleksanderkatan, line 48: I'd use only one digit after decimal, because the second one keeps flickering.
```suggestion
          frameTimeRef.current.innerText = avg.toFixed(1);
```

https://github.com/user-attachments/assets/a001b99a-03d4-4c14-9a18-715d085750fd


https://github.com/user-attachments/assets/4fa18feb-e781-4a19-9ddd-0fa0d8a3c2b4

apps/typegpu-docs/src/components/FpsCounter.tsx
```diff
@@ -0,0 +1,78 @@
...
+      if (!metrics.lastTime) {
+        metrics.lastTime = metrics.lastUpdate = time;
+        return;
+      }
+
+      metrics.frames.push(time - metrics.lastTime);
+      metrics.lastTime = time;
+
+      if (metrics.frames.length > POOLING_LENGTH) {
+        metrics.frames.shift();
```
aleksanderkatan, line 34: Just in case you want to feel a little less like a frontend developer, I made the observation that the average frame length is a total divided by length (and total can be calculated in O(1) by adding the new one and subtracting the old one), and I have here a simple queue implementation.

```ts
export class Queue<T> {
  #stacks: [T[], T[]];
  constructor(initialElements: T[]) {
    this.#stacks = [[...initialElements], []];
  }

  add(elem: T) {
    this.#stacks[0].push(elem);
  }

  remove(): T | undefined {
    if (this.#stacks[1].length === 0) {
      this.#relocate();
    }
    return this.#stacks[1].pop();
  }

  get length() {
    return this.#stacks[0].length + this.#stacks[1].length;
  }

  #relocate() {
    while (this.#stacks[0].length > 0) {
      this.#stacks[1].push(this.#stacks[0].pop() as T);
    }
  }
}
```

apps/typegpu-docs/src/components/FpsCounter.tsx
lursz: Resolved irl

apps/typegpu-docs/src/components/FpsCounter.tsx
lursz: Nice O(1) instead of O(n). But I think there is no need to overengineer this, this array has 60 elements, never more, never less and we have whole React to slow us down over our heads.  Besides, this impl forces me to drop reduce, and creates some memory overhead because of class and two stacks. What about circular buffer? Would this be the best of both worlds?

apps/typegpu-docs/src/components/FpsCounter.tsx
reczkok: Please capitalize the file name for consistency

PR #2097, opened 2026-01-22
feat: Lint plugin POC

aleksanderkatan:
Edit: We decided that we'll use a non-type-aware plugin instead.

This is an eslint plugin, thus it should be compatible with oxc.
The plugin currently has one rule, which detects int/int divisions. 
I did not do any package publishing stuff yet.

image

To test the CLI locally in the typegpu package:
- run `ni; nr build` in the lint plugin package,
- run `pnpm install -D eslint`, `pnpm install -D @typescript-eslint/parser` in typegpu package,
- add `"eslint-plugin-typegpu": "workspace:*"` dependency to typegpu dev dependencies in package.json,
- add an `eslint.config.js` file
```ts
import typegpu from 'eslint-plugin-typegpu';
import tsParser from '@typescript-eslint/parser';

export default [
  typegpu.configs.recommended,
  {
    files: ['**/*.ts', '**/*.tsx'],
    languageOptions: {
      parser: tsParser,
    },
  },
];
```
- run `ni; npx eslint` in typegpu package.

The plugin tests use vitest's `describe` and `it` under the hood, and are compatible with hot reload. Tests are wrapped into a `describe` to let vitest's static analysis know that this is in fact a test file.

Edit: 
I checked and the plugin is compatible with oxlint -- we need to install it and add `"jsPlugins": ["eslint-plugin-typegpu"],` to the `.oxlintrc.json` file, as well as list all rules that we want to enable. To my understanding, importing rule presets will become possible after issue https://github.com/oxc-project/oxc/issues/17527 is completed.


packages/eslint-plugin/src/rules/integerDivision.ts
```diff
@@ -0,0 +1,61 @@
...
+      BinaryExpression(node) {
+        if (node.operator !== '/') {
+          return;
+        }
+
+        if (node.parent?.type === 'CallExpression' && isIntCast(node.parent)) {
+          return;
+        }
+
+        if (isIntCast(node.left) && isIntCast(node.right)) {
```
reczkok: I think something like `12 / d.u32(5)` would be a common pattern (wrapping just the denominator and expecting integer division). I know coding agents tend to do that.

packages/eslint-plugin/tsup.config.ts
```diff
@@ -0,0 +1,12 @@
+import { defineConfig, type Options } from 'tsup';
+
+const config: Options = {
+  entry: ['src/index.ts'],
+  format: ['cjs', 'esm'],
```
cieplypolar:
```suggestion
  format: ['esm'],
```

packages/eslint-plugin/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "eslint-plugin-typegpu",
+  "version": "0.9.0",
+  "description": "An eslint plugin containing custom rules for TypeGPU.",
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
```
cieplypolar:
```suggestion
```

cieplypolar (commented):
LGTM!

PR #2100, opened 2026-01-23
impr/feat: Game Of Life on steroids and internal improvements

reczkok:
# Update 2
# All the library changes are now part of #2126 for reviewer convenience 

## - also we now handle the types of mixed operators using signatures instead of inside unify (which prevented vector validation)

Apart from total Game Of Life rewrite (featuring 3 implementations) this PR:
- Allows expected type to be a tuple of types (our type conversion system will pick the best one)
- Adds textureGather std function
- Concretizes dualImpl output to avoid fake abstracts
- Fixes a small bug that would mark all `~` operations as `bool`
- Adds `callableSchema` which is a lightweight version of `dualImpl` to avoid circular deps

### DIFF
--- a/apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
+++ b/apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
@@ -1,14 +1,12 @@
 import { randf } from '@typegpu/noise';
-import tgpu, {
-  type SampledFlag,
-  type StorageFlag,
-  type TgpuBindGroup,
-  type TgpuComputeFn,
-  type TgpuTexture,
+import tgpu, { common, d, std } from 'typegpu';
+import type {
+  SampledFlag,
+  StorageFlag,
+  TgpuBindGroup,
+  TgpuComputeFn,
+  TgpuTexture,
 } from 'typegpu';
-import { fullScreenTriangle } from 'typegpu/common';
-import * as d from 'typegpu/data';
-import * as std from 'typegpu/std';
 import {
   computeLayout,
   displayLayout,
@@ -28,7 +26,6 @@ const root = await tgpu.init({
 const hasTimestamp = root.enabledFeatures.has('timestamp-query');
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
 const context = root.configureContext({ canvas, alphaMode: 'premultiplied' });
-const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
 
 let gameSize = 64;
 const gameSizeUniform = root.createUniform(d.u32, gameSize);
@@ -66,9 +63,9 @@ let displayBindGroups: TgpuBindGroup<typeof displayLayout.entries>[];
 
 let even = 0;
 
-const randomInit = root['~unstable'].createGuardedComputePipeline((x, y) => {
+const randomInit = root.createGuardedComputePipeline((x, y) => {
   'use gpu';
-  randf.seed2(d.vec2f(x, y).div(d.f32(gameSizeUniform.$)).mul(timeUniform.$));
+  randf.seed2(d.vec2f(x, y) / d.f32(gameSizeUniform.$) * timeUniform.$);
   std.textureStore(
     computeLayout.$.next,
     d.vec2u(x, y),
@@ -99,13 +96,11 @@ const getBrushRadius = (radiusNorm: number, gameSize: number) => {
   return std.mix(halfPixelRadius, maxRadius, radiusNorm);
 };
 
-const handleDraw = root['~unstable'].createGuardedComputePipeline((x, y) => {
+const handleDraw = root.createGuardedComputePipeline((x, y) => {
   'use gpu';
-  const uv = d.vec2f(d.f32(x) + 0.5, d.f32(y) + 0.5).div(
-    d.f32(gameSizeUniform.$),
-  );
+  const uv = (d.vec2f(x, y) + 0.5) / d.f32(gameSizeUniform.$);
   const stroke = brushStrokeUniform.$;
-  randf.seed2(d.vec2f(x, y).div(d.f32(gameSizeUniform.$)).mul(timeUniform.$));
+  randf.seed2(d.vec2f(x, y) / d.f32(gameSizeUniform.$) * timeUniform.$);
   if (
     sdLine(uv, stroke.start, stroke.end) <=
       getBrushRadius(stroke.radius, gameSizeAccessor.$)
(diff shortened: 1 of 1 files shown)


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -137,7 +137,7 @@ function operatorToType<
   TR extends AnyData | UnknownData,
 >(lhs: TL, op: Operator, rhs?: TR): TL | TR | wgsl.Bool {
   if (!rhs) {
-    if (op === '!' || op === '~') {
```
reczkok: This was just not true

packages/typegpu/src/core/function/dualImpl.ts
```diff
@@ -112,7 +116,7 @@ export function dualImpl<T extends AnyFn>(
 
       return snip(
         options.codegenImpl(ctx, converted),
-        returnType,
+        concretize(returnType),
```
reczkok, line 119: This avoids a situation where we get all abstract inputs and mark the output as abstract even tho in WGSL it won't be

apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
aleksanderkatan, line 1: The brush in add mode does not always color the pixels under the cursor which makes it difficult to draw precisely (in both shots I held the LMB, they differ by brush radius). I think it is also offset by 0.5 of a pixel. Could we make brush radius unit to be pixels?

https://github.com/user-attachments/assets/adde785e-d6c5-4b3f-bacc-89bb5c3e7998

https://github.com/user-attachments/assets/90ba4112-f81d-4ad3-b9c8-e26ebef1c923

apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
```diff
@@ -1,177 +1,576 @@
...
+}
+requestAnimationFrame(frame);
 
 // #region Example controls & Cleanup
 
 export const controls = {
   size: {
     initial: '64',
-    options: [16, 32, 64, 128, 256, 512].map((x) => x.toString()),
+    options: [16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192].map((x) =>
```
aleksanderkatan: Size 16 does not work with bitpacked option.

```suggestion
    options: [32, 64, 128, 256, 512, 1024, 2048, 4096, 8192].map((x) =>
```

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -326,38 +326,40 @@ Consider using explicit conversions instead.`,
...
-      }' to type '${targetDataType.type}'`,
-    );
+    if (target === dataType) {
+      return snip(value, target, origin);
+    }
+
+    if (dataType.type === 'unknown') {
+      // Commit unknown to the expected type.
+      return snip(stitch`${snip(value, target, origin)}`, target, origin);
+    }
```
aleksanderkatan: Why do we commit to unknown only when there is exactly one allowed target data type?

apps/typegpu-docs/src/examples/simulation/game-of-life/shaders/bitpacked-compute.ts
```diff
@@ -0,0 +1,164 @@
...
+
+// deno-fmt-ignore
+const Neighbors8 = d.struct({
+  nw: d.u32, n: d.u32, ne: d.u32,
+  w:  d.u32,            e: d.u32,
+  sw: d.u32, s: d.u32, se: d.u32,
+});
+
+/**
+ * Gather a 3x3 neighborhood using 4 textureGather calls.
```
aleksanderkatan, line 21: We could replace 4 gathers with 2 gathers and 2 loads. Could this be faster?

apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
```diff
@@ -1,177 +1,576 @@
...
-  Reset: {
+  'zoom sensitivity': {
+    initial: 0.3,
+    min: 0.1,
+    max: 2,
+    step: 0.1,
+    onSliderChange: (value: number) => {
+      input.zoomSensitivity = value;
+    },
+  },
```
aleksanderkatan, line 548: I think we could remove this slider, it's not that useful to the user and the example already has many controls

packages/typegpu/src/std/texture.ts
```diff
@@ -504,6 +504,137 @@ export const textureDimensions = dualImpl({
...
+    ...args: GatherCubeArgs<T>
+  ): PrimitiveToLoadedType[T[typeof $internal]['type']];
+  <T extends textureCubeArray>(
+    ...args: GatherCubeArrayArgs<T>
+  ): PrimitiveToLoadedType[T[typeof $internal]['type']];
+  (...args: GatherDepth2dArgs): v4f;
+  (...args: GatherDepth2dArrayArgs): v4f;
+  (...args: GatherDepthCubeArgs): v4f;
+  (...args: GatherDepthCubeArrayArgs): v4f;
+  // (...args: TextureGatherCpuArgs): v4f | v4i | v4u;
```
aleksanderkatan: Do we need this general signature?

apps/typegpu-docs/src/examples/simulation/game-of-life/shaders/bitpacked-compute.ts
reczkok: Actually around 10% slower

packages/typegpu/src/tgsl/conversion.ts
reczkok: There is no valid scenario where the data type is unknown and more than one expected type is possible. We only reach this state during property or index access, where the expected type is determined by the accessed member and therefore singular. If, for any reason, multiple expected types happen, that should result in a conversion error - selecting one would require an arbitrary choice.

apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
reczkok: I don’t agree. This may work with the default scrolling behavior on macOS, but other operating systems such as Windows produce very different scroll delta values. As a result, zoom in examples like cubemap reflection become sunusable on my Windows PC, and I do not want this example to suffer from the same issue.

packages/typegpu/src/std/texture.ts
reczkok: Nope, I commented it out and forgot about it

apps/typegpu-docs/src/examples/simulation/game-of-life/index.ts
reczkok: Should be fixed now

aleksanderkatan: Works perfectly!

https://github.com/user-attachments/assets/928af2a2-6c4d-4f67-a315-67c210574090

PR #2105, opened 2026-01-26
feat: Add `Indirect` buffer usage and support `.dispatchWorkgroupsIndirect` API

reczkok:

### DIFF
--- a/apps/typegpu-docs/src/examples/algorithms/concurrent-chart/calculator.ts
+++ b/apps/typegpu-docs/src/examples/algorithms/concurrent-chart/calculator.ts
@@ -1,7 +1,6 @@
 import { initCache, prefixScan } from '@typegpu/concurrent-scan';
-import * as d from 'typegpu/data';
 import type { TgpuRoot } from 'typegpu';
-import * as std from 'typegpu/std';
+import { d, std } from 'typegpu';
 
 type SumResult = {
   success: boolean;
--- a/apps/typegpu-docs/src/examples/algorithms/concurrent-chart/meta.json
+++ b/apps/typegpu-docs/src/examples/algorithms/concurrent-chart/meta.json
@@ -1,5 +1,6 @@
 {
   "title": "Concurrent Chart",
   "category": "algorithms",
-  "tags": ["compute", "concurrent", "timestamp query"]
+  "tags": ["compute", "concurrent", "timestamp query"],
+  "dev": true
 }


packages/typegpu/src/data/offsetUtils.ts
```diff
@@ -0,0 +1,301 @@
...
+    for (let j = runStart; j <= i; j++) {
+      const runName = propNames[j];
+      if (!runName) {
+        continue;
+      }
+      const runInfo = offsets[runName] as PropOffsetInfo;
+      meta.set(runName, { offset: runInfo.offset, runEnd });
+    }
+    runStart = i + 1;
+  }
```
reczkok, line 220: I'm not quite sure we need this. We could just walk from accessed prop until we run out of requested memory or hit padding. This would make the API less flexible though (and worst case actually calculate more). This solution seemed nice for detecting contiguous memory and I thought maybe we could use it in the future (compiled partial write). I can simplify it if it's deemed too complex/expensive.

packages/typegpu/tests/computePipeline.test.ts
```diff
@@ -557,4 +559,138 @@ describe('TgpuComputePipeline', () => {
...
+      );
+
+      // biome-ignore lint/style/noNonNullAssertion: <it's fine>
+      expect(warnSpy.mock.calls[1]![0]).toMatchInlineSnapshot(
+        `"dispatchWorkgroupsIndirect: Starting at offset 44, only 8 contiguous bytes of data are available before padding. Dispatch requires 12 bytes (3 x u32). Reading across padding may result in undefined behavior."`,
+      );
+
+      pipeline.dispatchWorkgroupsIndirect(
+        deepBuffer,
+        (s) => s.nested.innerNested[0]?.yy as number,
```
iwoplaza: Let's change this API to accept a number, and create a user-facing API for calculating the starting offset of a nested value in a buffer. We can place a usage example in the JSDocs of `dispatchWorkgroupsIndirect` to signal the easiest usage of those APIs

packages/typegpu/src/data/offsetUtils.ts
```diff
@@ -0,0 +1,347 @@
...
+  const offsets = offsetsForProps(struct);
+  const propTypes = struct.propTypes as Record<string, AnyWgslData>;
+  const propNames = Object.keys(propTypes);
+
+  const meta = new Map<string, StructFieldMeta>();
+
+  let runStart = 0;
+  for (let i = 0; i < propNames.length; i++) {
+    const name = propNames[i];
+    if (!name) {
```
cieplypolar, line 191: Can user define a struct with the empty string as a key?

packages/typegpu/src/data/offsetUtils.ts
cieplypolar: I like it

packages/typegpu/tests/offsetUtils.test.ts
```diff
@@ -0,0 +1,140 @@
...
+  });
+
+  it('tracks offsets for nested structs inside arrays', () => {
+    const info = getOffsetInfoAt(
+      DeepStruct,
+      (s) => s.nested.innerNested[1]?.myVec.x as number,
+    );
+
+    expect(info.offset).toBe(128);
+    expect(info.contiguous).toBe(16);
```
cieplypolar:
```suggestion
    expect(info.contiguous).toBe(28);
```
Look at #2159

cieplypolar (commented):
Great use of accessors. However, the calculation of the longest contiguous block starting at given offset isn't correct. See #2159

packages/typegpu/src/data/offsetUtils.ts
aleksanderkatan: `isValidProp` does not yet filter this, but it should

packages/typegpu/tests/computePipeline.test.ts
```diff
@@ -613,4 +614,90 @@ describe('TgpuComputePipeline', () => {
...
+      // biome-ignore lint/style/noNonNullAssertion: <it's fine>
+      expect(warnSpy.mock.calls[2]![0]).toMatchInlineSnapshot(
+        `"dispatchWorkgroupsIndirect: Starting at offset 84, only 8 contiguous bytes are available before padding. Dispatch requires 12 bytes (3 x u32). Reading across padding may result in undefined behavior."`,
+      );
+
+      warnSpy.mockRestore();
+    });
+
+    it('does not warn when dispatch has sufficient contiguous data', ({ root, commandEncoder }) => {
+      const warnSpy = vi.spyOn(console, 'warn').mockImplementation(() => {});
```
iwoplaza: Could `using` be used here?

packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -65,6 +73,19 @@ export interface TgpuComputePipeline
     y?: number | undefined,
     z?: number | undefined,
   ): void;
+
+  /**
+   * Dispatches compute workgroups using parameters read from a buffer.
+   * The buffer must contain 3 consecutive u32 values (x, y, z workgroup counts).
+   * To get the correct offset within complex data structures, use `getOffsetInfoAt(...)`.
```
iwoplaza:
```suggestion
   * To get the correct offset within complex data structures, use `d.getOffsetInfoAt(...)`.
```

packages/typegpu/src/data/offsetUtils.ts
```diff
@@ -0,0 +1,348 @@
...
+
+/**
+ * A function that retrieves offset information for a specific primitive within a data schema.
+ * Example usage:
+ * ```ts
+ * const Boid = d.struct({
+ *  position: d.vec3f,
+ *  velocity: d.vec3f,
+ * });
+ * const offsetInfo = getOffsetInfoAt(Boid, (b) => b.velocity.y);
```
iwoplaza:
```suggestion
 * const offsetInfo = d.getOffsetInfoAt(Boid, (b) => b.velocity.y);
```

iwoplaza (approved):

PR #2107, opened 2026-01-26
impr: More shellless autonaming

aleksanderkatan:
Currently, functions are named by a different mechanism than everything else -- the name is one of the props assigned to typegpu metadata.

I tried to unify these mechanisms, but I came across two issues. 

Firstly, at this moment there is no way to get access to implementation-specific functions in `common.ts`. I needed to use `containsResourceConstructorCall`, which has different implementations in rollup and babel, and I couldn't do it without nasty ifs or rewriting the `common.ts` to be an abstract class.

Secondly, the general autonaming mechanism relies on a resource being assigned to a variable, so I couldn't do it in the case of function declarations (if we apply the autonaming first, nothing happens because there is no assignment, and if we transpile first, also nothing happens because now on the rhs of the variable declaration is a complex mess with metadata).

So instead I just modernised the function naming mechanism.


packages/unplugin-typegpu/src/common.ts
```diff
@@ -126,19 +126,53 @@ export function isShellImplementationCall(
...
+    return [node.key.name, node.value as ExpressionFor<T>];
+  } else if (
+    (node.type === 'ClassProperty' || node.type === 'PropertyDefinition') &&
+    node.value &&
+    node.key.type === 'Identifier'
+  ) {
+    return [node.key.name, node.value as ExpressionFor<T>];
+  }
+}
+
```
reczkok, line 171: I would love some comments with simple examples for the branches - you can quite quickly parse the condition but it would be nice to see at a glance

aleksanderkatan: sure

Issue #2031, opened 2026-01-09
(RFC) feat: Better defaults


iwoplaza:
# root.configureContext

These 6-or-so lines appear in pretty much every TypeGPU rendering example.

```ts
const context = canvas.getContext('webgpu') as GPUCanvasContext;

context.configure({
  device: root.device,
  format: presentationFormat,
  alphaMode: 'premultiplied',
  // ...
});
```

I propose we package it up as a single method: `configureContext`, which retrieves the webgpu context, and configures it with a few defaults:
- Passes `root.device` as `device` (no brainer)
- Passes `navigator.gpu.getPreferredCanvasFormat()` as `format`. We can debate, but I think this is a good default. The user can change this by passing a different format to the second options parameter.

```ts
const context = root.configureContext({
  canvas,
  alphaMode: 'premultiplied',
  // ...
});
```

# `navigator.gpu.getPreferredCanvasFormat()` as the default format

If we assume `navigator.gpu.getPreferredCanvasFormat()` to be the default format when rendering, the `createRenderPipeline` becomes even simpler.

```ts
const pipeline = root.createRenderPipeline({
  vertex: fullScreenTriangle,
  fragment: ({ uv }) => {
    'use gpu';
    return getGradientColor((uv.x + uv.y) / 2);
  },
  // No longer necessary:
  // targets: { format: presentationFormat },
});
```

# Default to `loadOp: 'clear', storeOp: 'store'`

This one might be controversial, but its by far the most frequent setup.

# Accept `context` in `withColorAttachment`

With this one and the previous default combined, the draw call could look like this:

```ts
pipeline.withColorAttachment({ view: context }).draw(3);
```

# Export namespaces from `typegpu`

As @aleksanderkatan suggested a while back in discussions #1497, we could export all namespaces from `typegpu` if we know they can still be tree-shaken after:

```ts
import tgpu from 'typegpu';
import * as d from 'typegpu/data';
import * as std from 'typegpu/std';
import * as common from 'typegpu/common';
// Turns into:
import { tgpu, d, std, common } from 'typegpu';
```

# The minimal hello world

After these changes, the minimal hello world could look like this:

```ts
import { tgpu, d, std, common } from 'typegpu';

const root = await tgpu.init();

const pipeline = root.createRenderPipeline({
  vertex: ({ $vertexIndex }) => {
    'use gpu';
[...]

PR #2116, opened 2026-01-27
feat: `root.configureContext`

aleksanderkatan:
See #2031 for reference.
This PR only contains the `root.configureContext`, we'll do render pipeline defaults in another PR.

### DIFF
--- a/apps/typegpu-docs/src/components/ExampleLayout.tsx
+++ b/apps/typegpu-docs/src/components/ExampleLayout.tsx
@@ -1,4 +1,4 @@
-import { useAtom, useSetAtom } from 'jotai';
+import { useSetAtom } from 'jotai';
 import type { ReactNode } from 'react';
 import { useId, useRef } from 'react';
 import CrossSvg from '../assets/cross.svg';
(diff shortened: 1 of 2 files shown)


apps/typegpu-docs/src/content/docs/fundamentals/roots.mdx
```diff
@@ -12,71 +12,90 @@ Roots are responsible for resource allocation and management. Whether you'd like
...
+### Destroying resources
+
+Calling `root.destroy()` will call `device.destroy()` to let the browser know that it can free up all the resources. 
+
+If the root was created via `tgpu.initFromDevice()`, this method does nothing.
+
+```ts
+root.destroy(); // <- frees up all the resources
+```
+
```
iwoplaza, line 91: Why is this now part of "Best practices"?

packages/typegpu/src/core/root/rootTypes.ts
```diff
@@ -578,6 +590,12 @@ export interface TgpuRoot extends Unwrapper {
    */
   readonly device: GPUDevice;
 
+  /**
+   * Creates and configures context for the provided canvas.
+   * Automatically sets the format to `navigator.gpu.getPreferredCanvasFormat()` if not provided.
+   */
```
iwoplaza, line 597: Can you add a note that this function throws if it couldn't configure the context?

iwoplaza:
```suggestion
   * @throws An error if no context could be obtained
   */
```

apps/typegpu-docs/src/content/docs/fundamentals/roots.mdx
aleksanderkatan: It just kind of feels like an example of good practice, because it is not really necessary.

> "Calling the destroy method on a [WebGPU interface](https://www.w3.org/TR/webgpu/#webgpu-interface) (such as [GPUDevice](https://www.w3.org/TR/webgpu/#gpudevice).[destroy()](https://www.w3.org/TR/webgpu/#dom-gpudevice-destroy) or [GPUBuffer](https://www.w3.org/TR/webgpu/#gpubuffer).[destroy()](https://www.w3.org/TR/webgpu/#dom-gpubuffer-destroy)) should be favored over relying on garbage collection if predictable release of allocated resources is needed."

"should be favored"

I can revert it though

PR #2125, opened 2026-01-29
fix: proper handling of varying nodes

cieplypolar:
### Changes:
- added example to check if varying nodes work
- added analyze method for `tgpuFnNode`, as a result we can properly resolve subnodes
- changed the conditions when we provide private variable for node

### Before:
ThreeJS handles varying nodes by creating a `varyings` struct. We incorrectly created private variable. This prevented the value from propagating correctly between shader modules.
```ts
const positionNode = t3.toTSL(() => {
  'use gpu';
  ...
  const newNormalLocal = d.vec3f(-derivative, 1.0, 0);
  vNormalAccessor.$.x = newNormalLocal.x;
  vNormalAccessor.$.y = newNormalLocal.y;
  vNormalAccessor.$.z = newNormalLocal.z;
  ...
});
```
Resolved to
```rust
var<private> var_2: vec3f; // data is trapped in this private var

fn fnName() -> vec3f {
  ...
  var newNormalLocal = vec3f(-(derivative), 1f, 0f);
  var_2.x = newNormalLocal.x;
  var_2.y = newNormalLocal.y;
  var_2.z = newNormalLocal.z;
  ...
}
```

### After:
The code now correctly targets the `varyings` struct
```rust
fn fnName() -> vec3f {
  ...
  var newNormalLocal = vec3f(-(derivative), 1f, 0f);
  varyings.vNormal.x = newNormalLocal.x;
  varyings.vNormal.y = newNormalLocal.y;
  varyings.vNormal.z = newNormalLocal.z;
  ....
}
```

The rest of this description is not relevant for the reviewers.

---

This part of `@typegpu/three` library is responsible for node resolution
```ts
if (this.var) {
  return this.var.$;
}

return tgpu['~unstable'].rawCodeSnippet(
  this.node.build(ctx.builder) as string,
  this.#dataType,
).$;
```
However, there is a problem. When we later read from the varying node
```ts
const transformedNormalAccessor = t3.fromTSL(
  TSL.transformNormalToView(vNormal),
  d.vec3f,
);

const normalNode = t3.toTSL(() => {
  'use gpu';
  return std.normalize(transformedNormalAccessor.$);
});
```
`this.node.build(ctx.builder)` returns `vNormal` instead of `varyings.vNormal`. 

#### ThreeJS behavior
As far as I understand the internals of ThreeJS, it traverses `material` from fragment shader to the vertex shader and modifies the nodes based on the context. For example, if we use `VaryingNode` in both the fragment and vertex shaders, the traversal process marks it with `needsInterpolation` upon visiting the node in the fragment shader. This prevents optimizations in the vertex shader. This prevents certain optimizations in the vertex shader. Consequently, the `VaryingNode` identifier is resolved as `varyings.name`.

I attach code snippets from ThreeJS repo below
```ts
// this property can be used to check if the varying can be optimized for a variable
varying.needsInterpolation || ( varying.needsInterpolation = ( builder.shaderStage === 'fragment' ) );
```
```ts
if ( node.isNodeVarying === true && node.needsInterpolation === true ) {
  if ( shaderStage === 'vertex' ) {
    return `varyings.${ node.name }`;
  }
}
```

#### TypeGPU (parametrized by usage of `VaryingNode`)
1. vertex shader - TypeGPU function, fragment shader - TSL
ThreeJS sees `VaryingNode` in the fragment shader, modifies it and `builder` infers the correct name.

2. vertex shader - TypeGPU function, fragment shader - TypeGPU function
Our `TgpuFunctionNode` does not implement `analyze` method. As a result, there is no way to mark `VaryingNode` used inside with `needsInterpolation` flag.

---

TODO:
- [x] varying nodes created from attributes should be private vars (they are handled as separate arguments; see `positionLocal`) - not sure about that
- [x] complex usage of varying nodes is still not working (when we read from varying node in 'use gpu' function)


packages/typegpu-three/src/typegpu-node.ts
```diff
@@ -206,7 +276,7 @@ export function toTSL(
 export class TSLAccessor<T extends d.AnyWgslData, TNode extends THREE.Node> {
   readonly #dataType: T;
 
-  readonly var: TgpuVar<'private', T> | undefined;
+  var: TgpuVar<'private', T> | undefined;
```
iwoplaza: I wouldn't want this to be update'able from the outside. Could you create a private `#var` element and a readonly getter `var`?

iwoplaza (approved):
Awesome changes! I submitted a little nitpick

PR #2126, opened 2026-01-30
impr: Refine std function and conversion handling

reczkok:
(Originally part of #2100)
This PR:
- tightens signatures of numeric std functions which provides errors faster and makes sure we have correct return types for the functions
- proposes a more robust solution for mixed vector/scalar ops using signatures instead of a condition in unification system (which made it impossible to validate vector types)
- adds texture gather std function (it's really nice)
- extends expected type to handle not only singular types but also tuples


packages/typegpu/src/core/function/createCallableSchema.ts
```diff
@@ -0,0 +1,89 @@
...
+import { tryConvertSnippet } from '../../tgsl/conversion.ts';
+import {
+  type DualFn,
+  isKnownAtComptime,
+  NormalState,
+  type ResolutionCtx,
+} from '../../types.ts';
+
+type MapValueToDataType<T> = { [K in keyof T]: BaseData };
+type AnyFn = (...args: never[]) => unknown;
```
cieplypolar: we have this type exported from `fnTypes.ts`

packages/typegpu/src/std/numeric.ts
```diff
@@ -62,6 +67,18 @@ const unaryIdentitySignature = (arg: BaseData) => {
...
+    const argRestricted = unify([arg], restrict);
+    if (!argRestricted) {
+      throw new SignatureNotSupportedError([arg], restrict);
+    }
+    return {
+      argTypes: argRestricted,
+      returnType: argRestricted[0] as BaseData,
+    };
+  };
+
```
aleksanderkatan: nit: I think we could use `variadicUnifyRestrictedSignature` and rename it to `unifyRestrictedSignature` instead of this, what do you think?

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -332,38 +337,40 @@ Consider using explicit conversions instead.`,
...
-    throw new WgslTypeError(
-      `Cannot convert value of type '${
-        String(snippet.dataType)
-      }' to type '${targetDataType.type}'`,
-    );
+    if (target === dataType) {
+      return snip(value, target, origin);
+    }
+
+    if (typeof dataType === 'symbol') {
```
aleksanderkatan: I could see some data types (Void?) becoming a symbol in the future.
```suggestion
    if (dataType === UnknownData) {
```

packages/typegpu/src/std/numeric.ts
```diff
@@ -843,7 +902,22 @@ function cpuMix<T extends AnyFloatVecInstance | number>(
 
 export const mix = dualImpl({
   name: 'mix',
-  signature: variadicUnifySignature,
+  signature: (...[e1, e2, e3]) => {
```
```suggestion
  signature: (e1, e2, e3) => {
```

packages/typegpu/src/data/numeric.ts
```diff
@@ -149,9 +160,9 @@ export const i32: I32 = Object.assign(i32Cast, {
   type: 'i32',
 }) as unknown as I32;
 
-const f32Cast = dualImpl({
+const f32Cast = callableSchema({
   name: 'f32',
-  signature: (arg) => ({ argTypes: arg ? [arg] : [], returnType: f32 }),
+  signature: (arg) => ({ argTypes: [arg ? [arg] : []], returnType: f32 }),
```
cieplypolar:
```suggestion
  signature: (arg) => ({ argTypes: arg ? [arg] : [], returnType: f32 }),
```
any reason for using 2d array?

packages/typegpu/src/std/numeric.ts
```diff
@@ -843,7 +890,22 @@ function cpuMix<T extends AnyFloatVecInstance | number>(
...
+          (e1 as unknown as Vec2f).primitive,
+        ]);
+      }
+      return { argTypes: [e1, e2, uarg[0] as AnyData], returnType: e1 };
+    }
+    const uargs = unify([e1, e2, e3], anyFloat);
+    if (!uargs) {
+      throw new SignatureNotSupportedError([e1, e2, e3], anyFloat);
+    }
+    return { argTypes: uargs, returnType: uargs[0] as AnyData };
```
cieplypolar:
```suggestion
    return { argTypes: uargs, returnType: uargs[0] as BaseData };
```

packages/typegpu/src/std/operators.ts
```diff
@@ -208,13 +236,7 @@ type ModOverload = {
...
 export const mod: ModOverload = dualImpl({
   name: 'mod',
-  signature: (...args) => {
-    const uargs = unify(args) ?? args;
-    return {
-      argTypes: uargs,
-      returnType: isNumericSchema(uargs[0]) ? uargs[1] : uargs[0],
-    };
-  },
+  signature: binaryArithmeticSignature,
```
cieplypolar:
```suggestion
  signature: binaryDivSignature,
```

cieplypolar: cannot perform module operation on matrices

packages/typegpu/tests/std/texture/textureGather.test.ts
```diff
@@ -0,0 +1,109 @@
...
+      tex2d_u32: { texture: d.texture2d(d.u32) },
+      tex2d_array: { texture: d.texture2dArray(d.i32) },
+      texcube_array: { texture: d.textureCubeArray() },
+      texdepth2d: { texture: d.textureDepth2d() },
+      texdepth2d_array: { texture: d.textureDepth2dArray() },
+
+      sampler: { sampler: 'non-filtering' },
+    });
+
+    expectTypeOf(() => {});
```
cieplypolar, line 23: does it do anything?

packages/typegpu/tests/std/texture/textureGather.test.ts
```diff
@@ -0,0 +1,109 @@
...
+        uv2d,
+      );
+      const gatherdepth2d_array = textureGather(
+        testLayout.$.texdepth2d_array,
+        testLayout.$.sampler,
+        uv2d,
+        idx,
+      );
+
+      if (false) {
```
cieplypolar, line 69: same as above

packages/typegpu/tests/examples/individual/tgsl-parsing-test.test.ts
cieplypolar, line 1: After looking at this test, I found out that we don't provide concrete return type for `arrayLength`.

Image

The following change to our dualImpl should fix that
```ts
export const arrayLength = dualImpl({
  name: 'arrayLength',
  signature: (arg) => {
    const ptrArg = isPtr(arg) ? arg : ptrFn(arg as StorableData);
    return ({
      argTypes: [ptrArg],
      returnType: u32, // change 1
    });
  },
  normalImpl: (a: unknown[] | ref<unknown[]>) =>
    isRef(a) ? a.$.length : a.length,
  codegenImpl(_ctx, [a]) {
    const length = sizeOfPointedToArray(a.dataType);
    return length > 0 ? `${length}u` : stitch`arrayLength(${a})`; // change 2
  },
});
```

iwoplaza: In my opinion, the type should stay an `abstractInt` when the length is known at comp-time, which is the same as somebody writing `4` in their shader code.

iwoplaza: I do agree with the second change though. Is the branch ever hit though?

packages/typegpu/src/data/numeric.ts
reczkok: nope

packages/typegpu/tests/std/texture/textureGather.test.ts
reczkok: yes but it's cursed

PR #2127, opened 2026-01-30
feat: RuleEnhancer & unwrapped POJOs rule

aleksanderkatan:
Blocked by #2097

The rule introduced in this PR requires context - it is only relevant when the current function has a 'use gpu' directive.
I saw three potential solutions:
1. Look for the directive when relevant using the ancestor chain.
2. Hold a global context with the list of directives in currently parsed function.
3. Keep and update the list of directives locally inside rule code.

While 1. seems like the simplest solution, I didn't like it due to potential long chains of lookups (options 2. and 3. always returns answers in constant time). 
Option 2. is, to my best knowledge, impossible, though I didn't find any concrete info on this. There seems to be no built-in api like that in eslint (the context only holds static informations), and the rules are supposed to be able to run in parallel. You can even run one rule on multiple threads (the number printed out is a "global" variable).
image
Maybe a file-scoped context could exist? I found no information on this. And as long as our rules are lightweight, I don't think we should care.

So I came up with a way of "enhancing" rules with data collectors. In short, we can first define a `RuleEnhancer` -- a visitor that collects some data and exposes access to it, and merge it into a visitor. This way code is clean and composable -- we can reuse the 'use gpu' lookup in a new rule without any effort.
Lookup version (1.):
```ts
export const unwrappedPojos = createRule({
  // other arguments
  create(context) {
    return {
      ObjectExpression(node) {
        if (!hasUseGpuDirective() /* implemented somewhere, potentially slow */) {
          return;
        }
        if (node.parent.type === 'CallExpression') {
          return;
        }
        context.report({
          node,
          messageId: 'unwrappedPojo',
          data: { snippet: context.sourceCode.getText(node) },
        });
      },
    };
  }),
});
```
'Enhancing' api (3.):
```ts
export const unwrappedPojos = createRule({
  // other arguments
  create: enhanceRule({ directives: directiveTracking }, (context, state) => {
    const { directives } = state;

    return {
      ObjectExpression(node) {
        if (!directives.insideUseGpu()) {
          return;
        }
        if (node.parent.type === 'CallExpression') {
          return;
        }
        context.report({
          node,
          messageId: 'unwrappedPojo',
          data: { snippet: context.sourceCode.getText(node) },
        });
      },
    };
  }),
});

```

Edit: I found `ASTUtils`, this could probably be done just as quick (but not as clean in the rule code) using this.

### DIFF
--- a/packages/eslint-plugin/src/enhanceRule.ts
+++ b/packages/eslint-plugin/src/enhanceRule.ts
@@ -72,7 +72,6 @@ function mergeVisitors(visitors: RuleListener[]): RuleListener {
     }
 
     merged[key] = (...args: unknown[]) => {
-      // biome-ignore lint/suspicious/useIterableCallbackReturn: those functions return void
       listeners.forEach((fn) => (fn as (...args: unknown[]) => void)(...args));
     };
   }


packages/eslint-plugin/src/enhancers/directiveTracking.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    'FunctionDeclaration:exit'() {
+      stack.pop();
+    },
+    'FunctionExpression:exit'() {
+      stack.pop();
+    },
+    'ArrowFunctionExpression:exit'() {
+      stack.pop();
+    },
```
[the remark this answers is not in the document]
aleksanderkatan: not worth it

packages/eslint-plugin/tsup.config.ts
cieplypolar, line 1: we migrated to `tsdown`, you can use `pnpm dlx tsdown migrate --dry-run` for smooth transition

packages/eslint-plugin/src/enhanceRule.ts
```diff
@@ -0,0 +1,81 @@
...
+      continue;
+    }
+
+    // Reverse order if node is an exit node
+    if (key.endsWith(':exit')) {
+      listeners.reverse();
+    }
+
+    merged[key] = (...args: unknown[]) => {
+      // biome-ignore lint/suspicious/useIterableCallbackReturn: those functions return void
```
iwoplaza: No longer biome

PR #2142, opened 2026-02-03
feat: Add `@typegpu/sort` scaffolding with simple bitonic sort implementation

reczkok:
## It also eats the concurrent-sum library

### DIFF
--- a/apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
+++ b/apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
@@ -182,7 +182,7 @@ function generateRandomArray() {
   const workgroupsTotal = Math.ceil(state.arraySize / WORKGROUP_SIZE);
   const [workgroupsX, workgroupsY, workgroupsZ] = decomposeWorkgroups(workgroupsTotal);
 
-  initSeed.write(Math.random() * 1000);
+  initSeed.write(Math.random());
 
   initPipeline.with(initBindGroup).dispatchWorkgroups(workgroupsX, workgroupsY, workgroupsZ);
 
@@ -190,14 +190,7 @@ function generateRandomArray() {
 }
 
 function render() {
-  renderPipeline
-    .withColorAttachment({
-      view: context.getCurrentTexture().createView(),
-      loadOp: 'clear',
-      storeOp: 'store',
-    })
-    .with(bindGroup)
-    .draw(3);
+  renderPipeline.withColorAttachment({ view: context }).with(bindGroup).draw(3);
 }
 
 const overlay = document.getElementById('sort-overlay') as HTMLDivElement;


apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    storage: d.arrayOf(d.u32),
+    access: 'readonly',
+  },
+});
+
+const fragmentFn = tgpu['~unstable'].fragmentFn({
+  in: { uv: d.vec2f },
+  out: d.vec4f,
+})((input) => {
+  const arrayLength = renderLayout.$.data.length;
```
iwoplaza: Super nitpicky, but maybe we could store a reference to data, then do `data.length` instead of arrayLength, and `data[idx]` instead of `renderLayout.$.data[idx]`
```suggestion
  const data = renderLayout.$.data;
```

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    .createBuffer(d.arrayOf(d.u32, state.arraySize))
+    .$usage('storage');
+
+  bindGroup = root.createBindGroup(renderLayout, {
+    data: buffer,
+  });
+
+  ascendingSorter = createBitonicSorter(root, buffer);
+  descendingSorter = createBitonicSorter(root, buffer, {
+    compare: (a, b) => {
```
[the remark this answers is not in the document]
cieplypolar: even whole options object could be cached

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    }
+    return isDescending ? arr[i - 1] >= val : arr[i - 1] <= val;
+  });
+  if (!isSorted) {
+    console.error('Sort verification failed!');
+    console.log('Result:', sorted);
+  }
+}
+
+generateRandomArray();
```
cieplypolar: you don't need to call this, since all controls get executed at init with default options

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
```diff
@@ -0,0 +1,190 @@
...
+    },
+  },
+  'Sort Order': {
+    initial: 'ascending',
+    options: ['ascending', 'descending'] as const,
+    onSelectChange: (value: SortOrderKey) => {
+      state.sortOrder = value;
+    },
+  },
+  Reshuffle: { onButtonClick: () => generateRandomArray() },
```
cieplypolar:
```suggestion
  Reshuffle: { onButtonClick: generateRandomArray },
```

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
```diff
@@ -0,0 +1,190 @@
...
+  Sort: { onButtonClick: () => sort() },
```
cieplypolar:
```suggestion
  Sort: { onButtonClick: sort },
```

packages/typegpu-sort/src/bitonic/bitonicSort.ts
cieplypolar, line 1: general remark, I don't remember the maximal size of wgsl buffer, but using only 1 dimension of compute grid seems limiting. I believe, we utilized all 3 dimensions during the development of the parallel scan.

packages/typegpu-sort/src/bitonic/bitonicSort.ts
```diff
@@ -0,0 +1,301 @@
...
+  const sortWorkgroups = Math.ceil(halfSize / WORKGROUP_SIZE);
+  const padWorkgroups = Math.ceil(paddedSize / WORKGROUP_SIZE);
+  const copyBackWorkgroups = Math.ceil(originalSize / WORKGROUP_SIZE);
+
+  let totalSteps = 0;
+  for (let k = 2; k <= paddedSize; k <<= 1) {
+    for (let j = k >> 1; j > 0; j >>= 1) {
+      totalSteps++;
+    }
+  }
```
aleksanderkatan: This for makes me feel a little uneasy, and `totalSteps` is only used to check if this step is the last one. Can we check if `j === 1 && k === paddedSize` instead?

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.html
aleksanderkatan, line 1: Example thumbnail is missing

packages/typegpu-sort/src/bitonic/bitonicSort.ts
reczkok: Made it less cursed

cieplypolar (approved):
Amazing work! Love the bitwise operations within the sorting kernel. 

Left some nits.

In the future, we could optimize this by using workgroup shared memory for example.

packages/typegpu-sort/src/bitonic/bitonicSort.ts
reczkok: The dimensions do not really matter. We should get good occupancy with 256 threads in a workgroup regardless if its 256 or 16x16.

aleksanderkatan: I think that it's not about the occupancy, but about the limit imposed by `maxComputeWorkgroupSizeX` and `maxComputeWorkgroupsPerDimension` limits (which, at worst, is only 2^24, so approx 16kk)

reczkok: Great catch! Added decomposition logic for big arrays and increased the buffer sizes in examples

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/decomposeWorkgroups.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const x = Math.min(total, MAX_WORKGROUPS_PER_DIMENSION);
+  const remainingAfterX = Math.ceil(total / x);
+
+  const y = Math.min(remainingAfterX, MAX_WORKGROUPS_PER_DIMENSION);
+  const remainingAfterY = Math.ceil(remainingAfterX / y);
+
+  const z = Math.min(remainingAfterY, MAX_WORKGROUPS_PER_DIMENSION);
+
+  if (Math.ceil(total / (x * y * z)) > 1) {
```
cieplypolar:
```suggestion
  if (total / (x * y * z) > 1) {
```
I think we don't need that

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
```diff
@@ -0,0 +1,300 @@
...
+const hasTimestampQuery = root.enabledFeatures.has('timestamp-query');
+const querySet = hasTimestampQuery ? root.createQuerySet('timestamp', 2) : null;
+
+const canvas = document.querySelector('canvas') as HTMLCanvasElement;
+const context = root.configureContext({ canvas });
+
+const presentationFormat = navigator.gpu.getPreferredCanvasFormat();
+
+const maxSide = Math.floor(Math.sqrt(maxBufferSize / 4));
+const minLog = Math.log2(4);
```
cieplypolar:
```suggestion
const minLog = 2; // log_2(4)
```

packages/typegpu-sort/README.md
```diff
@@ -0,0 +1,57 @@
+<div align="center">
+
+# @typegpu/sort
+
+🚧 **Under Construction** 🚧
+
```
aleksanderkatan, line 6:
```suggestion
```

I think this may scare people off

packages/typegpu-sort/package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "@typegpu/concurrent-scan",
-  "version": "0.10.0",
-  "description": "A concurrent scan module.",
+  "name": "@typegpu/sort",
+  "version": "0.9.0",
```
aleksanderkatan: Why does this not match the typegpu version?

packages/typegpu-sort/src/bitonic/bitonicSort.ts
```diff
@@ -0,0 +1,293 @@
...
+  dst: {
+    storage: d.arrayOf(d.u32),
+    access: 'mutable',
+  },
+  params: {
+    uniform: copyParamsType,
+  },
+});
+
+const copyPadKernel = tgpu['~unstable'].computeFn({
```
aleksanderkatan:
```suggestion
const copyPadKernel = tgpu.computeFn({
```
There are more unnecessary unstables

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.html
aleksanderkatan, line 1: The fadeout does not reset on sort press

https://github.com/user-attachments/assets/d03c8128-16d1-4853-b90b-4f26c1c2b7cc

packages/typegpu-sort/src/bitonic/slots.ts
```diff
@@ -0,0 +1,8 @@
+import tgpu, { d } from 'typegpu';
+
+/** Default comparison function: ascending order (a < b means a comes before b) */
+export const defaultCompare = tgpu.fn([d.u32, d.u32], d.bool)((a, b) => a < b);
+
+/** Slot for customizing the comparison function in bitonic sort.
+ * The function should return true if the first argument should come before the second. */
+export const compareSlot = tgpu.slot<(a: number, b: number) => boolean>(defaultCompare);
```
aleksanderkatan, line 8: We could allow `std.le`, `std.ge` here just like we support `std.add` in concurrent scan (the current signatures are missing `(number, number)` args), though I didn't see a noticeable performance increase

packages/typegpu-sort/README.md
reczkok: But I think it's only fair before we have any docs

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.html
reczkok: Should be solved

packages/typegpu-sort/src/bitonic/slots.ts
reczkok: Most probably every single driver will instantly optimize the function away so I guess we are fine (would be nice for UX but I woudl argue out of scope for now since the signature is not compatible)

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.ts
reczkok: I would leave the `arrayLength` variable just so we don't call the std function many times over but I will use the reference and also remove initLength (which I realized is totally redundant)

reczkok: What a beauty 😆 

image

PR #2145, opened 2026-02-04
feat: Add `ExampleOptions` to common

aleksanderkatan:
Changes:
- made `initial` mandatory in controls,
- changed vector slider and color picker to provide tgpu vectors, 
- added `defineControls` function, that returns the input but helps with typing,
- updated all examples to use this function,
- added `astro check` command to `test:types`.


apps/typegpu-docs/src/examples/common/defineControls.ts
reczkok, line 1: I created a patched version of this file that does two things:
- makes it so select works without explicit as const each time
- restricts `VectorSliderControlParam` to vector instances (which are also tuples)

The second change I think would require a little bit of a refactor to also restrict the types internally but should work and actually produce cleaner example code (since we do some weird casting/spreading).
Here is the reference code I used for testing (very superficial)

```ts
import type { v2f, v3f, v4f } from 'typegpu/data';

type SelectControlParam<T extends readonly [string, ...string[]]> = {
  initial?: NoInfer<T[number]>;
  options: T;
  onSelectChange: (newValue: NoInfer<T[number]>) => void;
};

type ToggleControlParam = {
  initial?: boolean;
  onToggleChange: (newValue: boolean) => void;
};

type SliderControlParam = {
  initial?: number;
  min?: number;
  max?: number;
  step?: number;
  onSliderChange: (newValue: number) => void;
};

type VectorSliderControlParam<T extends v2f | v3f | v4f> = {
  initial?: T;
  min: T;
  max: T;
  step: T;
  onVectorSliderChange: (newValue: T) => void;
};

type ColorPickerControlParam = {
  initial?: readonly number[];
  onColorChange: (newValue: readonly [number, number, number]) => void;
};

type ButtonControlParam = {
  onButtonClick: (() => void) | (() => Promise<void>);
};

type TextAreaControlParam = {
  initial?: string;
  onTextChange: (newValue: string) => void;
};

export function defineControls<const T extends Record<string, unknown>>(
  controls: {
    [Key in keyof T]:
      | false // short-circuit controls
      | SelectControlParam<
        T[Key] extends readonly [string, ...string[]] ? T[Key]
          : never
      >
      | ToggleControlParam
      | SliderControlParam
      | VectorSliderControlParam<
        T[Key] extends v2f | v3f | v4f ? T[Key]
          : never
      >
      | ColorPickerControlParam
      | ButtonControlParam
      | TextAreaControlParam;
  },
) {
  return controls;
}
```

PR #2148, opened 2026-02-09
fix: Better no color attachment handling

reczkok:
Some runtimes do not handle `[null]` as no color attachment which leads to stencil example breaking


iwoplaza (approved):
Amazing change 👏

PR #2151, opened 2026-02-09
docs: Ripple cube

reczkok:


apps/typegpu-docs/src/examples/simple/ripple-cube/background.ts
aleksanderkatan, line 1: Is this intended that zoom is basically binary?

https://github.com/user-attachments/assets/5762010e-081a-4356-b92a-04b7325e77e8

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
```diff
@@ -0,0 +1,324 @@
...
+    const px = [p.x, 1 - p.x, 1 + p.x, 2 - p.x, 2 + p.x];
+    const py = [p.y, 1 - p.y, 1 + p.y, 2 - p.y, 2 + p.y];
+    const pz = [p.z, 1 - p.z, 1 + p.z, 2 - p.z, 2 + p.z];
+
+    let shellD = d.f32(1e10);
+    for (let ix = 0; ix < 5; ix++) {
+      for (let iy = 0; iy < 5; iy++) {
+        for (let iz = 0; iz < 5; iz++) {
+          const q = d.vec3f(px[ix], py[iy], pz[iz]);
+          shellD = sdf.opSmoothUnion(
```
aleksanderkatan, line 110: Right now, we set the sphere radius we want to handle, and then we we check O(r^3) spheres. The spheres that are near 0 when their radius is r, are the spheres which have their center located roughly r from (0, 0, 0). Thus, there is O(r^2) spheres that need to be checked at any particular moment. However it is really complicated to find those on the fly. I wasn't able to optimize this.

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
```diff
@@ -0,0 +1,324 @@
...
+    const cellSize = 1 / GRID_SIZE;
+    const p = d.vec3f(x, y, z).add(0.5).mul(cellSize);
+
+    const r = timeAccess.$ * 0.15;
+    const px = [p.x, 1 - p.x, 1 + p.x, 2 - p.x, 2 + p.x];
+    const py = [p.y, 1 - p.y, 1 + p.y, 2 - p.y, 2 + p.y];
+    const pz = [p.z, 1 - p.z, 1 + p.z, 2 - p.z, 2 + p.z];
+
+    let shellD = d.f32(1e10);
+    for (let ix = 0; ix < 5; ix++) {
```
aleksanderkatan: How about we include a slider for the number of ripples? My mac handles 11 fine, and the effect is really cool

9:
Image

11:
Image

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
```diff
@@ -0,0 +1,324 @@
...
+      timeScale = v;
+    },
+  },
+};
+
+export function onCleanup() {
+  cancelAnimationFrame(animationFrame);
+  cameraResult.cleanupCamera();
+  sdfTexture.destroy();
+  perlinCache.destroy();
```
aleksanderkatan: Do these two lines make a difference if we destroy root right after anyway?
```suggestion
```

apps/typegpu-docs/src/examples/simple/ripple-cube/background.ts
reczkok: No but scroll is notoriously hard to manage between mouse and touchpad

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
reczkok: Neither did I in a way that would boost performance

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
reczkok: 11 is crazy :D Will include

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
reczkok: Prolly not

apps/typegpu-docs/src/examples/simple/ripple-cube/index.ts
reczkok: 11 manages to kill performance even on the more powerful phones so if I do include a way to extend it I need add a disclaimer

PR #2153, opened 2026-02-10
feat: More lint rules

aleksanderkatan:
- No `Math` methods.
- No uninitialized variables.

I decided to delegate the "No spread operators" rule to a different PR where we create a general rule for all unsupported JS syntax

Actually, maybe "No uninitialized variables" should be a part of #2299 as well?

### DIFF
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -940,20 +940,29 @@ Try 'return ${typeStr}(${str});' instead.
       const [_, condNode, consNode, altNode] = statement;
       const condition = this._typedExpression(condNode, bool);
 
-      const consequent =
-        condition.value === false ? undefined : this._block(blockifySingleStatement(consNode));
-      const alternate =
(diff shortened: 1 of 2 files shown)


packages/eslint-plugin/src/rules/math.ts
```diff
@@ -0,0 +1,42 @@
...
+    return {
+      CallExpression(node) {
+        if (!directives.insideUseGpu()) {
+          return;
+        }
+
+        if (
+          node.callee.type === 'MemberExpression' &&
+          node.callee.object.type === 'Identifier' &&
+          node.callee.object.name === 'Math'
```
reczkok: If someone has their own "Math" object this rule will fire anyway, right? Kinda an edge case just making sure

aleksanderkatan: That's correct, we technically could check if it is provided somehow, and now that I know more about ASTUtils, it shouldn't be too difficult

aleksanderkatan: Done, now only Math without any definitions is reported

PR #2157, opened 2026-02-11
impr: Make clouds use the `texture.write()` API and move from `rgba8unorm` to `r8unorm` since only one channel was used

reczkok:
Found it randomly while searching for `for` usages.

### DIFF
--- a/apps/typegpu-docs/src/examples/rendering/clouds/index.ts
+++ b/apps/typegpu-docs/src/examples/rendering/clouds/index.ts
@@ -38,13 +38,9 @@ const resolutionUniform = root.createUniform(
   d.vec2f(canvas.width, canvas.height),
 );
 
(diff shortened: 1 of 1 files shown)


apps/typegpu-docs/src/examples/rendering/clouds/index.ts
```diff
@@ -57,16 +53,10 @@ const sampler = root['~unstable'].createSampler({
 const noiseTexture = root['~unstable']
   .createTexture({
     size: [NOISE_TEXTURE_SIZE, NOISE_TEXTURE_SIZE],
-    format: 'rgba8unorm',
+    format: 'r8unorm',
```
reczkok, line 56: I think we only ever sample `.x` but please verify

PR #2159, opened 2026-02-12
fix: checking if struct/array element is contiguous

cieplypolar:
#### Changes
- sizeOf cache was expanded to include information for contiguousness and longest prefix (and renamed)
- original method was renamed to `computeMemoryLayout`
- `sizeOf`, `isContiguous` and `getLongestContiguousPrefix` are now exported from individual files
- tests for new methods and `getOffsetInfoAt`
- some oxlint adjustments

#### Motivation
Our current approach for calculating the longest contiguous block starting at given offset doesn't work

For example
image
We should return 28 instead of 16

This PR tries to address this issue and similar ones 
EDIT: hopefully, it is successful

### DIFF
--- a/packages/typegpu/src/data/schemaMemoryLayout.ts
+++ b/packages/typegpu/src/data/schemaMemoryLayout.ts
@@ -195,8 +195,6 @@ function computeMLOfWgslArray(data: WgslArray): SchemaMemoryLayout {
   let longestContiguousPrefix: number;
   if (isContiguous) {
     longestContiguousPrefix = size;
-  } else if (hasPadding && elementMemoryLayout.isContiguous) {
-    longestContiguousPrefix = elementMemoryLayout.size;
   } else {
     longestContiguousPrefix = elementMemoryLayout.longestContiguousPrefix;
   }
@@ -218,8 +216,6 @@ function computeMLOfDisarray(data: Disarray): SchemaMemoryLayout {
   let longestContiguousPrefix: number;
   if (isContiguous) {
     longestContiguousPrefix = size;
-  } else if (hasPadding && elementMemoryLayout.isContiguous) {
-    longestContiguousPrefix = elementMemoryLayout.size;
   } else {
     longestContiguousPrefix = elementMemoryLayout.longestContiguousPrefix;
   }


packages/typegpu/src/data/isContiguous.ts
```diff
@@ -0,0 +1,14 @@
+import type { AnyData } from './dataTypes.ts';
+import type { BaseData } from './wgslTypes.ts';
+import { getLayoutInfo } from './schemaMemoryLayout.ts';
+
+export function isContiguous(schema: BaseData): boolean {
+  return getLayoutInfo(schema, 'isContiguous');
+}
+
+/**
+ * Returns `true` if data represented by the `schema` don't have padding.
```
iwoplaza:
```suggestion
 * Returns `true` if data represented by the `schema` doesn't have padding.
```

packages/typegpu/src/data/schemaMemoryLayout.ts
```diff
@@ -0,0 +1,295 @@
...
+    ? Number.NaN
+    : data.elementCount * stride;
+
+  let longestContiguousPrefix: number;
+  if (isContiguous) {
+    longestContiguousPrefix = size;
+  } else if (hasPadding && elementMemoryLayout.isContiguous) {
+    longestContiguousPrefix = elementMemoryLayout.size;
+  } else {
+    longestContiguousPrefix = elementMemoryLayout.longestContiguousPrefix;
```
aleksanderkatan: Wouldn't this also work?
```suggestion
  } else {
    longestContiguousPrefix = elementMemoryLayout.longestContiguousPrefix;
```

packages/typegpu/src/data/schemaMemoryLayout.ts
```diff
@@ -0,0 +1,295 @@
...
+
+  const size = data.elementCount * stride;
+
+  let longestContiguousPrefix: number;
+  if (isContiguous) {
+    longestContiguousPrefix = size;
+  } else if (hasPadding && elementMemoryLayout.isContiguous) {
+    longestContiguousPrefix = elementMemoryLayout.size;
+  } else {
+    longestContiguousPrefix = elementMemoryLayout.longestContiguousPrefix;
```
aleksanderkatan:
```suggestion
  } else {
    longestContiguousPrefix = elementMemoryLayout.longestContiguousPrefix;
```

PR #2164, opened 2026-02-15
impr: Show some love to concurrent sum (maintenance)

reczkok:
This PR:
- **fixes caching bug (no caching happened before) by removing the computer "mode" (full scan vs total sum) from the cache key and creating the pipelines on demand**
- improves types for the binary op function (wider now - no problems with `std` and no `TgpuFn<any>`)
- simplified the timing logic by passing the ownership of the `QuerySet` to the user (the computer just writes the timestamps)
- removed some redundant shader builtins and logic
- general code cleanup (_slightly opinionated maybe_)

### DIFF
--- a/packages/typegpu-concurrent-scan/src/concurrentScan.ts
+++ b/packages/typegpu-concurrent-scan/src/concurrentScan.ts
@@ -13,54 +13,53 @@ import {
   operatorSlot,
   scanLayout,
   uniformAddLayout,
-  workgroupSize,
+  WORKGROUP_SIZE,
 } from './schemas.ts';
 import { scanBlock } from './compute/scan.ts';
 import { uniformOp } from './compute/applySums.ts';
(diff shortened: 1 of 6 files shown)


packages/typegpu-concurrent-scan/src/concurrentScan.ts
```diff
@@ -268,9 +273,9 @@ export function scan(
     operation: BinaryOp['operation'];
     identityElement: BinaryOp['identityElement'];
   },
-  timeCallback?: (timeTgpuQuery: TgpuQuerySet<'timestamp'>) => void,
+  querySet?: TgpuQuerySet<'timestamp'>,
```
reczkok, line 276: I'm not quite sure whether this should be in options or not so I put it where the callback was (but I'm open to changing it)

iwoplaza (approved):

PR #2165, opened 2026-02-15
fix: Disallow missing varyings in shell-less fragment input

iwoplaza:
Missing properties are properly erroring out:

Screenshot 2026-02-15 at 15 46 25


apps/typegpu-docs/src/examples/rendering/perlin-noise/index.ts
reczkok, line 1: If the purpose of the accessors isn’t instructional, I would remove them in favor of fixed buffers. I know this PR didn’t add them, but it does touch the file and adds yet another layer of abstraction.
What about something like this? 
```ts
const gridSizeUniform = root.createUniform(d.f32);
const timeUniform = root.createUniform(d.f32, 0);
const sharpnessUniform = root.createUniform(d.f32, 0.1);

const renderPipelineBase = root['~unstable']
  .pipe(perlinCacheConfig.inject(dynamicLayout.$));

const createRenderPipeline = (base: WithBinding) =>
  base.createRenderPipeline({
    vertex: common.fullScreenTriangle,
    fragment: ({ uv }) => {
      'use gpu';
      const suv = mul(gridSizeUniform.$, uv);

      const n = perlin3d.sample(d.vec3f(suv, timeUniform.$));

      // Apply sharpening function
      const sharp = sharpenFnSlot.$(n, sharpnessUniform.$);

      // Map to 0-1 range
      const n01 = sharp * 0.5 + 0.5;

      // Gradient map
      const dark = d.vec3f(0, 0.2, 1);
      const light = d.vec3f(1, 0.3, 0.5);
      return d.vec4f(mix(dark, light, n01), 1);
    },
    targets: { format: presentationFormat },
  });

const renderPipelines = {
  exponential: createRenderPipeline(
    renderPipelineBase
      .with(sharpenFnSlot, exponentialSharpen),
  ),
  tanh: createRenderPipeline(
    renderPipelineBase
      .with(sharpenFnSlot, tanhSharpen),
  ),
};
```
I don't know if we currently recommend any particular patterns for pipeline construction

reczkok (approved):

PR #2166, opened 2026-02-15
feat: Add .rgba swizzles to vectors

iwoplaza:


packages/typegpu/src/data/wgslTypes.ts
aleksanderkatan, line 1: There are many missing swizzles, for example `baba`

aleksanderkatan (changes requested):
Nice!

packages/typegpu/src/data/wgslTypes.ts
iwoplaza: We're changing the approach to minimize bundle size, so this will not be an issue

PR #2168, opened 2026-02-16
feat: Stabilize pipelines and entry functions

iwoplaza:


packages/unplugin-typegpu/test/auto-naming.test.ts
```diff
@@ -546,36 +547,14 @@ describe('[ROLLUP] auto naming', () => {
...
-        import * as d from 'typegpu/data';
-
-        let accessor = ((globalThis.__TYPEGPU_AUTONAME__ ?? (a => a))(tgpu['~unstable'].accessor(d.u32), "accessor"));
-              const cst = ((globalThis.__TYPEGPU_AUTONAME__ ?? (a => a))(tgpu.const(d.u32, 1), "cst"));
-
-              console.log(accessor, shell, fn, cst);
-        "
-      `);
-  });
-
```
aleksanderkatan, line 578: I think this test was accidentally removed instead of updated, since the (outdated) babel version is still here.

aleksanderkatan (commented):
Concurrent scan still uses the old api

PR #2176, opened 2026-02-17
feat: Overload operators for vectors

iwoplaza:
Our shader generator kind of always understood how to add vectors together, it just had two limitations:
- TypeScript didn't agree
- Running the functions in JS didn't work

**Changes:**
- TypeScript (or tsover) now agrees
- Support for running `use gpu` functions that contain overloaded operators in JS
- Down-leveling `a + b` into `a.add(b)` (etc.) in examples, and showing the down-leveled code to the user
  - Showing `tsover` code will require us to do more changes to the examples platform, we can do that after the release

### DIFF
--- a/apps/typegpu-docs/scripts/transform-overloads.ts
+++ b/apps/typegpu-docs/scripts/transform-overloads.ts
@@ -29,6 +29,8 @@ const assignmentOperators = [
   ts.SyntaxKind.SlashEqualsToken,
 ];
 
+const commutativeMethods = ['add', 'mul'];
+
 async function findTypeScriptFiles(dir: string): Promise<string[]> {
   const files: string[] = [];
 
@@ -92,7 +94,7 @@ function getOverloadPattern(
   }
 
   // For non-commutative operators, use the standard library function
-  if (methodName === 'div' || methodName === 'pow') {
+  if (!commutativeMethods.includes(methodName)) {
     return 'std.op(left, right)';
   }
 


packages/unplugin-typegpu/src/rollup-impl.ts
```diff
@@ -183,6 +184,44 @@ export const rollUpImpl = (rawOptions: Options) => {
...
+                  const rhs = magicString.sliceNode(right);
+                  magicString.overwriteNode(
+                    node,
+                    `${runtimeFn}(${lhs}, ${rhs})`,
+                  );
+                }
+              }
+            },
+          });
+
```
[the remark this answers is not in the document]
iwoplaza: Shush

apps/typegpu-docs/scripts/transform-overloads.ts
```diff
@@ -0,0 +1,294 @@
...
+  const leftType = checker.getTypeAtLocation(node.left);
+  const rightType = checker.getTypeAtLocation(node.right);
+
+  const overloadType = checker.__tsover__getOverloadReturnType(
+    node.left,
+    node.operatorToken.kind,
+    node.right,
+    leftType,
+    rightType,
+  );
```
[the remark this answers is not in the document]
iwoplaza: It's unlikely that this will change, but thanks

packages/typegpu/tests/tgsl/operatorOverloads.test.ts
aleksanderkatan, line 1: Have some free tests
```ts

describe('num op', () => {
  test('num +', () => {
    const main = () => {
      'use gpu';
      let result = 1 + d.vec3f(1, 2, 3) /* comptime */;
      return result;
    };

      "fn main() -> vec3f {
        var result = vec3f(2, 3, 4);
        return result;
      }"
    `);

    expect(String(main())).toMatchInlineSnapshot(`"vec3f(2, 3, 4)"`);
  });

  test('num -', () => {
    const main = () => {
      'use gpu';
      let result = 1 - d.vec3f(1, 2, 3) /* comptime */;
      return result;
    };

      "fn main() -> vec3f {
        var result = vec3f(0, -1, -2);
        return result;
      }"
    `);

    expect(String(main())).toMatchInlineSnapshot(`"vec3f(0, -1, -2)"`);
  });

  test('num *', () => {
    const main = () => {
      'use gpu';
      let result = 2 * d.vec3f(1, 2, 3) /* comptime */;
      return result;
    };

      "fn main() -> vec3f {
        var result = vec3f(2, 4, 6);
        return result;
      }"
    `);

    expect(String(main())).toMatchInlineSnapshot(`"vec3f(2, 4, 6)"`);
  });

  test('num /', () => {
    const main = () => {
      'use gpu';
      let result = 6 / d.vec3f(2, 3, 6) /* comptime */;
      return result;
    };

      "fn main() -> vec3f {
        var result = vec3f(3, 2, 1);
        return result;
      }"
    `);

    expect(String(main())).toMatchInlineSnapshot(`"vec3f(3, 2, 1)"`);
  });
});

```

apps/typegpu-docs/scripts/transform-overloads.ts
```diff
@@ -0,0 +1,288 @@
...
+  [ts.SyntaxKind.PlusToken]: 'add',
+  [ts.SyntaxKind.PlusEqualsToken]: 'add',
+  [ts.SyntaxKind.MinusToken]: 'sub',
+  [ts.SyntaxKind.MinusEqualsToken]: 'sub',
+  [ts.SyntaxKind.AsteriskToken]: 'mul',
+  [ts.SyntaxKind.AsteriskEqualsToken]: 'mul',
+  [ts.SyntaxKind.SlashToken]: 'div',
+  [ts.SyntaxKind.SlashEqualsToken]: 'div',
+  [ts.SyntaxKind.AsteriskAsteriskToken]: 'pow',
+  [ts.SyntaxKind.AsteriskAsteriskEqualsToken]: 'pow',
```
aleksanderkatan, line 22: modulo is one of the 5 operations allowed on mixed number/vector pairs, can we add it as well to tsover?

aleksanderkatan, line 22: pow operator is here, despite not working in tsover (typescript - error, js - returns NaN, wgsl - `Unsupported data types: vec3f, abstractFloat. Supported types are: f32, f16, abstractFloat, vec2f, vec3f, vec4f, vec2h, vec3h, vec4h`)

apps/typegpu-docs/package.json
```diff
@@ -4,8 +4,9 @@
   "version": "0.0.1",
   "private": true,
   "scripts": {
-    "dev": "astro dev",
-    "build": "astro check && astro build",
+    "transform-overloads": "find . -type f -name '*.tsnotover.ts' -delete && node scripts/transform-overloads.ts",
+    "dev": "pnpm run transform-overloads && astro dev",
+    "build": "pnpm run transform-overloads && astro check && astro build",
```
aleksanderkatan, line 9: Why do we need this setup? Does the plugin or ts fork not work on stackblitz and on the example page? In what cases will tsover users need to do the same?

Hot reload no longer updates example code, though it still updates what is run.

packages/typegpu/tests/tgsl/operatorOverloads.test.ts
aleksanderkatan, line 1: Please add tests for unary `-`, like `const result = -d.vec3f(1, 2, 3);`

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -128,6 +140,8 @@ function wrapInAutoName(
 }
 
 function functionVisitor(ctx: Context): TraverseOptions {
+  let useGpuDepth = 0;
```
aleksanderkatan: In what case is it valid for nested use gpu functions to exist? If it's not, I'd just have this be boolean and throw if set to true when already being true (unless we shouldn't throw in plugins?)

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -157,71 +189,126 @@ function functionVisitor(ctx: Context): TraverseOptions {
...
+        }
+      },
     },
 
-    CallExpression(path) {
-      const node = path.node;
+    FunctionDeclaration: {
+      enter(path) {
+        if (containsUseGpuDirective(path.node)) {
+          fnNodeToOriginalMap.set(path.node, types.cloneNode(path.node, true));
```
aleksanderkatan, line 257: If we actually want to support nested use gpu, then this line worsens the pessimistic complexity by a factor of O(n)

packages/unplugin-typegpu/test/use-gpu-directive.test.ts
```diff
@@ -284,6 +284,35 @@ describe('[BABEL] "use gpu" directive', () => {
...
   });
+
+  it('transforms numeric operations', async () => {
+    const code = `\
+      import tgpu from 'typegpu';
+
+      const main = (a, b) => {
+        'use gpu';
+        let c = a + b + 2;
+        c += 2 * b;
```
aleksanderkatan, line 298: Could you add a test where `c` is an accessed property?

packages/unplugin-typegpu/test/tgsl-transpiling.test.ts
```diff
@@ -256,8 +256,8 @@ describe('[ROLLUP] plugin for transpiling tgsl functions to tinyest', () => {
                   .computeFn({ in: { num: d.builtin.numWorkgroups }, workgroupSize: [1] })((($ => (globalThis.__TYPEGPU_META__ ??= new WeakMap()).set($.f = ((input) => {
                   const tmp = counter.$.x;
                   counter.$.x = counter.$.y;
-                  counter.$.y += tmp;
-                  counter.$.z += d.f32(input.num.x);
+                  counter.$.y = __tsover_add(counter.$.y, tmp);
+                  counter.$.z = __tsover_add(counter.$.z, d.f32(input.num.x));
```
aleksanderkatan, line 260: Babel and rollup behavior is different, is this okay? Why didn't babel change it?

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -43,11 +44,22 @@ function i(identifier: string): babel.Identifier {
...
+>();
+
 function functionToTranspiled(
   node: babel.ArrowFunctionExpression | babel.FunctionExpression,
   parent: babel.Node | null,
 ): babel.CallExpression {
-  const { params, body, externalNames } = transpileFn(node);
+  const { params, body, externalNames } = transpileFn(
+    fnNodeToOriginalMap.get(node) ?? node,
+  );
```
aleksanderkatan, line 62: Why do we keep this map, and transpile the original function?

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -43,11 +44,22 @@ function i(identifier: string): babel.Identifier {
...
+  | babel.ArrowFunctionExpression
+>();
+
 function functionToTranspiled(
   node: babel.ArrowFunctionExpression | babel.FunctionExpression,
   parent: babel.Node | null,
 ): babel.CallExpression {
-  const { params, body, externalNames } = transpileFn(node);
+  const { params, body, externalNames } = transpileFn(
+    fnNodeToOriginalMap.get(node) ?? node,
```
aleksanderkatan, line 61: Is this logic included in rollup?

apps/typegpu-docs/scripts/transform-overloads.ts
```diff
@@ -0,0 +1,288 @@
...
+    )
+  ) {
+    // Not overlaoded
+    return undefined;
+  }
+
+  // For non-commutative operators, use the standard library function
+  if (methodName === 'div' || methodName === 'pow') {
+    return 'std.op(left, right)';
+  }
```
aleksanderkatan: I think we should list commutative operators instead (there is fewer of those)
This way we would omit some potential mistakes, like forgetting that `sub` is not commutative.

apps/typegpu-docs/package.json
aleksanderkatan: Oh, I just now noticed
> Showing tsover code will require us to do more changes to the examples platform, we can do that after the release

So this is all a temporary setup?

apps/typegpu-docs/scripts/transform-overloads.ts
iwoplaza: Good idea, added 🫡

iwoplaza: It's supported in tsover, though I hesitated to overload the operators on our data types. That's mostly because the there is no `**` operator in WGSL so it wasn't as straight-forward to add as the other operations. We can add that one at a later date.

apps/typegpu-docs/package.json
iwoplaza: I haven't yet found a way that would make StackBlitz work with `tsover`, so this setup will continue to be used for that. To make `tsover` work in our example view we need to create a custom build of Monaco, which I just didn't have time to do.

I think we can roll out this functionality and let people use it, while working on supporting this in our examples in parallel.

packages/typegpu/tests/tgsl/operatorOverloads.test.ts
iwoplaza: There are no unary `-` overloads yet. I'll see if I will be able to add them before the release, but I would make it a separate PR.

packages/unplugin-typegpu/src/babel.ts
iwoplaza: The map is used to tinyest-ify the original code, without the swapped operator overloads. The rollup version performs the transform in two-passes, which is why it doesn't need to remember the old version of the function's code.

packages/unplugin-typegpu/test/tgsl-transpiling.test.ts
iwoplaza: Ah, I see. Babel only transforms operations in scopes that is marked with 'use gpu', and rollup transforms shelled code as well. This is sort of a gray area, because `tsover` is only valid in 'use gpu' scopes, so rollup enables those functions to run at runtime, even though they don't pass the type checker.

apps/typegpu-docs/package.json
aleksanderkatan: Sure! Will you add an issue for this?

iwoplaza: Here it is: https://github.com/software-mansion/TypeGPU/issues/2204

PR #2188, opened 2026-02-19
impr: block externals

cieplypolar:
Changes:
- block externals (analogous to function externals)
- more elegant "for ... of ..." loop generation


packages/typegpu/src/tgsl/shaderGenerator.ts
```diff
@@ -2,10 +2,11 @@ import type { Block, Expression, Statement } from 'tinyest';
 import type { Snippet } from '../data/snippet.ts';
 import type { GenerationCtx } from './generationHelpers.ts';
 import type { BaseData } from '../data/wgslTypes.ts';
+import type { ExternalMap } from '../../src/core/resolve/externals.ts';
```
iwoplaza: Tiny nitpick, it seems like we're going out of the src dir just to get back into it
```suggestion
import type { ExternalMap } from '../core/resolve/externals.ts';
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -51,6 +52,7 @@ import type { AnyFn } from '../core/function/fnTypes.ts';
 import { arrayLength } from '../std/array.ts';
 import { AutoStruct } from '../data/autoStruct.ts';
 import { mathToStd } from './math.ts';
+import type { ExternalMap } from '../../src/core/resolve/externals.ts';
```
iwoplaza: Same here
```suggestion
import type { ExternalMap } from '../core/resolve/externals.ts';
```

packages/typegpu/src/tgsl/shaderGenerator.ts
cieplypolar: No idea how I did that 😆

PR #2196, opened 2026-02-20
feat: Defaults for targets in `.createRenderPipeline` and `.withColorAttachments`

iwoplaza:


packages/typegpu/src/core/pipeline/connectTargetsToShader.ts
```diff
@@ -1,69 +1,46 @@
...
+      result.push({
+        ...matchingTarget,
+        format: matchingTarget?.format ??
+          (presentationFormat ??= navigator.gpu.getPreferredCanvasFormat()),
+      });
     }
-
-    result.push(matchingTarget);
+    return result;
   }
```
cieplypolar, line 39:
```suggestion
if (!isColorTargetState(targets)) {
  throw new Error(
    'Expected a single color target configuration, not a record.',
  );
}
```
Why did you delete this error?

apps/typegpu-docs/src/examples/simulation/wind-map/index.ts
```diff
@@ -247,10 +247,8 @@ const draw = () => {
   pipelines.fill
     .with(bindGroup)
     .withColorAttachment({
-      view: context.getCurrentTexture().createView(),
+      view: context,
       clearValue: [1, 1, 1, 1],
```
aleksanderkatan: While we're updating the api, can we make `clearValue` accept a `d.vec4f()` as well?

packages/typegpu/src/core/pipeline/connectTargetsToShader.ts
iwoplaza: Because we can't really verify if something is a regular record, or a color target state (since all properties are optional)

PR #2211, opened 2026-02-24
docs: Refresh docs for 0.10

### LINKED ISSUE #2195: docs: Texture docs resize info error
The resize disclaimer is in the wrong place:

Image

It should be here (along with the info that the texture need to have the `render` usage for resizing to work.

Image

iwoplaza:
closes #2195


reczkok:
Nice opportunity to address #2195

PR #2221, opened 2026-02-25
chore: Migrate to oxfmt

iwoplaza:


package.json
```diff
@@ -7,11 +7,11 @@
   "scripts": {
     "dev": "DEV=true pnpm --parallel -r dev",
     "dev:host": "DEV=true pnpm --parallel -r dev --host --mode https",
-    "fix": "oxlint --fix && deno fmt",
+    "fix": "oxfmt && oxlint --fix",
```
aleksanderkatan: Won't this order result in the need to call fmt again?

.vscode/settings.json
```diff
@@ -3,20 +3,5 @@
   "[astro]": {
     "editor.defaultFormatter": "astro-build.astro-vscode"
   },
```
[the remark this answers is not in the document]
aleksanderkatan: I guess, but CI will fail anyway and there is always `nr fix`

apps/typegpu-docs/src/components/design/Snackbar.tsx
```diff
@@ -3,23 +3,23 @@ export function Snackbar(props: { text: string }) {
 
   return (
     <div
-      className='absolute right-8 bottom-8 z-40 box-border flex max-w-[min(28rem,calc(100vw-4rem))] items-center gap-4 rounded-lg bg-red-100 p-4 text-gray-500'
-      role='alert'
+      className="absolute right-8 bottom-8 z-40 box-border flex max-w-[min(28rem,calc(100vw-4rem))] items-center gap-4 rounded-lg bg-red-100 p-4 text-gray-500"
```
aleksanderkatan, line 6: Some places (components?) now use " instead of ', do we care?

packages/typegpu/src/tgsl/consoleLog/deserializers.ts
```diff
@@ -76,22 +74,39 @@ const deserializerMap: DeserializerMap = {
...
+      toF(d[0]),
+      toF(d[1]),
+      toF(d[2]),
+      toF(d[4]),
+      toF(d[5]),
+      toF(d[6]),
+      toF(d[8]),
+      toF(d[9]),
+      toF(d[10]),
     ),
```
aleksanderkatan, line 84: I don't think the `// oxfmt-ignore` works. The docs (https://oxc.rs/docs/guide/usage/formatter/ignoring) say to use `// prettier-ignore`

.oxfmtrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "$schema": "./node_modules/oxfmt/configuration_schema.json",
+  "ignorePatterns": ["**/*.mdx", "**/*.md"],
```
aleksanderkatan, line 3: Some markdowns (like packages/typegpu-concurrent-scan/README.md) were updated despite of this, though only the ts blocks were affected, so I guess this is fine, but still not what I would expect

Now that I look at it, docs were not updated, can we try and apply oxfmt to them as well? (maybe in another PR)

aleksanderkatan, line 3: Is there an option to enable this in jsdoc as well?

package.json
iwoplaza: There were cases where the formatter broke some code, hence the swap, but as you point out, this would fail if oxlint applies a fix that needs reformatting. If the formatter breaks anything, we'll catch it

apps/typegpu-docs/src/components/design/Snackbar.tsx
iwoplaza: It's completely fine

.oxfmtrc.json
iwoplaza: There seems to be bugs in oxfmt when formatting .mdx files. Perhaps we can reproduce them and report them, but for now, I opted to disable them.

iwoplaza: Not sure

packages/typegpu/src/tgsl/consoleLog/deserializers.ts
iwoplaza: They do work, I just forgot to update these

PR #2232, opened 2026-03-01
docs: Remove "experimental" tag from examples that no longer use unstable APIs

iwoplaza:


apps/typegpu-docs/src/examples/rendering/perlin-noise/meta.json
```diff
@@ -1,5 +1,5 @@
 {
   "title": "Perlin Noise",
   "category": "rendering",
-  "tags": ["experimental"]
+  "tags": []
```
aleksanderkatan:
```suggestion
  "tags": ["noise"]
```

apps/typegpu-docs/src/components/ExampleCard.tsx
```diff
@@ -22,12 +22,12 @@ export function ExampleCard({ example }: { example: Example }) {
       </div>
       <div className="p-3">
         <h3 className="mb-2 font-semibold text-black text-lg">{example.metadata.title}</h3>
-        {example.metadata.tags && example.metadata.tags.length > 0 && (
+        {((example.metadata.tags && example.metadata.tags.length > 0) || example.metadata.dev) && (
```
aleksanderkatan, line 25: This changes nothing, right?
```suggestion
        {((example.metadata.tags) || example.metadata.dev) && (
```

iwoplaza: image

aleksanderkatan: I didn't see the `<div className="flex flex-wrap gap-1">` and thought it resulted in empty html anyway. The `?? []` below is really redundant though, right?

iwoplaza: The `?? []` applies when there are no tags, but `dev` is true

aleksanderkatan: Damn thinking is hard

PR #2239, opened 2026-03-03
feat: Resolution time benchmark github action

cieplypolar:
Changes:
- github action that compares pr, target, main function resolution times
- more samples in procedural.ts
- scramble seed per sample

Example comparison chart is below in the comments section.
- because I changed warmup step in `procedural.ts` in this PR, the charts might not align, but everything should be fine after the merge

TODO:
- [x] use `std.range` in `tgpu.unroll` when merged to main
- [x] remove coping `resolution-time` app when  merged to `release` branch
  included check if app exists on release

### DIFF
--- a/apps/resolution-time/package.json
+++ b/apps/resolution-time/package.json
@@ -11,7 +11,7 @@
     "typegpu": "workspace:*"
   },
   "devDependencies": {
-    "bun": "^1.3.10",
+    "bun": "1.3.10",
     "unplugin-typegpu": "workspace:*"
   }
 }
--- a/packages/typegpu/src/core/root/rootTypes.ts
+++ b/packages/typegpu/src/core/root/rootTypes.ts
@@ -835,12 +835,18 @@ export interface TgpuRoot extends Unwrapper, WithBinding {
    */
   createBuffer<TData extends AnyData>(
     typeSchema: ValidateBufferSchema<TData>,
-    initializer: (buffer: TgpuBuffer<TData>) => void,
+    // NoInfer is there to infer the schema type just based on the first parameter
+    initializer: (buffer: TgpuBuffer<NoInfer<TData>>) => void,
   ): TgpuBuffer<TData>;
   createBuffer<TData extends AnyData>(
     typeSchema: ValidateBufferSchema<TData>,
     // NoInfer is there to infer the schema type just based on the first parameter
     initial?: InferInput<NoInfer<TData>>,
+  ): TgpuBuffer<TData>;
+  createBuffer<TData extends AnyData>(
+    typeSchema: ValidateBufferSchema<TData>,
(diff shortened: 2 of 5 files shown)


.github/workflows/resolution-time.yml
```diff
@@ -0,0 +1,183 @@
...
+            async function findBotComment(issueNumber) {
+              if (!issueNumber) return null;
+              const comments = await github.rest.issues.listComments({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                issue_number: issueNumber,
+              });
+              return comments.data.find((comment) =>
+                comment.body.includes(botCommentIdentifier)
+              );
```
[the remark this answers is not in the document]
cieplypolar: It will work, since the call will return the first 30 comments and out desired comment will be there.

.github/workflows/resolution-time.yml
```diff
@@ -0,0 +1,176 @@
...
+          ref: release
+
+      - name: Copy resolution-time app to release checkout
+        run: |
+          if [ ! -d release-branch/apps/resolution-time ]; then
+            cp -r target-branch/apps/resolution-time release-branch/apps/resolution-time
+            echo "Copied resolution-time app from target branch to release checkout"
+          else
+            echo "resolution-time app already exists in release checkout"
+          fi
```
aleksanderkatan, line 44: So, we only copy the folder if it's missing, right? Will this step be removed after next release?

aleksanderkatan (approved):
Sure! ♐⁄

.github/workflows/resolution-time.yml
cieplypolar: yes, I included this `if statement` in case I forget about it 👻

.github/workflows/resolution-time.yml
```diff
@@ -0,0 +1,176 @@
...
+        working-directory: pr-branch
+        run: pnpm install --frozen-lockfile && cd apps/resolution-time/node_modules/bun && node install.js
+
+      - name: Install dependencies (target branch)
+        working-directory: target-branch
+        run: pnpm install --frozen-lockfile && cd apps/resolution-time/node_modules/bun && node install.js
+
+      - name: Install dependencies (release branch)
+        working-directory: release-branch
+        run: pnpm install --no-frozen-lockfile && cd apps/resolution-time/node_modules/bun && node install.js
```
iwoplaza: I think we could go without the manual `node install.js` if we approved bun's postinstall script.

iwoplaza: Here's the PR that adds it:
https://github.com/software-mansion/TypeGPU/pull/2364

cieplypolar: done

PR #2267, opened 2026-03-06
docs: Accessors

iwoplaza:


apps/typegpu-docs/src/content/docs/fundamentals/accessors.mdx
```diff
@@ -0,0 +1,497 @@
...
+
+const getDoubleRed = tgpu
+  .fn(getColor)
+  .with(colorAccess, d.vec3f(1, 0, 0))
+  .with(multiplierAccess, 2);
+```
+
+```wgsl
+fn getColor() -> vec3f {
+  return vec3f(1, 0, 0) * 2f;
```
aleksanderkatan: We'd compute this at comptime, this is for illustration purposes, right?

apps/typegpu-docs/src/content/docs/fundamentals/accessors.mdx
```diff
@@ -0,0 +1,497 @@
...
+| | Accessors | Slots | Bind Groups |
+|---|---|---|---|
+| **What can be bound** | GPU resources (buffers, variables, textures, constants, functions, raw values) | Any JavaScript value | Buffers, textures, samplers |
+| **Schema-aware** | Yes — schema is declared upfront, TypeGPU validates the binding | No | Yes (layout defines types) |
+| **Field access through `.$`** | Yes — `accessor.$.field` works for struct schemas | No (a slot returning a struct doesn't support field access) | N/A |
+| **Mutation support** | `tgpu.mutableAccessor` | No | Yes (storage buffers) |
+| **When resolved** | Compile time (shader generation) | Compile time (shader generation) | Runtime (before draw/dispatch) |
+| **Primary use case** | Parameterising GPU functions over resources | Parameterising shader *logic* (callbacks, numbers, config) | Swapping resources between draw calls without recompiling |
+
+**Choose slots** when you want to inject arbitrary JavaScript values or TypeGPU functions that affect shader logic — for example, a gravity function or a boolean flag.
```
aleksanderkatan: Maybe instead we should choose slots when accessors cannot do the job?

apps/typegpu-docs/src/content/docs/fundamentals/accessors.mdx
```diff
@@ -0,0 +1,497 @@
...
+
+**Choose accessors** when the placeholder must be a GPU resource with a data schema, especially when you need deep field access or mutation, or when you want a library to expose a resource customization point.
```
aleksanderkatan: Can we focus more on the customization and code decoupling here? I don't think that deep field access is more important than that.

apps/typegpu-docs/src/content/docs/fundamentals/accessors.mdx
```diff
@@ -0,0 +1,497 @@
...
+
+**Choose bind groups** when you need to swap resources at runtime without changing the compiled shader — the classic WebGPU pattern for per-frame or per-draw-call resource binding.
+
+:::note
+Accessors and slots operate at **compile time**: the WGSL output reflects which buffer or variable was bound.
```
aleksanderkatan, line 497: Actually, why is that? What stops the following from working?
```ts
    const slot = tgpu.slot(5);
    const fn = tgpu.fn([])(() => {
      return slot.$;
    });
    console.log(fn());
```

apps/typegpu-docs/src/content/docs/fundamentals/accessors.mdx
iwoplaza: Good catch, updating it

apps/typegpu-docs/src/content/docs/fundamentals/accessors.mdx
```diff
@@ -0,0 +1,501 @@
...
+```
+
+```wgsl
+// Generated WGSL
+fn getColor() -> vec3f {
+  return vec3f(2, 0, 0);
+}
+```
+
+Notice how, because the values are known at compile time, the result is precomputed and inlined directly into the WGSL function.
```
cieplypolar, line 163:
```suggestion
:::note
Notice how, because the values are known at compile time, the result is precomputed and inlined directly into the WGSL function.
:::
```

PR #2271, opened 2026-03-09
chore: Tweaks for working with worktrees

iwoplaza:


.gitignore
```diff
@@ -31,3 +31,6 @@ yarn-error.log*
 .DS_Store
 .idea
 .husky
+
+# agents
+.claude/worktrees
```
reczkok: we could make this just .claude

PR #2276, opened 2026-03-11
fix/feat: bit operation on vectors

### LINKED ISSUE #2266: bug: unnecessary conversion to common type for `<<` and `>>` operators
This code
```ts
const repro = () => {
  'use gpu';
  const x = d.i32(8);
  return x << d.u32(2);
};

console.log(tgpu.resolve([repro]));
```
resolves to
```rust
fn repro() -> i32 {
  const x = 8i;
  return (x << 2i);
}
```
which is invalid wgsl :((

cieplypolar:
Changes:
- fixed early `convertToCommonType` call for bit operations
- added `std.bitShiftLeft/Right`
- added infix bit operators
- tests

Closes #2266.

### DIFF
--- a/packages/typegpu/src/std/operators.ts
+++ b/packages/typegpu/src/std/operators.ts
@@ -10,7 +10,6 @@ import {
   type BaseData,
   type mBaseForVec,
   type vBaseForMat,
-  type AnyUnsignedVecInstance,
   isFloat32VecInstance,
   isInteger32VecInstance,
   isMat,
@@ -18,6 +17,7 @@ import {
   isUint32VecInstance,
   isVec,
   isVecInstance,
+  vecIToVecU,
 } from '../data/wgslTypes.ts';
 import { SignatureNotSupportedError } from '../errors.ts';
 import { unify } from '../tgsl/conversion.ts';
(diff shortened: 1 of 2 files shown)


packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1528,12 +1545,18 @@ export function isMat(value: unknown): value is Mat2x2f | Mat3x3f | Mat4x4f {
...
 }
 
-export function isFloat32VecInstance(
-  element: number | AnyVecInstance | AnyMatInstance,
-): element is AnyFloat32VecInstance {
+export function isFloat32VecInstance(element: unknown): element is AnyFloat32VecInstance {
   return isVecInstance(element) && ['vec2f', 'vec3f', 'vec4f'].includes(element.kind);
 }
 
+export function isInteger32VecInstance(value: unknown): value is v2u | v2i | v3u | v3i | v4u | v4i {
```
iwoplaza, line 1552: Tiny nit
```suggestion
export function isInt32VecInstance(value: unknown): value is v2u | v2i | v3u | v3i | v4u | v4i {
```

packages/typegpu/src/data/wgslTypes.ts
```diff
@@ -1528,12 +1545,18 @@ export function isMat(value: unknown): value is Mat2x2f | Mat3x3f | Mat4x4f {
...
+  return isVecInstance(value) && /[iu]$/.test(value.kind);
+}
+
+export function isUnsignedVecInstance(value: unknown): value is v2u | v3u | v4u {
```
iwoplaza: Tiny nit
```suggestion
export function isUint32VecInstance(value: unknown): value is v2u | v3u | v4u {
```

packages/typegpu/src/std/operators.ts
```diff
@@ -279,3 +283,111 @@ export const neg = dualImpl({
...
+  }
+
+  let rhsType: BaseData;
+  if (isVec(lhsUnified)) {
+    const cc = lhsUnified.componentCount;
+    const vecU = cc === 2 ? vec2u : cc === 3 ? vec3u : vec4u;
+    const rhsUnified = unify([rhs], [u32, vecU])?.[0];
+    if (!rhsUnified) {
+      throw new SignatureNotSupportedError([rhs], [u32, vecU]);
+    }
```
aleksanderkatan, line 311: This could be simplified, right?
```suggestion
    const rhsUnified = unify([rhs], [u32, vec2u, vec3u, vec4u])?.[0];
    if (!rhsUnified) {
      throw new SignatureNotSupportedError([rhs], [u32, vec2u, vec3u, vec4u]);
    }
```

packages/typegpu/src/data/wgslTypes.ts
cieplypolar: it would be misleading, because guard include unsigned vectors

packages/typegpu/src/std/operators.ts
cieplypolar: not really, we don't want to accept rhs of different arity

PR #2280, opened 2026-03-11
chore: Transform overloads before test:types

iwoplaza:
- Matching `.tsnotover.tsx` as well as `.tsnotover.ts` files, preparing for the upcoming @typegpu/react package.
- Running transform-overloads before type tests to make sure all files are up to date locally.
- Running the `dev` script on the docs only, as no other packages have a `dev` command (they used to)


aleksanderkatan (approved):
While we're at this, can we update `tsconfig.base.json` to exclude `packages/typegpu/tsconfig.test.json` to include `tsnotover` files? VSCode's LSP keeps indexing these files

Image

Image

PR #2282, opened 2026-03-12
feat: More convenient TgpuVertexFn.AutoIn and TgpuVertexFn.AutoOut types

iwoplaza:


packages/typegpu/src/core/function/tgpuFragmentFn.ts
```diff
@@ -118,6 +119,8 @@ export declare namespace TgpuFragmentFn {
   // readable, and refactoring to use a builtin argument is too much hassle.
   type In = Record<string, BaseData>;
   type Out = Record<string, BaseData> | BaseData;
+  type AutoIn<T extends AnyAutoCustoms> = AutoFragmentIn<T>;
+  type AutoOut<T extends AnyAutoCustoms = AnyAutoCustoms> = AutoFragmentOut<T>;
```
cieplypolar, line 123: is there a reason why `out` has a default type and `in` does not?

iwoplaza: `AutoOut` is best used as a type suggestion / conformance test using `satisfies`, which `AutoIn` is meant to be used as a type annotation for the function parameter:

```ts
function foo(input: TgpuVertexFn.AutoIn<{ pos: d.v3f, normal: d.v3f }>) {
  // ...
  return { $position: position, uv, } satisfies TgpuVertexFn.AutoOut;
}

// Properly inferred from the function definition 🎉
type Foo = ReturnType<typeof foo>; // { $position: d.v4f, uv: d.v2f }
```

We want to encourage people typing `AutoIn` explicitly, while not requiring duplication when using `satisfies`. If both had explicit type parameters, then, you would have to do the following:

```ts
function foo(input: TgpuVertexFn.AutoIn<{ pos: d.v3f, normal: d.v3f }>) {
  // ...

  // Unnecessary duplication of `uv`
  return { $position: position, uv, } satisfies TgpuVertexFn.AutoOut<{ uv: d.v2f }>;
}
```

packages/typegpu/src/core/function/tgpuVertexFn.ts
```diff
@@ -83,6 +84,8 @@ export interface TgpuVertexFn<
 export declare namespace TgpuVertexFn {
   type In = BaseData | Record<string, BaseData>;
   type Out = Record<string, BaseData>;
+  type AutoIn<T> = _AutoVertexIn<T>;
```
cieplypolar, line 87:
```suggestion
  type AutoIn<T extends Record<string, InferGPU<BaseIOData> | TgpuVertexAttrib>> = _AutoVertexIn<T>;
```
I noticed that `T` here has no constraints, is there a particular reason? I checked the `confetti` example with my change, it seemed to work

packages/typegpu/src/core/function/tgpuVertexFn.ts
```diff
@@ -7,6 +7,7 @@ import { $getNameForward, $internal, $resolve } from '../../shared/symbols.ts';
 import type { Prettify } from '../../shared/utilityTypes.ts';
 import type { ResolutionCtx, SelfResolvable } from '../../types.ts';
 import { shaderStageSlot } from '../slot/internalSlots.ts';
+import type { _AutoVertexIn, AnyAutoCustoms, AutoVertexOut } from './autoIO.ts';
```
cieplypolar, line 10:
```suggestion
import type { AutoVertexInWithAttribs, AnyAutoCustoms, AutoVertexOut } from './autoIO.ts';
```
I would consider renaming it to be more descriptive at the first glance

PR #2283, opened 2026-03-12
docs: Add 3d mode to genetic racing (along with track editor)

reczkok:
I don't think I will merge it since it makes the example too complex for my liking (will probably make a separate 3d example in the future) but I think it's cool to play with so feel free


iwoplaza:
Konrad said he'd do it later, from scratch

PR #2291, opened 2026-03-13
chore: Move individual example tests to `apps/typegpu-docs`

iwoplaza:
It makes more sense for example tests to be in the `typegpu-docs`, and with vitest projects, that can become a reality 😎

### DIFF
--- a/packages/typegpu-testing-utility/package.json
+++ b/packages/typegpu-testing-utility/package.json
@@ -8,7 +8,6 @@
     "url": "git+https://github.com/software-mansion/TypeGPU.git"
   },
   "type": "module",
-  "sideEffects": false,
   "exports": "./src/index.ts",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


packages/typegpu-testing-utility/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "name": "typegpu-testing-utility",
+  "version": "0.0.0",
+  "private": true,
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/software-mansion/TypeGPU.git"
+  },
+  "type": "module",
+  "sideEffects": false,
```
aleksanderkatan:
```suggestion
```

This package has side effects, though it probably does not matter since we do not publish it.

aleksanderkatan (approved):
Great! 📚🧪

PR #2235, opened 2026-03-02
feat: Lint rule for invalid assignment

aleksanderkatan:

PR #2299, opened 2026-03-19
feat: Lint rule for unsupported js

### LINKED ISSUE #2298: feat: Lint rule for all unsupported JS syntax

aleksanderkatan:
~~Blocked by #2235 since both PRs need to update directive tracking enhancer.~~

I tried to be thorough because this may be beneficial for agents writing typegpu functions. 
I went through all possible node types, here is the current state of things.

Valid:
- ArrayExpression
- BlockStatement
- BreakStatement
- CallExpression
- ConditionalExpression
- ContinueStatement
- ForOfStatement
- ForStatement
- Identifier
- IfStatement
- MemberExpression
- ObjectExpression
- ReturnStatement
- ThisExpression
- WhileStatement

Reported:
- ArrowFunctionExpression
- AssignmentPattern
- AwaitExpression
- ChainExpression
- ClassDeclaration
- ClassExpression
- DoWhileStatement
- ForInStatement
- FunctionDeclaration
- FunctionExpression
- NewExpression
- PrivateIdentifier
- SpreadElement
- SequenceExpression
- SwitchStatement
- TemplateLiteral
- ThrowStatement
- TryStatement
- YieldExpression

Conditionally reported:
- AssignmentExpression (only if it's `!=` or other unsupported)
- BinaryExpression (only if it's `==` or other unsupported)
- Literal (only if it's regex literal)
- LogicalExpression (only if `??`)
- Property (only if it's computed, methods are already reported by function expression)
- UnaryExpression (only if unsupported)
- UpdateExpression (only if prefix)
- VariableDeclaration (only if declares multiple variables or if it's a `var`)
- VariableDeclarator (only if uses variable destructuring `const [a] = ...`, `const { a } = ...`)

Skipped because there is no need to report:
- AccessorProperty (class is already reported)
- ArrayPattern (already reported by VariableDeclarator)
- CatchClause (try is already reported)
- ClassBody (class is already reported)
- MethodDefinition (class is already reported)
- ObjectPattern (already reported by VariableDeclarator)
- PropertyDefinition (class is already reported)
- RestElement (functions are already reported)
- SwitchCase (switch is already reported)
- TaggedTemplateExpression (template literal is already reported)
- TemplateElement (template literal is already reported)

Skipped because I don't think we care:
- \<all JSX node types\>
- \<all TS node types\>
- \<all import, export node types\>
- DebuggerStatement
- Decorator
- EmptyStatement
- LabeledStatement
- Program
- StaticBlock
- Super
- WithStatement


packages/eslint-plugin/docs/rules/no-unsupported-syntax.md
```diff
@@ -0,0 +1,64 @@
...
+  const arr = [1, 2, 3];
+  for (const item of arr) {
+    a += item;
+  }
+  return a;
+} 
+```
+
+Note that it is possible that TypeGPU starts/stops supporting JS syntax from version to version.
+Make sure that the minor versions of `typegpu` and `eslint-plugin-typegpu` match.
```
cieplypolar:
```suggestion
> [!NOTE]
> Note that it is possible that TypeGPU starts/stops supporting JS syntax from version to version.
> Make sure that the minor versions of `typegpu` and `eslint-plugin-typegpu` match.
```

> [!NOTE]
> Note that it is possible that TypeGPU starts/stops supporting JS syntax from version to version.
> Make sure that the minor versions of `typegpu` and `eslint-plugin-typegpu` match.

packages/eslint-plugin/tests/rules/noUnsupportedSyntax.test.ts
```diff
@@ -0,0 +1,345 @@
...
+        code: "const fn = async () => { 'use gpu'; await foo(); }",
+        errors: [
+          {
+            messageId: 'unexpected',
+            data: { snippet: 'await foo()', syntax: 'await expression' },
+          },
+        ],
+      },
+      {
+        code: "const fn = async () => { 'use gpu'; if (1 == 2) { return true; } else { return false; } }",
```
cieplypolar:
```suggestion
        code: "const fn = (i: number) => { 'use gpu'; return i == 2; }",
```

packages/eslint-plugin/tests/rules/noUnsupportedSyntax.test.ts
```diff
@@ -0,0 +1,345 @@
...
+            data: { snippet: "new Error('x')", syntax: "'new' expression" },
+          },
+        ],
+      },
+      {
+        code: "const fn = () => { 'use gpu'; try { } catch(e) { } }",
+        errors: [
+          {
+            messageId: 'unexpected',
+            data: { snippet: 'try { } catch(e) { }', syntax: 'try-catch statement' },
```
cieplypolar, line 290: what about `finally`?

packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
```diff
@@ -0,0 +1,252 @@
...
+        }
+        report(node, 'yield expression');
+      },
+    };
+  }),
+});
+
+const unsupportedUnaryOps = ['+', 'typeof', 'void', 'delete'];
+
+const unsupportedBinaryOps = ['==', '>>>', 'in', 'instanceof', '|>'];
```
cieplypolar:
```suggestion
const unsupportedBinaryOps = ['==', '!=', '>>>', 'in', 'instanceof', '|>'];
```
I am wondering why we throw for `==` in wgsl generator, but pass `!=`

packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
cieplypolar, line 1: Plugin does not report `**=`, `??=`, `&&=`, `||=`, `>>>=`, `<<<=` and probably more

packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
```diff
@@ -0,0 +1,252 @@
...
+        node,
+        messageId: 'unexpected',
+        data: { snippet: context.sourceCode.getText(node), syntax },
+      });
+    }
+
+    return {
+      ArrowFunctionExpression(node) {
+        if (
+          directives.getDirectiveStack().length >= 2 &&
```
cieplypolar:
```suggestion
          stack.length >= 2 &&
```
I know it only returns reference, but we don't have to call it 2 times.

cieplypolar (commented):
Great feature! I left some comments that should be addressed before merge.

packages/eslint-plugin/tests/rules/noUnsupportedSyntax.test.ts
aleksanderkatan: In order for `finally` to appear, a `try` must appear first, and it will already be reported

aleksanderkatan: It's a part of try statement.
image

cieplypolar (approved):

reczkok (approved):
Cool

PR #2300, opened 2026-03-20
impr: Small improvements for future concurrent-scan work

reczkok:
use unroll, extract elements per thread (for easier testing), add benchmark, failed test logging 
I experimented with workgroup size offset to avoid bank conflicts but from my testing it seemed to improve occupancy but did not improve performance so I dropped it for now

### DIFF
--- a/packages/typegpu-sort/src/scan/compute/applySums.ts
+++ b/packages/typegpu-sort/src/scan/compute/applySums.ts
@@ -1,5 +1,11 @@
 import tgpu, { d } from 'typegpu';
-import { operatorSlot, uniformOpLayout, WORKGROUP_SIZE } from '../schemas.ts';
+import {
+  ELEMENTS_PER_THREAD,
+  ELEMENTS_RANGE,
+  operatorSlot,
+  uniformOpLayout,
+  WORKGROUP_SIZE,
+} from '../schemas.ts';
 
 export const uniformOp = tgpu.computeFn({
   workgroupSize: [WORKGROUP_SIZE],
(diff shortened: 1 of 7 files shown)


packages/typegpu-sort/src/scan/schemas.ts
```diff
@@ -1,6 +1,8 @@
 import tgpu, { d } from 'typegpu';
 
 export const WORKGROUP_SIZE = 256;
+export const ELEMENTS_PER_THREAD = 8;
+export const ELEMENTS_RANGE = Array.from({ length: ELEMENTS_PER_THREAD }, (_, i) => i);
```
aleksanderkatan: Can we use `std.range(ELEMENTS_PER_THREAD)` here?

packages/typegpu-sort/src/scan/compute/applySums.ts
```diff
@@ -10,11 +16,10 @@ export const uniformOp = tgpu.computeFn({
 })(({ gid, wid }) => {
   const globalIdx = gid.x;
   const workgroupId = wid.x;
-  const baseIdx = globalIdx * 8;
+  const baseIdx = globalIdx * ELEMENTS_PER_THREAD;
   const opValue = uniformOpLayout.$.sums[workgroupId];
 
-  // TODO: use `tgpu.unroll(8)`
-  for (let i = d.u32(0); i < 8; i++) {
+  for (const i of tgpu.unroll(ELEMENTS_RANGE)) {
```
aleksanderkatan: I think this will break when aligned to main
```suggestion
  for (const i of tgpu.unroll(std.range(ELEMENTS_RANGE))) {
```

aleksanderkatan (approved):
Sure! ◎📡

iwoplaza:
Superseded by #2756

PR #2309, opened 2026-03-24
impr: Better `withPerformanceCallback` API

reczkok:
Since I saw a lot of people call `withPerformanceCallback` per frame (which leads to too many `querySet` creations) this PR moves the actual `querySet` allocation from priors to core. It also does not throw when `timestamp-query` feature is not enabled (it becomes a no-op which I think is more comfortable) - instead it warns (to be discussed if we want that). I also fixed the terrible tests I wrote which modified the mock per test (so when one test failed the others did too)

### DIFF
--- a/packages/typegpu/tests/renderPipeline.test.ts
+++ b/packages/typegpu/tests/renderPipeline.test.ts
@@ -6,8 +6,10 @@ import tgpu, {
   common,
   d,
   MissingBindGroupsError,
+  type TgpuFragmentFn,
   type TgpuFragmentFnShell,
   type TgpuRenderPipeline,
+  type TgpuVertexFn,
   type TgpuVertexFnShell,
 } from '../src/index.js';
 import { $internal } from '../src/shared/symbols.ts';
@@ -1591,7 +1593,7 @@ describe('root.createRenderPipeline', () => {
       attribs: { vertexIndex: vertexLayout.attrib },
       vertex: ({ vertexIndex, $vertexIndex }) => {
         'use gpu';
-        return { $position: d.vec4f() };
+        return { $position: d.vec4f(vertexIndex, $vertexIndex, 0, 1) };
       },
       fragment: () => {
         'use gpu';
@@ -1619,7 +1621,7 @@ describe('root.createRenderPipeline', () => {
       },
       fragment: ({ position }) => {
         'use gpu';
-        return d.vec4f();
+        return position;
       },
       targets: { format: 'rgba8unorm' },
     });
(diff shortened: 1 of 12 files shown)


packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -1007,6 +1011,13 @@ class RenderPipelineCore implements SelfResolvable {
     return 'renderPipelineCore';
   }
 
+  get performanceCallbackQuerySet() {
+    if (!this.options.root.enabledFeatures.has('timestamp-query')) {
+      return undefined;
+    }
+    return (this.#performanceCallbackQuerySet ??= this.options.root.createQuerySet('timestamp', 2));
```
cieplypolar, line 1030: I think this is a breaking change.

Now, performance callback will always use 'the core' query set, not the one provided with `withTimestampWrites`.

Before:
Image

packages/typegpu/tests/computePipeline.test.ts
```diff
@@ -141,23 +141,24 @@ describe('TgpuComputePipeline', () => {
...
       expect(() => {
-        root.createComputePipeline({ compute: entryFn }).withPerformanceCallback(callback);
-      }).toThrow(
-        'Performance callback requires the "timestamp-query" feature to be enabled on GPU device.',
+        const before = root.createComputePipeline({ compute: entryFn });
+        const after = before.withPerformanceCallback(callback);
+        // no-op
+        expect(after).toBe(before);
+      }).not.toThrow();
+      expect(consoleWarnSpy.mock.calls[0]?.[0]).toMatchInlineSnapshot(
```
cieplypolar:
```suggestion
      expect(consoleWarnSpy).toHaveBeenCalledWith(
```
I think it is more readable

packages/typegpu/tests/renderPipeline.test.ts
```diff
@@ -520,17 +520,17 @@ describe('root.withVertex(...).withFragment(...)', () => {
...
-        'Performance callback requires the "timestamp-query" feature to be enabled on GPU device.',
+        const before = root.createRenderPipeline({
+          vertex: vertexFn,
+          fragment: fragmentFn,
+        });
+        const after = before.withPerformanceCallback(callback);
+        // no-op
+        expect(after).toBe(before);
+      }).not.toThrow();
+      expect(consoleWarnSpy.mock.calls[0]?.[0]).toMatchInlineSnapshot(
```
cieplypolar: same as above

packages/typegpu/src/core/pipeline/renderPipeline.ts
cieplypolar: Also, there is no test checking that.

packages/typegpu-testing-utility/src/extendedIt.ts
```diff
@@ -172,6 +172,7 @@ export const it = base
 
     onCleanup(() => {
       vi.unstubAllGlobals();
+      vi.restoreAllMocks();
```
cieplypolar:
```suggestion
```
The behavior has changed in Vitest 4.
https://vitest.dev/api/vi.html#vi-restoreallmocks

PR #2314, opened 2026-03-25
feat: std.range()

iwoplaza:
Adds `std.range(b, e, s)` which works at compile-time and returns an array of numbers, starting from `b`, ending at `e`, stepping by `s`. This works well with `tgpu.unroll`

### DIFF
--- a/packages/typegpu/src/std/range.ts
+++ b/packages/typegpu/src/std/range.ts
@@ -28,8 +28,14 @@ export const range = comptime<
     end = start;
     start = 0;
   }
-  if (step === 0) {
-    throw new Error('Step cannot be zero');
+  if (!Number.isInteger(start)) {
+    throw new Error(`'start' must be an integer, got ${start}`);
+  }
+  if (!Number.isInteger(end)) {
+    throw new Error(`'end' must be an integer, got ${end}`);
+  }
+  if (!Number.isInteger(step) || step === 0) {
+    throw new Error(`'step' must be a non-zero integer, got ${step}`);
   }
 
   const result: TgpuRange = [] as unknown as TgpuRange;
--- a/packages/typegpu/tests/std/range.test.ts
+++ b/packages/typegpu/tests/std/range.test.ts
@@ -40,14 +40,6 @@ test('std.range - negative step', () => {
   expect(JSON.stringify(std.range(3, -1, -1))).toMatchInlineSnapshot(`"[3,2,1,0]"`);
 });
 
-// fractional / floating-point step
-test('std.range - float step', () => {
-  expect(JSON.stringify(std.range(0, 1, 0.25))).toMatchInlineSnapshot(`"[0,0.25,0.5,0.75]"`);
-  expect(std.range(0, 1, 0.25)).toHaveLength(4);
-  expect(std.range(0, 1, 0.25)[0]).toBe(0);
-  expect(std.range(0, 1, 0.25)[3]).toBe(0.75);
-});
-
 test('std.range - returns empty array when step direction mismatches range direction', () => {
   // positive range, negative step
   expect(JSON.stringify(std.range(0, 5, -1))).toMatchInlineSnapshot(`"[]"`);
@@ -57,7 +49,15 @@ test('std.range - returns empty array when step direction mismatches range direc
 
 // error cases
 test('std.range - throws on zero step', () => {
-  expect(() => std.range(0, 5, 0)).toThrow('Step cannot be zero');
+  expect(() => std.range(0, 5, 0)).toThrowErrorMatchingInlineSnapshot(
+    `[Error: 'step' must be a non-zero integer, got 0]`,
+  );
+});
+
+test('std.range - float step', () => {
+  expect(() => std.range(0, 1, 0.25)).toThrowErrorMatchingInlineSnapshot(
+    `[Error: 'step' must be a non-zero integer, got 0.25]`,
+  );
 });
 
 describe('on the GPU', () => {


apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -308,15 +308,81 @@ import tgpu, { d } from 'typegpu';
...
+    {
+      processNeighbor(cell + vec2i(0, -1));
+    }
+    // (and so on...)
+  }
+  // (and so on ...)
+}
+```
+
+If you'd like to unroll over a range of numbers, use `tgpu.unroll(n)`.
```
aleksanderkatan: Maybe we could mimic the Python's [range](https://docs.python.org/3/library/stdtypes.html#typesseq-range) interface? It shouldn't be too difficult and will make unroll more flexible

apps/typegpu-docs/src/examples/algorithms/genetic-racing/index.ts
```diff
@@ -117,8 +117,8 @@ const senseRaycast = (pos: d.v2f, angle: number, offset: number) => {
   'use gpu';
   const dir = d.vec2f(std.cos(angle + offset), std.sin(angle + offset));
   let hitT = d.f32(1);
-  for (const step of tgpu.unroll([1, 2, 3, 4, 5, 6, 7, 8])) {
-    const t = d.f32(step / 8);
+  for (const step of tgpu.unroll(8)) {
+    const t = d.f32((step + 1) / 8);
```
aleksanderkatan: 🤔 
```suggestion
  for (const step of tgpu.unroll(1, 9)) {
    const t = d.f32((step) / 8);
```

packages/typegpu/src/data/array.ts
```diff
@@ -80,6 +80,6 @@ const WgslArrayImpl = {
   type: 'array',
 
   toString(this: WgslArray): string {
-    return `arrayOf(${this.elementType}, ${this.elementCount})`;
+    return `arrayOf(${String(this.elementType)}, ${this.elementCount})`;
```
aleksanderkatan, line 83: Does this change anything? The element type cannot be a symbol, right?
```suggestion
    return `arrayOf(${this.elementType}, ${this.elementCount})`;
```

packages/typegpu/tests/unroll.test.ts
```diff
@@ -11,6 +11,12 @@ describe('tgpu.unroll', () => {
     expect(x).toBe(arr);
   });
 
+  it('called outside the gpu function with a length, creates a 0..n-1 array', () => {
+    const x = tgpu.unroll(3);
+
+    expect(x).toEqual([0, 1, 2]);
+  });
+
```
aleksanderkatan: Do we want to consider this valid unroll usage? What was our consensus

packages/typegpu/tests/unroll.test.ts
```diff
@@ -23,16 +29,33 @@ describe('tgpu.unroll', () => {
...
+      }"
+    `);
+  });
+
+  it('called inside the gpu function but outside of forOf with length creates a 0..n-1 array', () => {
+    const f = () => {
+      'use gpu';
+      const a = tgpu.unroll(3);
+      return a[0];
+    };
```
aleksanderkatan: Users may be confused that
```ts
const f = () => {
  'use gpu';
  const a = tgpu.unroll(3);
  for (const b of a) {}
};
```
and
```ts
const f = () => {
  'use gpu';
  for (const b of tgpu.unroll(3)) {}
};
```
result in different behavior if we do not explain it clearly

packages/typegpu/src/data/array.ts
iwoplaza: It's just to silence the linter

packages/typegpu/tests/unroll.test.ts
iwoplaza: Nah, I still think we should throw. I will do that in the next PR though 🫡

packages/typegpu/src/data/array.ts
aleksanderkatan: I don't seem to have any lint issues here, but I guess it doesn't matter that much

packages/typegpu/src/core/unroll/tgpuUnroll.ts
```diff
@@ -17,23 +20,102 @@ export class UnrollableIterable implements SelfResolvable {
...
+
+  const impl = jsImpl as unknown as DualFn<typeof jsImpl> & { [$internal]: true };
 
   setName(impl, 'unroll');
   impl.toString = () => 'unroll';
   impl[$internal] = true;
   impl[$gpuCallable] = {
     call(_ctx, [value]) {
+      if (isKnownAtComptime(value) && typeof value.value === 'number') {
+        const elementCount = value.value;
```
cieplypolar:
```suggestion
        const elementCount = value.value;
        if (elementCount <= 0) {
           throw new Error('Cannot unroll loop of <= 0 iterations.')
        }
```
It will be more descriptive than `Cannot unroll loop. Length of iterable is unknown at comptime.`

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -308,15 +308,81 @@ import tgpu, { d } from 'typegpu';
...
+
+```wgsl
+// Generated WGSL
+
+fn processNeighbors(cell: vec2i) {
+  // unrolled iteration #0
+  {
+    // unrolled iteration #0
+    {
+      processNeighbor(cell + vec2i(-1, -1));
```
aleksanderkatan, line 332: (Another PR Idea) Maybe we could come up with a more general block simplifying rules, that dedent blocks containing exactly 1 ast node that does not introduce new variables (function call, if/else, another block...)

apps/typegpu-docs/src/content/docs/fundamentals/utils.mdx
```diff
@@ -308,15 +308,81 @@ import tgpu, { d } from 'typegpu';
...
+
+```ts
+// (...)
+const FBM_OCTAVES = 3;
+
+function fbm(pos: d.v3f): number {
+  'use gpu';
+  let sum = d.f32();
+
+  // i = 0, 1, 2
```
aleksanderkatan: If size is known, `[0, 1, 2]` is quicker. I'd modify this to suggest that `FBM_OCTAVES` is only known at comptime


```suggestion
const FBM_OCTAVES = 3; // user-provided value

function fbm(pos: d.v3f): number {
  'use gpu';
  let sum = d.f32();

```

packages/typegpu/tests/std/range.test.ts
```diff
@@ -0,0 +1,119 @@
...
+  expect(JSON.stringify(std.range(3, -1, -1))).toMatchInlineSnapshot(`"[3,2,1,0]"`);
+});
+
+// fractional / floating-point step
+test('std.range - float step', () => {
+  expect(JSON.stringify(std.range(0, 1, 0.25))).toMatchInlineSnapshot(`"[0,0.25,0.5,0.75]"`);
+  expect(std.range(0, 1, 0.25)).toHaveLength(4);
+  expect(std.range(0, 1, 0.25)[0]).toBe(0);
+  expect(std.range(0, 1, 0.25)[3]).toBe(0.75);
+});
```
aleksanderkatan: Python does not support this, but I guess we can!

packages/typegpu/tests/std/range.test.ts
```diff
@@ -0,0 +1,119 @@
+import { test } from 'typegpu-testing-utility';
+import { describe, expect } from 'vitest';
+import { d, std } from '../../src/index.js';
+import tgpu from '../../src/index.js';
+
+// range(n) — single argument, generates [0, n)
+test('std.range - single arg', () => {
+  expect(JSON.stringify(std.range(4))).toMatchInlineSnapshot(`"[0,1,2,3]"`);
+  expect(JSON.stringify(std.range(0))).toMatchInlineSnapshot(`"[]"`);
+  expect(JSON.stringify(std.range(1))).toMatchInlineSnapshot(`"[0]"`);
```
aleksanderkatan, line 10:
```suggestion
  expect(JSON.stringify(std.range(1))).toMatchInlineSnapshot(`"[0]"`);
  expect(JSON.stringify(std.range(-1))).toMatchInlineSnapshot(`"[]"`);
```

packages/typegpu/src/std/range.ts
```diff
@@ -0,0 +1,42 @@
+import { comptime } from '../core/function/comptime.ts';
+
+export interface TgpuRange extends Array<number> {
+  start: number;
+  end: number;
+  step: number;
+}
+
+export const range = comptime((start: number, end?: number, step: number = 1): TgpuRange => {
```
aleksanderkatan: Range is not explicitly documented in docs, and it is not in WGSL spec, so I think that we should include pretty extensive docs here, otherwise people may be hesitant to use it. I think that we should also note that this returns an array, not an iterator

packages/typegpu/src/std/range.ts
```diff
@@ -0,0 +1,42 @@
...
+      result.push(i);
+    }
+  }
+  if (step < 0) {
+    for (let i = start; i > end; i += step) {
+      result.push(i);
+    }
+  }
+  return result;
+});
```
aleksanderkatan, line 60: This way, when you write a one-argument range, the argument's name is no longer `start`
```suggestion
}) as TgpuComptime<
  ((start: number, end: number, step?: number) => TgpuRange) & ((end: number) => TgpuRange)
>;
```

PR #2315, opened 2026-03-25
docs: Use Vite v8 for lower memory consumption

iwoplaza:


apps/typegpu-docs/src/components/shaderhunt/ChallengesSignupPopover.astro
```diff
@@ -1,135 +1,246 @@
...
-          }
-        } catch (error) {
-          alert('Network error. Please try again.');
+    const popovers = document.querySelectorAll(
+        '[data-challenges-signup-popover="true"]',
+    );
+
+    class DialogBox {
+        /** @type {HTMLDialogElement} */
+        dialogElement;
```
aleksanderkatan: Did we need to get rid of typescript syntax from this file? If so, why?

iwoplaza: There is still some mismatch between how Astro uses Vite, and Vite 8, so it was just crashing during dev that it couldn't recognize TypeScript syntax.

iwoplaza: I think it would be worth it to rewrite this component in React anyway

iwoplaza: It seems that moving the logic to a separate .ts file doesn't break

apps/typegpu-docs/src/examples/exampleContent.ts
```diff
@@ -25,7 +25,7 @@ function noCacheImport<T>(importFn: () => Promise<T>): Promise<T> {
   const [url, isRelative] = extractUrlFromViteImport(importFn);
 
   if (!url) {
-    throw new Error(`Could not no-cache-import using ${importFn}`);
+    throw new Error(`Could not no-cache-import using ${String(importFn)}`);
```
aleksanderkatan: Which lint rule tells you to do this?

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -950,34 +950,33 @@ describe('wgslGenerator', () => {
...
-  //       for (var i_1 = 0u; i_1 < 4; i_1++) {
-  //         let i_2 = arr[i_1];
-  //         {
-  //           res += i32(i_2);
-  //         }
-  //       }
-  //     }"
-  //   `);
-  // });
+  it('handles "for ... of ..." loop variable name when there is shadowning', ({ root }) => {
```
aleksanderkatan:
```suggestion
  it('handles "for ... of ..." loop variable name when there is shadowing', ({ root }) => {
```

aleksanderkatan (approved):
Sure!! 🧻⬇️

PR #2342, opened 2026-04-01
fix: Color picker aligned with actual color in 'Smoky Triangle'

cieplypolar:
Changes:
- enabled extended toneMapping in the `smoky triangle` example
- added intensity slider to example controls


cieplypolar:
While reading some article I found formula for that correction. I'll leave it here to remember about it.
image

aleksanderkatan (approved):
I don't understand what's going on, but it looks good!
Also, why into main and not into release?

apps/typegpu-docs/src/components/ControlPanel.tsx
```diff
@@ -127,13 +127,17 @@ function VectorSliderRow<T extends d.v2f | d.v3f | d.v4f>({
 function ColorPickerRow({
   label,
   initial,
+  multiplier = 1,
```
aleksanderkatan: Where is this parameter set?

cieplypolar: oops, I swear I reverted that change

apps/typegpu-docs/src/examples/rendering/smoky-triangle/index.ts
```diff
@@ -73,12 +77,18 @@ const pipeline = root.pipe(perlinCache.inject()).createRenderPipeline({
...
   },
+  targets: { format: 'rgba16float' },
 });
 
 const canvas = document.querySelector('canvas') as HTMLCanvasElement;
-const context = root.configureContext({ canvas, alphaMode: 'premultiplied' });
+const context = root.configureContext({
+  canvas,
+  alphaMode: 'premultiplied',
+  format: 'rgba16float',
```
aleksanderkatan: Firefox has a regression

TypeError: GPUCanvasContext.configure: Canvas texture format `rgba16float` is not yet supported. Subscribe to <https://bugzilla.mozilla.org/show_bug.cgi?id=1834395> for updates on its development in Firefox.

iwoplaza (approved):

PR #2345, opened 2026-04-03
dx: Lint rule for imports

aleksanderkatan:
This PR introduces a new internal lint plugin, along with ~~two rules~~ one rule.

Rule `noUselessPathSegments` - mirrors [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-useless-path-segments.md) rule that is not yet implemented in [oxlint](https://oxc.rs/docs/guide/usage/linter/rules.html?sort=name&dir=asc&scope=import), it looks for unnecessary import segments like `../../src/file.txt` -> `../file.txt`.

~~Rule `noLongImports` - forbids imports starting with `../..` in examples (except `../../common`), since they are most likely an import mishap like `import { vec2f } from '../../../../../../packages/typegpu/src/data/vector.ts';` that won't work on stackblitz anyway.~~


packages/eslint-plugin-internal/tests/rules/noUselessPathSegments.test.ts
```diff
@@ -0,0 +1,66 @@
+import { describe } from 'vitest';
+import { ruleTester } from '../utils/ruleTester.ts';
+import { noUselessPathSegments } from '../../src/rules/noUselessPathSegments.ts';
+import path from 'path';
+
+const filename = path.join(process.cwd(), 'packages', 'typegpu', 'tests', 'buffer.test.ts');
```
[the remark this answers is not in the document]
aleksanderkatan: true, though this really should not matter for this test

packages/eslint-plugin-internal/src/rules/noLongImports.ts
```diff
@@ -0,0 +1,32 @@
...
+    },
+    schema: [],
+  },
+  defaultOptions: [],
+
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        const importPath = node.source.value;
+        if (importPath.startsWith('../../') && !importPath.startsWith('../../common/')) {
```
reczkok, line 22: Doesn't this introduce a pretty high chance of false positives?

aleksanderkatan: We only enable this rule on example files, where this will not work due to stackblitz pathing requirements

packages/eslint-plugin-internal/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "test": "vitest"
+  },
+  "dependencies": {
+    "@typescript-eslint/utils": "^8.57.2"
+  },
+  "devDependencies": {
+    "@types/node": "catalog:types",
+    "@typescript-eslint/rule-tester": "^8.57.2",
+    "eslint": "^9.39.2",
+    "typescript": "^5.9.3",
```
cieplypolar:
```suggestion
    "typescript": "catalog:types",
```

packages/eslint-plugin-internal/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "vitest": "^4.0.17"
```
cieplypolar:
```suggestion
    "vitest": "catalog:test"
```

packages/eslint-plugin-internal/src/rules/noLongImports.ts
cieplypolar: why not `../`?

packages/eslint-plugin-internal/src/rules/noUselessPathSegments.ts
```diff
@@ -0,0 +1,52 @@
...
+    },
+    schema: [],
+  },
+  defaultOptions: [],
+
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        const importPath = node.source.value;
+        if (!importPath.startsWith('.')) {
```
cieplypolar, line 23: we should report the case when import is only the filename without `./` prefix
for example: `import * as s from 'something.ts'`

packages/eslint-plugin-internal/tests/rules/noUselessPathSegments.test.ts
```diff
@@ -0,0 +1,66 @@
...
+
+describe('noUselessPathSegments', () => {
+  ruleTester.run('noUselessPathSegments', noUselessPathSegments, {
+    valid: [
```
cieplypolar, line 10: i would add checks for other quotes like " and maybe `

cieplypolar (changes requested):
Great work! I left some comments that should be addressed before merge.

packages/eslint-plugin-internal/tests/rules/noUselessPathSegments.test.ts
aleksanderkatan: \` is actually invalid JS, and there already was an invalid test for ", so I just added `'import item from "./file.ts";'`

packages/eslint-plugin-internal/src/rules/noUselessPathSegments.ts
aleksanderkatan: It turns out that now folders in examples are not a problem anymore. It seems to work both with Monaco and with StackBlitz, I don't know when it happened. So I opted to just remove this rule.

Edit: the mishaps also stopped happening at some point, maybe it was the change from `import * as d` to `import { d }`

cieplypolar (approved):

PR #2346, opened 2026-04-03
feat: unary operator `!` support and `std.not`

cieplypolar:
Changes:
- support for unary operator !
  - JS truthiness check
- `std.not`
  - mainly for boolean vectors
  - closely mimics WGSL
  - generally should be preferred (unary operator ! checks if number is null on GPU side)
- tests

Differences between them:
image


packages/typegpu/src/std/boolean.ts
```diff
@@ -164,19 +169,81 @@ export const ge = dualImpl({
...
+      return vec2b(!value.x, !value.y);
+    }
+    if (value.length === 3) {
+      return vec3b(!value.x, !value.y, !value.z);
+    }
+    return vec4b(!value.x, !value.y, !value.z, !value.w);
+  }
+
+  return !value;
+}
```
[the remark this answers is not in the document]
cieplypolar: it's fine

packages/typegpu/src/std/boolean.ts
```diff
@@ -164,19 +169,86 @@ export const ge = dualImpl({
...
+    const returnType = isVec(arg) ? correspondingBooleanVectorSchema(arg) : bool;
+    return {
+      argTypes: [arg],
+      returnType,
+    };
+  },
   normalImpl: cpuNot,
-  codegenImpl: (_ctx, [arg]) => stitch`!(${arg})`,
+  codegenImpl: (_ctx, [arg]) => {
+    if (isKnownAtComptime(arg)) {
```
iwoplaza: Should be unreachable, because if "arg" was known at comptime, then normalImol would have been called instead. Is that not the case?

iwoplaza (approved):
Awesome push forward! 💜🙏

packages/typegpu/src/std/boolean.ts
cieplypolar: WOW, I didn't know that `DualImpl` calls `normalImpl` if the arguments are known at comptime.  Very cool!

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -2068,4 +2069,144 @@ describe('wgslGenerator', () => {
...
+      pos: d.vec2f,
+      vel: d.vec2f,
+    });
+
+    const slot = tgpu.slot<d.Infer<typeof Boid>>({ pos: d.vec2f(), vel: d.vec2f() });
+    const accessor = tgpu.accessor(d.vec4u, d.vec4u(1, 8, 8, 2));
+
+    const f = () => {
+      'use gpu';
+      if (!!slot.$ && !!accessor.$) {
```
aleksanderkatan, line 2162: Can we use expected type or something so that there is no need to do `!!`? Right now there is no tgpu-side error when using non-boolean values in ifs

packages/typegpu/tests/std/boolean/not.test.ts
```diff
@@ -1,11 +1,158 @@
...
+      "fn f() -> i32 {
+        if (((false && true) && false)) {
+          return 1;
+        }
+        return -1;
+      }"
+    `);
+  });
+
+  it('mimics JS on non-primitive values', ({ root }) => {
```
aleksanderkatan, line 132: Isn't this test duplicated in aonther file? Was it meant to use `std.not`?

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -157,6 +157,35 @@ function operatorToType<
...
+      return snip(`!${argStr}`, bool, 'runtime');
+    }
+    if (wgsl.isNumericSchema(dataType)) {
+      const resultStr = `!bool(${argStr})`;
+      const nanGuardedStr = // abstractFloat will be resolved as comptime
+        dataType.type === 'f32'
+          ? `(((bitcast<u32>(${argStr}) & 0x7fffffff) > 0x7f800000) || ${resultStr})`
+          : dataType.type === 'f16'
+            ? `(((bitcast<u32>(${argStr}) & 0x7fff) > 0x7c00) || ${resultStr})`
+            : resultStr;
```
aleksanderkatan: What's going on here?

aleksanderkatan (approved):
Sure!

packages/typegpu/src/tgsl/wgslGenerator.ts
cieplypolar: Unfortunately, NaN is treated differently in JS and WGSL.
JS: !NaN -> true
WGSL: !NaN -> false

Looking at the IEEE754 specs I can easily detect NaNs.

f32 -> u32 bitcast is almost free

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
cieplypolar: Actually, we can. I'll do it.

packages/typegpu/src/std/boolean.ts
```diff
@@ -13,12 +13,17 @@ import {
...
+  isNumericSchema,
+  isVec,
+  isVecBool,
   isVecInstance,
   type v2b,
   type v3b,
   type v4b,
 } from '../data/wgslTypes.ts';
 import { unify } from '../tgsl/conversion.ts';
+
```
reczkok:
```suggestion
```

reczkok (approved):
Cool

PR #2355, opened 2026-04-08
impr: Depreacte the `partialWrite` API and add `patch` API that better matches the `write` features

reczkok:
This PR:
- deprecates the `partialWrite` in favor of the new `patch` API with extended capabilities 
- simplifies internal helpers (made them non-generic since we don't really need that)
- fixed redundant per write allocation (since buffers have their own internal array buffers)

The main API differences are:
- now `patch` has similar permissiveness as the new `write` API (accepts tuples and typed arrays)
- ***finally***  can pass a full array without gymnastics
- for partial array writes we now use a record with number fields (makes it so much easier to recognize and IMO cleaner to use)

We could now move the entire `patch` API to use compiled writers everywhere (due to compiled writers handling start/end offsets) but I want to get this into the 0.11 release so for now let's say out of scope - I left most internals untouched. Now we use them for leaf values (since they handle padding better) - this is to be discussed since this means creating a writer per primitive data type

Here are benchmark results (for what they're worth):
image

Important to mention that they use the old API so we are paying conversion costs for each call

### DIFF
--- a/packages/typegpu/src/data/compiledIO.ts
+++ b/packages/typegpu/src/data/compiledIO.ts
@@ -1,5 +1,4 @@
 import { roundUp } from '../mathUtils.ts';
-import type { InferInput } from '../shared/repr.ts';
 import { alignmentOf } from './alignmentOf.ts';
 import { isDisarray, isUnstruct } from './dataTypes.ts';
 import { offsetsForProps } from './offsets.ts';
@@ -17,16 +16,15 @@ export const EVAL_ALLOWED_IN_ENV: boolean = (() => {
   }
 })();
 
-const compiledWriters = new WeakMap<
-  wgsl.BaseData,
-  (
-    output: DataView,
-    offset: number,
-    value: unknown,
-    littleEndian?: boolean,
-    endOffset?: number,
-  ) => void
->();
+export type CompiledWriter = (
+  output: DataView,
+  offset: number,
+  value: unknown,
+  littleEndian?: boolean,
+  endOffset?: number,
+) => void;
+
+const compiledWriters = new WeakMap<wgsl.BaseData, CompiledWriter>();
 
 const typeToPrimitive = {
   u32: 'u32',
@@ -285,30 +283,14 @@ export function buildWriter(
   return go(node, offsetExpr, valueExpr, depth);
 }
 
-export function getCompiledWriterForSchema<T extends wgsl.BaseData>(
-  schema: T,
-):
-  | ((
-      output: DataView,
-      offset: number,
-      value: InferInput<T>,
-      littleEndian?: boolean,
-      endOffset?: number,
-    ) => void)
-  | undefined {
+export function getCompiledWriter(schema: wgsl.BaseData): CompiledWriter | undefined {
   if (!EVAL_ALLOWED_IN_ENV) {
-    console.warn('This environment does not allow eval - using default writer as fallback');
     return undefined;
   }
 
-  if (compiledWriters.has(schema)) {
-    return compiledWriters.get(schema) as (
-      output: DataView,
-      offset: number,
-      value: InferInput<T>,
-      littleEndian?: boolean,
-      endOffset?: number,
-    ) => void;
+  const cached = compiledWriters.get(schema);
+  if (cached) {
+    return cached;
   }
 
   try {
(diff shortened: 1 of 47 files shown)


packages/typegpu/tests/buffer.test.ts
```diff
@@ -566,22 +566,22 @@ describe('TgpuBuffer', () => {
     expect(rawBuffer).toBeDefined();
 
     expect(device.mock.queue.writeBuffer.mock.calls).toStrictEqual([
-      [rawBuffer, 8, new Uint8Array([255, 127, 255, 127]), 0, 4],
```
reczkok, line 569: This is the difference between floor (typed binary) and round (compiled writer) (1)

packages/typegpu/src/shared/repr.ts
```diff
@@ -44,7 +45,7 @@ export type InferInput<T> =
 
 /**
  * Extracts a sparse/partial inferred representation of a resource.
- * Used by the `buffer.writePartial` API.
+ * Used by the deprecated `buffer.writePartial` API.
```
iwoplaza: Can we also deprecate the `Partial` types?

packages/typegpu/tests/buffer.test.ts
iwoplaza: Could you elaborate?

packages/typegpu/src/data/compiledIO.ts
```diff
@@ -285,30 +283,15 @@ export function buildWriter(
...
-      value: InferInput<T>,
-      littleEndian?: boolean,
-      endOffset?: number,
-    ) => void)
-  | undefined {
+export function getCompiledWriter(schema: wgsl.BaseData): CompiledWriter | undefined {
   if (!EVAL_ALLOWED_IN_ENV) {
     console.warn('This environment does not allow eval - using default writer as fallback');
     return undefined;
   }
```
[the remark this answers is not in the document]
reczkok: Out of scope but good advice

packages/typegpu/src/data/partialIO.ts
```diff
@@ -69,63 +93,69 @@ export function getWriteInstructions<TData extends wgsl.BaseData>(
...
+    if (compiledWriter) {
+      compiledWriter(new DataView(buf), offset, value, isLittleEndian, offset + leafSize);
     } else {
-      const leafSize = sizeOf(node);
       writer.seekTo(offset);
-      writeData(writer, node, partialValue as Infer<T>);
-
-      segments.push({ start: offset, end: offset + leafSize, padding });
+      writeData(writer, node, value);
     }
```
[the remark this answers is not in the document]
reczkok: good call

packages/typegpu/src/data/partialIO.ts
```diff
@@ -69,63 +93,69 @@ export function getWriteInstructions<TData extends wgsl.BaseData>(
...
-          offset + idx * elementSize,
-          elementSize - sizeOf(arrSchema.elementType),
-        );
+      const sparse = value as Record<string, unknown>;
+      for (const key of Object.keys(sparse)) {
+        const idx = Number(key);
+        if (!Number.isNaN(idx)) {
+          collect(arrSchema.elementType, sparse[key], offset + idx * elementSize, elementPadding);
+        }
       }
```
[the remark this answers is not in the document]
reczkok: It's fine

packages/typegpu/tests/buffer.test.ts
reczkok: `0.5 * 65535 = 32767.5` -> round (what the compiled writer serializers use) gives 32768 (encoded as `unorm16 [0, 128]`) -> The old implementation must've truncated and give 32767 (encoded as `unorm16 [255 127]`)

iwoplaza (approved):
Awesome API 👏

packages/typegpu/tests/buffer.test.ts
```diff
@@ -944,6 +930,172 @@ describe('TgpuBuffer (InferInput)', () => {
...
+      items: [
+        [1, 2, 3],
+        [4, 5, 6],
+        [7, 8, 9],
+        [10, 11, 12],
+      ],
+    });
+
+    // Full replacement with TypedArray
+    arrBuf.patch({ items: new Float32Array(48) });
```
cieplypolar:
```suggestion
    arrBuf.patch({ items: new Float32Array(64) });
```

cieplypolar (approved):
Very nice, very cool :shipit:

PR #2359, opened 2026-04-10
fix: Improvement to argument usage tracking

iwoplaza:
These are mostly refactors that push the GLSL Generator towards being able to generate function declarations, but they in turn improve on a few things:
- Not emitting a struct definition for an unused input parameter.
- Not accidentally shadowing a global variable named `_arg_0`, `_arg_1`, etc.
- Not using a regex to test if an argument is being used.
- [tests] Not testing the internals of our generator system, as that makes it overly rigid, hard to evolve


packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -140,7 +144,17 @@ export function createFnCore(implementation: Implementation, fnAttribute = ''):
...
 
-        ctx.addDeclaration(`${fnAttribute}fn ${id}${header}${body}`);
+        let attributes = '';
+        if (functionType === 'compute') {
+          attributes = `@compute @workgroup_size(${workgroupSize?.join(', ')}) `;
+        } else if (functionType === 'vertex') {
+          attributes = `@vertex `;
+        } else if (functionType === 'fragment') {
+          attributes = `@fragment `;
+        }
```
aleksanderkatan, line 154: nit: we could extract this to a helper

packages/typegpu/src/data/snippet.ts
```diff
@@ -74,14 +74,9 @@ export interface Snippet {
...
 
-export interface ResolvedSnippet {
+export interface ResolvedSnippet extends Snippet {
   readonly value: string;
-  /**
-   * The type that `value` is assignable to (not necessary exactly inferred as).
-   * E.g. `1.1` is assignable to `f32`, but `1.1` itself is an abstract float
-   */
   readonly dataType: BaseData;
-  readonly origin: Origin;
```
aleksanderkatan, line 84: Maybe we could update `snip` to return ResolvedSnippet if the origin is missing? Then we could simplify e.g. the following
```ts
      return snip(
        stitch`array(${item.map((element) => this.resolve(element))})`,
        UnknownData,
        /* origin */ 'runtime',
      ) as ResolvedSnippet;
```
to
```ts
      return snip(
        stitch`array(${item.map((element) => this.resolve(element))})`,
        UnknownData,
      );
```

Actually, why does `dataType` type not match? Shouldn't this interface allow for `UnknownData`?

aleksanderkatan (commented):
Looks good!

packages/typegpu/src/data/snippet.ts
iwoplaza: It's technically not valid for an already resolved snippet to be of an unknown type, because that means it's unrepresentable in WGSL.

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -1464,27 +1049,6 @@ describe('wgslGenerator', () => {
     `);
   });
 
-  it('generates correct code for for loops with single statements', () => {
```
cieplypolar, line 1467: Did we delete this test because later on we check for...of wgsl code generation?

cieplypolar (approved):
Soo COOL! :shipit:

reczkok (approved):
Very nice

PR #2363, opened 2026-04-11
refactor: Naming cleanup

iwoplaza:
- **(🚀 feature)** Exposed more APIs that can be used by alternative shader generators (`@typegpu/gl`)
- **(🔧 fix)** Moved the pushing of the function and block scope further up, so that the arg_n parameters that we're reserving are scoped to the function.
- **(🧹 cleanup)** Made the API more clear as to what it's achieving
  - `getUniqueName` -> `makeUniqueIdentifier(..., 'global')`
  - `makeNameValid` -> `makeUniqueIdentifier(..., 'block')`
- **(🧹 cleanup)** Removed unused `namespace.on` functionality (previously used by `@typegpu/three`).
- **(🧹 cleanup)** Instead of duplicating the scope layers logic between the item state stack and name registry, the naming logic is merged into the ResolutionCtx.


packages/typegpu/src/types.ts
```diff
@@ -330,8 +332,27 @@ export interface ResolutionCtx {
...
+   * @param primer The basis for the unique identifier. Depending on the strategy, or
+   *               the names already taken, this may be modified to ensure uniqueness.
+   * @param scope The scope in which to generate the identifier. 'global' means
+   *              the identifier is meant to be unique across the entire program, while
+   *              'block' means it cannot shadow any existing identifiers visible from
+   *              within the current block. After the block is popped, any identifiers
+   *              defined within it are no longer visible.
+   * @returns an identifier that is unique within the given scope
+   */
+  makeUniqueIdentifier(primer: string | undefined, scope: 'global' | 'block'): string;
```
aleksanderkatan, line 345: I'd add a note in docs that this reserves the identifier. Actually, do we need to expose the `reserveIdentifier`? It is only used internally.

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -375,19 +402,54 @@ export class ResolutionCtxImpl implements ResolutionCtx {
...
     this.#logGenerator = opts.root ? new LogGeneratorImpl(opts.root) : new LogGeneratorNullImpl();
     this.#namespaceInternal = opts.namespace[$internal];
   }
 
-  getUniqueName(resource: object): string {
-    return getUniqueName(this.#namespaceInternal, resource);
+  isIdentifierTaken(name: string): boolean {
+    return (
+      this.#namespaceInternal.takenGlobalIdentifiers.has(name) ||
+      !!this._itemStateStack.isIdentifierTakenLocally(name)
```
aleksanderkatan:
```suggestion
      this._itemStateStack.isIdentifierTakenLocally(name)
```

aleksanderkatan (approved):
Sure!

aleksanderkatan:
I'm not sure if I like going away from `NameRegistry`, I know we duplicated interfaces, but it was more modular that way

packages/typegpu/src/types.ts
iwoplaza: `.reserveIdentifier` is used by a PR that's stacked on top of this one, #2360

iwoplaza:
> I'm not sure if I like going away from NameRegistry, I know we duplicated interfaces, but it was more modular that way

@aleksanderkatan It was more modular that way, and I wrote it with the assumption that there were going to be more ways to name things, but deciding what to name something now requires much more context than before, and we don't foresee any new naming strategy. Imo we can revisit modularizing it if we find that we need to introduce more name strategies.

reczkok (approved):
Looks good

PR #2365, opened 2026-04-13
chore: Pin bun version to decrease change of getting PWNED

iwoplaza:


aleksanderkatan (approved):
better safe than sorry!

PR #2368, opened 2026-04-14
chore: Manual publish workflow

iwoplaza:
This workflow publishes packages that have versions not yet on NPM, and that aren't marked as "private" in their package.json file.

**Changes:**
- Renamed `SKIP_ALL_CHECKS` to `SKIP_TESTS`, as now verifying the publish tag has to be explicitly disabled

**When nothing changed**:
image

**Patch bump**:
image


reczkok (approved):
Yep

PR #2369, opened 2026-04-14
chore: Disable lfs on CI

iwoplaza:
We don't use LFS anymore, seems redundant


cieplypolar (approved):
Byee

aleksanderkatan (approved):
Yay! LFS no more

reczkok (approved):

PR #2373, opened 2026-04-14
fix: Temporary std.range polyfill for resolution benchmark

iwoplaza:


iwoplaza:
Already addressed by https://github.com/software-mansion/TypeGPU/pull/2370

PR #2374, opened 2026-04-14
impr: Better plugin types

aleksanderkatan:
Not only does it simplify config, but now we can also see what rules are actually present in the preset.

image


iwoplaza (approved):
Mega cool

cieplypolar (approved):
:shipit:

PR #2377, opened 2026-04-14
fix: Add missing cool factor and remove explicit experimental tags

reczkok:
Cool factor missing -> added
`experimental` tag is no longer a thing - we auto detect that from `~unstable` usage -> removed


iwoplaza (approved):

aleksanderkatan (approved):
yep

cieplypolar (approved):
:shipit:

PR #2381, opened 2026-04-14
[READONLY] impr: More flexible bool constructor (truthiness check)

cieplypolar:
Changes:
- bool constructor mimics JS
- separate `NODE.LogicalExpr`
- automatically casts value to bool if expected type is bool
- fix NaN check for `f16`
image


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -379,6 +363,32 @@ ${this.ctx.pre}}`;
 
       const rhsExpr = this._expression(rhs);
 
+      // they are not know at comptime
```
reczkok, line 366:
```suggestion
      // they are not known at comptime
```

reczkok (approved):
Cool

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -2234,6 +2232,239 @@ describe('wgslGenerator', () => {
...
+        return res;
+      });
+
+      expect(tgpu.resolve([testFn])).toMatchInlineSnapshot(`
+        "fn testFn(n: i32, f: f32) -> i32 {
+          var res = -1;
+          if (bool(n)) {
+            res = 1i;
+          }
+          if ((((bitcast<u32>(f) & 0x7fffffff) <= 0x7f800000) && bool(f))) {
```
aleksanderkatan, line 2280: `f` appears multiple times in the code, I don't like that.
```ts
    it('does not change behavior', () => {
      const calls = tgpu.privateVar(d.u32);
      const g = tgpu.fn(
        [],
        d.f32,
      )(() => {
        'use gpu';
        calls.$ += 1;
        return 1;
      });

      const f = () => {
        'use gpu';
        if (g()) {
        }
      };

        "var<private> calls: u32;

        fn g() -> f32 {
          calls += 1u;
          return 1f;
        }

        fn f() {
          if ((((bitcast<u32>(g()) & 0x7fffffff) <= 0x7f800000) && bool(g()))) {

          }
        }"
      `);
    });
```
A fix I see is using a tgpu.fn, instead of inlining this code

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -2234,6 +2232,239 @@ describe('wgslGenerator', () => {
...
+        }
+        if (v) {
+          res = 2;
+        }
+        if (a) {
+          res = 3;
+        }
+        if (std.atomicLoad(a)) {
+          res = 4;
+        }
```
aleksanderkatan, line 2309: nit: this is primitive

aleksanderkatan (changes requested):
Nice! We just need to address the float -> bool conversion. If you think this should be a different pr, I'll approve this one

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
cieplypolar: Solved with clever trick. Thank you for your help @aleksanderkatan

iwoplaza:
We actually want the bool constructor to be less flexible

PR #2390, opened 2026-04-17
feat: Mutability tracking

### LINKED ISSUE #2356: feat: Mutability tracking for better `let` and `const` declarations
Our approach with emitting `let` and `const` in variable declarations is currently very defensive, meaning unless we can determine at the declaration site that the variable won't be reassigned or mutated, we emit `var`.

```ts
const a = 1; // It's a primitive (immutable), so: const a = 1i;
const b = d.vec3f(); // It's an object (mutable), so: var b = vec3f();
```
If we instead deferred emitting the `'var'|'let'|'const'` qualifiers until we reached the end of the function, we could analyze if the value is being mutated in any way.

```ts
function main() {
  'use gpu';
  const b = d.vec3f();
  
  b.x = 1.0; // we note that `b` was mutated
} // <- all statements are processed, now we can emit the qualifier for `b`
```

## Possible implementation
Instead of emitting `'var'|'let'|'const'`, we can emit `#VAR_0#`, `#VAR_1#` and so on, that we then replace after generating the function body.

```wgsl
fn main() {
  #VAR_0# b = vec3f();
  
  b.x = 1f;
}
```

aleksanderkatan:
Closes #2356 

We could keep and propagate info about unmodified consts (WGSL `const`) to handle cases like the below, but I think it is outside the scope of this PR 
```ts
const v = d.vec2f();
const v1 = v;
const v2 = v1;
const a = v2.x
```

I don't really like how this is implemented, but I see no better way. Maybe we could mark at implicit pointer dereference in `convertToCommonType` to avoid some of the ref/unroll calls, but this requires better origin tracking.


packages/typegpu/src/tgsl/conversion.ts
cieplypolar, line 239:
```suggestion
  return snippet;
```

cieplypolar: I know that I wrote this fragment of code, but now I think this check in unnecessary.

aleksanderkatan: I don't understand, what can be simplified? Action 'none' does not mean return as-is, but it also replaces the dataType with targetType

aleksanderkatan:
We should probably test atomics as well, atomic load takes address of its argument

edit: nvm, you cannot declare an atomic in a function

packages/typegpu/tests/mutabilityTracking.test.ts
```diff
@@ -0,0 +1,620 @@
+import { describe, expect, it } from 'vitest';
+import * as d from '../src/data/index.ts';
+import tgpu, { std } from '../src/index.js';
```
iwoplaza: Since this is a public API test, could you import from official entry-points? I would like this to be our pattern going forward.

```suggestion
import tgpu, { d, std } from 'typegpu';
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -252,7 +252,7 @@ ${this.ctx.pre}}`;
   }
 
   public blockVariable(
-    varType: 'var' | 'let' | 'const',
+    varType: 'var' | 'let' | 'const' | undefined,
```
iwoplaza: I think we should use a `'DEFERRED'` enum value, or something similar, to make the intent of this new union member more explicit.

iwoplaza (approved):
This is so good! 💜

packages/typegpu/src/tgsl/wgslGenerator.ts
reczkok:
```suggestion
    varType: 'var' | 'let' | 'const' | 'deferred',
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -287,9 +287,13 @@ ${this.ctx.pre}}`;
...
   }
 
+  /**
+   * Creates a variable declaration string.
+   * `keyword` may be a placeholder filled in later.
+   */
   protected emitVarDecl(
     pre: string,
-    keyword: 'var' | 'let' | 'const',
+    keyword: string,
```
reczkok:
```suggestion
    keyword: 'var' | 'let' | 'const' | `#VAR_${string}#`,
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1032,7 +1058,7 @@ ${this.ctx.pre}else ${alternate}`;
     }
 
     if (statement[0] === NODE.let || statement[0] === NODE.const) {
-      let varType: 'var' | 'let' | 'const' = 'var';
+      let varType: 'var' | 'let' | 'const' | undefined;
```
reczkok:
```suggestion
      let varType: 'var' | 'let' | 'const' | 'deferred' = 'deferred';
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1137,9 +1164,18 @@ ${this.ctx.pre}else ${alternate}`;
       const snippet = this.blockVariable(varType, rawId, concretize(dataType), eq.origin);
       const rhsSnippet = tryConvertSnippet(this.ctx, eq, dataType, false);
       const rhsStr = this.ctx.resolve(rhsSnippet.value, rhsSnippet.dataType).value;
+
+      let emittedVarType: string | undefined = varType;
+      if (emittedVarType === undefined) {
```
reczkok:
```suggestion
      let emittedVarType: 'var' | 'let' | 'const' | 'deferred' | `#VAR_${string}#` = varType;
      if (emittedVarType === 'deferred') {
```

reczkok (commented):
Looks, good - I would just consider increasing the type safety a little (look at commnets) since it's not too cumbersome here and will make it easier to maintain

packages/typegpu/src/tgsl/wgslGenerator.ts
reczkok: `#VAR_${string}#` could also be aliased as something easier to parse at a glance and more modifiable

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: I went with `'<deferred>'`

reczkok (approved):

PR #2394, opened 2026-04-20
fix(unplugin-typegpu): Polyfill process in unplugin-typegpu

iwoplaza:
When `unplugin-typegpu/babel` is imported in the browser, it fails because `@babel/types` expects `process` to be defined ☠️. This slight change fixes this issue.


packages/unplugin-typegpu/src/core/polyfill-process.ts
```diff
@@ -0,0 +1,8 @@
+// The babel plugin can be imported in a browser, which doesn't have `process` defined
```
aleksanderkatan, line 1: Is this our or Babel's fault? Is there any issue for this?

iwoplaza: It's expected to use `@babel/types` in node only, so this isn't really an issue for them. We want to allow the use of our plugin in the browser though (sometimes).

reczkok (approved):
Is Good

PR #2399, opened 2026-04-21
impr: Genetic Racing racing lines and small refactor

reczkok:


apps/typegpu-docs/src/examples/algorithms/genetic-racing/index.ts
```diff
@@ -1,6 +1,6 @@
-import { tgpu, common, d, std } from 'typegpu';
+import tgpu, { common, d, std } from 'typegpu';
```
iwoplaza:
```suggestion
import { tgpu, common, d, std } from 'typegpu';
```

iwoplaza (approved):
I skimmed through, but it looks good!

PR #2400, opened 2026-04-21
chore: Shaderbits

iwoplaza:
This PR is meant to remove circular dependencies between `/data` and the rest of the codebase by not allowing any imports from `/data` in other files, meaning we won't be using our public API schemas in internal code-gen.


iwoplaza:
Not worth it

PR #2404, opened 2026-04-22
feat: Better resolution error hints

### LINKED ISSUE #2367: feat: Better resolution error hints in dev
Currently, we don't have access to the original source code written by users when generating WGSL, we only have tinyest, so when we print hints about what the dev could do to fix a bug, we emit them in WGSL and hope it's similar enough to the source:

There are cases though where it falls apart:
Image

The original source where this happens is the following on line 151:

Image

Ideally we would show:
```
- fn:preInitParticle: 'particles.$[i] = particle;' is invalid, because references cannot be assigned.
----
Try 'particles.$[i] = ParticleData(particle);' to copy the value instead.
----
```

## Possible solution

Just as we transform JS into tinyest, we can transform tinyest back into JS 🚀. We can make sure that outside of Dev mode, it's treeshakeable, and we don't provide hints in production.

aleksanderkatan:
There are some things that we cannot reproduce:
- whitespaces are not present in the AST, tough luck,
- TS types aren't either,
- same with parentheses, so I put them defensively, though I skip them in some places. 
- Some nodes are just not in the tinyest yet, like function definitions.

An example of extra parentheses:
```ts
const noise = va + u.x * (vb - va) + u.y * (vc - va) + u.x * u.y * (va - vb - vc + vd);
```
becomes
```ts
const noise = ((va + (u.x * (vb - va))) + (u.y * (vc - va))) + ((u.x * u.y) * (((va - vb) - vc) + vd));
```


iwoplaza (approved):
Music to my eyes 😭. Beautiful work

PR #2407, opened 2026-04-24
impr: Better error on JS array runtime access

aleksanderkatan:


cieplypolar (approved):
Looks good

iwoplaza (approved):

packages/typegpu/src/tgsl/accessIndex.ts
aleksanderkatan:
```suggestion
import { isDisarray, MatrixColumnsAccess } from '../data/dataTypes.ts';
```

packages/typegpu/src/tgsl/accessIndex.ts
```diff
@@ -101,12 +101,16 @@ export function accessIndex(target: Snippet, indexArg: Snippet | number): Snippe
...
+    if (isKnownAtComptime(index)) {
+      // No idea what the type is, so we act on the snippet's value and try to guess
+      return coerceToSnippet(
+        // oxlint-disable-next-line typescript/no-explicit-any -- we're inspecting the value, and it could be any value
+        (target.value as any)[index.value as number],
+      );
+    } else {
+      return undefined; // this will throw an error in wgslGenerator
+    }
   }
```
iwoplaza: We can let this fallthrough to the last return, which has the same functionality:

```suggestion
  if ((isKnownAtComptime(target) && isKnownAtComptime(index))) {
    // No idea what the type is, so we act on the snippet's value and try to guess
    return coerceToSnippet(
      // oxlint-disable-next-line typescript/no-explicit-any -- we're inspecting the value, and it could be any value
      (target.value as any)[index.value as number],
    );
```

iwoplaza (approved):
Left a nit. Thanks for this fix!

packages/typegpu/src/tgsl/accessIndex.ts
aleksanderkatan: ah, of course, thanks!

PR #2411, opened 2026-04-27
feat: Expose schema read-write APIs

### LINKED ISSUE #2410: feat: Expose lower-level schema read and write APIs
Defining schemas is core to TypeGPU, and so far, we can do the following with schemas:
- Query their size (`d.sizeOf(...)`)
- Query their alignment (`d.alignmentOf(...)`)
- Plug them into a buffer or shorthand (`root.createBuffer(...)`)
- Plug them into shader definitions like variables and constants (tgpu.privateVar, tgpu.const)
- Use them in shells to constrict input types
- (among many more)

There is one functionality that I feel like is needlessly hiding beneath our `buffer.write` and `buffer.read` APIs, and that is using a schema to write our data instance objects (d.v3f, number, etc...) into an ArrayBuffer provided by the user.

```ts
import { d, writeToArrayBuffer, readFromArrayBuffer } from 'typegpu';

const Boid = d.struct({
  pos: d.vec3f,
  vel: d.vec3f,
});

const buffer = new ArrayBuffer(d.sizeOf(Boid));
writeToArrayBuffer(buffer, Boid, { pos: [1, 2, 3], vel: [0, 1, 0] }, /* { offset: ... } */);
const boid = readFromArrayBuffer(buffer, Boid);
```

My main use case right now adding the same functionality we have for our buffers to the WebGL fallback, but I can imagine this API being useful beyond that of course.

A lower-level patch utility would also be pretty useful:

```ts
import { d, patchArrayBuffer } from 'typegpu';

const Boid = d.struct({
  pos: d.vec3f,
  vel: d.vec3f,
});

const BoidsArray = d.arrayOf(Boid, 10);
const buffer = new ArrayBuffer(d.sizeOf(BoidsArray));
patchArrayBuffer(buffer, BoidsArray, { 5: { pos: [1, 2, 3], vel: [0, 1, 0] } });
```

aleksanderkatan:
Changes:
- extract `write`, `patch` and `read` functions for ArrayBuffers,
- calculate offsets in all cases (so that mapped writes are correct).


reczkok (approved):
Very cool

apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
```diff
@@ -539,6 +539,53 @@ import ListItem from '../../../components/ListItem.astro';
 </ListItem>
 :::
 
+## ArrayBuffer IO
```
iwoplaza: I would move this documentation to the `Schemas` guide, and refer to it here as in "This is what we do under the hood, if you'd like to do this yourself, you can split it up into two API calls"

iwoplaza (approved):

apps/typegpu-docs/src/content/docs/fundamentals/buffers.mdx
aleksanderkatan: I worded it differently, see if it's okay

apps/typegpu-docs/src/content/docs/apis/buffers.mdx
```diff
@@ -236,6 +236,11 @@ If you pass an unmapped buffer, the data will be written to the buffer using `GP
 If you passed your own buffer to the `root.createBuffer` function, you need to ensure it has the `GPUBufferUsage.COPY_DST` usage flag if you want to write to it using the `write` method.
 :::
 
+:::tip
+To write to, patch, or read from an `ArrayBuffer` directly, TypeGPU provides [a dedicated API](/TypeGPU/fundamentals/data-schemas/#arraybuffer-io) for that purpose.
```
iwoplaza:
```suggestion
To write to, patch, or read from an `ArrayBuffer` directly, TypeGPU provides [a dedicated API](/TypeGPU/apis/data-schemas/#arraybuffer-io) for that purpose. 
```

iwoplaza (approved):
Just change the link in the docs

PR #2412, opened 2026-04-27
impr: Use `f16` and better subgroup shader in MNIST example

reczkok:


apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.ts
aleksanderkatan, line 212: Could we also include status for f16? I don't think there is a way to know whether the shader runs on f16 or f32 at this moment

aleksanderkatan (approved):
Nice!

PR #2425, opened 2026-04-29
feat: Expand property access in externals

### LINKED ISSUE #2019: feat: Expand property access when generating externals
This change could make code written using the library more tree-shakeable (to be verified by tree-shaking tests), as well as make things like [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements) accessible in shaders.

#### Before

The shader cannot access `#buffer`, because private elements can only be accessed in the context of a method.

```ts
create() {
  const foo = () => {
    'use gpu';
    return this.#buffer.$ * 5;
  };
  // externals: () => ({ 'this': this })
}
```

The private element can be accessed easily, as it's just an object with string props of matching names.

#### After
```ts
create() {
  const foo = () => {
    'use gpu';
    return this.#buffer.$ * 5;
  };
  // externals: { 'this': { '#buffer': { '$': () => this.#buffer.$ } } }
}
```

Proof of private elements accessing through externals: [TypeScript Playground](https://www.typescriptlang.org/play/?target=99#code/GYVwdgxgLglg9mABBATgUwIZTQZQBYYAmaKAFGgB7YpgYA2OEcADmgFyIBKaTKhAPAGcoKGGADmAGkTgA1mDgB3MAD4AlB268Bw0ROlyFylYgDeAKESJ0UEDUQB5AEYArHlAB0wFHAC2AUTARGDRBUmc3aA80INFQ8ioSWgYmVjUPXwxmUlIAbVk0AE9pADd6EDQAXTVEAF4TCysrGGBEUihC1jhWsroKutraxAByUEhYBGGaxqarAHo5xABBCAhQwTFxRCg8NERmH2ZLWes0W3t8otLytFI1SuOrAF9H0-OkS+LkdCxcAmIyL1+hhBFweHA+EJgvoZGB5EpVPcANzHJ5qNQol7mCB0EGggASaDodDgZmOAGImOAoBwwCBfE4SCjjkwwLoQNAIXcybMdjBBB5KXBqXVEABGABMKOe5hZP2w3JmVlZgjgdDQHhJ4lIqEw2HwRBIpFMIz5gmGHBNwyF1ItbRq9W2eH5gqpQUQTw96OlHvMWJVUEQu2JpKGYDQikQhJDdxRwZJHl1v1jQA)

## Algorithm

During external gathering, if an unknown identifier is found, it's not immediately added to the external names pool. Instead of gathering external names, it's a recursive structure of the following type:

```ts
type ExternalNameData = Record<string, ExternalNameData>;
```

It's up to the developer to figure out an algorithm that would allow us to gather such info, but my guess would be to:
- When we've encountered an unrecognized top-level identifier, we iterate through that node's ancestors as well as the top-level node.
- If the current node is not a *valid member access* (see below), then we store an entry in the map with the value equal to the access path (e.g. `"foo.bar"`), signifying that this value is accessed on it's own, not it's children.
- If the current node is a *valid member access* (see below), and there is no entry for this access with a string value, we proceed. Otherwise we abort.
- We create an entry with the value `{ }` only if there's no entry, and continue with the next node in the chain.

## Valid member access
We only want to expand member accesses that are statically verifiable, so the access has to be:
- non-computed (so foo.bar, not foo['bar']).

## Example
Let's go through the following example and see the map step by step.

Code:
```ts
const foo = () => {
  'use gpu';
  const result = this.#buffer.$ * this.config.multiplier;
  const config = Config(this.config);
  // ...
};
```

Initial map:
```js
{}
```

After processing `this.#buffer.$`:
```js
{
  'this': {
    '#buffer': {
      '$': 'this.#buffer.$',
    },
  },
}
```

After processing `this.config.multiplier`:
```js
{
  'this': {
    '#buffer': {
      '$': 'this.#buffer.$',
    },
    'config': {
      'multiplier': 'this.config.multiplier',
    },
  }
}
```

After processing `Config(this.config)` (notice we remove the access to 'multiplier', as we need access to the whole config object):
```js
{
  'this': {
    '#buffer': {
      '$': 'this.#buffer.$',
    },
    'config': 'this.config',
  }
}
```

### LINKED ISSUE #2504: chore: Remove externalNames from unplugin
It is now a part of #2425 

Blocked by #2502, #2505
`externalNames` make transformed function code contain redundancy, since it can always be retrieved via Object.keys from externals.

```ts
  const metadata = `{
    v: ${FORMAT_VERSION},
    ast: ${embedJSON(ast)}, // <- redundancy here
      .join(', ')}}),
  }`;
```

aleksanderkatan:
Closes #2504
Closes #2019

See the new doc for description on how now externals work. 

The cut-off at '.$' actually changes some behavior: with .$ in externals, number slots resolve to abstract numbers instead of concrete.


aleksanderkatan:
FLAT

aleksanderkatan:
@pullfrog review

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,261 @@
...
+
+There are two important reasons why we need to delay the external computation and cannot just generate entries like `"ext.p1.p2": ext.p1.p2`.
+
+The first reason is externals recapture:
+```ts
+let buffer;
+let pipeline;
+const computeFn = tgpu.computeFn({ workgroupSize: [1] })(() => { 
+  'use gpu'; buffer.$ += 1; 
+}); // { buffer: buffer } would now become { buffer: undefined }
```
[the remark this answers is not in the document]
aleksanderkatan: No, this is an explanation of why we do not do that. With a thunk, this would work.

aleksanderkatan:
@pullfrog review

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -198,7 +198,7 @@ class ItemStateStackImpl implements ItemStateStack {
 
         const external = layer.externalMap[id];
         if (isNamable(external) && getName(external) === undefined) {
-          setName(external, id);
+          setName(external, id.split('.').at(-1) as string);
```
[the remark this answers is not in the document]
aleksanderkatan: Yeah this is what I meant

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+- auto-naming (`globalThis.__TYPEGPU_AUTONAME__`),
+- function metadata assignment (`globalThis.__TYPEGPU_META__`),
+- operator overloading (`__tsover_add` and similar).
+
+## Auto-naming
+
+Most TypeGPU resources can be named via the `$name` method.
+While this is completely optional, a name can make debugging easier, and it also serves as a primer for identifiers in resulting WGSL.
+
+Unplugin-typegpu traverses the AST of each bundled file, and matches most common patterns where a default name for a resource (recognized by constructor name) can be inferred.
```
iwoplaza:
```suggestion
`unplugin-typegpu` traverses the AST of each bundled file, and matches most common patterns where a default name for a resource (recognized by constructor name) can be inferred.
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+const createBuffer = () => {
+  return root.createBuffer(d.u32);
+}
+```
+:::
+
+
+## Function metadata assignment
+
+Function metadata is what makes it possible to parse JS functions into WGSL.
```
iwoplaza:
```suggestion
Function metadata is what makes it possible to generate WGSL from JavaScript code. 
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+}
+```
+:::
+
+
+## Function metadata assignment
+
+Function metadata is what makes it possible to parse JS functions into WGSL. 
+The metadata consists of the metadata version, the function name, the AST ([abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree)), and externals (the captured scope of the function),
+Only functions marked with `'use gpu'` are processed.
```
iwoplaza:
```suggestion
Only functions marked with `'use gpu'` are processed, as well as those immediately passed into `tgpu.fn` shells. 
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+### AST
+
+The plugin uses the [tinyest-for-wgsl](https://github.com/software-mansion/TypeGPU/blob/main/packages/tinyest-for-wgsl/README.md) 
+package to parse the function AST into our custom AST format called [tinyest](https://github.com/software-mansion/TypeGPU/blob/main/packages/tinyest/README.md).
+Our `wgslGenerator` traverses this AST and generates code snippets.
+
+This could theoretically be done in runtime from the function's `toString`, but it would require bundling a JavaScript parser with TypeGPU, and catching externals would be really difficult.
```
iwoplaza:
```suggestion
This could theoretically be done at runtime from the function's `toString`, but it would require bundling a JavaScript parser with TypeGPU, and make catching externals impossible.
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+buffer = root.createMutable(d.u32); // externals do not update
+```
+The same scenario may occur when a function references resources that are recreated (e.g. a resized buffer).
+
+The second reason is the fact that some of our getters depend on the current resolution mode.
+`const.$` returns the value of the const when called outside of resolution, but during resolution it returns a different object with `[$ownSnippet]` and `[$resolve]` symbols.
+
+#### Step 3: runtime
+
+In runtime, the thunks are changed to getters for convenience.
```
iwoplaza:
```suggestion
At runtime, the thunks are changed to getters for convenience.
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+```
+
+This is the reason why in unplugin we prepend each external with `() =>` 
+instead of just prepending the entire object - this lets us only evaluate the externals that were actually used.
+
+:::
+
+## Operator overloading
+
+For full [tsover](https://github.com/software-mansion/tsover/tree/main) experience, inside the function code, overloaded operators are swapped with calls to functions like `__tsover_add`, defined by [tsover-runtime](https://github.com/software-mansion/tsover/blob/main/packages/tsover-runtime/README.md).
```
iwoplaza:
```suggestion
To allow [tsover](https://github.com/software-mansion/tsover/tree/main) to work as we expect when the function is called in JavaScript, operators used inside the function body are swapped with calls to functions like `__tsover_add`, defined by [tsover-runtime](https://github.com/software-mansion/tsover/blob/main/packages/tsover-runtime/README.md).
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+This makes it possible to add vectors together without referencing `std.mul` and such.
+
+:::note
+This change occurs only in the actual code.
+Generated tinyest metadata is not modified.
+:::
+
+:::caution
+Unplugin only transforms operators inside of 'use gpu' functions.
```
iwoplaza:
```suggestion
`unplugin-typegpu` only transforms operators inside of 'use gpu' functions.
```

packages/unplugin-typegpu/src/babel.ts
```diff
@@ -9,13 +9,35 @@ import {
...
+        .reduce<t.Expression>((obj, prop) => t.memberExpression(obj, t.identifier(prop)), base);
+
+      return t.objectProperty(
+        t.stringLiteral(key),
+        t.arrowFunctionExpression([], propAccess),
+        false,
+      );
+    }),
+  );
+}
```
iwoplaza, line 39: Let's just remove `externalNames` from v2 of metadata

packages/tinyest-for-wgsl/src/externals.ts
```diff
@@ -0,0 +1,50 @@
...
+  return ctx.stack.some((scope) => scope.declaredNames.includes(name));
+}
+
+/**
+ * Keeps the set of nodes visited by `tryFindExternalChain`.
+ * This helps optimize code like `ext().x.y.z.t`:
+ * instead of traversing chains `.x.y.z.t`, `.x.y.z`, `.x.y` and `.x`,
+ * we only traverse the first one and then return early.
+ */
+const visitedMap: WeakMap<Context, Set<JsNode>> = new WeakMap();
```
iwoplaza: Why is this map a WeakMap with Context as the key, and not a property on the Context itself?

iwoplaza (approved):
Looks great! 🕺🪩
Just had one question/nitpick to resolve before merging

packages/tinyest-for-wgsl/src/externals.ts
aleksanderkatan: No real reason, I just thought that this would separate code better since it is not used elsewhere, and it's just an optimization of `tryFindExternalChain`. I can put it on the context so that it's less confusing, especially since `ignoreExternalDepth` is already there

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+let pipeline;
+const computeFn = tgpu.computeFn({ workgroupSize: [1] })(() => { 
+  'use gpu'; buffer.$ += 1; 
+}); // { buffer: buffer } would now become { buffer: undefined }
+buffer = root.createMutable(d.u32); // externals do not update
+```
+The same scenario may occur when a function references resources that are recreated (e.g. a resized buffer).
+
+The second reason is the fact that some of our getters depend on the current resolution mode.
+`const.$` returns the value of the const when called outside of resolution, but during resolution it returns a different object with `[$ownSnippet]` and `[$resolve]` symbols.
```
cieplypolar:
```suggestion
`tgpu.const` `.$` call returns its value when called outside of resolution, but during resolution it returns a different object with `[$ownSnippet]` and `[$resolve]` symbols.
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+Also, during resolution, deciding whether a given object comes from externals is difficult. Getters let us forget about this problem.
+Without this change, the `wgslGenerator` would throw when encountering a function not marked with 'use gpu'.
+
+When the `wgslGenerator` cannot find an identifier in scope, 
+it accesses the externals, which return the correct code snippets and automatically add any used definitions to the context, 
+as this is the behavior of resource prop access during resolution.
+
+:::caution[Fragility of external prop access]
+There are cases where external prop access performed in during resolution modifies the context.
```
cieplypolar:
```suggestion
There are cases where external prop access performed during resolution modifies the context.
```

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+
+This is the reason why in unplugin we prepend each external with `() =>` 
+instead of just prepending the entire object - this lets us only evaluate the externals that were actually used.
+
+:::
+
+## Operator overloading
+
+To allow [tsover](https://github.com/software-mansion/tsover/tree/main) to work as we expect when the function is called in JavaScript, operators used inside the function body are swapped with calls to functions like `__tsover_add`, defined by [tsover-runtime](https://github.com/software-mansion/tsover/blob/main/packages/tsover-runtime/README.md).
+This makes it possible to add vectors together without referencing `std.mul` and such.
```
cieplypolar:
```suggestion
This makes it possible to add vectors together without referencing `std.add` and such.
```

packages/typegpu/src/resolutionCtx.ts
cieplypolar:
```suggestion
          setName(external, id.replace(/\./g, '_'));
```
wouldn't it be easier to debug if we have whole chain?

packages/typegpu/tests/tgsl/shellless.test.ts
```diff
@@ -472,7 +472,7 @@ describe('shellless', () => {
   });
 
   it('throws a descriptive error when an accessor argument is not dereferenced', () => {
-    const myAccess = tgpu.accessor(d.f32);
+    const myAccess = tgpu.accessor(d.f32, 2);
```
cieplypolar, line 475: Did your changes cause error that accessor doesn't have a default value?

cieplypolar (commented):
❤️ ❤️

packages/typegpu/tests/tgsl/shellless.test.ts
aleksanderkatan: At some point they did (not now though), these tests threw because slot was not filled, so I left them like this

packages/typegpu/src/resolutionCtx.ts
aleksanderkatan: Maybe. We can't do this for all externals though, `EXT[0].buffer` will still be named `buffer`

cieplypolar: I see

cieplypolar (approved):

apps/typegpu-docs/src/content/docs/advanced/explaining-the-magic.mdx
```diff
@@ -0,0 +1,312 @@
...
+  return root.createBuffer(d.u32);
+}
+```
+:::
+
+
+## Function metadata assignment
+
+Function metadata is what makes it possible to generate WGSL from JavaScript code. 
+The metadata consists of the metadata version, the function name, the AST ([abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree)), and externals (the captured scope of the function),
```
reczkok:
```suggestion
The metadata consists of the metadata version, the function name, the AST ([abstract syntax tree](https://en.wikipedia.org/wiki/Abstract_syntax_tree)), and externals (the captured scope of the function).
```

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -2,28 +2,11 @@ import type * as babel from '@babel/types';
 import type * as acorn from 'acorn';
 import * as tinyest from 'tinyest';
 import { FuncParameterType } from 'tinyest';
+import type { Context, JsNode, TranspilationResult } from './types';
```
reczkok:
```suggestion
import type { Context, JsNode, TranspilationResult } from './types.ts';
```

packages/unplugin-typegpu/src/core/factory.ts
```diff
@@ -32,20 +32,22 @@ function embedJSON(jsValue: unknown) {
     .replace(/\u2029/g, '\\u2029');
 }
 
+function externalsToString(externals: Externals): string {
+  const entries = Array.from(externals, (key) => `"${key}": () => ${key}`);
+  return `{ ${entries.join(', ')} }`;
```
reczkok: nit: Could we kill the spaces in empty cases?

packages/tinyest-for-wgsl/src/externals.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Checks if the provided node is an external chain access.
+ * @example
+ * tryFindExternalChain(ctx, node`ext`); // 'ext'
+ * tryFindExternalChain(ctx, node`ext.p.q`); // 'ext.p.q'
+ * tryFindExternalChain(ctx, node`ext.p.q().r`); // undefined
+ * tryFindExternalChain(ctx, node`local.p.q`); // undefined
+ * tryFindExternalChain(ctx, node`ext.$.q`); // undefined
+ */
+export function tryFindExternalChain(ctx: Context, node: JsNode): string | undefined {
```
reczkok, line 16: For most nodes this will go all the way though right? Couldn't we invert the logic so unless one of the nodes (identifier, this, member) we don't even add it to the Set?

reczkok (approved):
Looks good, great docs! Left some nits

packages/tinyest-for-wgsl/src/externals.ts
aleksanderkatan: The function only traverses down when it actually may be an external chain. See some of my debug below. Do you just want to optimize how many nodes we keep in this set? 



Calling `tryFindExternalChain`...
  Visiting node type: BlockStatement
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: VariableDeclaration
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: MemberExpression
  Visiting node type: Identifier
Found external chain: ext.value

Calling `tryFindExternalChain`...
  Visiting node type: VariableDeclaration
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: MemberExpression
  Visiting node type: MemberExpression
  Visiting node type: Identifier
Found external chain: ext.config.multiplier

Calling `tryFindExternalChain`...
  Visiting node type: VariableDeclaration
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: MemberExpression
  Visiting node type: MemberExpression
  Visiting node type: Identifier
Found external chain: ext.config.zero

Calling `tryFindExternalChain`...
  Visiting node type: VariableDeclaration
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: MemberExpression
  Visiting node type: MemberExpression
  Visiting node type: Identifier
Found external chain: ext.config.multiplier

Calling `tryFindExternalChain`...
  Visiting node type: BlockStatement
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: VariableDeclaration
No external chain found.

Calling `tryFindExternalChain`...
  Visiting node type: MemberExpression
  Visiting node type: Identifier
Found external chain: ext.value

aleksanderkatan: Actually, we only travel down member expressions, so it would suffice to just keep those (at least until we update externals again)

packages/unplugin-typegpu/src/core/factory.ts
aleksanderkatan: The rest of metadata uses no spaces (`ast: {"params":[],"body":[0,[]]},`), so I just killed them altogether for consistency

packages/tinyest-for-wgsl/src/types.ts
```diff
@@ -0,0 +1,37 @@
...
+  externalNames: Externals;
+  /** Used to signal to identifiers that they should not treat their resolution as possible external uses. */
+  ignoreExternalDepth: number;
+  /**
+   * Keeps the set of nodes visited by `tryFindExternalChain`.
+   * This helps optimize code like `ext().x.y.z.t`:
+   * instead of traversing chains `.x.y.z.t`, `.x.y.z`, `.x.y` and `.x`,
+   * we only traverse the first one and then return early.
+   */
+  visitedNodes: Set<babel.MemberExpression | acorn.MemberExpression>;
```
iwoplaza, line 23: Could this be a `WeakSet` instead to not leak memory?

```suggestion
  visitedNodes: WeakSet<babel.MemberExpression | acorn.MemberExpression>;
```

aleksanderkatan: How does it leak memory? It's no longer global

aleksanderkatan:
@aleksanderkatan we need to add a note that typescript types are completely ignored in our magic

PR #2426, opened 2026-04-29
fix: Keep internal accessor slot unnamed, even after code transforms

iwoplaza:
It seems that the package bundling process caused the import rename to be reverted back to `slot`, so we wrap it in an IIFE instead, which achieves the same effect


aleksanderkatan (approved):
Sure! 🎰🏷️

reczkok (approved):

cieplypolar (approved):

PR #2438, opened 2026-05-04
impr: `console` and `Math` method reference checks

aleksanderkatan:
Prerequisite for #2425

This PR removes the `=== console` and `=== Math` checks from `wgslGenerator.ts`, since they will fail when we expand externals to e.g. `{ 'console' : { 'log': /* getter */ console.log } }`


iwoplaza (approved):
Great change, thanks! 🙌

cieplypolar (approved):
:shipit:

PR #2439, opened 2026-05-04
impr: Infix dispatch handling

aleksanderkatan:
Prerequisite for #2425

This PR changes the way infix dispach is handled in preparation for externals like `{ 'vec' : { 'mul': /* getter */ vec.mul } }`

### DIFF
--- a/packages/typegpu/src/core/function/comptime.ts
+++ b/packages/typegpu/src/core/function/comptime.ts
@@ -2,7 +2,7 @@ import { WgslTypeError } from '../../errors.ts';
 import { setName, type TgpuNamable } from '../../shared/meta.ts';
 import { $getNameForward, $gpuCallable, $internal } from '../../shared/symbols.ts';
 import { coerceToSnippet } from '../../tgsl/generationHelpers.ts';
-import { type DualFn, isKnownAtComptime } from '../../types.ts';
+import { type DualFn, isKnownAtComptime, NormalState } from '../../types.ts';
 
 type AnyFn = (...args: never[]) => unknown;
 
@@ -48,7 +48,7 @@ export function comptime<T extends (...args: never[]) => unknown>(func: T): Tgpu
   impl.toString = () => 'comptime';
   impl[$getNameForward] = func;
   impl[$gpuCallable] = {
-    call(_ctx, args) {
+    call(ctx, args) {
       if (!args.every((s) => isKnownAtComptime(s))) {
         throw new WgslTypeError(
           `Called comptime function with runtime-known values: ${args
@@ -58,7 +58,12 @@ export function comptime<T extends (...args: never[]) => unknown>(func: T): Tgpu
         );
       }
 
-      return coerceToSnippet(func(...(args.map((s) => s.value) as never[])));
+      ctx.pushMode(new NormalState());
+      try {
+        return coerceToSnippet(func(...(args.map((s) => s.value) as never[])));
+      } finally {
+        ctx.popMode();
+      }
     },
   };
   impl.$name = (label: string) => {
--- a/packages/typegpu/src/data/index.ts
+++ b/packages/typegpu/src/data/index.ts
@@ -5,48 +5,20 @@
 // NOTE: This is a barrel file, internal files should not import things from this file
 
 import { Operator } from 'tsover-runtime';
-import { type InfixOperator, infixOperators } from '../tgsl/accessProp.ts';
 import { MatBase } from './matrix.ts';
 import { VecBase } from './vectorImpl.ts';
-
-function assignInfixOperator<T extends typeof VecBase | typeof MatBase>(
-  object: T,
-  operator: InfixOperator,
-  operatorSymbol: symbol,
-) {
-  // oxlint-disable-next-line typescript/no-explicit-any -- anything is possible
-  const proto = object.prototype as any;
-  const opImpl = infixOperators[operator] as (lhs: unknown, rhs: unknown) => unknown;
-
-  proto[operator] = function (this: unknown, other: unknown): unknown {
-    return opImpl(this, other);
-  };
-
-  proto[operatorSymbol] = (lhs: unknown, rhs: unknown): unknown => {
-    return opImpl(lhs, rhs);
-  };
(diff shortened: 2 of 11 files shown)


packages/typegpu/src/tgsl/accessProp.ts
```diff
@@ -191,15 +182,12 @@ export function accessProp(target: Snippet, propName: string): Snippet | undefin
...
-      // Not a valid swizzle
-      return undefined;
-    }
-
+  if (
+    isVec(target.dataType) &&
+    propLength >= 1 &&
+    propLength <= 4 &&
+    /^[xyzw]+$|^[rgba]+$/.test(propName)
+  ) {
```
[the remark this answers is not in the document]
aleksanderkatan: idc

packages/typegpu/src/data/index.ts
```diff
@@ -5,48 +5,39 @@
...
+    value: opImpl,
+  });
 
-  proto[operatorSymbol] = (lhs: unknown, rhs: unknown): unknown => {
-    return opImpl(lhs, rhs);
-  };
+  Object.defineProperty(base.prototype, operator, {
+    get() {
+      return infixDispatch(this, opImpl);
+    },
```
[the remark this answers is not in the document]
aleksanderkatan: I don't think it's avoidable, and vectors are already not performant in JS anyways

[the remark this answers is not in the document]
iwoplaza: We could *technically* branch on whether or not we're generating a shader, return what it is currently when we are, otherwise return a function that is the same across instances and actually uses 'this'

packages/typegpu/tests/tgsl/infixOperators.test.ts
```diff
@@ -70,115 +70,191 @@ describe('wgslGenerator', () => {
...
   });
 
-  it('precomputes adds on known values', () => {
-    const testFn = tgpu.fn([])(() => {
-      const v1 = d.vec3f(1, 2, 3).add(5);
-      const v2 = d.vec3f(1, 2, 3).add(d.vec3f(3, 2, 1));
-    });
+  it('works when application is deferred', ({ root }) => {
+    const v = d.vec2f(1, 2).mul;
+    const u = tgpu.comptime(() => d.vec2f(3, 4).mul);
```
iwoplaza: I don't think that's something we want to support, at least not officially.

aleksanderkatan: I wanted to test whether `buffer.$.mul` would work, and I see no other way than to use comptime.
This is more or less what will happen with new externals.

packages/typegpu/src/data/index.ts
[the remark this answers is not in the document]
iwoplaza:
```ts
function jsInfix(this: unknown, arg) {
  return opImpl(this, arg);
}

Object.defineProperty(base.prototype, operator, {
    get() {
      if (inCodegenMode()) {
        return infixDispatch(this, opImpl);
      }
      return jsInfix;
    },
});
```

packages/typegpu/src/data/index.ts
```diff
@@ -21,9 +22,20 @@ function assignInfixOperator<T extends typeof VecBase | typeof MatBase>(
     value: opImpl,
   });
 
+  // To optimize infix operators on JS side, this function,
```
aleksanderkatan:
```suggestion
  // To optimize infix operators on JS side,
```

packages/typegpu/src/tgsl/infixDispatch.ts
```diff
@@ -0,0 +1,75 @@
...
+export type InfixOperator = (typeof infixOperators)[InfixOperatorName];
+
+/**
+ * InfixDispatch is only used in codegen mode.
+ * `lhs` may either be Numeric or Snippet.
+ * @example
+ * const external = d.vec2u(1);
+ * const fn = () => {
+ *    'use gpu';
+ *    external.mul(2); // lhs is Numeric
```
cieplypolar, line 30:
```suggestion
 *    external.mul(2); // lhs is Value
```
IMO vector is not Numeric. Maybe call it Value?

cieplypolar (approved):
Looks good and clean 🔥

packages/typegpu/src/tgsl/infixDispatch.ts
aleksanderkatan: I don't like `Value` name, it's too general for me (why isn't `vec2b` or a struct a value?)
How about `Tensor`?

packages/typegpu/src/core/function/comptime.ts
```diff
@@ -58,7 +58,10 @@ export function comptime<T extends (...args: never[]) => unknown>(func: T): Tgpu
         );
       }
 
-      return coerceToSnippet(func(...(args.map((s) => s.value) as never[])));
+      ctx.pushMode(new NormalState());
+      const result = coerceToSnippet(func(...(args.map((s) => s.value) as never[])));
+      ctx.popMode();
+      return result;
```
iwoplaza:
```suggestion
      ctx.pushMode(new NormalState());
      try {
        return coerceToSnippet(func(...(args.map((s) => s.value) as never[])));
      } finally {
        ctx.popMode();
      }
```

packages/typegpu/src/tgsl/infixDispatch.ts
```diff
@@ -0,0 +1,75 @@
...
+    this.operator = operator;
+  }
+}
+
+export function isInfixDispatch(o: unknown): o is InfixDispatch {
+  return o instanceof InfixDispatch;
+}
+
+/**
+ * This function is used on vex/mat prototypes.
```
iwoplaza:
```suggestion
 * This function is used on vec/mat prototypes.
```

iwoplaza (approved):

packages/typegpu/src/core/function/comptime.ts
aleksanderkatan: Why? In what case is it necessary?

PR #2447, opened 2026-05-06
fix: Struct fields not suggested

### LINKED ISSUE #2445: bug: Struct fields no longer suggested in `root.createFoo` APIs
I noticed that struct fields are no longer suggested when asking the IDE for available field names in the initial value:

Image

It does however suggest field names of functions... so as an experiment, I flipped the order of the overloads, having the plain value form first, initializer second:

Image

And it fixed the issue:

Image

It would be great to add regression tests for this using `attest`

aleksanderkatan:
For all 4 buffer constructors (buffer, uniform, mutable, readonly) there were 4 overloads:
1. One with init data.
2. One with an initializer.
3. One with either init data or an initializer.
4. One with a raw buffer.

The issue #2445 suggests reordering 1 and 2, though I instead removed them. 
I couldn't merge 3 and 4 because it would hint GPUBuffer methods when initializing a buffer with a struct

### DIFF
--- a/packages/typegpu/src/core/root/rootTypes.ts
+++ b/packages/typegpu/src/core/root/rootTypes.ts
@@ -836,12 +836,12 @@ export interface TgpuRoot extends Unwrapper, WithBinding {
   createBuffer<TData extends AnyData>(
     typeSchema: ValidateBufferSchema<TData>,
     // NoInfer is there to infer the schema type just based on the first parameter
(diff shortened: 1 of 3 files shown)


iwoplaza (approved):
Thanks a lot! Nice to see the non host-shareable errors get a lot cleaner as a byproduct 🚀

Copilot left a few nice suggestions, I left a 👍 where I agree.

PR #2470, opened 2026-05-08
@typegpu/geometry Improve line rendering APIs

deluksic:
Renamed and reordered the APIs a bit, and allow customization of arrow caps.

- `lineSegmentVariableWidth(vertexIndex, A, B, C, D, MAX_JOIN_COUNT)` -> `polylineVariableWidth(A, B, C, D, vertexIndex, MAX_JOIN_COUNT)`
- new API: `lineVariableWidth(A, B, vertexIndex, MAX_JOIN_COUNT)`

With new `arrowCapParamsSlot`, one can change arrow cap shape (length, width, slant).

image

image


packages/typegpu/src/core/texture/texture.ts
```diff
@@ -243,7 +243,7 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
           this.#texture = branch.device.createTexture({
             label: getName(this) ?? '<unnamed>',
             format: props.format,
-            size: props.size,
+            size: [...props.size],
```
deluksic: I was getting some type issues in my strict codebase, not sure if you want to fix it upstream.

packages/typegpu/src/tgsl/shaderGenerator.ts
```diff
@@ -13,7 +13,7 @@ export interface ShaderGenerator {
   initGenerator(ctx: GenerationCtx): void;
 
   functionDefinition(options: FunctionDefinitionOptions): string;
-  typeInstantiation(schema: BaseData, args: readonly Snippet[]): ResolvedSnippet;
+  typeInstantiation(schema: BaseData, args: readonly (Snippet | undefined)[]): ResolvedSnippet;
```
deluksic: same here

packages/typegpu/src/core/texture/texture.ts
iwoplaza: Are these errors brought up by a linter, or by the type checker? 👀

deluksic: I was installing the library by doing `file:..` installation from local disk to test the implementation, and I don't know why, but typescript started type checking the dependency and found this to be an issue. Not sure how strict typegpu project ts is, but I had `strict` and `noUncheckedIndexedAccess`.

Here specifically, it complained that `props.size` is readonly and can't be assigned to `size: number[]` which is mutable. Makes sense, but that's also why I don't use readonly haha. You really need to splatter it everywhere.

packages/typegpu/src/tgsl/shaderGenerator.ts
deluksic: Reverted

packages/typegpu/src/core/texture/texture.ts
deluksic: reverted, it was caused by typescript version 6.0.0 shipping slightly different webgpu types than `@webgpu/types` package.

iwoplaza:
@pullfrog Review this please 🙏

iwoplaza (approved):
LGTM! Thanks a lot for these changes 🐸 💚

PR #2474, opened 2026-05-09
docs: Remake the `fluid-with-atomics` example

reczkok:
Rework of the old fluid-with-atomics example:
- move to shell-less functions and clean up the helpers
- move the rendering from quad per tile to fullscreen triangle (like 10x performance boost or so)
- move the drawing from cpu to gpu

### DIFF
--- a/apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
+++ b/apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
@@ -680,7 +680,6 @@ export const controls = defineControls({
 
 export function onCleanup() {
   cancelAnimationFrame(animationFrame);
-  destroyGridState();
   root.destroy();
 }
 
--- a/apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
+++ b/apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
@@ -23,11 +23,6 @@ const MAX_WATER_LEVEL_UNPRESSURIZED = tgpu.const(d.u32, 0xff);
 const MAX_WATER_LEVEL = tgpu.const(d.u32, (1 << 24) - 1);
 const MAX_PRESSURE = tgpu.const(d.u32, 12);
 
-const CELL_EMPTY = tgpu.const(d.u32, CellKind.empty);
-const CELL_WALL = tgpu.const(d.u32, CellKind.wall);
-const CELL_SOURCE = tgpu.const(d.u32, CellKind.source);
-const CELL_DRAIN = tgpu.const(d.u32, CellKind.drain);
-
 const BrushConfig = {
   wall: { cellKind: CellKind.wall, waterAmount: 0 },
   source: { cellKind: CellKind.source, waterAmount: 0 },
@@ -110,17 +105,17 @@ const getWaterLevel = (coord: d.v2u): number => {
 
 const isFlowBlocked = (coord: d.v2u): boolean => {
   'use gpu';
-  return !isInBounds(coord) || getFlags(coord) === CELL_WALL.$;
+  return !isInBounds(coord) || getFlags(coord) === CellKind.wall;
 };
 
 const isDrainTarget = (coord: d.v2u): boolean => {
   'use gpu';
-  return !isInBounds(coord) || getFlags(coord) === CELL_DRAIN.$ || isBoundary(coord);
+  return !isInBounds(coord) || getFlags(coord) === CellKind.drain || isBoundary(coord);
 };
 
 const canStoreWater = (coord: d.v2u): boolean => {
   'use gpu';
-  return !isDrainTarget(coord) && getFlags(coord) !== CELL_WALL.$;
+  return !isDrainTarget(coord) && getFlags(coord) !== CellKind.wall;
 };
 
 const getTargetWaterLevel = (coord: d.v2u): number => {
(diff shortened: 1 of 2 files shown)


apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -107,249 +178,314 @@ const getStableStateBelow = tgpu.fn(
...
-    // compute
-    computePipeline.dispatchWorkgroups(
-      options.size / options.workgroupSize,
-      options.size / options.workgroupSize,
-    );
+function createGridState(size: number) {
+  const cellCount = size * size;
+  const flagsBuffer = root.createBuffer(FlagsArray(cellCount)).$usage('storage');
+  const currentWaterBuffer = root.createBuffer(WaterArray(cellCount)).$usage('storage');
+  const nextWaterBuffer = root.createBuffer(AtomicWaterArray(cellCount)).$usage('storage');
```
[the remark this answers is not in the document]
reczkok: Pure hallucination btw
https://www.w3.org/TR/webgpu/?utm_source=chatgpt.com#security-uninitialized

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -1,104 +1,175 @@
...
+const WATER_BRUSH_LEVEL = 100;
+const SOURCE_RATE = tgpu.const(d.u32, 20);
 const MAX_WATER_LEVEL_UNPRESSURIZED = tgpu.const(d.u32, 0xff);
 const MAX_WATER_LEVEL = tgpu.const(d.u32, (1 << 24) - 1);
 const MAX_PRESSURE = tgpu.const(d.u32, 12);
 
+const CELL_EMPTY = tgpu.const(d.u32, CellKind.empty);
+const CELL_WALL = tgpu.const(d.u32, CellKind.wall);
+const CELL_SOURCE = tgpu.const(d.u32, CellKind.source);
+const CELL_DRAIN = tgpu.const(d.u32, CellKind.drain);
```
iwoplaza: Why wrap these in a const, instead of using `CellKind.*` directly?

reczkok: I remember running into some implicit conversion issues, but I will check if that is still the case

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -367,52 +503,40 @@ const startDrawing = (erase: boolean) => {
...
+  drawFrame();
 };
 
+canvas.addEventListener('contextmenu', (event) => {
+  if (event.target === canvas) {
+    event.preventDefault();
+  }
+});
+
 canvas.onmousedown = (event) => {
```
aleksanderkatan, line 553: nit: could you handleDrawing on click/press as well? This would make it easier to place precise pixels

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -107,249 +178,314 @@ const getStableStateBelow = tgpu.fn(
...
+  applyFlow(coord, target, flow);
+  remainingWater.$ -= flow;
+};
+
+const simulationPipeline = root.createGuardedComputePipeline((x: number, y: number) => {
+  'use gpu';
+  const coord = d.vec2u(x, y);
+  if (handleCellFlags(coord)) {
     return;
   }
```
aleksanderkatan, line 291: Is no simulation on boundary intentional?

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -107,249 +178,314 @@ const getStableStateBelow = tgpu.fn(
...
-      .with(vertexInstanceLayout, currentStateBuffer)
-      .draw(4, options.size ** 2);
-  };
+function stepSimulation() {
+  if (!gridState) {
+    return;
+  }
+
+  simulationPipeline.with(gridState.bindGroup).dispatchThreads(options.size, options.size);
+  gridState.currentWaterBuffer.copyFrom(gridState.nextWaterBuffer);
```
aleksanderkatan, line 429: Do we intentionally avoid double buffering in this example?

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -1,104 +1,175 @@
...
-  }
-}
+const getIndex = (coord: d.v2u): number => {
+  'use gpu';
+  return coord.y * simParams.$.resolution.x + coord.x;
+};
+
+const isInBounds = (coord: d.v2u): boolean => {
+  'use gpu';
+  return coord.x < simParams.$.resolution.x && coord.y < simParams.$.resolution.y;
```
aleksanderkatan, line 83: This function is asymetric without `coord.x > 0 && coord.y > 0`, is this intentional?

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
```diff
@@ -107,249 +178,314 @@ const getStableStateBelow = tgpu.fn(
...
+  if (brushDistance > brush.radius) {
+    return;
   }
 
-  const normalized = std.min(input.cell / d.f32(0xff), 1);
+  const index = getIndex(d.vec2u(x, y));
+  if (brush.erasing !== 0) {
+    brushStateLayout.$.flags[index] = CELL_EMPTY.$;
+    brushStateLayout.$.currentWater[index] = 0;
+    std.atomicStore(brushStateLayout.$.nextWater[index], 0);
```
aleksanderkatan, line 326: Do we modify both buffers to avoid copying entire `nextWater` back to `currentWater`?

aleksanderkatan (approved):
Nice! So fast

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
reczkok: Yup

apps/typegpu-docs/src/examples/simulation/fluid-with-atomics/index.ts
reczkok: Yup, but copy might actually be faster in some cases - I would say it's fast enough so we don't really need to worry

PR #2476, opened 2026-05-10
fix(unplugin-typegpu): Inline magic-string-ast helpers to still support CJS-only environments

iwoplaza:
image

`magic-string-ast` exports only ESM, but `magic-string` does both CJS and ESM.

### DIFF
--- a/packages/unplugin-typegpu/src/core/factory.ts
+++ b/packages/unplugin-typegpu/src/core/factory.ts
@@ -1,5 +1,5 @@
 import defu from 'defu';
-import { generateTransform, MagicStringAST } from 'magic-string-ast';
+import MagicString from 'magic-string';
 import { getBabelParserOptions, getLang } from 'ast-kit';
 import type { UnpluginBuildContext, UnpluginContext, UnpluginFactory } from 'unplugin';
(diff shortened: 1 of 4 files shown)


cieplypolar (approved):
Looking good 🚀

packages/unplugin-typegpu/src/core/factory.ts
```diff
@@ -101,21 +103,33 @@ function replaceWithAssignmentOverload(
...
+  slice(this: UnpluginPluginState, node: NodeLocation): string {
+    return this.magicString.slice(node.start ?? 0, node.end ?? 0);
+  },
+  remove(this: UnpluginPluginState, node: NodeLocation): void {
+    this.magicString.remove(node.start ?? 0, node.end ?? 0);
+  },
+  overwrite(this: UnpluginPluginState, node: NodeLocation, content: string): void {
+    this.magicString.overwrite(node.start ?? 0, node.end ?? 0, content);
+  },
+};
```
reczkok, line 131: Since we are basically recreating the old `sliceNode`, `removeNode`, `overwriteNode` APIs we could name them as such

PR #2481, opened 2026-05-11
chore: Bump @ungap/structured-clone in response to deprecation notice (and potential CVE)

iwoplaza:


package.json
```diff
@@ -76,7 +76,8 @@
       "@types/node": "catalog:types",
       "typescript": "catalog:types",
       "three": "catalog:example",
-      "@typescript-eslint/utils": "^8.57.2"
+      "@typescript-eslint/utils": "^8.57.2",
+      "@ungap/structured-clone": "1.3.1"
```
aleksanderkatan: Should we note somewhere why this is here despite not being directly used? Or is the existence of this PR enough?

aleksanderkatan (approved):

package.json
cieplypolar: Why not `override`?

PR #2482, opened 2026-05-11
fix: Accept readonly tuples and arrays in `d.InferInput`

### LINKED ISSUE #2465: bug: Accept readonly tuples in `d.InferInput`

aleksanderkatan:
Did we have a reason for defining `InferInput` like this
```ts
export type InferInput<T> =
  | Infer<T>
  | (T extends { readonly [$inRepr]: infer TRepr } ? TRepr : never);
```
instead of this?
```ts
export type InferInput<T> = T extends { readonly [$inRepr]: infer TRepr } ? TRepr : Infer<T>;
```

This seems less intuitive, but there seems to be a test preventing the change: `d.Infer<T> should be assignable to d.InferInput<T>` (why?)

### DIFF
--- a/packages/typegpu/tests/buffer.test.ts
+++ b/packages/typegpu/tests/buffer.test.ts
@@ -800,7 +800,9 @@ describe('TgpuBuffer', () => {
   }) => {
     const buffer = root.createBuffer(d.arrayOf(d.u16, 32));
 
-    expectTypeOf(buffer.write).parameter(0).toEqualTypeOf<number[] | Uint16Array | ArrayBuffer>();
+    expectTypeOf(buffer.write)
+      .parameter(0)
+      .toEqualTypeOf<number[] | readonly number[] | Uint16Array | ArrayBuffer>();
   });
 
   it('should write an arrayOf(u16) buffer from a Uint16Array', ({ root, device }) => {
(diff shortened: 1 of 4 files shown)


packages/typegpu/tests/buffer.test.ts
```diff
@@ -800,7 +800,9 @@ describe('TgpuBuffer', () => {
   }) => {
     const buffer = root.createBuffer(d.arrayOf(d.u16, 32));
 
-    expectTypeOf(buffer.write).parameter(0).toEqualTypeOf<number[] | Uint16Array | ArrayBuffer>();
+    expectTypeOf(buffer.write)
+      .parameter(0)
+      .toEqualTypeOf<number[] | readonly number[] | Uint16Array | ArrayBuffer>();
```
aleksanderkatan, line 805: This is forced by `Infer` being assignable to `InferInput`

iwoplaza:
> This seems less intuitive, but there seems to be a test preventing the change: d.Infer<T> should be assignable to d.InferInput<T> (why?)

@aleksanderkatan In a practical sense, TypeScript isn't always able to guarantee that if an argument is typed as `d.Infer<T>`, then it will be accepted in the `.write` method of a buffer, for example.
Intuitively, everything that is read from a schema, should be writable back to it.

iwoplaza (approved):
Yup!

reczkok (approved):
Less casting yay

PR #2489, opened 2026-05-12
feat: `std.copy`

### LINKED ISSUE #2475: feat: `std.copy` API
Instead of:

```ts
function foo(vec: d.v2f | d.v3f) {
  'use gpu';
  const copied = vec.kind === 'vec2f' ? d.vec2f(vec) : d.vec3f(vec);
}
```

we could do:

```ts
function foo(vec: d.v2f | d.v3f) {
  'use gpu';
  const copied = std.copy(vec);
}
```

This becomes more useful the more possible types there are, as well as in cases where the type is generic.

aleksanderkatan:


apps/typegpu-docs/src/content/docs/apis/data-schemas.mdx
```diff
@@ -545,38 +545,46 @@ fn myFn(){
 }
 ```
 
-In JavaScript, `s2` is a reference to `s1` and therefore `s1` is modified when `s2` is modified.
-On the WGSL side, we make `s2` a reference to `s1` and modifying `s2.n` is translated to modifying `(*s2).n`.
-
-Schemas can be used to create deep copies of values.
+In both cases, modifying `s2` results in a change to `s1`.
+This is intuitive to JS developers, but it may be also problematic in some cases:
```
reczkok: I kinda liked the old more in-depth explanation, maybe something like:
```
Even though modifying `s2` results in a change to `s1` in both cases, it happens in two entirely different ways:

- In JavaScript, `s2` is a reference to `s1`, so modifying `s2` also modifies `s1`.
- On the WGSL side, we make `s2` a pointer to `s1`, so modifying `s2.n` is translated to modifying `(*s2).n`.

To force an explicit copy in both cases, use a schema when the exact type is known:
// ... example

Or use `std.copy` when the schema may vary:
// ... example

```
Something along these lines, wdyt?

reczkok (commented):
Looks good, I would add some gpu tests for the actual `std.copy` use case (polymorphic functions)

apps/typegpu-docs/src/content/docs/apis/data-schemas.mdx
iwoplaza: I also think that we should recommend using explicit schemas whenever possible, as that makes the code more readable at a glance, and only use `std.copy` when that's not possible.

aleksanderkatan: I updated the docs, see if they are better now. Also, wouldn't this entire section be more fitted for the `Functions` section?

I also added a polymorphic test, though I'm not sure what more to test there, since GPU `std.copy` is just identity.

reczkok (approved):
Great

PR #2490, opened 2026-05-12
refactor: Better origin tracking

iwoplaza:
This changes the behavior of origin tracking a bit in a few important ways:

## Origins are kept even for naturally ephemeral values
Before this PR, accessing the prop `x` of a vector defines as a private variable had the origin `runtime`:

```ts
const state = tgpu.privateVar(d.vec3f);

function foo() {
  'use gpu';
  state.$; // 'private'
  state.$.x; // 'runtime'
}
```

This was fine, because it meant it was easier to detect whether something can be used to initialize a variable without taking a pointer or not, but it made it impossible to track whether assignment was legal:

```ts
const OFFSET = tgpu.const(d.vec3f, 1);

function foo() {
  'use gpu';
  OFFSET.$.x += 1;
}
```

Since `OFFSET.$.x` has the origin `'runtime'`, there is no way for the system to determine that this assignment is illegal. To fix this, the system now tracks origins even for naturally ephemeral values.

## Not tracking constant variables
The following TypeScript code:
```ts
const a = 1;
const b = a;
```
Previously generates the following WGSL:
```ts
const a = 1f;
const b = a;
```
That's because `const` statements initialized with values with the `'constant'` origin that were naturally ephemeral were also granted the `'constant'` origin. This however made a few things unclear:
1. If a snippet has the `'constant'` origin, we have no way of knowing whether it's a reference to a constant variable, or an ephemeral constant value.
2. We plan on deferring giving statements either 'let', 'const' or 'var' depending on how they're used in the rest of the program (#2390), so we will not be able to give these snippets an origin that reflects their constant-ness, since we just don't know it yet.

Because of this, I opted to always generate snippets with the `'local'` origin for new definitions that don't reference other values. If we do want to handle this case somehow in the future, I would opt for adding a new `local-constant` origin, or something similar.


packages/typegpu/tests/tgsl/wgslGenerator.test.ts
aleksanderkatan, line 1: Some cases are not tested and not handled by wgslGenerator.

TS error, generates invalid code (in analogous situation with const we throw an error):
```ts
  it('forbids assignment to uniforms', ({ root }) => {
    const myUniform = root.createUniform(d.vec2u);
    const myFn = () => {
      'use gpu';
      myUniform.$.x = 1;
    };

    expect(() => tgpu.resolve([myFn])).toThrowErrorMatchingInlineSnapshot();
  });
```

No TS error, generates invalid code:
```ts
  it('forbids assignment to uniform props', ({ root }) => {
    const myUniform = root.createUniform(d.vec2u);
    const myFn = () => {
      'use gpu';
      myUniform.$.x = 1;
    };

    expect(() => tgpu.resolve([myFn])).toThrowErrorMatchingInlineSnapshot();
  });
```

Analogous tests for readonlys:
```ts
  it('forbids assignment to readonlys', ({ root }) => {
    const myReadonly = root.createReadonly(d.vec2u);
    const myFn = () => {
      'use gpu';
      // @ts-expect-error
      myReadonly.$ = d.vec2u();
    };

    expect(() => tgpu.resolve([myFn])).toThrowErrorMatchingInlineSnapshot();
  });

  it('forbids assignment to readonly props', ({ root }) => {
    const myReadonly = root.createReadonly(d.vec2u);
    const myFn = () => {
      'use gpu';
      myReadonly.$.x = 1;
    };

    expect(() => tgpu.resolve([myFn])).toThrowErrorMatchingInlineSnapshot();
  });
```

While we're at it, we can also include these two:
```ts
  it('forbids assignment to consts', () => {
    const c = tgpu.const(d.vec2u, d.vec2u(1, 2));
    const testFn = () => {
      'use gpu';
      // @ts-expect-error
      c.$ = d.vec2u();
    };

    expect(() => tgpu.resolve([testFn])).toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:testFn
      - fn*:testFn(): 'c.$ = d.vec2u()' is invalid, because the left side is a constant.]
    `);
  });

  it('forbids assignment to const props', () => {
    const c = tgpu.const(d.vec2u, d.vec2u(1, 2));
    const testFn = () => {
      'use gpu';
      c.$.x = 1;
    };

    expect(() => tgpu.resolve([testFn])).toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:testFn
      - fn*:testFn(): 'c.$.x = 1' is invalid, because the left side is a constant.]
    `);
  });
```

packages/typegpu/src/core/constant/tgpuConstant.ts
```diff
@@ -119,11 +114,7 @@ class TgpuConstImpl<TDataType extends BaseData> implements TgpuConst<TDataType>,
 
     ctx.addDeclaration(`const ${id}: ${resolvedDataType} = ${resolvedValue};`);
 
-    return snip(
-      id,
-      this.dataType,
-      isNaturallyEphemeral(this.dataType) ? 'constant' : 'constant-tgpu-const-ref',
-    );
+    return snip(id, this.dataType, 'constant-tgpu-const-ref');
```
aleksanderkatan: Does this change modify any runtime behavior?

packages/typegpu/src/data/snippet.ts
```diff
@@ -4,34 +4,36 @@ import { DEV } from '../shared/env.ts';
...
+  // TGPU.CONST REFERENCES
+  // A reference to a tgpu.const().$ value (which is frozen), that is treated by WGSL as a 'constant'.
+  // This is the usual case, read about 'runtime-tgpu-const-ref' to know when this doesn't apply.
+  // Turns into a `const` when assigned to a variable
+  | 'constant-tgpu-const-ref'
+  // A reference to a tgpu.const().$ value (which is frozen), that is NOT treated by WGSL as a 'constant'.
+  // This can happen if say a constant is accessed with a runtime-known index. WGSL doesn't treat it like
+  // a 'constant' anymore, but it's still a frozen value in JS, so we must treat it like it's immutable.
+  // Turns into a `let` when assigned to a variable
+  | 'runtime-tgpu-const-ref';
```
aleksanderkatan: Actually, what's "-ref" about tgpu-const origin? To stay consistent, shouldn't the suffix either be removed, or 'argument' and others made into 'argument-ref' etc.?

packages/typegpu/src/data/ref.ts
```diff
@@ -172,11 +172,10 @@ export function derefSnippet(snippet: Snippet): Snippet {
   }
 
   const innerType = snippet.dataType.inner;
-  const origin = isNaturallyEphemeral(innerType) ? 'runtime' : snippet.origin;
 
   if (snippet.value instanceof RefOperator) {
-    return snip(stitch`${snippet.value.snippet}`, innerType, origin);
+    return snip(stitch`${snippet.value.snippet}`, innerType, snippet.origin);
```
aleksanderkatan, line 197: A free test (maybe it's unrelated to the change)
```ts
  it('allows for d.ref to a const prop', () => {
    const myConst = tgpu.const(d.vec2u, d.vec2u());

    const modify = (n: d.ref<number>) => {
      'use gpu';
      n.$++;
    };

    const main = () => {
      'use gpu';
      const dref = d.ref(myConst.$.x); // initialized by the const value, not a ref to it
      modify(dref);
    };

        "const myConst: vec2u = vec2u();
  
        fn modify(n: ptr<function, u32>) {
          (*n)++;
        }
  
        fn main() {
          var dref = myConst.x;
          modify((&dref));
        }"
      `);
  });
```

aleksanderkatan, line 197: This test throws a wrong error (again, probably was the case before)
```ts
  it('throws on d.ref to a const ref', () => {
    const myConst = tgpu.const(d.vec2u, d.vec2u());

    const modify = (n: d.ref<d.v2u>) => {
      'use gpu';
      n.$.x++;
    };

    const main = () => {
      'use gpu';
      modify(d.ref(myConst.$));
    };

    expect(() => tgpu.resolve([main])).toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:main
      - fn*:main(): d.ref() created with primitive types must be stored in a variable before use]
    `);
  });
```

packages/typegpu/src/tgsl/accessIndex.ts
```diff
@@ -27,31 +20,18 @@ export function accessIndex(target: Snippet, indexArg: Snippet | number): Snippe
...
-    } else if (target.origin === 'runtime-tgpu-const-ref') {
-      // Runtime refs keep their ref semantics unless the element is ephemeral only
-      origin = isElementNatEph ? 'runtime' : 'runtime-tgpu-const-ref';
-    } else if (!isTargetEphemeral && !isElementNatEph) {
-      // Stable containers can forward their origin information
-      origin = target.origin;
-    } else if (isIndexConstant && target.origin === 'constant') {
-      // Plain constants indexed with constants stay constant
-      origin = 'constant';
+      origin = index.origin === 'constant' ? 'constant-tgpu-const-ref' : 'runtime-tgpu-const-ref';
```
aleksanderkatan: This should work, correct?
```ts
  it('allows for tgpu.const-tgpu.const index access', () => {
    const myConstArray = tgpu.const(d.arrayOf(d.u32, 2), [1, 2]);
    const myConstIndex = tgpu.const(d.u32, 1);

    const fn = () => {
      'use gpu';
      return myConstArray.$[myConstIndex.$];
    };

    expect(extractSnippetFromFn(fn).origin).toBe('constant-tgpu-const-ref');
  });
```

```suggestion
      origin =
        index.origin === 'constant' || index.origin === 'constant-tgpu-const-ref'
          ? 'constant-tgpu-const-ref'
          : 'runtime-tgpu-const-ref';
```

packages/typegpu/src/tgsl/accessIndex.ts
```diff
@@ -27,31 +20,18 @@ export function accessIndex(target: Snippet, indexArg: Snippet | number): Snippe
...
+    } else if (target.origin === 'constant') {
+      // Ephemeral constants indexed with constants stay constant, otherwise they become runtime-known
+      origin = index.origin === 'constant' ? 'constant' : 'runtime';
```
aleksanderkatan: same here, though it's more difficult to test: #2512
```suggestion
      origin =
        index.origin === 'constant' || index.origin === 'constant-tgpu-const-ref'
          ? 'constant'
          : 'runtime';
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1031,119 +1147,12 @@ ${this.ctx.pre}if (${condition}) ${consequent}
...
-  - Try 'const ${rawId} = ${rhsStr}' if you won't reassign '${rawId}' later.
-  -----`,
-              );
-            }
-          }
-        }
-      }
+    if (statement[0] === NODE.let) {
+      return this._letStatement(statement);
+    }
```
aleksanderkatan, line 1198: Why not split this only after `stmtType` becomes relevant? Otherwise we introduce non insignificant code duplication

aleksanderkatan (changes requested):
Nice work!

I leave some tests. Updating `validateSnippetMutation` should be enough to make them pass.

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: It is relevant though 👀. It's mostly about intent, and the separation, at least to me, makes it a lot easier to follow the logic.

packages/typegpu/src/core/constant/tgpuConstant.ts
iwoplaza: It's not supposed to 👀

packages/typegpu/src/data/ref.ts
iwoplaza: Fixed 🫡

iwoplaza: Not legal anymore

aleksanderkatan (approved):
Sure!

packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
```diff
@@ -25,13 +25,13 @@ export interface TgpuRawCodeSnippet<TDataType extends BaseData> {
 }
 
 // The origin 'function' refers to values passed in from the calling scope, which means
-// we would have access to this value anyway. Same goes for 'argument' and 'this-function',
+// we would have access to this value anyway. Same goes for 'argument' and 'local-def',
 // the values literally exist in the function we're writing.
 //
-// 'constant-ref' was excluded because it's a special origin reserved for tgpu.const values.
+// '*-tgpu-const-ref' were excluded because they're a special origin reserved for tgpu.const values.
```
cieplypolar:
```suggestion
// '*-immutable-def' were excluded because they're a special origin reserved for tgpu.const values.
```

packages/typegpu/tests/constant.test.ts
```diff
@@ -159,4 +160,47 @@ describe('tgpu.const', () => {
       }"
     `);
   });
+
+  it('forbids assignment to consts', () => {
```
cieplypolar, line 164: could you add test for assignment to `runtime-immutable-def` origin

cieplypolar (approved):
Sooo clean. I love it. :shipit:

PR #2493, opened 2026-05-12
[REPRODUCTION] fix: Babel should hoist exported function statements

iwoplaza:


iwoplaza:
Done as part of #1353

PR #2497, opened 2026-05-13
docs: Going parallel guide

reczkok:


apps/typegpu-docs/astro.config.mjs
```diff
@@ -122,12 +122,14 @@ export default defineConfig({
             {
               label: 'Your first GPU program',
               slug: 'fundamentals/your-first-gpu-program',
-              badge: { text: 'new' },
+            },
+            {
+              label: 'Going parallel with reusable resources',
+              slug: 'fundamentals/compute-shaders',
             },
```
iwoplaza, line 131: We can bring back the "new" labels I think, since that was just a test to see how it will look like in the future

iwoplaza (approved):
Meearrggee! :shipit:
The only thing I would change is the `new` labels

PR #2500, opened 2026-05-14
chore(@typegpu/react): Warn when `disableWorklets` was changed while a component was mounted

iwoplaza:
TODO:
- See if disabling worklets makes more sense than enabling them explicitly


iwoplaza:
Stale

PR #2502, opened 2026-05-15
impr: Better metadata versioning

### LINKED ISSUE #2432: impr: Better metadata versioning
When `unplugin-typegpu` embeds metadata into source code, it does so with a single-number version:

```ts
const metadata = `{
  v: ${FORMAT_VERSION},
  ast: ${embedJSON(ast)},
    .join(', ')}}),
}`;
```

`FORMAT_VERSION` is a number imported from `tinyest`, and directly relates to possible changes in the _tinyest_ representation, which would allow us to distinguish how to interpret the metadata if we choose to update that format in the future. There are cases where we want to update other parts of the metadata though, like the recent #2019 workload.

## Proposal

I propose we reinterpret this `v` property as the version of the metadata as a whole, not just `tinyest`. We can interpret `1` as being the current form, and use `2` when emitting the changed externals format introduced in #2019.

```ts
const METADATA_FORMAT_VERSION = 2;

const metadata = `{
  v: ${METADATA_FORMAT_VERSION},
  ast: ${embedJSON(ast)},
  externals: /* ... */,
}`;
```

We can then have a `normalizeMetadata` function in the `typegpu` package that accepts both versions, and returns a common data structure that is used by the rest of the program.

```ts
interface RawMetadata1 {
  v: 1;
  // ...
}

interface RawMetadata2 {
  v: 2;
  // ...
}

type RawMetadata = RawMetadata1 | RawMetadata2;

interface Metadata {
  // ...
}

function normalizeMetadata(meta: RawMetadata): Metadata {
  if (meta.v === 1) {
    // ...
  }

  if (meta.v === 2) {
    // ...
  }

  throw new Error(`Unrecognized TypeGPU metadata format: ${JSON.stringify(meta)}`);
}
```

aleksanderkatan:
I did some additional rewriting. Most importantly, `updateMetadata` is now removed, since it was only used for setting name, and there are now two new global maps `nameMap`, `metadataMap` in addition to `globalThis.__TYPEGPU_META__`.

- `nameMap` holds all names given to typegpu resources (if `globalThis.__TYPEGPU_AUTONAME__` is defined, so is `nameMap`).
- `globalThis.__TYPEGPU_META__` now only holds raw function metadata.
- `metadataMap` holds normalized function metadata, without the name. Also, the function's entry in `globalThis.__TYPEGPU_META__` is cleared after normalization.

Why?
We needed a new map to store normalized metadata, because it may be queried multiple times, and without a map it would be recalculated every time. 

Wile creating this new map, I wanted to 'consume' the metadata as soon as `setName` happened (otherwise, we'd need to merge multiple normalized metadatas), but there were issues with externals referencing values not yet defined. To avoid spaghetti, I split this into `nameMap` and `metadataMap`.

This also gives a clearer mental model, since now metadata is equivalent with 'use gpu' functions (`tgpu.const` no longer has metadata with all fields except `name` missing, no more optional on every function metadata field).

Further work (another ticket): remove `externalNames: string[]` from typegpu, and a few versions later also from unplugin (it is completely unnecessary, or at least it seems so)


aleksanderkatan:
I forgot to change `v` name, I'll do it in a moment

packages/typegpu/tests/internal/metadataNormalization.test.ts
iwoplaza, line 1: I know this is not yet a pattern that has been applied to the repo, but over time I would like to move tests that touch internal APIs into an `/internal` folder, could you move this test to `packages/typegpu/tests/internal/metadataNormalization.test.ts`?

iwoplaza (approved):
Looks good! Thanks a lot 💜

packages/typegpu/tests/internal/metadataNormalization.test.ts
aleksanderkatan: done

iwoplaza (approved):
A second approve 👍

cieplypolar (approved):

PR #2505, opened 2026-05-15
chore: Don't use `extrenalNames`

aleksanderkatan:
Blocked by #2502
`externalNames` are redundant, and this check in `fnCore` will pass for every metadata created by unplugin.

```ts
  const metadata = `{
    v: ${FORMAT_VERSION},
    ast: ${embedJSON(ast)}, // <- redundancy here
      .join(', ')}}),
  }`;
```


iwoplaza (approved):
YIP YIP YIP

PR #2507, opened 2026-05-16
fix: Keep entry function argument names as close the the originals

iwoplaza:
Frameworks like Pixi.js expect the argument to be names a certain way in order to detect them, and us adding `_arg_` to every argument name makes that impossible.


aleksanderkatan (approved):
Sure!

PR #2517, opened 2026-05-19
feat: bitcastF32toU32

cieplypolar:
Let's wait for #2615.

TODO:
- remember to add `sideEffect: false`


packages/typegpu/src/std/bitcast.ts
```diff
@@ -65,3 +69,36 @@ export const bitcastU32toI32 = dualImpl({
...
+    const uargs = unify(arg, [f32]) ?? arg;
+    return {
+      argTypes: uargs,
+      returnType: isVec(uargs[0])
+        ? uargs[0].type === 'vec2f'
+          ? vec2u
+          : uargs[0].type === 'vec3f'
+            ? vec3u
+            : vec4u
+        : u32,
```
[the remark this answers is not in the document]
iwoplaza: Whatever they said 👍

iwoplaza (approved):
I would add a few more dataType checks like the AI said, but other than that, you have my stamp of approval :shipit:

packages/typegpu/src/std/bitcast.ts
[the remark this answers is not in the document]
cieplypolar: I will also correct the existing bitcasts in this file 🫡

packages/typegpu/src/data/numberOps.ts
```diff
@@ -27,3 +27,9 @@ export function bitcastU32toI32Impl(n: number): number {
   dataView.setUint32(0, n, true);
   return dataView.getInt32(0, true);
 }
+
+export function bitcastF32toU32Impl(n: number): number {
+  const dataView = new DataView(new ArrayBuffer(4));
```
aleksanderkatan: We could create just one dataView shared by these functions, this would make this way more performant (we already do this in numeric.ts)

Image

packages/typegpu/src/std/bitcast.ts
```diff
@@ -65,3 +76,40 @@ export const bitcastU32toI32 = dualImpl({
...
+  codegenImpl: (_ctx, [n]) => {
+    return isVec(n.dataType)
+      ? stitch`bitcast<vec${n.dataType.componentCount}u>(${n})`
+      : stitch`bitcast<u32>(${n})`;
+  },
+  signature: (...arg) => {
+    const uargs = unify(arg, f32AllowedSchemas);
+    if (!uargs) {
+      throw new SignatureNotSupportedError(arg, f32AllowedSchemas);
+    }
```
aleksanderkatan: I'm not sure if the unify's behavior is exactly what we want here.
f32 -> f32 // sure
abstractFloat -> f32 // sure
ptr\<f32\> -> f32 // sure
u32 -> f32 // surely this was an user error
Can we throw in the last case?

aleksanderkatan (approved):
Nice! 🫦🩹

packages/typegpu/src/data/numberOps.ts
cieplypolar: sure

packages/typegpu/src/std/bitcast.ts
cieplypolar: will be done

aleksanderkatan (approved):

packages/typegpu/src/std/bitcast.ts
```diff
@@ -1,17 +1,23 @@
...
+import { unifyStrict } from '../tgsl/conversion.ts';
+import { SignatureNotSupportedError } from '../errors.ts';
 
-export type BitcastU32toF32Overload = ((value: number) => number) &
-  ((value: v2u) => v2f) &
-  ((value: v3u) => v3f) &
-  ((value: v4u) => v4f);
+type BitcastU32toF32Overload = <T extends number | v2u | v3u | v4u>(
+  value: T,
+) => T extends v2u ? v2f : T extends v3u ? v3f : T extends v4u ? v4f : number;
```
iwoplaza, line 18: Why this change? How does it behave when a value is a union of say `v2u | number`? I'd love to see tests for this

cieplypolar:
@pullfrog review this PR

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -321,6 +321,25 @@ export function unify<T extends (BaseData | UnknownData)[] | []>(
   };
 }
 
+export function unifyStrict<T extends (BaseData | UnknownData)[] | []>(
+  inTypes: T,
+  restrictTo?: BaseData[],
+): { [K in keyof T]: BaseData } | undefined {
+  if (inTypes.some((type) => type === UnknownData)) {
+    return undefined;
+  }
```
iwoplaza, line 337: Could you add a doc comment explaining how this function is different from `unify`?

packages/typegpu/src/std/bitcast.ts
cieplypolar: done, test added

aleksanderkatan (approved):

PR #2518, opened 2026-05-19
feat: isBeingTranspiled + getTargetShaderLanguage

cieplypolar:
| context | isBeingTranspiled | getTargetShaderLanguage |
| :--- | :--- | :--- |
| comptime/lazy (during resolution) | `false` | `wgsl` |
|comptime in js | `false` | `undefined` |
| js | `false` | `undefined` |
| resolution | `true` | `wgsl` |
| simulate | `false` | `undefined` |


iwoplaza:
I'll just note down the mechanism we talked about today:

## isBeingTranspiled
`std.isBeingTranspiled()` returns true if the direct callee is being transpiled. Mechanically, this is equivalent to a `() => false` function that has a `$gpuCallable` method that produces `true`.

## getTargetShaderLanguage
> Signature: `std.getTargetShaderLanguage(): string | undefined`

The function is supposed to return `'wgsl'` when a resolution ctx is available, and `undefined` otherwise. I'll add the mechanism for changing the shader language name later on.

The only question is, what od we want `std.getTargetShaderLanguage()` to return in SIMULATE mode? If `undefined`, which is technically true, then we have no way of checking if we're in SIMULATE mode or not. But maybe it's for the better. @reczkok @aleksanderkatan @cieplypolar thoughts?

cieplypolar:
> I'll just note down the mechanism we talked about today:
> 
> ## isBeingTranspiled
> `std.isBeingTranspiled()` returns true if the direct callee is being transpiled. Mechanically, this is equivalent to a `() => false` function that has a `$gpuCallable` method that produces `false`.
> 
> ## getTargetShaderLanguage
> > Signature: `std.getTargetShaderLanguage(): string | undefined`
> 
> The function is supposed to return `'wgsl'` when in CODEGEN mode, and `undefined` otherwise. I'll add the mechanism for changing the shader language name later on.
> 
> The only question is, what od we want `std.getTargetShaderLanguage()` to return in SIMULATE mode? If `undefined`, which is technically true, then we have no way of checking if we're in SIMULATE mode or not. But maybe it's for the better. @reczkok @aleksanderkatan @cieplypolar thoughts?

`isBeingTranspiled` should return true in the `$gpuCallable` impl, right?

Also, I think `getTargetShaderLanguage` in this form does not differ from `isBeingTranspiled`. IMO, it should return `wgsl` if `resolutionCtx` exists. 

Your version:
| context | isBeingTranspiled | getTargetShaderLanguage |
| :--- | :--- | :--- |
| comptime/lazy | `false` | `undefined` |
| js | `false` | `undefined` |
| resolution | `true` | `wgsl` |
| simulate | `false` | `undefined` |

My version:
| context | isBeingTranspiled | getTargetShaderLanguage |
| :--- | :--- | :--- |
| comptime/lazy | `false` | `wgsl` |
| js | `false` | `undefined` |
| resolution | `true` | `wgsl` |
| simulate | `false` | `undefined` |

@iwoplaza @aleksanderkatan @reczkok

iwoplaza:
@cieplypolar You're totally right, I updated the comment to match your version.

cieplypolar:
Also, if we ever want to change the SIMULATE env behavior, it will be straightforward; simple check if `getExecMode().type === 'simulate'`

By the way, we should consider returning `simulate-wgsl` not just `simulate`.

aleksanderkatan:
> By the way, we should consider returning `simulate-wgsl` not just `simulate`.

WIll there ever be `simulate-gl`?

cieplypolar:
> > By the way, we should consider returning `simulate-wgsl` not just `simulate`.
> 
> WIll there ever be `simulate-gl`?

Honestly, I don't know. Now, `simulate` returns `undefined`.

packages/typegpu/src/std/environment.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Returns `true` when the direct callee is being transpiled for GPU, otherwise `false`.
+ *
+ * @example
+ * const f = () => {
+ *   'use gpu';
+ *   return isBeingTranspiled() ? 1 : 0;
+ * };
+ *
+ * f() // returns 0, but resolved WGSL looks like this:
```
aleksanderkatan: nit
```suggestion
 * f(); // returns 0, but resolved WGSL looks like this:
```

packages/typegpu/src/std/environment.ts
```diff
@@ -0,0 +1,62 @@
...
+ * };
+ *
+ * f() // returns false, but resolved WGSL looks like this:
+ *
+ * fn f() -> bool {
+ *   return true;
+ * }
+ *
+ * @note
+ * Inside `lazy`, it always returns `wgsl`.
```
aleksanderkatan, line 51: Until now I didn't know lazy is not callable in JS. What no docs does to a man. Can we mention it here?

```suggestion
 * Inside `lazy`, since it is only accessible during resolution, it always returns `wgsl`.
```

aleksanderkatan (approved):
Good job! 🐑ℹ️✝︎

The target branch is merged already, you can change the target to main.

packages/typegpu/tests/std/environment.test.ts
```diff
@@ -0,0 +1,172 @@
+import { it } from 'typegpu-testing-utility';
+import { expect, describe } from 'vitest';
+
+import tgpu, { d, std } from '../../src/index.js';
```
iwoplaza:
```suggestion
import tgpu, { d, std } from 'typegpu';
```

packages/typegpu/src/std/environment.ts
```diff
@@ -0,0 +1,62 @@
...
+import { snip } from '../data/snippet.ts';
+import { getExecMode, getResolutionCtx } from '../execMode.ts';
+import { $gpuCallable } from '../shared/symbols.ts';
+import type { DualFn } from '../types.ts';
+
+const impl = (() => false) as DualFn<() => boolean>;
+impl.toString = () => 'isBeingTranspiled';
+impl[$gpuCallable] = {
+  call(_ctx, _args) {
+    return snip(true, bool, 'constant');
```
iwoplaza: Could we use `coerceToSnippet` here? This will allow future snippet properties to be properly applied, like side effects.

```suggestion
    return coerceToSnippet(true);
```

iwoplaza (approved):
Yup yup!

packages/typegpu/src/std/environment.ts
iwoplaza: I don't want that to always be the case. Ofc we need to talk about how to handle memoizing things when outside of resolution, and how to implement slots, but I'd love to have this working in JS as well.

PR #2527, opened 2026-05-20
chore: Update TypeGPU dependencies

iwoplaza:
The dependency range for `tinyest` is populated by pnpm as we expect:

image


aleksanderkatan (approved):
Sure!

PR #2531, opened 2026-05-21
chore: Don't test internals of wgslGenerator

### LINKED ISSUE #2509: chore: Don't test the internals of the WgslGenerator
Instead, use:
- `expectSnippetOf`
- `expectDataTypeOf`
- `expect(tgpu.resolve([...])`

Also, make all WgslGenerator functions that are only used internally `protected`, and prefix them with an `_`.

aleksanderkatan:
Not sure if this wa the exact scope of this issue, but now tests in `tgsl` folder do not any typegpu internals. Tests that were specifically about internals were remade to use `expectSnippetOf` ect., or moved to the 'internal' folder.


packages/typegpu/tests/tgsl/conversion.test.ts
```diff
@@ -1,354 +1,224 @@
...
 
-  it('handles implicit casts', () => {
-    const res = getBestConversion([d.i32, d.f32]);
-    expect(res?.targetType).toBe(d.f32);
-    expect(res?.actions).toEqual([
-      { sourceIndex: 0, action: 'cast', targetType: d.f32 },
-      { sourceIndex: 1, action: 'none' },
-    ]);
-    expect(res?.hasImplicitConversions).toBe(true);
+  it('performs implicit casts and warns', () => {
```
iwoplaza, line 27: If it warns, can we mock and test for `console.warn` here?

packages/typegpu/tests/tgsl/conversion.test.ts
```diff
@@ -1,354 +1,224 @@
...
+        d: bool,
+      }
+
+      fn fn_1() {
+        let s = Struct(1f, 2i, vec2f(1), true);
+      }"
+    `);
   });
 
   it('maps values requiring implicit casts and warns', () => {
```
iwoplaza, line 147: Same here, could we test for warnings?

packages/typegpu/tests/utils/parseResolved.ts
```diff
@@ -43,7 +43,7 @@ class ExtractingGenerator extends WgslGenerator {
   }
 }
 
-export function extractSnippetFromFn(cb: () => unknown): Snippet {
+export function extractSnippetFromFn(cb: TgpuFn | (() => unknown)): Snippet {
```
iwoplaza, line 46: Is `TgpuFn` not assignable to `() => unknown`?

packages/typegpu/tests/internal/dualImpl.test.ts
```diff
@@ -1,14 +1,13 @@
 import { describe, expect, it } from 'vitest';
 import { dualImpl, MissingCpuImplError } from '../../src/core/function/dualImpl.ts';
-import { Void } from '../../src/data/wgslTypes.ts';
-import tgpu from '../../src/index.js';
+import tgpu, { d } from 'typegpu';
```
iwoplaza: Could we make all internal tests import only from source, and not from the potentially built package? So in this case, importing from '../../../src/index.js' would be valid.

packages/typegpu/tests/utils/parseResolved.ts
aleksanderkatan: It is not if it takes any arguments. Just like regular functions, we can always pass a function that needs fewer arguments than expected by the type, but not more.
image
Alternatively, I could accept `AnyFn`, but that would allow shellless with arguments which can't be resolved.

packages/typegpu/tests/internal/dualImpl.test.ts
aleksanderkatan: All relevant files should now import source

iwoplaza (approved):
Awesome work 👏

PR #2532, opened 2026-05-21
feat: Limited support for the runtime ternary operator

### LINKED ISSUE #2530: feat: Ternary operator support in simple non-comptime cases
If we can ensure that both branches of a ternary operator are side-effect free with heuristics, we can emit a `select()` call instead.

Maybe the heuristic essentially boils down to "Is there a call expression in here?"

> [!NOTE]
> We can also do proper side-effect analysis to allow side-effect free functions to be used in ternary operators. We can store this as another kind of flag on the snippet type.
> 
> If there are more cases where this sort of analysis might be useful, we can prioritize this.

iwoplaza:
- Adds a new flag to the snippet, which needs to be explicitly set whenever we can guarantee that it contains a side-effect free expression
- Uses that flag to emit `std.select` from a ternary operator if both branches produce scalars or vectors, and are free of side-effects.

**TODO in subsequent PRs:**
- [ ] Detect if a function can produce side-effects, which should boil down to detecting if all of its statements are side-effect free.
- [ ] Mark more operations as side-effect free, like prop access, some standard functions (e.g. `atomicStore` is not side-effect free), etc.

### DIFF
--- a/packages/typegpu/src/data/snippet.ts
+++ b/packages/typegpu/src/data/snippet.ts
@@ -92,6 +92,7 @@ export interface Snippet {
    */
   readonly dataType: BaseData | UnknownData;
   readonly origin: Origin;
+  readonly possibleSideEffects: boolean;
 }
 
 export interface ResolvedSnippet extends Snippet {
@@ -105,11 +106,18 @@ class SnippetImpl implements Snippet {
   readonly value: unknown;
   readonly dataType: BaseData | UnknownData;
   readonly origin: Origin;
+  readonly possibleSideEffects: boolean;
 
-  constructor(value: unknown, dataType: BaseData | UnknownData, origin: Origin) {
+  constructor(
+    value: unknown,
+    dataType: BaseData | UnknownData,
+    origin: Origin,
+    possibleSideEffects: boolean,
+  ) {
     this.value = value;
     this.dataType = dataType;
     this.origin = origin;
+    this.possibleSideEffects = possibleSideEffects;
   }
 }
 
@@ -121,12 +129,23 @@ export function isSnippetNumeric(snippet: Snippet) {
   return isNumericSchema(snippet.dataType);
 }
 
-export function snip(value: string, dataType: BaseData, origin: Origin): ResolvedSnippet;
-export function snip(value: unknown, dataType: BaseData | UnknownData, origin: Origin): Snippet;
+export function snip(
+  value: string,
+  dataType: BaseData,
+  origin: Origin,
+  possibleSideEffects?: boolean,
+): ResolvedSnippet;
 export function snip(
   value: unknown,
   dataType: BaseData | UnknownData,
   origin: Origin,
+  possibleSideEffects?: boolean,
+): Snippet;
+export function snip(
+  value: unknown,
+  dataType: BaseData | UnknownData,
+  origin: Origin,
+  possibleSideEffects: boolean = true,
 ): Snippet | ResolvedSnippet {
   if (DEV && isSnippet(value)) {
     // An early error, but not worth checking every time in production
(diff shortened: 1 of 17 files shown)


packages/typegpu/src/std/boolean.ts
```diff
@@ -386,9 +425,27 @@ function cpuSelect<T extends number | boolean | AnyVecInstance>(
 export const select = dualImpl({
   name: 'select',
   signature: (f, t, cond) => {
-    const [uf, ut] = unify([f, t]) ?? ([f, t] as const);
+    const [uf, ut] = unify([f, t], validSelectBranchTypes) ?? ([f, t] as const);
```
cieplypolar, line 428: I think it doesn't change anything, because unify won't throw, unify will return undefined and we will fallback to `[f, t]`

packages/typegpu/src/data/snippet.ts
```diff
@@ -138,5 +146,26 @@ export function snip(
...
+}
+
+export function noSideEffects(snippet: ResolvedSnippet): ResolvedSnippet;
+export function noSideEffects(snippet: Snippet): Snippet;
+export function noSideEffects(snippet: Snippet): Snippet {
+  return new SnippetImpl(
+    snippet.value,
+    snippet.dataType,
+    snippet.origin,
+    /* possibleSideEffects */ false,
```
cieplypolar:
```suggestion
  return withSideEffects(false, snippet);
```
Or even just bind  `withSideEffects` with first argument set to `false`

packages/typegpu/tests/tgsl/sideEffects.test.ts
```diff
@@ -0,0 +1,77 @@
...
+  });
+
+  test('vectors created from indexed arrays', () => {
+    expectSideEffects(() => {
+      'use gpu';
+      const arr = [1, 2, 3];
+      return d.vec3f(arr[0]!);
+    }).toEqual(false);
+  });
+});
```
cieplypolar, line 81:
```ts
test('vector from accessor', () => {
  const v = tgpu.accessor(d.vec3f, d.vec3f());
  expectSideEffects(() => {
    'use gpu';
    return v.$;
  }).toEqual(false);
});
```

it does not work for accessors, but I see it might be difficult.

cieplypolar (commented):
Very cool feature 💜. I left some comments to address before merging.

packages/typegpu/tests/tgsl/sideEffects.test.ts
iwoplaza: We're going to be whitelisting more and more snippets with future PRs, but I added accessor support 🫡

reczkok (approved):

packages/typegpu/src/data/snippet.ts
```diff
@@ -138,5 +146,21 @@ export function snip(
...
+export function withSideEffects(possibleSideEffects: boolean, snippet: Snippet): Snippet;
+export function withSideEffects(possibleSideEffects: boolean, snippet: Snippet): Snippet {
+  return new SnippetImpl(snippet.value, snippet.dataType, snippet.origin, possibleSideEffects);
+}
+
+export function noSideEffects(snippet: ResolvedSnippet): ResolvedSnippet;
+export function noSideEffects(snippet: Snippet): Snippet;
+export function noSideEffects(snippet: Snippet): Snippet {
+  return withSideEffects(/* possibleSideEffects */ false, snippet);
+}
```
aleksanderkatan, line 186: Can we add `possibleSideEffects` argument to `snip` and just use that by default? 
Currently, there are zero places where `withSideEffects` is NOT called with a `snip`, and there is only one place where `noSideEffects` isn't called with a `snip()`, and it can easily be refactored to fit that api.

aleksanderkatan (approved):
Great feature! 🏃‍♂️‍➡️⏲️🔱

PR #2534, opened 2026-05-21
feat: common.attachAutoResizer

iwoplaza:
This PR adds a new common helper called `attachAutoResizer` which adapts the canvas' internal resolution to match its physical pixel size, with additional safe-guards. It's a web only helper, not to be used in React Native apps.

The examples were updated to use this helper, which replaces the need for the example environment to handle the canvas resizing. This also makes the examples much more portable, as resizing a canvas can be tricky to get right, and now it's solved in every example.

### DIFF
--- a/apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.html
+++ b/apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.html
@@ -1,4 +1,4 @@
-<canvas></canvas>
+<canvas class="w-full h-full" />
 <div id="sort-overlay" hidden>
   <div id="sort-spinner"></div>
   <span id="sort-status"></span>
--- a/apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.html
+++ b/apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.html
@@ -1,6 +1,6 @@
 <div class="loading">🤖 Downloading the model...</div>
-<div class="example">
-  <canvas></canvas>
+<div class="example gap-4 flex-col md:flex-row">
+  <canvas class="w-80 h-80"></canvas>
   <div class="predictions-container">
     <div class="bars-container">
       <div class="bar">O</div>
@@ -36,7 +36,7 @@
   }
 
   .loading.loaded ~ .example {
-    display: contents;
+    display: flex;
   }
 
   .predictions-container {
--- a/apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.html
+++ b/apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.html
@@ -1,6 +1,6 @@
 <div class="loading">🤖 Downloading the model...</div>
 <div class="example gap-4 flex-col md:flex-row">
-  <canvas class="w-80 h-80"></canvas>
+  <canvas class="w-80 h-80" />
   <div class="predictions-container">
     <div class="bars-container">
       <div class="bar">O</div>
--- a/apps/typegpu-docs/src/examples/rendering/xor-dev-runner/index.html
+++ b/apps/typegpu-docs/src/examples/rendering/xor-dev-runner/index.html
@@ -1,4 +1,4 @@
-<canvas data-fit-to-container></canvas>
+<canvas class="w-full h-full" />
 <p class="absolute px-2 py-1 rounded-xl top-2 mx-auto text-lg text-center text-white bg-black/50">
   Port of "Runner" by XorDev (<a
     class="text-purple-300"
--- a/apps/typegpu-docs/src/examples/tests/log-test/index.ts
+++ b/apps/typegpu-docs/src/examples/tests/log-test/index.ts
@@ -288,10 +288,10 @@ export const controls = defineControls({
   },
 });
 
-const detachAutoResizer = common.attachAutoResizer({ root, canvas });
+const autoResizer = common.attachAutoResizer({ root, canvas });
 
 export function onCleanup() {
-  detachAutoResizer();
+  autoResizer.detach();
   root.destroy();
 }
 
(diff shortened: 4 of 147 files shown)


packages/typegpu/src/common/attachAutoResizer.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+  });
+
+  try {
+    observer.observe(canvas, { box: 'device-pixel-content-box' });
+  } catch {
+    observer.observe(canvas, { box: 'content-box' });
+  }
+
+  return () => {
```
deluksic: What do you think about returning a `{ detach: () => {} }` so its self-documenting?

deluksic: then the rest of the code will read `autoResizer.detach()`

packages/typegpu/src/common/attachAutoResizer.ts
```diff
@@ -0,0 +1,42 @@
+import type { TgpuRoot } from '../core/root/rootTypes.ts';
+
+export interface AttachAutoResizerOptions {
+  root: TgpuRoot;
+  canvas: HTMLCanvasElement;
+  onResize?(): void;
```
deluksic: `onResize` will usually want to do something with the new canvas size, so might as well pass it in

iwoplaza: I think devs usually have access to the canvas they passed in, and they can access `canvas.width` and `canvas.height`. That way we can avoid an additional array allocation on each resize.

packages/typegpu/src/common/attachAutoResizer.ts
iwoplaza: Great idea, changed it 🫡

iwoplaza:
@pullfrog Review this one please 🙏

packages/typegpu/src/common/attachAutoResizer.ts
```diff
@@ -0,0 +1,48 @@
+import type { TgpuRoot } from '../core/root/rootTypes.ts';
+
+export interface AttachAutoResizerOptions {
+  root: TgpuRoot;
+  canvas: HTMLCanvasElement;
+  onResize?: (() => void) | undefined;
```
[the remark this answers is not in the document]
deluksic: This is not true. In codebases with `exactOptionalPropertyTypes`, having `| undefined` has meaning. If you omit `undefined` you are explicitly saying `{ onResize: undefined }` can't be sent in.

apps/typegpu-docs/src/examples/rendering/xor-dev-runner/index.html
```diff
@@ -1,4 +1,4 @@
-<canvas data-fit-to-container></canvas>
+<canvas class="w-full h-full" />
 <p class="absolute px-2 py-1 rounded-xl top-2 mx-auto text-lg text-center text-white bg-black/50">
```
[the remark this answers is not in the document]
aleksanderkatan: From what I google, the clanker is right, and this is equivalent to just not closing the canvas element at all. This appears in multiple places, try regex replacing `<canvas(.*)/>` with `<canvas$1></canvas>`

apps/typegpu-docs/src/examples/algorithms/bitonic-sort/index.html
```diff
@@ -1,4 +1,4 @@
-<canvas></canvas>
+<canvas class="w-full h-full" />
```
aleksanderkatan, line 1: Is this what we want?
Image

apps/typegpu-docs/src/examples/tests/uniformity/index.ts
```diff
@@ -83,14 +83,17 @@ export const controls = defineControls({
   },
 });
 
-const resizeObserver = new ResizeObserver(() => {
-  canvasRatioUniform.write(canvas.width / canvas.height);
-  redraw();
+const autoResizer = common.attachAutoResizer({
```
aleksanderkatan, line 86: Were three examples intentionally excluded from this change?

apps/typegpu-docs/src/examples/tests/texture-test/index.ts
aleksanderkatan, line 1: This example didn't use to stretch

apps/typegpu-docs/src/examples/tests/sdf-docs-examples/index.ts
aleksanderkatan, line 1: This example didn't use to stretch

apps/typegpu-docs/src/examples/tests/rc-docs-examples/index.ts
aleksanderkatan, line 1: This example didn't use to stretch

apps/typegpu-docs/src/examples/tests/log-test/index.ts
```diff
@@ -288,7 +288,10 @@ export const controls = defineControls({
   },
 });
 
+const autoResizer = common.attachAutoResizer({ root, canvas });
```
aleksanderkatan, line 291: This example shouldn't have a resizer.

apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.html
aleksanderkatan, line 1: This example is now broken

aleksanderkatan (changes requested):
Here are some things I noticed:

Many examples (mostly orbit camera ones, but not all of those) throw when resizing the page to be so small that only example list is displayed.
http://localhost:4321/TypeGPU/examples#example=rendering--os-awards a FiniteMathAssumption error is thrown (before, it had wgsl errors), it seems that it works regardless
http://localhost:4321/TypeGPU/examples#example=simulation--gravity -||-
http://localhost:4321/TypeGPU/examples#example=rendering--pom -||-
http://localhost:4321/TypeGPU/examples#example=rendering--xor-dev-runner -||-
http://localhost:4321/TypeGPU/examples#example=rendering--function-visualizer -||-
http://localhost:4321/TypeGPU/examples#example=rendering--phong-reflection -||-
http://localhost:4321/TypeGPU/examples#example=rendering--render-bundles -||-
http://localhost:4321/TypeGPU/examples#example=rendering--render-bundles-with -||-
http://localhost:4321/TypeGPU/examples#example=simple--ripple-cube -||-, also has many WGSL warnings
http://localhost:4321/TypeGPU/examples#example=simple--mesh-skinning crashes

http://localhost:4321/TypeGPU/examples#example=algorithms--jump-flood-distance started flashing black when resizing

PR #2540, opened 2026-05-24
feat: npx typegpu

iwoplaza:


packages/typegpu/bin.mjs
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env node
+import { spawn } from 'node:child_process';
+import pkg from './package.json' with { type: 'json' };
+
+const versionPattern = /^(\d+)\.(\d+)\.(\d+)/;
+
+const result = versionPattern.exec(pkg.version);
+const [_, major, minor] = result;
+console.log(major, minor);
```
aleksanderkatan:
```suggestion
```

packages/typegpu/package.json
```diff
@@ -77,7 +78,7 @@
     "types": "./dist/index.d.ts"
   },
   "scripts": {
-    "build": "tsdown",
+    "build": "tsdown && cp bin.mjs dist/bin.mjs",
```
aleksanderkatan: Is there no more elegant way to do this?

aleksanderkatan (approved):
Looks great!

cieplypolar (approved):

packages/typegpu/package.json
iwoplaza: There is actually, I'm now using the `copy` config property from tsdown 🫡

PR #2545, opened 2026-05-24
fix(@typegpu/cli): remove default name suggestion, leave just placeholder

### LINKED ISSUE #2542: fix(@typegpu/cli): Remove the default name suggestion, leave just placeholder
Image

zichen0116:
## What

Remove `initialValue` from the `p.text()` prompts in `getProjectDirectory` and `getPackageName`, keeping only the `placeholder` (grey hint text).

## Why

Currently both prompts pre-fill the input field with a default value (`tgpu-project`). This makes it easy to accidentally accept the default without thinking about a good project name. Showing only the placeholder as a hint encourages users to choose their own name while still providing a suggestion.

## Changes

- `getProjectDirectory`: Removed `initialValue` from `p.text()` options; empty input falls back to the default value
- `getPackageName`: Removed `initialValue` from `p.text()` options; empty input falls back to the default value
- Renamed parameter from `initialValue` to `defaultValue` for clarity

Fixes #2542

### DIFF
--- a/packages/typegpu-cli/src/utils/inputs.ts
+++ b/packages/typegpu-cli/src/utils/inputs.ts
@@ -9,11 +9,10 @@ function isValidPackageName(packageName: string) {
   return /^(?:@[a-z\d][a-z\d\-._]*\/)?[a-z\d][a-z\d\-._]*$/.test(packageName.trim());
 }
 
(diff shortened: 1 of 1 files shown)


aleksanderkatan:
Issue closed by a different PR

PR #2546, opened 2026-05-24
fix(@typegpu/cli): show install output and don't auto-start dev server

### LINKED ISSUE #2544: fix(@typegpu/cli): Ergonomic tweaks
- I'd like to see the output of the package installation process, right now I am just waiting for a while
- Let's not start the dev server immediately, just tell the user what to do next

zichen0116:
## What

Two ergonomic improvements to the `tgpu` create command:

1. **Show installation output**: Replace the spinner-based `pmInstall` with interactive mode so users can see `npm`/`pnpm` install progress instead of staring at a spinner with no feedback.

2. **Don't auto-start dev server**: Replace the combined "Install with X and start now?" prompt with just "Install dependencies with X?". After scaffolding (and optional install), always show the next steps so users know what to do without the CLI hijacking their terminal.

## Why

- When running `tgpu` to create a new project, the installation step shows a spinner but no actual output, leaving users wondering if anything is happening.
- The current flow auto-starts the dev server (`pnpm dev`) which takes over the terminal. Users should be told what to do next instead.

### `packages/typegpu-cli/src/utils/pm.ts`
- `pmInstall`: Replaced spinner wrapper with `p.log.step()` + interactive `runCommand` call so install output is visible to the user

### `packages/typegpu-cli/src/create.ts`
- Changed prompt from "Install with X and start now?" to "Install dependencies with X?"
- Removed auto-start of dev server (`pmRun(pm, ['dev'])`)
- Always show next-step instructions in the outro, adjusting based on whether install was run
- Removed unused `pmRun` import

Fixes #2544

### DIFF
--- a/packages/typegpu-cli/src/create.ts
+++ b/packages/typegpu-cli/src/create.ts
@@ -57,16 +57,22 @@ export async function createProject(cwd: string) {
   const installCmd = resolveCommand(pm, 'install', []);
   const runCmd = resolveCommand(pm, 'run', ['dev']);
 
-  let msg = 'Done!\n';
-  msg += `   To get started run:\n\n`;
+  const steps: string[] = [];
+  const shouldCd = (!shouldInstall && !!installCmd) || !!runCmd || !!cdPath;
+  if (shouldCd && cdPath) {
+    steps.push(`   cd ${cdPath}`);
+  }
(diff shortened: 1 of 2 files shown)


iwoplaza (approved):
Thanks! 🤖

PR #2550, opened 2026-05-25
fix(@typegpu/cli): Show install output and don't auto-start dev server

iwoplaza:


aleksanderkatan (approved):
Looks good!

packages/typegpu-cli/src/create.ts
```diff
@@ -46,25 +46,33 @@ export async function createProject(cwd: string) {
...
-  const installAndRun = await confirmStep(`Install with ${pm} and start now?`, true);
+  const shouldInstall = await confirmStep(`Install dependencies with ${pm}?`, true);
 
-  if (installAndRun) {
+  if (shouldInstall) {
     process.chdir(root);
     pmInstall(pm);
-    pmRun(pm, ['dev']);
-    return;
   }
```
aleksanderkatan, line 54: Why do we ask? Can't we just install? The user will have one step less to think about.

iwoplaza: It's what's done by other tools in this space, and I actually appreciate it, because sometimes I want to just explore the template, or install it with a different package manager than I used for creating the template.

aleksanderkatan: If you select any packages, npm add will install anyways though

PR #2556, opened 2026-05-25
feat: Select ecosystem packages in CLI

### LINKED ISSUE #2555: feat: Select ecosystems packages to install in typegpu-cli

aleksanderkatan:
I did some additional refactoring and small changes.

### DIFF
--- a/packages/typegpu-cli/src/steps/typegpu.ts
+++ b/packages/typegpu-cli/src/steps/typegpu.ts
@@ -21,6 +21,7 @@ export async function ensureTypegpu(pm: Agent, pkg: PackageJsonWithDeps): Promis
 export async function askForPkgs(pm: Agent, pkg: PackageJsonWithDeps) {
   const options = typegpuPkgs.filter((entry) => !hasDependency(pkg, entry.value));
   if (options.length === 0) {
+    p.log.info('All typegpu ecosystem packages are already installed.');
     return;
   }
   const packages = await selectPkgs(options);
--- a/packages/typegpu-cli/src/steps/typegpu.ts
+++ b/packages/typegpu-cli/src/steps/typegpu.ts
@@ -1,10 +1,11 @@
 import type { Agent } from 'package-manager-detector';
 import * as p from '@clack/prompts';
 
-import { hasDependency } from '../utils/pkg.ts';
+import { hasDependency, typegpuPkgs } from '../utils/pkg.ts';
 import type { PackageJsonWithDeps } from '../utils/types.ts';
 import { pmAdd } from '../utils/pm.ts';
 import { confirmStep } from '../utils/prompts.ts';
+import { selectPkgs } from '../utils/inputs.ts';
 
 export async function ensureTypegpu(pm: Agent, pkg: PackageJsonWithDeps) {
   if (hasDependency(pkg, 'typegpu')) {
@@ -14,4 +15,13 @@ export async function ensureTypegpu(pm: Agent, pkg: PackageJsonWithDeps) {
   if (!(await confirmStep('Install typegpu?'))) return;
   pmAdd(pm, ['typegpu'], false);
   // no p.log.success because pmAdd already logs it
+}
+
+export async function askForPkgs(pm: Agent, pkg: PackageJsonWithDeps) {
+  const options = typegpuPkgs.filter((entry) => !hasDependency(pkg, entry.value));
+  if (options.length === 0) {
+    return;
+  }
+  const packages = await selectPkgs(options);
+  pmAdd(pm, packages, false);
 }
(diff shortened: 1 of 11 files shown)


iwoplaza:
> Should I hardcode to install newest 0.11.x version, or is just latest okay?

Since the CLI versioning is supposed to match the targeted TypeGPU major & minor, I would fetch the latest versions of the ecosystem packages with the matching major & minor of the CLI package.

I also think we can minimize the amount of bookkeeping we need to do by using something like `unplugin-macros` or the `comptime` package to fetch all other packages from the `/packages` folder, keep only those that are scoped to the `@typegpu/` org, filter out those that are marked private, as well as remove some blacklisted packages (for example @typegpu/cli). The list would be precomputed and embedded in the final JS bundle.

packages/typegpu-cli/src/utils/inputs.ts
```diff
@@ -43,3 +43,12 @@ export async function getPackageName(initialValue: string) {
...
   return packageName.trim();
 }
+
+export async function selectPkgs(options: { value: string; hint: string }[]) {
+  const packages = await p.multiselect({
+    message: "Pick packages to add ('space' to select, 'enter' to confirm):",
+    options: options,
+    required: false,
+  });
+  return typeof packages === 'symbol' ? [] : packages;
```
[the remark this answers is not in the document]
cieplypolar: I agree with copilot. If user change mind about ecosystem packages, they can unselect all the packages and press enter (probably)

packages/typegpu-cli/src/steps/typegpu.ts
```diff
@@ -15,3 +16,12 @@ export async function ensureTypegpu(pm: Agent, pkg: PackageJsonWithDeps) {
   pmAdd(pm, ['typegpu'], false);
   // no p.log.success because pmAdd already logs it
 }
+
+export async function askForPkgs(pm: Agent, pkg: PackageJsonWithDeps) {
+  const options = typegpuPkgs.filter((entry) => !hasDependency(pkg, entry.value));
+  if (options.length === 0) {
+    return;
```
cieplypolar: maybe log that all ecosystem packages are installed

packages/typegpu-cli/src/utils/pm.ts
```diff
@@ -43,6 +43,10 @@ export function pmAdd(pm: Agent, pkgs: string[], dev: boolean) {
   if (!cmd) {
     failAndExit(`Cannot resolve add command for ${pm}`);
   }
+  if (pkgs.length === 0) {
```
cieplypolar: i would check that before entering `pmAdd`

cieplypolar (commented):
Nice addition 💚

packages/typegpu-cli/src/steps/typegpu.ts
aleksanderkatan: I can do that, though I don't think this is realistically possible for a project to use all of those

packages/typegpu-cli/src/utils/pkg.ts
```diff
@@ -1,8 +1,29 @@
...
 
-export function hasDependency(pkg: PackageJsonWithDeps, name: string) {
+export function hasDependency(pkg: PackageJson, name: string) {
   const deps = pkg.dependencies ?? {};
   const devDeps = pkg.devDependencies ?? {};
   const peerDeps = pkg.peerDependencies ?? {};
   return name in deps || name in devDeps || name in peerDeps;
 }
+
+export const VERSION = '^0.11.0';
```
aleksanderkatan, line 10: Hardcoded. If we wish, we can create a map from package to version in future.

packages/typegpu-cli/src/utils/files.ts
```diff
@@ -49,7 +50,11 @@ export async function prepareDirectory(cwd: string, projectDir: string) {
   return dir;
 }
 
-export function copyTemplate(templateDir: string, projectDir: string, packageName: string) {
+export async function scaffoldProject(
```
aleksanderkatan, line 53: I dislike how this function became async, but we cannot ask for packages to add before reading the JSON (otherwise, we might suggest packages already present in the preset)

aleksanderkatan:
Now the dependencies are added directly to package.json, without use of the package manager (thus no install happens)

packages/typegpu-cli/src/utils/types.ts
```diff
@@ -1,16 +1,12 @@
 import { type } from 'arktype';
 
-export const PackageJsonWithDepsSchema = type({
+export const PackageJsonSchema = type({
+  name: 'string',
   'dependencies?': 'Record<string, string>',
   'devDependencies?': 'Record<string, string>',
   'peerDependencies?': 'Record<string, string>',
 });
```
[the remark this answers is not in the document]
aleksanderkatan: do we care?

cieplypolar (approved):
Nice

PR #2561, opened 2026-05-26
fix: Check for context instead of comptime mode in comptime

reczkok:


iwoplaza:
I added a test showcasing that this isn't a complete solution, we need to iterate on this

cieplypolar:
> I added a test showcasing that this isn't a complete solution, we need to iterate on this

Look at #2571. Maybe it is good enough

packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -198,7 +219,7 @@ export class TgpuMutableAccessorImpl<T extends BaseData>
   }
 
   get $(): InferGPU<T> {
-    if (inCodegenMode()) {
+    if (getResolutionCtx()) {
```
cieplypolar:
```suggestion
    const ctx = getResolutionCtx();
    if (!ctx) {
      throw new Error(
        '`tgpu.mutableAccessor` relies on GPU resources and cannot be accessed outside of a compute dispatch or draw call',
      );
    }

    if (ctx.mode.type !== 'codegen') {
      const slotValue = ctx.unwrap(this.slot);

      if (
        typeof slotValue !== 'function' &&
        !hasTinyestMetadata(slotValue) &&
        !isTgpuFn(slotValue)
      ) {
        return slotValue as unknown as InferGPU<T>;
      }

      ctx.pushMode(new CodegenState());
      try {
        return this[$gpuValueOf];
      } finally {
        ctx.popMode('codegen');
      }
    }

    return this[$gpuValueOf];
```

iwoplaza:
Not planned, as we decided accessors should only work in codegen mode

PR #2562, opened 2026-05-26
fix(cli): leave package name prompt empty

### LINKED ISSUE #2542: fix(@typegpu/cli): Remove the default name suggestion, leave just placeholder

Will-thom:
## Summary
- remove the pre-filled initial value from the package name prompt
- keep the generated name visible as placeholder only

Fixes #2542

## Tests
- corepack pnpm exec tsc --p packages/typegpu-cli/tsconfig.json --noEmit
- corepack pnpm oxfmt --check packages/typegpu-cli/src/utils/inputs.ts

Note: local Node is v22.21.1, while the repo asks for >=24.0.0; commands completed with the engine warning.

### DIFF
--- a/packages/typegpu-cli/src/utils/inputs.ts
+++ b/packages/typegpu-cli/src/utils/inputs.ts
@@ -31,7 +31,6 @@ export async function getPackageName(initialValue: string) {
   const packageName = await p.text({
     message: 'Package name:',
     placeholder: initialValue,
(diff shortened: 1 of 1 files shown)


aleksanderkatan:
The CLI will create the project in a wrong folder when the prompt is left empty because the placeholder is not used by default, it is just visual help for the user

PR #2568, opened 2026-05-26
feat: vite + react template

cieplypolar:
TODO:
- [x] add template to create script
- [x] automate template selector colors (typegpu signature gradient)


packages/typegpu-cli/templates/template-vite-react/src/App.tsx
```diff
@@ -0,0 +1,108 @@
+import React, { useMemo } from 'react';
```
aleksanderkatan, line 1: `import React` is no longer necessary since React 17, is there a reason to still import it?
```suggestion
import { useMemo } from 'react';
```

packages/typegpu-cli/src/create.ts
```diff
@@ -10,12 +10,31 @@ import { askForAgentSkills } from './steps/skills.ts';
...
+  },
+  {
+    value: 'vite-react',
+    label: 'Vite + React',
   },
-];
+] as const;
+
+const coloredLabelsTemplates = PROJECT_TEMPLATES.map((template, i) => {
+  const n = PROJECT_TEMPLATES.length;
```
aleksanderkatan: nit:
```suggestion
  const t = i / (PROJECT_TEMPLATES.length - 1);
```

aleksanderkatan (approved):
Sure! 🚨🎰

packages/typegpu-cli/templates/template-vite-react/src/App.tsx
cieplypolar: oxlint --check is not happy with that

iwoplaza (approved):

PR #2571, opened 2026-05-27
fix: comptime can read and work with accessors

cieplypolar:


packages/typegpu/src/core/slot/accessor.ts
```diff
@@ -174,13 +175,33 @@ export class TgpuAccessorImpl<T extends BaseData>
...
+    if (!ctx) {
+      throw new Error(
+        '`tgpu.accessor` relies on GPU resources and cannot be accessed outside of a compute dispatch or draw call',
+      );
     }
 
-    throw new Error(
-      '`tgpu.accessor` relies on GPU resources and cannot be accessed outside of a compute dispatch or draw call',
-    );
+    if (getExecMode().type !== 'codegen') {
```
reczkok: Since we already have the context, we might as well
```suggestion
    if (ctx.mode.type !== 'codegen') {
```

reczkok (approved):
I'm still conceptually pretty scared by combining comptime with accessors but it is very useful

PR #2572, opened 2026-05-28
fix: Handle `NaN` and `Infinity`

### LINKED ISSUE #2535: fix: Handle `NaN` and `infinity`
Can be written in plain JS, cannot appear during shader compilation in WGSL.
Furthermore, they are later appended with 'f'/'u'/'i', which can result in accidentally compiling code.
```ts
    const fn = () => {
      'use gpu';
      const NaNf = 0;
      const b = d.f32(NaN);
    };

      "fn fn_1() {
        const NaNf = 0;
        const b = NaNf;
      }"
    `);
```

Solution 1 (sane):
throw when encountered

Solution 2 (insane):
Replace NaN and infinity by calls to the following:
```ts
const divide = (a: number, b: number) => {
  'use gpu';
  return a / b;
};

const getNaN = () => {
  'use gpu';
  return divide(0, 0);
};

const getInfinity = () => {
  'use gpu';
  return divide(1, 0);
};
```

aleksanderkatan:
Calls to schemas and vectors with NaN or Infinity throw an error.
If a snippet with NaN or Infinity is resolved, an error is thrown.
On JS side, NaNs and Infinities may still appear as a result of function calls, like `std.div(1, 0)` or `std.bitcastU32toF32(0x7F800000)`


packages/typegpu/src/data/numeric.ts
```diff
@@ -136,6 +142,9 @@ const f32Cast = callableSchema({
     if (typeof v === 'boolean') {
       return v ? 1 : 0;
     }
+    if (!Number.isFinite(v)) {
+      throw new Error(`Cannot cast '${v}' to f32 because of the Finite Math Assumption.`);
+    }
```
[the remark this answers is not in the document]
aleksanderkatan: For consistency's sake I think it's best to throw in all constructors

packages/typegpu/src/data/numeric.ts
```diff
@@ -48,6 +48,9 @@ const u32Cast = callableSchema({
     if (typeof v === 'boolean') {
       return v ? 1 : 0;
     }
+    if (!Number.isFinite(v)) {
+      throw new Error(`Cannot cast '${v}' to u32 because of the Finite Math Assumption.`);
```
cieplypolar:
```suggestion
      throw new FiniteMathAssumptionError(v, u32)
```
or something like that?

cieplypolar (approved):
Sure. Left one nit.

packages/typegpu/src/data/numeric.ts
aleksanderkatan: Nice idea, will do

packages/typegpu/src/errors.ts
```diff
@@ -212,3 +212,13 @@ export class SignatureNotSupportedError extends Error {
...
+
+export class FiniteMathAssumptionError extends Error {
+  constructor(value: number, type: BaseData) {
+    super(
+      `Cannot convert value '${value}' to type ${type.type} because of the Finite Math Assumption.`,
+    );
+
+    Object.setPrototypeOf(this, FiniteMathAssumptionError.prototype);
+  }
+}
```
reczkok, line 224: Very clean approach but I would rethink the message (I know that this is the actual name from the spec but it won't mean much to readers IMO - either a more descriptive message or a link to the spec?)

reczkok (approved):
Great work

packages/typegpu/src/errors.ts
aleksanderkatan: I added a link to both errors

cieplypolar (approved):
Once again ✅

PR #2574, opened 2026-05-28
fix (@typegpu/cli): fix expo template

cieplypolar:
All package managers work


iwoplaza (approved):
Looks good, what was the issue in particular that caused `npm` to fail?

cieplypolar:
> Looks good, what was the issue in particular that caused `npm` to fail?

I assume that yarn and pnpm handle transitive dependencies differently than npm, and `npx expo install --fix` helped a lot.
Also, for some reason, only pnpm implicitly downloads `eslint` package required by our plugin.

PR #2597, opened 2026-06-10
feat: React prerendering test app

cieplypolar:
Blocked by #2606.

- [ ] we can add this test to CI if needed
   before doing that, we must fix `@typegpu/react` so it doesn't throw on the server side
- [x] useUniform also needs fixing
image

EDIT: my fault, didn't prefix env variables with `NEXT_PUBLIC_`


apps/react-prerendering-test/components/Shader.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { useMemo } from 'react';
+import tgpu, { common, d, std } from 'typegpu';
+import { perlin2d } from '@typegpu/noise';
+import { hexToOklab, oklabToRgb } from '@typegpu/color';
+
+export default function Shader() {
+  const root = useRoot();
+  const time = useUniform(d.f32);
+
+  function noise(v: d.v2f) {
```
iwoplaza: I'd prefer if the functions, accessors and all other resources that don't depend on the root or canvas be defined outside of the component, so that they're not recreated on every React re-render

apps/react-prerendering-test/checkSsr.mts
iwoplaza, line 1: I think it'd be nice if this script exited with an error code `1` if any of the nested steps fail.

iwoplaza (approved):
Thanks a lot! Left a few nits

apps/react-prerendering-test/checkSsr.mts
cieplypolar: sure

PR #2598, opened 2026-06-11
fix: Make numeric names valid

### LINKED ISSUE #2592: fix: Make numeric names valid
This may actually happen when using externals, since we name unnamed items based on the external key.

```ts
    const myBuf = root.createMutable(d.u32).$name('0');
    const fn = () => {
      'use gpu';
      myBuf.$ = 2;
    };

      "@group(0) @binding(0) var<storage, read_write> 0: u32;

      fn fn_1() {
        0 = 2u;
      }"
    `);
```

aleksanderkatan:


packages/typegpu/src/nameUtils.ts
```diff
@@ -359,14 +359,18 @@ export const builtins = new Set([
...
+/**
+ * Sanitizes the primer so that it is compliant with WGSL guidelines.
+ * This primer is not necessarily a valid identifier yet, as it may still collide with other idents or reserved keywords.
+ */
 /*#__NO_SIDE_EFFECTS__*/
 export function sanitizePrimer(primer: string | undefined) {
   if (primer) {
     const base = primer
       .replaceAll(/\s/g, '_') // whitespaces
       .replaceAll(/[^\w\d]/g, ''); // removing illegal characters
```
cieplypolar:
```suggestion
      .replaceAll(/[^\w]/g, ''); // removing illegal characters
```
`\w` includes digits

packages/typegpu/src/nameUtils.ts
```diff
@@ -413,11 +417,11 @@ export function validateIdentifier(ident: string): ValidationResult {
       error: `Identifiers cannot start with double underscores.`,
     };
   }
-  const prefix = ident.split('_')[0] as string;
-  if (bannedTokens.has(prefix) || builtins.has(prefix)) {
```
cieplypolar, line 417: We can't simply delete this check.
```ts
import { expect } from 'vitest';
import { it } from 'typegpu-testing-utility';
import tgpu, { d } from '../../src/index.js';

it('reserved keyword in compute fn', () => {
  const f = tgpu.computeFn({
    in: { loop: d.builtin.globalInvocationId },
    workgroupSize: [1],
  })`{
      let x = loop.x;
    }`;

    "@compute @workgroup_size(1) fn f(@builtin(global_invocation_id) loop: vec3u) {
          let x = loop.x;
        }"
  `);
});

it('reserved keyword in fragment fn', () => {
  const f = tgpu.fragmentFn({
    in: { loop: d.builtin.position },
    out: d.vec4f,
  })`{
      return vec4f(loop.x);
    }`;

    "@fragment fn f(@builtin(position) loop: vec4f) -> @location(0)  vec4f {
          return vec4f(loop.x);
        }"
  `);
});
```
The unused argument pruning will save as, but we should validate input props with more attention.

aleksanderkatan: I added an explicit check for this, see if you see this solution fit

iwoplaza: Nice! I assume that it still generally works because builtins and banned tokens are considered "taken", so they're being skipped when searching for a unique identifier

packages/typegpu/src/nameUtils.ts
```diff
@@ -413,11 +417,11 @@ export function validateIdentifier(ident: string): ValidationResult {
...
+  // see: https://www.w3.org/TR/WGSL/#syntax-ident_pattern_token
+  if (!/^(([_\p{XID_Start}][\p{XID_Continue}]+)|([\p{XID_Start}]))$/u.test(ident)) {
     return {
       success: false,
-      error: `Identifiers cannot start with reserved keywords.`,
+      error: `Identifier is not compliant with the WGSL guideline.`,
```
iwoplaza: Wording nit
```suggestion
      error: `Not compliant with WGSL guidelines.`,
```

iwoplaza (approved):

cieplypolar (approved):

PR #2600, opened 2026-06-11
docs: Lazy loading examples

iwoplaza:
Changes:
- Example source code is lazily-loaded, only when the example code is actually required to be viewed in the code editor. Same goes for the source code of `typegpu` and all packages that the editor needs for proper type inference.
- The CodeEditor component itself is loaded lazily, Monaco editor is quite beefy
- Example tags are computed ahead of time thanks to the `comptime` package (I had to patch it up a bit to work with Astro).

4G Throttling:
|Before|After|
|-|-|
|<video src="https://github.com/user-attachments/assets/b8fe134d-70ba-4c67-b429-e21160ee4a88" />|<video src="https://github.com/user-attachments/assets/403ba280-3ce8-4101-9b1e-fbeaf52e4ef0" />|


reczkok (approved):
Looks good :shipit:

aleksanderkatan (commented):
Looks great! I only skimmed the changes

PR #2604, opened 2026-06-12
feat: Asynchronous pipeline creation

### LINKED ISSUE #2501: feat: Add support for asynchronous pipeline creation
I feel like an API like that would be nice to have and citing [MDN](https://developer.mozilla.org/en-US/docs/Web/API/GPUDevice/createComputePipelineAsync), creating pipelines asynchronously **"is generally preferable"**
Image
We currently create all pipelines and resources lazily synchronously but maybe the future is asynchronous? I feel like that could enable us to distribute the work nicely (for example when creating multiple pipelines one can begin asynchronously being created by WebGPU API while we begin resolving the next)

aleksanderkatan:
Changes:
- refactor `ComputePipelineCore`,
- add new API for `initAsync` and `initSync`,
- update some examples to use `initAsync`,
- add tests, update docs.

The PERF is set to true only in browser tests, which currently do not work.
Also, I think that the measurements are just incorrect.
Still, I opted to leave them as is in a separate class. I tested manually the results before and after on Stable Fluids (7 resolves), and they were similar, so I don't think any logic changed.


apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -242,6 +242,29 @@ tgpu.resolve([innerPipeline]);
...
+const pipeline = root.createComputePipeline({
+  compute: mainCompute,
+});
+const myPipelines = [root.createComputePipeline({
+  compute: mainCompute,
+})];
+// ---cut---
+pipeline.initSync();
+
+await Promise.all(myPipelines.map((pipeline) => pipeline.initAsync()));
```
iwoplaza: I'd make the snippets a bit easier to read, I think people are familiar with how to await multiple promises at once, so we can skip doing that ourselves here for the sake of readability
```suggestion
pipeline.initSync();
// or
await pipeline.initAsync();
```

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -242,6 +242,29 @@ tgpu.resolve([innerPipeline]);
 ```
 :::
 
+## Initialization
+
+Pipeline initialization involves resolving the pipeline code, creating the shader module, and creating the underlying WebGPU pipeline.
+This happens automatically the first time the pipeline is executed via `.draw`, `.dispatchWorkgroups`, or a similar method.
+The `initSync` and `initAsync` methods let you perform this work ahead of time, avoiding a stall on first execution.
```
iwoplaza: I'm missing the explanation of when one should use `initSync` as opposed to `initAsync` and vice versa. On the other hand, it's non-trivial to explain it briefly 🤔

Maybe:


```suggestion
Pipeline initialization involves resolving the pipeline code, creating the shader module, and creating the underlying WebGPU pipeline.
This happens automatically the first time the pipeline is executed via `.draw`, `.dispatchWorkgroups`, or a similar method.
The `initSync` function lets you jump-start that initialization on the WebGPU process early. To await until the initialization actually finishes and avoid a stall on first execution, you can use the `initAsync` function instead.
```

apps/typegpu-docs/src/examples/image-processing/background-segmentation/index.ts
```diff
@@ -113,6 +113,7 @@ let blurBindGroups: TgpuBindGroup<typeof blurLayout.entries>[];
 const prepareModelInputPipeline = root
   .with(paramsAccess, paramsUniform)
   .createGuardedComputePipeline(prepareModelInput);
+prepareModelInputPipeline.pipeline.initSync();
```
iwoplaza: I'd prefer if we had the `initSync` and `initAsync` methods on guarded compute pipelines as well. This would make the helper functions that catch errors a lot simpler on both the type level and at runtime, since they wouldn't have to branch on whether something is a guarded pipeline or not, they would just call the `initSync` or `initAsync` function.

apps/typegpu-docs/src/examples/simulation/gravity/index.ts
```diff
@@ -113,6 +113,8 @@ const renderPipeline = root
     },
   });
 
+await Promise.all([computeCollisionsPipeline.initAsync(), computeGravityPipeline.initAsync()]);
```
iwoplaza: I'd prewarm these promises early:

```suggestion
// Initializing all used pipelines
const pipelinePromises = [
  computeCollisionsPipeline.initAsync(),
  computeGravityPipeline.initAsync(),
  skyBoxPipeline.initAsync(),
  renderPipeline.initAsync(),
];
```

And await them right before requesting the first frame:
```ts
// Only requesting the first frame after all pipelines have been initialized
await Promise.all(pipelinePromises);
requestAnimationFrame(frame);
```

On another note, why aren't we preinitializing the render pipelines?

aleksanderkatan: we aren't because they do not yethave initAsync/initSync methods - as I mentioned, I wanted to wait with them for an initial review so as not to do the same fixes twice

packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -352,82 +382,150 @@ class ComputePipelineCore implements SelfResolvable {
...
             label: `${getName(this) ?? '<unnamed>'} - Pipeline Layout`,
             bindGroupLayouts: usedBindGroupLayouts.map((l) => this.root.unwrap(l)),
           }),
           compute: { module },
+        })
+        .then((pipeline) => {
+          this.#memo = { pipeline, usedBindGroupLayouts, catchall, logResources };
+          this.#performanceTracker.measureCompile(device);
+          this.#initAsyncPromise = undefined;
+        });
```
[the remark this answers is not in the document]
aleksanderkatan: People probably won't be calling resolve after one fails, but I might as well

apps/typegpu-docs/src/examples/image-processing/background-segmentation/index.ts
aleksanderkatan: Done

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
aleksanderkatan: I updated this paragraph and added commented snippets, tell me what you think

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -242,6 +242,64 @@ tgpu.resolve([innerPipeline]);
 ```
 :::
 
+## Initialization
+
+Pipeline initialization involves resolving the pipeline code, creating the shader module, and creating the underlying WebGPU pipeline.
+This happens automatically the first time the pipeline is executed via `.draw`, `.dispatchWorkgroups`, or a similar method.
+The `initSync` method lets you start the initialization early.
+To wait until initialization actually finishes on the device (fully avoiding a stall on first execution), use `initAsync` instead.
```
reczkok, line 250: I would stress here (and **especially** in the AI Skill) that this is not that important for small/mid sized shaders - only makes sense when you have a bigger application or shaders that are very big or have very many of them) - AI is prone to overusing solutions like those.

apps/typegpu-docs/src/examples/algorithms/matrix-next/index.ts
```diff
@@ -107,6 +111,7 @@ function updateInputDisplays() {
 
 let isComputing = false;
 
+await Promise.all(pipelinePromises);
```
reczkok: I feel like we can await it even later (might make no difference tho)

reczkok (commented):
Looks nice, I like the performance tracking approach

aleksanderkatan:
@pullfrog review

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
aleksanderkatan: Wouldn't it be better to just not mention it in the skill at all? Can we discuss it in #2650 ?

iwoplaza: I am okay with the note added below

iwoplaza (commented):
Looks good, I'd just remove all uses of `initSync` and `initAsync` from examples 👌

aleksanderkatan:
@iwoplaza `initSync` on radiance example actually makes a difference, so I think it's good to leave it in (I added `initSync`, `initAsync` to library runners). I will remove it from other examples though.
Before:

https://github.com/user-attachments/assets/2e41fd3f-37a9-42a4-9caf-f925b75d1bab

After:

https://github.com/user-attachments/assets/abc03330-8a51-42f9-9eb3-b38a0ff1d7e4

Note: in order for the pipelines to not be cached, I needed to added a randomized `let` to every function's code.

packages/typegpu-sdf/src/jumpFlood.ts
```diff
@@ -255,6 +255,9 @@ export type Executor = {
   with(bindGroup: TgpuBindGroup): Executor;
   /** Clean up GPU resources created by this executor. */
   destroy(): void;
+
+  initSync(): void;
```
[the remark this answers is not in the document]
iwoplaza: I agree with the frog

packages/typegpu-radiance-cascades/src/runner.ts
```diff
@@ -57,6 +57,9 @@ export type RadianceCascadesExecutor<TOutput extends OutputResource = OutputText
   with(bindGroup: TgpuBindGroup): RadianceCascadesExecutor<TOutput>;
   destroy(): void;
   readonly output: TOutput;
+
+  initSync(): void;
```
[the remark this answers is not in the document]
iwoplaza: I agree with the frog

apps/typegpu-docs/src/examples/algorithms/mnist-inference/index.ts
```diff
@@ -102,6 +102,7 @@ const pipelines = {
     ? root.createComputePipeline({ compute: subgroupCompute })
     : null,
 };
+const pipelinePromises = [pipelines.default.initAsync(), pipelines.subgroup?.initAsync()];
```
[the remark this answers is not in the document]
aleksanderkatan: I left one usage of initAsync so that we have a coverage of the functionality in examples (initSync is used in radiance cascades example)

apps/typegpu-docs/src/examples/rendering/radiance-cascades-drawing/index.ts
```diff
@@ -205,6 +205,8 @@ const drawInteraction = createDrawInteraction({
   },
 });
 
+radianceRunner.initSync();
```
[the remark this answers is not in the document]
aleksanderkatan: Intentional, this example benefits a lot from this init

aleksanderkatan:
Well, it passes locally
Edit: a test was importing from source. Weird that it worked locally.

iwoplaza (approved):
Let's gooo

PR #2606, opened 2026-06-12
fix(unplugin-typegpu): Transform preserves exports

cieplypolar:
- [x] rollup does not work with default export


packages/unplugin-typegpu/test/use-gpu-directive.test.ts
```diff
@@ -1205,36 +1205,391 @@ test('hoists exported marked function statements', async () => {
...
+          return {};
+        }
+      }) && $.f)({});
+      console.log(add);
+      console.log(mul);
+
+      /** ADD */
+      export { add };
+      /** MUL */
+      export { mul };"
```
iwoplaza: The only slight nitpick is that the doc comments are associated with the export statements, instead of the function definitions.

packages/unplugin-typegpu/test/use-gpu-directive.test.ts
```diff
@@ -1205,36 +1205,391 @@ test('hoists exported marked function statements', async () => {
...
+          externalNames: []
+        },
+        externals: () => {
+          return {};
+        }
+      }) && $.f)({});
+      console.log(add);
+
+      /** ADD */
+      export default add;"
```
iwoplaza: Same here

iwoplaza (approved):
Thanks for the fixes! If there's a way to reassociate the comments with the function definitions, then that would be great

packages/unplugin-typegpu/test/use-gpu-directive.test.ts
cieplypolar: good point, done

iwoplaza (approved):
Thanks for the latest changes! :shipit:

PR #2616, opened 2026-06-16
[RFC] chore: Don't name externals in mergeExternals

aleksanderkatan:
Currently, whether an unnamed item is autonamed in `mergeExternals` depends on its depth in plugin-generated externals:
```ts
  it('...', () => {
    const c1 = (() => tgpu.const(d.vec2u, d.vec2u(1)))(); // unnamed
    const ext = { c2: (() => tgpu.const(d.vec2u, d.vec2u(2)))() /* unnamed */ };
    const fn = () => {
      'use gpu';
      const a = c1.$;
      const b = ext.c2.$;
    };

      "const c1: vec2u = vec2u(1);

      const item: vec2u = vec2u(2);

      fn fn_1() {
        const a = c1;
        const b = item;
      }"
    `);
  });
```
As seen in #2593, an actual deep pass for proper autonaming is not only difficult, but also hits the performance due to how long `isResolvable` takes.

I think removing this may be a good step in order to make our externals both correct and performant.


packages/typegpu/tests/rawFn.test.ts
```diff
@@ -539,6 +539,42 @@ describe('tgpu.fn with raw wgsl and missing types', () => {
...
+    `);
+  });
+
+  it("resolved externals's names stay the same", () => {
+    const c1 = (() => tgpu.const(d.vec2u, d.vec2u(1)))(); // unnamed
+    tgpu.resolve([tgpu.fn([])`() { let a = myConst; }`.$uses({ myConst: c1 })]);
+    tgpu.resolve([tgpu.fn([])`() { let a = otherName; }`.$uses({ otherName: c1 })]);
+
+    expect(getName(c1)).toMatchInlineSnapshot(`"myConst"`);
+  });
```
aleksanderkatan, line 577: This makes me wonder if we should do anything at all, though I don't think it's that much of an issue

apps/typegpu-docs/tests/individual-example-tests/shifting-gradient.test.ts
iwoplaza, line 1: This was a very useful case for ambient naming (through external keys), as `time` is the result of calling `useUniform`, a helper API that's not instrumented by unplugin-typegpu to be auto-named.

What we could do instead, is add a separate mechanism of ambient naming to 'use gpu' shader functions. When an external is retrieved using `getById`, and it is namable, we can name it. Same whenever we access a prop, we can name the value using the key if its namable. Wdyt?

aleksanderkatan: Great idea! I thought we could just add more cases to plugin autonaming, but this seems to do the job

packages/typegpu/tests/rawFn.test.ts
iwoplaza: I think it's fine, I think it's unusual for a thing to be references by other names in different WGSL templates, and even if it is, it's still a valid name for it.

iwoplaza (approved):
I like these changes! 💜

PR #2622, opened 2026-06-17
fix(unplugin-typegpu): expose built artifacts via $built$ exports

### LINKED ISSUE #2605: fix: `unplugin-typegpu` exports .ts file when imported from `workspace:*`
We cannot test if `unplugin-typegpu` works correctly when used in `babel.config.js`.
The reason is that we locally export only .ts file.

Expected:
```js
module.exports = (api) => {
  api.cache(true);
  return {
    presets: ['next/babel'],
    plugins: ['unplugin-typegpu/babel'], // cannot parse .ts file
  };
};
```

Current:
```js
module.exports = (api) => {
  api.cache(true);
  return {
    presets: ['next/babel'],
    plugins: ['./node_modules/unplugin-typegpu/dist/babel.js'],
  };
};
```

mvanhorn:
## Summary

Adds `$built$` sibling entries to the dev `exports` map in `packages/unplugin-typegpu/package.json` so consumers can opt into the compiled `dist/*.js` artifacts instead of the raw TypeScript source. Every existing subpath (`.`, `rollup`, `babel`, `bun`, `esbuild`, `farm`, `rolldown`, `rolldown-browser`, `rspack`, `vite`, `webpack`) gains a matching `<subpath>/$built$` entry resolving to `./dist/<name>.js` (and `./dist/<name>.d.ts` for types). The original `./src/*.ts` entries are left untouched, and `publishConfig.exports` is unchanged.

This mirrors the exact mechanism already used by the sibling `typegpu` package, which exposes `$built$` variants of each export pointing at its compiled `dist` output.

## Why this matters

Today the dev `exports` map points every subpath directly at `./src/*.ts`. When the package is consumed outside the dev/monorepo resolution (or when a consumer simply wants the compiled output), there is no way to opt into the built artifacts, so the raw `.ts` source is imported. Issue #2605 requests exactly this, and the maintainer confirmed the direction:

> "We could use the same mechanism we use for the `typegpu` package, and expose the built package artifacts under `unplugin-typegpu/babel/$built$`."

## Testing

- `package.json` remains valid JSON.
- `pnpm --filter unplugin-typegpu build` (tsdown) produces every referenced `dist` file; no `$built$` entry points at a non-existent artifact (all 11 subpaths cross-checked against the tsdown entry list).
- Runtime resolution verified: `require('unplugin-typegpu/babel/$built$')` resolves to the compiled `./dist/babel.js`.
- `pnpm --filter unplugin-typegpu test:types` passes; `oxlint` and `oxfmt --check` report no issues on the changed file.

Fixes #2605

### DIFF
--- a/packages/unplugin-typegpu/package.json
+++ b/packages/unplugin-typegpu/package.json
@@ -29,16 +29,60 @@
   "exports": {
     "./package.json": "./package.json",
     ".": "./src/index.ts",
+    "./$built$": {
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.js"
+    },
     "./rollup": "./src/rollup.ts",
+    "./rollup/$built$": {
+      "types": "./dist/rollup.d.ts",
+      "default": "./dist/rollup.js"
+    },
     "./babel": "./src/babel.ts",
+    "./babel/$built$": {
+      "types": "./dist/babel.d.ts",
+      "default": "./dist/babel.js"
+    },
     "./bun": "./src/bun.ts",
+    "./bun/$built$": {
+      "types": "./dist/bun.d.ts",
+      "default": "./dist/bun.js"
+    },
     "./esbuild": "./src/esbuild.ts",
+    "./esbuild/$built$": {
+      "types": "./dist/esbuild.d.ts",
+      "default": "./dist/esbuild.js"
+    },
     "./farm": "./src/farm.ts",
+    "./farm/$built$": {
+      "types": "./dist/farm.d.ts",
+      "default": "./dist/farm.js"
+    },
     "./rolldown": "./src/rolldown.ts",
+    "./rolldown/$built$": {
+      "types": "./dist/rolldown.d.ts",
+      "default": "./dist/rolldown.js"
+    },
     "./rolldown-browser": "./src/rolldown-browser.ts",
+    "./rolldown-browser/$built$": {
+      "types": "./dist/rolldown-browser.d.ts",
+      "default": "./dist/rolldown-browser.js"
+    },
     "./rspack": "./src/rspack.ts",
+    "./rspack/$built$": {
+      "types": "./dist/rspack.d.ts",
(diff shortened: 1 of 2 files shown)


iwoplaza (changes requested):
The exports only include the ESM export, I'd like them to export both the ESM and CJS exports, just like the published package

mvanhorn:
Done in f201454 - each $built$ export now provides explicit import (ESM) and require (CJS) conditions, matching the published package layout. Verified with a local tsdown build that dist emits .cjs/.d.cts alongside .js/.d.ts for every subpath.

mvanhorn:
Two open threads from the Copilot reviewer on `packages/unplugin-typegpu/package.json`, both I think resting on assumptions that do not hold for this repo:

1. On the `$built$` subpath naming. This is not new API surface invented here. `packages/typegpu/package.json` already ships `$built$` subpaths, so this PR is following the existing convention rather than establishing one. Renaming to `./built` would make unplugin-typegpu inconsistent with the core package, which seems like the worse trade. Happy to rename both together if you would rather move off the convention, but that feels like a separate change.

2. On `dist/` not being guaranteed in the published tarball. This package is published from inside `dist/`, not from the package root. `tgpu-dev-cli prepack` (which `prepublishOnly` runs) writes a transformed `dist/package.json` and rewrites every `./dist/*` path to `./*` on the way (`packages/tgpu-dev-cli/prepack.mjs:71-73`), then sets `private: false`, clears `scripts`, drops `devDependencies`, and copies the README in. So `dist/` is the tarball root at publish time and there is no `files` or `.npmignore` gap to close.

Re-verified the branch at a9d0196: all 11 `$built$` exports resolve for ESM, CJS, types, and the default fallback; the `.js`, `.cjs`, `.d.ts`, and `.d.cts` artifacts are generated; package build and typecheck pass; 84 focused tests pass; repo-wide lint and format pass.

iwoplaza (approved):

PR #2625, opened 2026-06-17
docs: Triplanar mapping and split-view component

reczkok:
## **IMPORTANT**: 2000 lines of this PR is the obj
It is very nice and simple :)


reczkok:
@pullfrog review please

apps/typegpu-docs/src/examples/rendering/triplanar-mapping/thumbnail.png
cieplypolar: Thumbnail shows material that is not available in the example

aleksanderkatan, line 1: I don't see this particular material available

apps/typegpu-docs/src/examples/rendering/triplanar-mapping/index.ts
```diff
@@ -0,0 +1,379 @@
...
+const albedoView = pbrTexture.createView(d.texture2d(d.f32), {
+  format: 'rgba8unorm-srgb',
+  baseArrayLayer: 0,
+  arrayLayerCount: 1,
+});
+const narmView = pbrTexture.createView(d.texture2dArray(d.f32), {
+  baseArrayLayer: 1,
+  arrayLayerCount: 4,
+});
+const Layer = { normal: 0, ao: 1, roughness: 2, metallic: 3 } as const;
```
aleksanderkatan, line 52: Sanity check, do I understand this correctly that this 4-channel, 5-layer texture only uses the first channel from layers 3, 4 and 5? Could this theoretically be a 3-layer texture if the PBR image was in different format?

aleksanderkatan (approved):
Great example! 43:20

apps/typegpu-docs/src/examples/rendering/triplanar-mapping/index.ts
reczkok: Yes, the new texture write API will accommodate this use case and allow to pack those into one texture with no offline conversions

PR #2627, opened 2026-06-17
fix: Throw descriptive error when recursion is detected in shellless function

cieplypolar:


iwoplaza (approved):
Looks good! Thanks for catching this and fixing it 🐝💛

reczkok (approved):

PR #2635, opened 2026-06-19
bump: tsover v6.0.2

iwoplaza:


packages/typegpu/src/core/texture/texture.ts
```diff
@@ -243,7 +243,8 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
           this.#texture = branch.device.createTexture({
             label: getName(this) ?? '<unnamed>',
             format: props.format,
-            size: props.size,
+            // The WebGPU types accept only mutable arrays, which is too loosely typed
+            size: props.size as number[],
```
aleksanderkatan, line 247: Just to be safe
```suggestion
            size: [...props.size],
```

aleksanderkatan (approved):
Sure!!

packages/typegpu/src/core/texture/texture.ts
iwoplaza: To be safe about what? 👀

PR #2636, opened 2026-06-19
chore: Skip dist/ typecheck

iwoplaza:
This was causing TypeScript to overflow the stack when running `pnpm run test:unit-and-attest`


reczkok (approved):
Yes please

PR #2638, opened 2026-06-19
test: Split tests into internal and not-internal tests

iwoplaza:
Blocked by #2642


packages/typegpu/tests/data/atomic.test.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, expect, expectTypeOf, it } from 'vitest';
-import * as d from '../../src/data/index.ts';
+import * as d from 'typegpu/data';
```
aleksanderkatan: Is this import intentionally like this instead of `import { d } from 'typegpu';`? Will we refactor this in an upcoming PR?

packages/typegpu/tests/internal/resolve.test.ts
aleksanderkatan, line 1: Many of these tests could be run on the built package

packages/typegpu/tests/unplugin/autoname.test.ts
aleksanderkatan, line 1: I'm not sure if I like what happened to this test. We plan on making `getName` public, right? Can we revert to its original state and put it in internals, with TODO to move it back here?

packages/typegpu/tests/vector.test.ts
```diff
@@ -66,8 +64,8 @@ describe('vec2f', () => {
 
     const buffer = new ArrayBuffer(d.sizeOf(d.vec2f));
 
-    writeData(new BufferWriter(buffer), d.vec2f, vec);
-    expect(readData(new BufferReader(buffer), d.vec2f)).toStrictEqual(vec);
+    writeToArrayBuffer(buffer, d.vec2f, vec);
```
aleksanderkatan, line 66: All these tests now use compiled writers instead of `writeData`, though I guess it's alright

aleksanderkatan (commented):
Great work! Those aren't yet run on the built package, right?

iwoplaza:
> Great work! Those aren't yet run on the built package, right?

@aleksanderkatan  Nope, that will be a separate PR

packages/typegpu/tests/vector.test.ts
iwoplaza: I made a task to more thoroughly test both the compiled and non-compiled writers: https://github.com/software-mansion/TypeGPU/issues/2658

packages/typegpu/tests/buffer.test.ts
```diff
@@ -1,15 +1,12 @@
...
 import { describe, expect, expectTypeOf, vi } from 'vitest';
-import * as common from '../src/common/index.ts';
-import * as d from '../src/data/index.ts';
-import { sizeOf } from '../src/data/sizeOf.ts';
-import type { ValidateBufferSchema, ValidUsagesFor } from '../src/index.js';
-import { getName } from '../src/shared/meta.ts';
-import type { InferPatch, IsValidBufferSchema, IsValidUniformSchema } from '../src/shared/repr.ts';
-import type { TypedArray } from '../src/shared/utilityTypes.ts';
+import * as common from 'typegpu/common';
+import * as d from 'typegpu/data';
```
aleksanderkatan: nit
```suggestion
import { d, common } from 'typegpu';
```

aleksanderkatan (approved):
I skimmed it, looks good!

Issue #2454, opened 2026-05-07
impr: Unify `TgpuMutable` and `TgpuBufferMutable` (and the `Readonly` versions)


reczkok:
Intuitively calling:
```ts
const countMutable = root.createBuffer(d.u32).$usage('storage').as('mutable');
```
should be equivalent to calling:
```ts
const countMutable = root.createMutable(d.u32);
```
But the types are entirely different

PR #2639, opened 2026-06-21
feat: accept buffer shorthands directly in createBindGroup

### LINKED ISSUE #2477: feat: Allow buffer shorthands (usages) directly in bind groups
In a continued effort to make shorthands easier to use, I think we should consider allowing buffer shorthands (and usages) as an input to `root.createBindGroup`

mvanhorn:
## Summary

Lets buffer shorthands (`root.createUniform` / `root.createMutable` / `root.createReadonly`) be passed directly as entries to `root.createBindGroup(...)`, without first unwrapping them back to the underlying buffer usage.

```ts
const uniform = root.createUniform(d.vec3f);

// before: had to unwrap to the buffer/usage
root.createBindGroup(layout, { foo: uniform.buffer });

// now: the shorthand is accepted directly
root.createBindGroup(layout, { foo: uniform });
```

Changes are purely additive on the input side:

- `LayoutEntryToInput` now accepts `TgpuUniform` for uniform entries and `TgpuMutable`/`TgpuReadonly` for storage entries (and `TgpuBufferShorthand` in the widest-type branch), reusing the existing `MemIdentity`/`UnwrapRuntimeConstructor` machinery so the same type guarantees apply.
- `TgpuBindGroupImpl.unwrap` detects shorthands via the existing `isBufferShorthand` guard and resolves them to their underlying `buffer`, so the rest of the bind-group machinery is unchanged.

All previously accepted entry types (raw buffer usages, `GPUBuffer`, textures, samplers) continue to work unchanged. This covers the direct bind-group acceptance path only and does not depend on #2454.

## Why this matters

Closes #2477. The shorthands were introduced to make buffers easier to use, but bind groups still required unwrapping them by hand, which was an inconsistency the issue (and the maintainer discussion) flagged. Accepting them directly removes that friction and matches the ergonomics shorthands were meant to provide.

## Testing

Added unit tests in `bindGroupLayout.test.ts` covering:

- a `createUniform` shorthand passed directly for a uniform entry, producing the same resolved bind group as the unwrapped buffer
- `createMutable` and `createReadonly` shorthands accepted for storage entries

Verified locally (in `packages/typegpu`):

- `pnpm run test:types` — passes
- `pnpm exec vitest run --project=!browser bindGroupLayout` — 36/36 pass
- `oxlint --type-aware` + `oxfmt --check` on the touched files — clean

Fixes #2477

### DIFF
--- a/packages/typegpu/src/tgpuBindGroupLayout.ts
+++ b/packages/typegpu/src/tgpuBindGroupLayout.ts
@@ -1,5 +1,12 @@
 import { isBuffer, type TgpuBuffer, type UniformFlag } from './core/buffer/buffer.ts';
 import {
+  isBufferShorthand,
+  type TgpuBufferShorthand,
+  type TgpuMutable,
+  type TgpuReadonly,
+  type TgpuUniform,
+} from './core/buffer/bufferShorthand.ts';
+import {
   isUsableAsUniform,
   type TgpuBufferMutable,
   type TgpuBufferReadonly,
@@ -341,6 +348,7 @@ export type LayoutEntryToInput<T extends TgpuLayoutEntry | null> =
   TgpuLayoutEntry | null extends T
     ?
         | TgpuBuffer<AnyWgslData>
+        | TgpuBufferShorthand<AnyWgslData>
         | GPUBuffer
         | TgpuSampler
         | GPUSampler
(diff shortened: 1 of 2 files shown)


aleksanderkatan:
Done in #2727

PR #2641, opened 2026-06-22
impr: Extend the `.write` texture API to be more capable

### LINKED ISSUE #2662: [docs] textures: Writing arrays of images to a 3D texture crashes
## Docs

This may be me who hasn't understood yet how it works but it seems that the ["Writing arrays of images"](https://docs.swmansion.com/TypeGPU/apis/textures/#writing-arrays-of-images) docs example doesn't work :

```ts
const texture3d = root.createTexture({
  size: [256, 256, 3],
  format: 'rgba8unorm',
  dimension: '3d',
}).$usage('sampled');

texture3d.write([imageBitmap1, imageBitmap2, imageBitmap3]); // ↓
```
 Image

WebGPU's `copyExternalImageToTexture` (used internally by `.write()`) only accepts 2D textures as destination — `dimension: '3d'` is invalid by spec, regardless of usage flags.

## Workaround

Drop `dimension: '3d'`. The default creates a 2D-array texture, which works:

```ts
const texture3d = root.createTexture({
  size: [256, 256, 3],
  format: 'rgba8unorm',
}).$usage('sampled', 'render');

texture3d.write([imageBitmap1, imageBitmap2, imageBitmap3]); // ✅
```

The docs example should be updated to use a 2D-array, or include a note that `dimension: '3d'` cannot be written from image sources.

## Environment

- TypeGPU `0.11.2`
- Arc / macOS

reczkok:
- [x] Decide whether to attempt `3d` texture support or throw a clear error (supported)

- `texture.write(image)` now takes: `origin`, `size`, `sourceOrigin`, `sourceSize`, `mipLevel`, `filter`, `flipY`, `premultipliedAlpha`, `colorSpace`
- implicit resampling on size mismatch is gone, mismatches now throw unless you pass `fit: 'stretch'` (scale) or `fit: 'clip'` (1:1 copy of the overlapping region)
- new `texture.writeAsync(blob)` decodes a Blob and writes it, with `fit: 'stretch'` the resize happens during decode so there is no pipeline on our side
- new `common.writeChannels(texture, { r: { source, from }, ... })` packs image sources into individual channels of one texture (lovely for pbr)
- new `texture.clear(color)` overload, and `copyFrom` / raw data writes accept region + mip options
- image writes now require the `'render'` usage flag (previously browser level fail)
- sRGB stays sRGB during resampling, wide formats stage in `rgba16float`, blit pipelines are cached per device

### fit behavior matrix

Target region = `size` option, or mip size minus origin by default.

| Case | no `fit` | `fit: 'stretch'` | `fit: 'clip'` |
| --- | --- | --- | --- |
| source == target | direct copy | direct copy (no-op) | direct copy (no-op) |
| source smaller | throws | blit, source scaled up to fill target | patch copy at origin, rest untouched |
| source larger | throws | blit, source scaled down to fit target | top-left part of source copied 1:1 |
| `origin` set, no `size` | throws unless source matches remaining region | source stretched over remaining region | patch copy at origin |
| `origin` + `size` set | throws unless source matches `size` | source scaled to fill `size` at `origin` | 1:1 copy at `origin`, cut to `min(source, size)` |
| `mipLevel` set | same rules against that mip's size | blit into that mip | clip against that mip |
| `writeAsync(blob)` | same rules after decode | resized during decode, no blit | decoded full, clipped on write |
| `writeChannels` | same rules | stretched channel blit | clipped channel blit |


packages/typegpu/src/common/writeChannels.ts
```diff
@@ -0,0 +1,90 @@
...
+  defaultWriteSize,
+  normalizeImageWrite,
+  validateFit,
+  type TextureChannel,
+  type TextureChannelWriteLayout,
+  type TextureImageWrite,
+  type TextureWriteOptions,
+} from '../core/texture/textureWrite.ts';
+import type { RenderFlag } from '../core/texture/usageExtension.ts';
+import { $internal } from '../shared/symbols.ts';
```
iwoplaza, line 13: Code in `typegpu/common` should not import from library internals (except for maybe `typegpu/~internal`), maybe there's a way to expose these through the public API, or in the extreme case, duplicate them?

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -167,12 +194,28 @@ export interface TgpuTexture<TProps extends TextureProps = any> extends TgpuNama
...
+    options?: TextureWriteOptions,
+  ): void;
+  /** Writes raw texel data to the texture */
+  write(
+    source: ArrayBuffer | TypedArray | DataView,
+    options?: number | TextureRawWriteOptions,
+  ): void;
+  /** Decodes an image blob and writes it to the texture. Requires the `'render'` usage flag */
+  writeAsync(source: Blob, options?: TextureBlobWriteOptions): Promise<void>;
   // TODO: support copies from GPUBuffers and TgpuBuffers
```
iwoplaza, line 214: Is there an issue for this? 👀

packages/typegpu/src/common/writeChannels.ts
reczkok: #2985

packages/typegpu/src/core/texture/texture.ts
reczkok: #2986

Issue #2645, opened 2026-06-23
[RFC] impr: Weak type checking on binary expression operands


cieplypolar:
It all comes down to this list:

```ts
const parenthesizedOps = [
  '==',
  '!=',
  '===',
  '!==',
  '<',
  '<=',
  '>',
  '>=',
  '<<',
  '>>',
  '+',
  '-',
  '*',
  '/',
  '%',
  '|',
  '^',
  '&',
  '&&',
  '||',
];
```

Currently, we don't really check the types of the operands. For example:

```ts
 const eq = tgpu.fn([Boid, BoidOnSteroids])((x, y) => {
  'use gpu';
  const r = x === y;
});
```
will happily compile.

Right now, we rely on the fact that wgsl compiler will throw a descriptive error.
I was fine with this approach for a long time, but...

### No 1. 
```
const x = d.vec3f();
const y = x;

const eq = () => {
  'use gpu';
  const _r = x === y;
};
```
The above function will inline `true` in place of the equality comparison.
However, if `x` and `y` were runtime known values, the return type of the function would have been `d.vec3b` (since that't how wgsl treats `==` between vectors).

### No 2.
Delegating error handling to the wgsl compiler quickly makes things difficult to debug.
For instance, out ternary operator polyfill uses `select` underneath. It does not check if the `condition` operand can be converted to a boolean, it assumes that typescript will prevent users from doing so. 

While typescript sometimes catches this, an equality comparison between runtime known vectors results in boolean vector. This completely changes the behavior of `select`. Furthermore, users end up with no idea  why there is a type mismatch in a `select` expression they didn't even use.

### No 3.
Conceptually, the wgsl generator should always produce valid code.

---

### Proposition:
Strictly check operand types to closely match the wgsl specification.

PR #2647, opened 2026-06-23
impr: Strict type checking on logical and relational binary expressions operands

### LINKED ISSUE #2702: [RFC] impr: && and || operators behavior
In JS, operators `&&` and `||` are tricky. For example, `&&` returns first falsy value.

Currently, we support short-circuiting of `&&` and `||` operators, but we force the result to be `bool`. We can extend this.

```ts
import { tgpu, d } from 'typegpu';

const f = () => {
	'use gpu';
	
	const a = falsyComptimeNumber() && runtimeVector();
	//    ^ number | vector
    // lhs is falsy at comptime
	// in generated wgsl: a is number

	const b = truthyComptimeNumber() && runtimeVector();
	//    ^ number | vector
    // rhs is truthy at comptime
	// in generated wgsl: b is vector
		
	if (falsyComptimeNumber() && runtimeNumber()) // comptime cast to bool 
	if (truthyComptimeNumber() && runtimeVector()) // tryConvertSnippet triggered, it will throw
	if (comptime1() && comptime2()) // easy comptime bool
	
	// if lhs is runtime then none of these apply, we force them to booleans
}
```

cieplypolar:
Part of the #2645.

Addressed: <, >, >=, <=, ===, !==, || and &&.

Also, forced ternary operator to check if condition is boolean.

Closes #2702.


iwoplaza:
Just a few notes from what we talked about in person:
- If the left operand is known at comptime, we don't have to require the right side to be of type boolean since it becomes the whole expression
- If the left operand is runtime known, then we should require both sides to be of type boolean in order to not confuse devs on behavior.

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan, line 1: There is a slight inconsistency between `if (...)` and `if (true && ...)` 

```ts

  it('...', () => {
    const Boid = d.struct({ prop: d.u32 });
    const myAccess = tgpu.accessor(Boid);

    const fn1 = () => {
      'use gpu';
      if (myAccess.$) {
      }
    };

    expect(() => tgpu.resolve([tgpu.fn(fn1).with(myAccess, { prop: 1 })]))
      .toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:fn1
      - fn*:fn1(): Cannot convert value of type 'struct:Boid' to any of the target types: [bool]]
    `);
    expect(() => tgpu.resolve([tgpu.fn(fn1).with(myAccess, tgpu.const(Boid, { prop: 1 }))]))
      .toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:fn1
      - fn*:fn1(): Cannot convert value of type 'struct:Boid' to any of the target types: [bool]]
    `);

    const fn2 = () => {
      'use gpu';
      if (true && myAccess.$) {
      }
    };

    expect(tgpu.resolve([tgpu.fn(fn2).with(myAccess, { prop: 1 })])).toMatchInlineSnapshot(`
      "fn fn2() {
        {

        }
      }"
    `);
    expect(() =>
      tgpu.resolve([tgpu.fn(fn2).with(myAccess, tgpu.const(Boid, { prop: 1 }))]),
    ).toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:fn2
      - fn*:fn2(): Cannot convert value of type 'struct:Boid' to any of the target types: [bool]]
    `);
```

In `if(X)` we return `this._typedExpression(Xnode, bool);` (a throw in the case above)
In `if(true && X)` we return `snip(!!X.value, bool, 'constant', false)` (true in the case above)

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -332,41 +340,72 @@ ${this.ctx.pre}}`;
...
-          return snip(!!rhsExpr.value, bool, 'constant', false);
+        if (rhsExpr.dataType === UnknownData) {
+          throw new WgslTypeError(`Right-hand side of '${op}' is of unknown type`);
         }
 
         // we can skip lhs
-        const convRhs = tryConvertSnippet(this.ctx, rhsExpr, bool, false);
-        const rhsStr = this.ctx.resolveSnippet(convRhs).value;
-        return snip(rhsStr, bool, 'runtime', convRhs.possibleSideEffects);
+        return rhsExpr;
```
aleksanderkatan: No cast to bool check & call here?

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -332,41 +340,72 @@ ${this.ctx.pre}}`;
...
+      }
+
+      if (!wgsl.isBool(lhsExpr.dataType) || !wgsl.isBool(rhsExpr.dataType)) {
+        throw new WgslTypeError(
+          `Logical expression '${op}' requires boolean operands. Got '${String(lhsExpr.dataType)}' and '${String(rhsExpr.dataType)}'.`,
+        );
+      }
+
+      const lhsStr = this.ctx.resolve(lhsExpr.value, lhsExpr.dataType).value;
+      const rhsStr = this.ctx.resolve(rhsExpr.value, rhsExpr.dataType).value;
```
aleksanderkatan:
```suggestion
      const lhsStr = this.ctx.resolveSnippet(lhsExpr).value;
      const rhsStr = this.ctx.resolveSnippet(rhsExpr).value;
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -332,41 +340,72 @@ ${this.ctx.pre}}`;
...
+
+      // hardcoded parentheses - operators not present in `parenthesizedOps`
```
aleksanderkatan, line 490: Why were they removed from there?

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -381,24 +420,26 @@ ${this.ctx.pre}}`;
...
-        isKnownAtComptime(rhsExpr)
-      ) {
+      const stdBinaryRelationalOp = binaryRelationalOpToStdMap[op];
+      if (stdBinaryRelationalOp && isKnownAtComptime(lhsExpr) && isKnownAtComptime(rhsExpr)) {
         const left = lhsExpr.value;
         const right = rhsExpr.value;
+
+        switch (op) {
+          case '===':
+            return snip(left === right, bool, 'constant', false);
```
aleksanderkatan, line 526: I get it that we want string and other comparisons to work, but it's still funny that comparing `d.vec3f()` with `d.vec3f()` will result with `false`. Maybe we should put in docs?

aleksanderkatan (commented):
Yay! Some consistency!

packages/typegpu/src/tgsl/wgslGenerator.ts
cieplypolar: because now && and || are in separate branch, no need for additional lookup

packages/typegpu/src/tgsl/wgslGenerator.ts
cieplypolar: Fair

packages/typegpu/src/tgsl/wgslGenerator.ts
cieplypolar: `tgpu.const` is not treated as comptime-known. I think this is correct, as we agreed it opts out of comptime execution.

packages/typegpu/src/tgsl/wgslGenerator.ts
cieplypolar: Sure, I'll add docs

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: It was here to be a baseline. The inconsistency itself is the following (prepending a condition with `true &&` should not change behavior):
```ts
  it('...', () => {
    const Boid = d.struct({ prop: d.u32 });
    const myAccess = tgpu.accessor(Boid);

    const fn1 = () => {
      'use gpu';
      if (myAccess.$) {
      }
    };

    expect(() => tgpu.resolve([tgpu.fn(fn1).with(myAccess, { prop: 1 })]))
      .toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:fn1
      - fn*:fn1(): Cannot convert value of type 'struct:Boid' to any of the target types: [bool]]
    `);

    const fn2 = () => {
      'use gpu';
      if (true && myAccess.$) {
      }
    };

    expect(tgpu.resolve([tgpu.fn(fn2).with(myAccess, { prop: 1 })])).toMatchInlineSnapshot(`
      "fn fn2() {
        {

        }
      }"
    `);
```
if (X) tries to cast to bool and throws,
if (true && X) returns snip(!!rhsExpr.value, bool, 'constant', false)

cieplypolar: I see

iwoplaza (approved):
Looks good! 🍞

PR #2648, opened 2026-06-24
fix: External traversal for raw WGSL implemented functions

### LINKED ISSUE #2620: fix: External traversal for raw wgsl implemented functions
There is a lot of unnecessary traversal. We could instead find usages of top level externals, and traverse accesses from there until either the prop chain ends, or we come across something resolvable.

aleksanderkatan:
Before: given an external, if resolvable then replace occurrences with resolved, otherwise look which keys appear in the code with a prop following them, and call recursively with `external.prop`.

Now: given externals, look for prop chains starting with external keys, then follow them until either a resolvable appears, or the chain ends.

Other changes:
- I removed the warn when an unused external is found. I think that it may occur that a given set of externals is applied to all functions to avoid having to list them separately or to allow for user provided raw WGSL implementations, and it was unsilencable. Tell me if you think it's worth adding it back (it will be tricky, we will have to traverse externals for that).
- Resolved wgsl is now ordered differently. Before, externals were added in order they were provided, now there is only one regex match call, and the order matches the order of appearance.


aleksanderkatan:
@pullfrog review

packages/typegpu/src/core/resolve/externals.ts
```diff
@@ -87,52 +84,54 @@ export function replaceExternalsInWgsl(
...
     }
 
-    console.warn(
-      `During resolution, the external '${externalName}' has been omitted. Only TGPU resources, 'use gpu' functions, primitives, and plain JS objects can be used as externals.`,
-    );
+    let resolved = cache.get(currentItem);
+    if (resolved === undefined) {
+      resolved = ctx.resolve(currentItem).value;
+      cache.set(currentItem, resolved);
+    }
```
iwoplaza: The context has an internal cache, what happens if you remove this cache?

aleksanderkatan: It threw, though now that I think about that, those were recursive functions errors, maybe I should take a closer look

aleksanderkatan: I assumed this is the intended behavior, since before the change, each item was resolved exactly once

aleksanderkatan: #2677

packages/typegpu/src/core/resolve/externals.ts
```diff
@@ -87,52 +84,45 @@ export function replaceExternalsInWgsl(
...
-                innerAcc,
-              )
-            : innerAcc,
-        acc,
-      );
+      if (typeof currentItem !== 'object' || currentItem === null || i === chain.length - 1) {
+        console.warn(
+          `During resolution, the external '${chain.slice(0, i + 1).join('.')}' has been omitted. Only TGPU resources, 'use gpu' functions, primitives, and plain JS objects can be used as externals.`,
+        );
+        return match;
```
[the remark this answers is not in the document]
aleksanderkatan: Do we care?

packages/typegpu/src/core/resolve/externals.ts
```diff
@@ -66,18 +66,18 @@ export function addReturnTypeToExternals(
...
-    `(?<![\\w\\$_.])${name.replaceAll('.', '\\.').replaceAll('$', '\\$')}(?![\\w\\$_])`,
-    'g',
-  );
-}
+const anyIdent = /([$_\p{XID_Start}][$\p{XID_Continue}]*)/u; // WGSL ident, modified to include $
+const anyPropChain = new RegExp(`(${anyIdent.source})(\\.${anyIdent.source})*`, 'ug');
+const boundedPropChain = new RegExp(
+  `(?<![\\p{XID_Continue}\\$.])${anyPropChain.source}(?![\\p{XID_Continue}\\$])`,
+  'ug',
+);
```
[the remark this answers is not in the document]
iwoplaza: This seems to be a hallucination. Nevertheless, I would add some internal tests for this Regex to verify that it in fact works.

Image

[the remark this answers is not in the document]
aleksanderkatan: will do

packages/typegpu/tests/internal/externals.test.ts
```diff
@@ -276,3 +281,58 @@ describe('external name collisions', () => {
...
+        `"p1.p2.p3.p4.p5"`,
+      );
+      expect(matches(' p1.$ ', boundedPropChain)).toMatchInlineSnapshot(`"p1.$"`);
+    });
+
+    it('does not match non prop chains', () => {
+      expect(matches(' 1.0 ', boundedPropChain)).toMatchInlineSnapshot(`""`);
+      expect(matches(' .5f ', boundedPropChain)).toMatchInlineSnapshot(`""`);
+      expect(matches(' .prop ', boundedPropChain)).toMatchInlineSnapshot(`""`);
+      expect(matches(' p1 . p2 ', boundedPropChain)).toMatchInlineSnapshot(`"p1 p2"`);
```
aleksanderkatan, line 331: Is this how it should be?

iwoplaza: Technically it works in JS, and in WGSL, so why not
Image

Image

iwoplaza: Aah, sorry, I got it backwards. Honestly, I wouldn't worry about it too much

iwoplaza (approved):
Very awesome! 🌸

PR #2655, opened 2026-06-29
chore: Automatically publishing packages that have changed since release as PR packages

iwoplaza:
**Changes**:
- Gathering packages that have changed compared to the `release` branch, builds them, and publishes them through `pkg.pr.new`
- Made `@typegpu/gl` a public package, so that it can be published both on npm and on pkg-pr-new.

### DIFF
--- a/.github/workflows/pkg-pr.yml
+++ b/.github/workflows/pkg-pr.yml
@@ -14,6 +14,9 @@ jobs:
       - name: Checkout your repository using git
         uses: actions/checkout@v4
 
+      - name: Fetch release ref
+        run: git fetch --no-tags --depth=1 origin release:release
+
       - name: Print commit id, message and tag
         run: |
           git show -s --format='%h %s'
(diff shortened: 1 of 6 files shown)


scripts/publish-pkg-pr.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  // Building the packages
+
+  const buildArgs = toPublish.map((pkg) => `--filter ${pkg.name}`);
+  await $`pnpm run ${buildArgs} prepublishOnly --skip-publish-tag-check`.env({
+    ...process.env,
+    SKIP_TESTS: 'true',
+  });
+
+  const publishArgs = toPublish.map((pkg) => `'./packages/${pkg.name}'`);
```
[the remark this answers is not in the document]
iwoplaza: './packages/typegpu-gl' exists as a path though...

.github/workflows/pkg-pr.yml
```diff
@@ -14,6 +14,9 @@ jobs:
       - name: Checkout your repository using git
         uses: actions/checkout@v4
 
+      - name: Fetch release ref
+        run: git fetch --no-tags --depth=1 origin release:release
```
aleksanderkatan, line 18: Wouldn't it make more sense to publish based on target branch? I think this would simplify logic, as well as reduce unnecessary publishes

cieplypolar (approved):
Looks good

.github/workflows/pkg-pr.yml
iwoplaza: When you want to test out a nightly package, it usually depends on other changes already introduced in `main`. In this setup, you can go to a PR you want to test out, and just get all the necessary packages without having to jump around other PRs

aleksanderkatan (approved):
Sure

PR #2663, opened 2026-06-30
chore: Remove `bound` prop from layout

aleksanderkatan:
Its types reference buffer usages, so I'd rather just remove them instead of updating.


aleksanderkatan:
@pullfrog review

reczkok (approved):
lovely

cieplypolar (approved):
Looks good

PR #2664, opened 2026-06-30
test: Test built package in CI

iwoplaza:


.github/workflows/ci.yml
```diff
@@ -28,7 +28,6 @@ jobs:
 
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
+
       - name: Test
         run: pnpm test
-      - name: Build docs
```
iwoplaza, line 33: Docs are now built as part of the `test` script

iwoplaza:
@pullfrog review

reczkok (approved):

package.json
```diff
@@ -22,7 +22,7 @@
     "dev": "DEV=true pnpm run --filter typegpu-docs dev",
     "dev:host": "DEV=true pnpm run --filter typegpu-docs dev --host --mode https",
     "fix": "oxlint -c oxlint.config.ts --fix && oxfmt",
-    "test": "pnpm run test:types && pnpm run test:style && pnpm run test:unit-and-attest && pnpm run test:circular-deps",
+    "test": "pnpm run test:types && pnpm run test:style && pnpm run -r build && TEST_BUILT=1 pnpm run test:unit-and-attest && pnpm run test:circular-deps",
```
aleksanderkatan: I think it may be confusing in a couple of weeks why `nr test` runs on built package, and say `nr test:unit` or `nr test:types` does not. Can we create a new script for this? For example `test:built`

aleksanderkatan (changes requested):
I don't think we can just forget about internal tests in CI. See #2668, it passes despite broken tests

package.json
iwoplaza: Good points, thanks for the review. I updated 🫡

aleksanderkatan (approved):
Great, this approach looks much cleaner! 📝🛠️

packages/typegpu-testing-utility/src/extendedIt.ts
```diff
@@ -138,10 +138,39 @@ export const it = base
...
+        }));
+      },
+      get devicePromise(): Promise<GPUDevice> {
+        return (devicePromise ??= this.stallPromise.then(() => device));
+      },
+      get stallResolve() {
+        void this.stallPromise; // ensuring the promise is initialized
+        return stallResolve;
+      },
+    };
```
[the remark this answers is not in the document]
iwoplaza: Fixtures are scoped per test case by default, so this cache is only alive for the duration of a single test

cieplypolar (approved):
Looks good

PR #2674, opened 2026-06-30
chore (@typegpu/cli): Bump deps of expo template

cieplypolar:


packages/typegpu-cli/src/create.ts
aleksanderkatan, line 1: Not sure when this happened, but now the controls are displayed and we can remove them from this step:

Image

aleksanderkatan, line 1: Is this expected after `npm i`?
Image

iwoplaza: It's the same for me, we can remove them

cieplypolar: `audit` yes, speaking of `uuid`, I believe it is not dependency that I put in expo template

cieplypolar: sure, I'll leave the bottom one

aleksanderkatan (approved):

reczkok (approved):

iwoplaza (approved):

PR #2676, opened 2026-06-30
feat (@typegpu/cli): TypeGPU ASCII logo

cieplypolar:
image

Width: 31 characters

Let me know, if you want flag to disable it 🙏 .


aleksanderkatan (approved):
🪵🎨

packages/typegpu-cli/src/create.ts
```diff
@@ -16,7 +16,7 @@ import {
 } from './options.ts';
 
 const GRADIENT_START = [0.831, 0.553, 1.0] as const;
-const GRADIENT_END = [0.32, 0.4, 0.95] as const;
+const GRADIENT_END = [0.0, 0.349, 0.874] as const;
```
aleksanderkatan: Why this change?

cieplypolar: better aligned with gradient from svg

iwoplaza:
I'm still a bit iffy on the size that the logo has to be to be readable when using this method.

Could you try to insert this ASCII sequence of the plum outline, color it using the gradient, and see how that looks?

```
 /|
 \|____
 , `\  \
/ `\ \  \
\   `\| /
 \_____/
```

aleksanderkatan (approved):

packages/typegpu-cli/src/create.ts
```diff
@@ -72,6 +70,8 @@ export async function createProject(cwd: string, options?: CreateProjectOptions)
 
   p.log.step(`Scaffolding project in ${projectName}...`);
 
+  const root = await prepareDirectory(cwd, projectName, { interactive: !nonInteractive });
```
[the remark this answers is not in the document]
cieplypolar: but we don't want to create directory if user cancels the template choice prompt

iwoplaza:
The no-color versions seems to be offset in some rows

image

cieplypolar:
> The no-color versions seems to be offset in some rows
> 
> image

fixed 🖖

reczkok (approved):
Very cool

PR #2691, opened 2026-07-03
feat(@typegpu/color): Tonemapping functions

### LINKED ISSUE #1730: feat: Add tonemapping functions to `@typegpu/color`
Here are some great sources:
https://mini.gmshaders.com/p/tonemaps
https://bruop.github.io/tonemapping/

Aquel32:
What was added:
- Most common tonemapping functions such as `aces`, `hable`, `reinhard` and `neutral`.
- A new dev example (`tonemapping`) in the docs.
- Dedicated WGSL code validation test.

Closes #1730


cieplypolar:
Running `pnpm run fix` will resolve most of the issues. It starts by running `oxlint --fix`. It will list problems that cannot be automatically fixed. Once you have corrected those, a second run with format everything using `oxfmt` according to our repository standard 🫡 .

packages/typegpu-color/src/tonemapping.ts
```diff
@@ -0,0 +1,83 @@
...
+  return saturate(hableCurve(rgb) / hableCurve(W));
+});
+
+export const reinhard = tgpu.fn(
+  [vec3f],
+  vec3f,
+)((rgb) => {
+  'use gpu';
+
+  return saturate(rgb / (vec3f(1.0) + rgb));
```
aleksanderkatan, line 53: nit
```suggestion
  return saturate(rgb / (1 + rgb));
```

packages/typegpu-color/src/tonemapping.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const hable = tgpu.fn(
+  [vec3f],
+  vec3f,
+)((rgb) => {
+  'use gpu';
+
+  const W = vec3f(11.2);
+
+  return saturate(hableCurve(rgb) / hableCurve(W));
```
aleksanderkatan, line 44: I found one implementation that multiplies rgb by 2 here, which one is correct?
https://github.com/GPUOpen-LibrariesAndSDKs/Cauldron/blob/master/src/VK/shaders/tonemappers.glsl

packages/typegpu-color/src/tonemapping.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  const startCompression = 0.8 - 0.04;
+  const desaturation = 0.15;
+
+  const x = min(rgb.r, min(rgb.g, rgb.b));
+  const offset = select(0.04, x - 6.25 * x * x, x < 0.08);
+
+  let color = rgb - offset;
+
+  const peak = max(color.r, max(color.g, color.b));
```
aleksanderkatan, line 70: nit: our max supports this
```suggestion
  const peak = max(color.r, color.g, color.b);
```

packages/typegpu-color/src/tonemapping.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  if (peak < startCompression) {
+    return saturate(color);
+  }
+
+  const d = 1.0 - startCompression;
```
aleksanderkatan, line 76: nit: Using `1.0` instead of `1` does not change how we interpret numbers, in fact it may even be misleading since all of those are abstract ints, they just get converted to floats based on how they are used

Image

apps/typegpu-docs/src/examples/image-processing/tonemapping/meta.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Tonemapping",
+  "category": "image-processing",
+  "tags": ["ecosystem", "color", "tonemapping"],
+  "dev": true,
```
aleksanderkatan, line 5: I think it may be a good showcase, and it may make your functions more discoverable
Not every example needs to be 2k lines long
```suggestion
```

apps/typegpu-docs/src/examples/image-processing/tonemapping/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  },
+  'Point Light Y': {
+    initial: 0.5,
+    min: 0,
+    max: 1,
+    step: 0.01,
+    onSliderChange: (value: number) => {
+      pointLightY.write(value);
+    },
+  },
```
aleksanderkatan, line 151: You could use vector slider if you wanted to
Also, this would allow merging `pointLightX` and `pointLightY` buffers into one without using partial writes/keeping CPU state

```suggestion
  'Point Pos': {
    initial: d.vec2f(0.5),
    min: d.vec2f(),
    max: d.vec2f(1),
    step: d.vec2f(0.01),
    onVectorSliderChange: (value) => {
      pointLightX.write(value.x);
      pointLightY.write(value.y);
    },
  },
```

apps/typegpu-docs/src/examples/image-processing/tonemapping/index.ts
```diff
@@ -0,0 +1,162 @@
...
+
+let sideBySide = false;
+
+const mainFragment = tgpu.fragmentFn({
+  in: { uv: d.vec2f },
+  out: d.vec4f,
+})(({ uv }) => {
+  'use gpu';
+
+  const brightness = pointLightScale.$ / std.length(uv - d.vec2f(pointLightX.$, pointLightY.$));
```
aleksanderkatan, line 23: really nit: this would make it more verbose
```suggestion
  const brightness = pointLightScale.$ / std.distance(uv, d.vec2f(pointLightX.$, pointLightY.$));
```

apps/typegpu-docs/src/examples/image-processing/tonemapping/index.ts
```diff
@@ -0,0 +1,162 @@
...
+  fragment: mainFragment,
+});
+
+const sideBySideFragment = tgpu.fragmentFn({
+  in: { uv: d.vec2f },
+  out: d.vec4f,
+})(({ uv }) => {
+  'use gpu';
+
+  const rows = 3;
```
aleksanderkatan: Rows is set to 3, but this actually handles 2 rows

apps/typegpu-docs/src/examples/image-processing/tonemapping/index.ts
aleksanderkatan, line 1: If you find this view interesting, here's a code for it (no pressure to include it, I was just curious to how it would look like)

Image

Image

```ts
const mappings = {
  0: (v: d.v3f) => {
    'use gpu';
    return d.vec3f(v);
  },
  1: aces,
  2: hable,
  3: reinhard,
  4: neutral,
} as const;
const COUNT = Object.keys(mappings).length;
const timeUniform = root.createUniform(d.f32);

const mainFragment = tgpu.fragmentFn({
  in: { uv: d.vec2f },
  out: d.vec4f,
})(({ uv }) => {
  'use gpu';

  const brightness = pointLightScale.$ / std.distance(uv, d.vec2f(pointLightX.$, pointLightY.$));
  let color = pointLightColor.$ * brightness;

  const z = uv - 0.5;
  const angle = std.atan2(z.y, z.x) + Math.PI;
  const segment = (Math.PI * 2) / COUNT;

  for (const i of tgpu.unroll(std.range(COUNT))) {
    if (i * segment <= angle && angle < (i + 1) * segment) {
      color = mappings[i as keyof typeof mappings](color);
    }
  }

  return d.vec4f(color, 1.0);
});
```

Also, here's how it looks when moving automatically:

https://github.com/user-attachments/assets/cdf9b06a-3224-4725-be9f-fc6768e46527

aleksanderkatan (approved):
Great changes! 🎵🗺️

PR #2692, opened 2026-07-05
feat: Support `TRANSIENT_ATTACHMENT` for Textures and add an escape hatch

reczkok:
This PR also adds `setFlags` for textures, it's not `addFlags` like for the buffers because i believe for the escape hatch we should give full control (for example `addFlags` would've not been enough for users wanting to use TRANSIENT_ATTACHMENT before we officially support it since it is exclusive with the default copy flags, I believe buffers should also expose such a method in the future)


packages/typegpu/tests/texture.test.ts
```diff
@@ -154,6 +154,55 @@ describe('TgpuTexture', () => {
...
+    ).toThrow("Transient texture usage cannot be combined with 'sampled' or 'storage'.");
+  });
+
+  it('sets raw WebGPU usage flags exactly', ({ root, device }) => {
+    const texture = root
+      .createTexture({
+        size: [512, 512],
+        format: 'rgba8unorm',
+      })
+      .$setFlags(GPUTextureUsage.RENDER_ATTACHMENT);
```
iwoplaza: Since this is an escape hatch, maybe a name like `overrideFlags` would make that more apparent by just looking at the use of this API?
```suggestion
      .$overrideFlags(GPUTextureUsage.RENDER_ATTACHMENT);
```

reczkok: Yeah, I like that idea

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -142,6 +142,9 @@ export interface TgpuTexture<TProps extends TextureProps = any> extends TgpuNama
   $usage<T extends AllowedUsages<TProps>[]>(
     ...usages: T
   ): this & UnionToIntersection<LiteralToExtensionMap[T[number]]>;
+  $overrideFlags(
+    flags: GPUTextureUsageFlags,
+  ): this & UnionToIntersection<LiteralToExtensionMap['sampled' | 'storage' | 'render']>;
```
iwoplaza: I feel like this is a bit clearer and involves less TypeScript magic
```suggestion
  ): this & StorageFlag & SampledFlag & RenderFlag;
```

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -265,22 +269,52 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
...
     this.usableAsSampled ||= hasSampled;
-    this.usableAsRender ||= hasRender;
+    this.usableAsRender ||= hasRender || hasTransient;
 
     return this as this & UnionToIntersection<LiteralToExtensionMap[T[number]]>;
   }
 
+  $overrideFlags(
+    flags: GPUTextureUsageFlags,
+  ): this & UnionToIntersection<LiteralToExtensionMap['sampled' | 'storage' | 'render']> {
```
iwoplaza:
```suggestion
  ): this & StorageFlag & SampledFlag & RenderFlag {
```

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -265,22 +269,52 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
...
+    this.#flags = flags;
+    this.#flagsOverridden = true;
+    this.usableAsSampled = true;
+    this.usableAsStorage = true;
+    this.usableAsRender = true;
+    return this as this &
+      UnionToIntersection<LiteralToExtensionMap['sampled' | 'storage' | 'render']>;
```
iwoplaza:
```suggestion
    return this as this & StorageFlag & SampledFlag & RenderFlag;
```

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -265,22 +269,52 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
...
     const hasRender = usages.includes('render');
+    const hasTransient = usages.includes('transient');
+    const hasTransientConflict = hasSampled || hasStorage;
+
+    if (
+      (hasTransient && (hasTransientConflict || this.usableAsSampled || this.usableAsStorage)) ||
+      (this.#flags & GPUTextureUsage.TRANSIENT_ATTACHMENT && hasTransientConflict)
+    ) {
+      throw new Error("Transient texture usage cannot be combined with 'sampled' or 'storage'.");
+    }
```
iwoplaza: I'm wondering if it would be simpler to first save the value of `#flags`, apply the requested flags, and if the is a conflict, revert and throw. This way, we wouldn't have to check both the current flags as well as the would-be flags.

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -265,22 +273,54 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
...
+      this.#flags & GPUTextureUsage.TRANSIENT_ATTACHMENT &&
+      this.#flags & (GPUTextureUsage.TEXTURE_BINDING | GPUTextureUsage.STORAGE_BINDING)
+    ) {
+      this.#flags = previousFlags;
+      throw new Error("Transient texture usage cannot be combined with 'sampled' or 'storage'.");
+    }
+
+    if (hasTransient) {
+      this.#flags = GPUTextureUsage.TRANSIENT_ATTACHMENT | GPUTextureUsage.RENDER_ATTACHMENT;
+    }
```
iwoplaza, line 304: I think you're doing this already above on line 289

```ts
this.#flags |= hasTransient
  ? GPUTextureUsage.TRANSIENT_ATTACHMENT | GPUTextureUsage.RENDER_ATTACHMENT
  : 0;
```

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -265,22 +273,53 @@ class TgpuTextureImpl<TProps extends TextureProps> implements TgpuTexture<TProps
...
+
+    const hasTransientUsage =
+      hasTransient || !!(this.#flags & GPUTextureUsage.TRANSIENT_ATTACHMENT);
+    const hasSampledOrStorageUsage = !!(
+      nextFlags &
+      (GPUTextureUsage.TEXTURE_BINDING | GPUTextureUsage.STORAGE_BINDING)
+    );
+
+    if (hasTransientUsage && hasSampledOrStorageUsage) {
+      throw new Error("Transient texture usage cannot be combined with 'sampled' or 'storage'.");
```
aleksanderkatan, line 303: There are more requirements for transient flag to be a valid, like the texture being 2d, and having mipmap count 1, do we rely on webgpu errors for these?

aleksanderkatan (approved):
Sure! 🦑🪂

packages/typegpu/src/core/texture/texture.ts
reczkok: Yeah we did up to this point for other stuff

cieplypolar (approved):

PR #2700, opened 2026-07-09
fix: Restrict using slot for strings

### LINKED ISSUE #2498: fix: Restrict using `tgpu.slot` for injecting raw shader code
Currently we treat slots that carry strings like WGSL-code containers:

```ts
const fooSlot = tgpu.slot<string>();

const bar = tgpu.fn([d.f32], d.f32)`(x) {
  return fooSlot;
}`.$uses({ fooSlot });

const shader = tgpu.resolve([bar.with(fooSlot, 'x + 5')]);
```
Generates:
```wgsl
fn bar(x: f32) -> f32 {
  return x + 5;
}
```

In JS-implemented functions, string-carrying slots have more uses than that, for example for comptime checks:

```ts
const modeSlot = tgpu.slot<'fast' | 'precise'>();

function bar() {
  'use gpu';
  if (modeSlot.$ === 'fast') {
    // ...
  } else {
    // ...
  }
}
```

The same WGSL-injection functionality can be achieved using `tgpu.rawCodeSnippet`, and provides more context when used from a JS-implemented function.

## Solution
I suggest we just throw an Error when a string of type `UnknownData` is about to be injected into source code.

aleksanderkatan:
Includes changes from #2703 by @iwoplaza


aleksanderkatan:
@pullfrog review

packages/typegpu/src/core/function/fnCore.ts
```diff
@@ -127,7 +128,10 @@ export function createFnCore(
 
           this.setExternals('args', {
             in: Object.fromEntries(
-              entryInput.positionalArgs.map((a) => [a.schemaKey, a.schemaKey]),
+              entryInput.positionalArgs.map((a) => [
+                a.schemaKey,
+                new ResolvableString(a.schemaKey),
```
iwoplaza, line 133: I feel like we could use a raw code snippet here, instead of creating a new `ResolvableString` type

aleksanderkatan: I guess we could. We would need to lift the limitation about 'argument' origin not being allowed. And then, the data type, origin and side effects would be ignored anyway.

iwoplaza: That's true. Maybe `ResolvableString` is a better solution in this case after all.

iwoplaza (approved):
👏👏👏

Issue #2697, opened 2026-07-08
test: More robust snippet testing


iwoplaza:
If we need it, there's a way to more granularly test expressions and statements.

**Testing expressions**
```ts
import { CAPTURE, extractSnippetsFromFn } from 'typegpu-testing-utility';

// ...

function foo() {
  'use gpu';
  return 5 + (CAPTURE(6 + 7));
  //                  ^^^^^
}

expect(extractSnippetsFromFn(foo)[0]!.dataType.type).toMatchInlineSnapshot('abstractInt');
```

**Testing statements**
```ts
import { CAPTURE_FOLLOWING, extractSnippetsFromFn } from 'typegpu-testing-utility';

// ...

function foo() {
  'use gpu';
  CAPTURE_FOLLOWING();
  return 5 + 6 + 7;
  // ^^^^^^^^^^^^^^
}

expect(extractSnippetsFromFn(foo)[0]!.dataType.type).toMatchInlineSnapshot('void');
```

PR #2704, opened 2026-07-09
test: Enforce only argument-less functions as input to expectDataType

iwoplaza:
`'use gpu'` functions passed to are meant to create and return snippets to be inspected


aleksanderkatan:
Why? This way, we can never inspect a snippet of type argument

reczkok (approved):

iwoplaza:
> Why? This way, we can never inspect a snippet of type argument

@aleksanderkatan That's true. I'm generally not super happy with how the current testing API changes the meaning of `return`, and I didn't want to extend it further. Inspecting arguments will be possible with the #2697 APIs though!

aleksanderkatan (approved):

PR #2711, opened 2026-07-12
feat: Early return pruning

iwoplaza:
Improvements:
- Early return in comptime
- Cosmetic improvements in generated code

### DIFF
--- a/apps/typegpu-docs/src/content/docs/apis/utils.mdx
+++ b/apps/typegpu-docs/src/content/docs/apis/utils.mdx
@@ -231,18 +231,14 @@ Generates:
 
 fn processNeighbors(cell: vec2i) {
   // unrolled iteration #0
-  {
-    // unrolled iteration #0
-    {
-      processNeighbor(cell + vec2i(-1, -1));
-    }
-    // unrolled iteration #1
-    {
-      processNeighbor(cell + vec2i(0, -1));
-    }
-    // (and so on...)
-  }
-  // (and so on ...)
+  // unrolled iteration #0 / #0
+  processNeighbor(cell + vec2i(-1, -1));
+  // unrolled iteration #0 / #1
+  processNeighbor(cell + vec2i(0, -1));
+  // unrolled iteration #0 / #2
+  processNeighbor(cell + vec2i(1, -1));
+  // ---
+  // (and so on...)
 }
 ```
 
@@ -275,17 +271,12 @@ Generates:
 fn fbm(pos: vec3f) -> f32 {
   var sum = 0f;
   // unrolled iteration #0
-  {
-    sum += noise3d(pos * 1.4f) * 1f;
-  }
+  sum += noise3d(pos * 1.4f) * 1f;
   // unrolled iteration #1
-  {
-    sum += noise3d(pos * 2.8f) * 0.5f;
-  }
+  sum += noise3d(pos * 2.8f) * 0.5f;
   // unrolled iteration #2
-  {
-    sum += noise3d(pos * 5.6f) * 0.25f;
-  }
+  sum += noise3d(pos * 5.6f) * 0.25f;
+  // ---
   return sum;
 }
 ```
@@ -324,17 +315,12 @@ for (const foo of tgpu.unroll([1, 2, 3])) {
 Generates:
 ```wgsl
 // unrolled iteration #0
-{
-  result += 1u;
-}
+result += 1u;
 // unrolled iteration #1
-{
-  result += 2u;
-}
+result += 2u;
 // unrolled iteration #2
-{
-  result += 3u;
-}
+result += 3u;
+// ---
 ```
 :::note
 `std.range` falls into the primitive array expression category.
@@ -355,17 +341,14 @@ for (const foo of tgpu.unroll([b1, b2])) {
 Generates:
 ```wgsl
 // unrolled iteration #0
-{
-  let boo = (&b1);
-  res = (res + b1.vel);
-  (*boo).pos = vec2i();
-}
+let boo = (&b1);
+res = (res + b1.vel);
+(*boo).pos = vec2i();
 // unrolled iteration #1
-{
-  let boo = (&b2);
-  res = (res + b2.vel);
-  (*boo).pos = vec2i();
-}
+let boo = (&b2);
+res = (res + b2.vel);
+(*boo).pos = vec2i();
+// ---
 ```
 
 :::caution
@@ -384,17 +367,12 @@ for (const foo of tgpu.unroll(d.vec3f(v))) {
 Generates:
 ```wgsl
 // unrolled iteration #0
-{
-  res = res + v[0u];
-}
+res = res + v[0u];
 // unrolled iteration #1
-{
-  res = res + v[1u];
-}
+res = res + v[1u];
 // unrolled iteration #2
-{
-  res = res + v[2u];
-}
+res = res + v[2u];
+// ---
 ```
 
 **Array expression of struct field names**
@@ -417,17 +395,12 @@ Generates:
 // ...
 
 // unrolled iteration #0
-{
-  result += 1u;
-}
+result += 1u;
 // unrolled iteration #1
-{
-  result += 2u;
-}
+result += 2u;
 // unrolled iteration #2
-{
-  result += 3u;
-}
+result += 3u;
+// ---
 
 // ...
 ```
@@ -452,17 +425,12 @@ Generates:
 var arr = array<i32, 3>(1, 2, 3);
 var res = 0f;
 // unrolled iteration #0
-{
-  res += f32(arr[0u]);
-}
+res += f32(arr[0u]);
 // unrolled iteration #1
-{
-  res += f32(arr[1u]);
-}
+res += f32(arr[1u]);
 // unrolled iteration #2
-{
-  res += f32(arr[2u]);
-}
+res += f32(arr[2u]);
+// ---
 ```
 
 **Buffer, `const`, `accessor`, `comptime`, `lazy`, etc.**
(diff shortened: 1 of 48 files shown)


packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -226,33 +242,71 @@ export class WgslGenerator implements ShaderGenerator {
...
+      if (willInline ? this.ctx.blockDepth <= 3 : this.ctx.blockDepth <= 2) {
+        body = body.replace(/[ ]*return;\s*$/, '');
+      }
+
+      if (body === '') {
+        return { code: '', controlFlow };
+      }
+
+      if (willInline) {
+        return { code: body.trim().replaceAll('\n  ', '\n'), controlFlow };
```
```suggestion
        return { code: body.trim().replaceAll(`\n${INDENT[1]}`, '\n'), controlFlow };
```
I case we change `INDENT` array.

cieplypolar (approved):
Awesome 🚀

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -208,10 +208,26 @@ const usageToVarTemplateMap: Record<VariableScope | BindableBufferUsage, string>
   readonly: 'storage, read',
 };
 
+interface ResolvedStatement {
+  code: string;
+  /**
+   * True if the statement (or statements) in `code` define variables that would
+   * be scoped to the nearest block.
+   */
+  definesInNearestScope?: boolean | undefined;
```
aleksanderkatan: Are there any statements that are not variable definitons that define? Could we just name this `isDefinition`?

aleksanderkatan: Why can this be undefined? If I'm not mistaken, `RawCodeSnippet`s are always expressions, and `define`s don't generate statements.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -208,10 +208,26 @@ const usageToVarTemplateMap: Record<VariableScope | BindableBufferUsage, string>
...
+  /**
+   * If not undefined, the execution of the statement (or statements) in `code`
+   * results in jumping to another part of the code. This information can be
+   * used to prune unreachable statements.
+   */
+  controlFlow?: 'return' | 'break' | 'continue' | undefined;
```
aleksanderkatan: I think this prop needs a better name and description. The primary code construct used for manipulating the control flow of a program, `if`, is not included here.

Also, this description does not clarify the transitiveness of this property - if a statement includes one of these deep inside, does it count?
```ts
if (...) {
  continue;
} else {
  return;
}
```
The above would have two control flow operations (excluding the `if`), yet this prop only allows one.

aleksanderkatan: Is any particular control flow relevant anywhere? Or is it just the presence of it?

aleksanderkatan: Are `continue` and `break` dropped when we encounter an enclosing `for`/`while` statement? Why or why not?

iwoplaza: I updated the name and description 🫡

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: I wanted to be more specific about what we care about in the name itself. A block can contain definitions (like an if statement), but it doesn't cause a definition in the nearest scope.

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: I did want to differentiate between different types of control flow, but for now, only `return` has any special handling. I wanted to future proof this property.

iwoplaza: Not currently, I didn't want to focus on it as it's something that rarely happens, though I think we could

cieplypolar (approved):
Coool :shipit:

packages/typegpu/tests/unroll.test.ts
aleksanderkatan, line 1: nit: tgpuUnroll.ts still has outdated jsdocs

apps/typegpu-docs/src/content/docs/apis/utils.mdx
```diff
@@ -458,17 +444,14 @@ for (const foo of tgpu.unroll([b1, b2])) {
...
 // unrolled iteration #1
-{
-  let boo = (&b2);
-  res = (res + b2.vel);
-  (*boo).pos = vec2i();
-}
+let boo = (&b2);
+res = (res + b2.vel);
+(*boo).pos = vec2i();
+// ---
```
aleksanderkatan, line 459: invalid code (lets with repeating idents)

I'd add a note here that we include a block if it's needed, because just looking code examples above raises red flags for me

apps/typegpu-docs/src/content/docs/apis/utils.mdx
aleksanderkatan, line 440: res is not defined

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -174,9 +174,7 @@ describe('WgslGenerator', () => {
         var arr = array<i32, 3>(1, 2, 3);
         for (var i = 0u; i < 3u; i += 1u) {
           let foo = arr[i];
-          {
-            continue;
-          }
+          continue;
```
aleksanderkatan, line 177: This continue looks prunable, will it be prined in the future?

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -2236,4 +2169,90 @@ describe('WgslGenerator', () => {
...
+    }
+
+    expect(tgpu.resolve([main])).toMatchInlineSnapshot(`
+      "fn main() {
+        const hello = 3;
+      }"
+    `);
+  });
+
+  it('prunes based on early return', ({ root }) => {
```
aleksanderkatan, line 2191: Free test
```ts
  it('prunes based on early return even when more statements are present', ({ root }) => {
    const counter = tgpu.privateVar(d.u32);
    function main() {
      'use gpu';
      const hello = 1 + 2;
      if (true) {
        const bye = 3;
        return;
      }
      // This line will never be generated, meaning the private
      // var `counter` will never be declared
      counter.$++;
    }

      "fn main() {
        const hello = 3;
        {
          const bye = 3;
          return;
        }
      }"
    `);
  });

```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -237,33 +244,81 @@ export class WgslGenerator implements ShaderGenerator {
...
-${body}
-${this.ctx.pre}}`;
+
+      const willInline = allowInlining && !definesInNearestScope;
+
+      // Omitting the 'return;' at the end of the statement list if
+      // the 'return;' would be placed in the function body outside
+      // of any nested block.
+      if (this.ctx.blockDepth === functionInitialBlockDepth) {
+        body = body.replace(/[ ]*return;\s*$/, '');
```
aleksanderkatan: I'm actually surprised this does not break with raw snippets, have a free test
```ts
  it('does not break with raw code snippets', () => {
    const snip = tgpu['~unstable'].rawCodeSnippet('// no need to do anything, just return', d.Void);

    const fn = () => {
      'use gpu';
      if (true) {
        snip.$;
        return;
      }
    };

      "fn fn_1() {
        // no need to do anything, just return;
      }"
    `);
  });
```
Can we do this just in case?
```suggestion
        body = body.replace(/[ ]+return\s*;\s*$/, '');
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1283,12 +1345,15 @@ Try 'return ${typeStr}(${str});' instead.
...
         concretize(refSnippet.dataType as wgsl.BaseData) as wgsl.StorableData,
       );
-      return stitch`${this.ctx.pre}var ${varName} = ${tryConvertSnippet(
-        this.ctx,
-        refSnippet,
-        refSnippet.dataType as wgsl.AnyWgslData,
-        false,
-      )};`;
+      return {
+        code: stitch`${this.ctx.pre}var ${varName} = ${tryConvertSnippet(
```
aleksanderkatan, line 1349: Why doesn't this go through emitVarDecl?

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1399,23 +1471,33 @@ Try 'return ${typeStr}(${str});' instead.
...
   }
 
-  protected _statement(statement: tinyest.Statement): string {
+  protected _statement(statement: tinyest.Statement): ResolvedStatement {
     if (typeof statement === 'string') {
       const id = this._identifier(statement);
       const resolved =
         id.value !== undefined && id.value !== null ? this.ctx.resolveSnippet(id).value : '';
-      return resolved ? `${this.ctx.pre}${resolved};` : '';
+      return { code: resolved ? `${this.ctx.pre}${resolved};` : '', definesInNearestScope: false };
```
aleksanderkatan, line 1485: Not sure how we handle raw code snippets, but at this moment, this test breaks:
```ts
  it('does not break with raw code snippets', () => {
    const snip = tgpu['~unstable'].rawCodeSnippet('const a = 1', d.Void);

    const fn = () => {
      'use gpu';
      for (const a of tgpu.unroll([1, 2, 3])) {
        snip.$;
      }
    };

      "fn fn_1() {
        // unrolled iteration #0
        const a = 1;
        // unrolled iteration #1
        const a = 1;
        // unrolled iteration #2
        const a = 1;
        // ---
      }"
    `);
  });
```

aleksanderkatan (changes requested):
Great job!

Our resolution starts becoming too smart for our tests. Maybe we should consider having some internal settings for disabling comptime evaluation, branch pruning, return pruning etc.?

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
iwoplaza: Yeppers

PR #2716, opened 2026-07-13
dx: More robust snippet testing

### LINKED ISSUE #2697: test: More robust snippet testing

aleksanderkatan:


aleksanderkatan:
@pullfrog review

packages/typegpu-testing-utility/src/capture.ts
```diff
@@ -0,0 +1,46 @@
+import { UnknownData, WgslGenerator, type Snippet } from 'typegpu/~internal';
+import * as tinyest from 'tinyest';
+import { tgpu, type TgpuFn } from 'typegpu';
+import { dualImpl } from 'typegpu/~internal';
```
cieplypolar:
```suggestion
import { UnknownData, WgslGenerator, type Snippet, dualImpl } from 'typegpu/~internal';
import * as tinyest from 'tinyest';
import { tgpu, type TgpuFn } from 'typegpu';
```

packages/typegpu/tests/internal/capturedSnippets.test.ts
cieplypolar, line 1:
```ts
  it('works with typedExpression - struct', () => {
    const Boid = d.struct({
      pos: d.vec3f,
    });

    const fn = tgpu.fn(
      [],
      Boid,
    )(() => {
      'use gpu';
      return CAPTURE({ pos: d.vec3f() });
    });

    const captured = captureSnippets(fn);
    expect(captured[0]?.value).toMatchInlineSnapshot(`"Boid(vec3f())"`);
  });

  it('works with typedExpression - numeric', () => {
    const fn = tgpu.fn(
      [],
      d.u32,
    )(() => {
      'use gpu';
      return CAPTURE(d.f32(1.67));
    });

    const captured = captureSnippets(fn);
    expect(captured[0]?.value).toMatchInlineSnapshot(`1.6699999570846558`);
  });
```
Could you add these tests?
The second test needs to be fixed.

aleksanderkatan: So I think we decided that this is intended behavior, and that we want to remove expected type stack, right? I created an issue for that: #2726

cieplypolar (approved):

PR #2720, opened 2026-07-13
docs: "Jelly Slider" code refactor

iwoplaza:
These changes are meant to make it easier to embed the Jelly Slider on the new home page. If it's successful, I'll do the same to other examples we might want to place there 🫡


apps/typegpu-docs/src/examples/rendering/jelly-slider/scene.ts
```diff
@@ -0,0 +1,874 @@
+import * as sdf from '@typegpu/sdf';
+import tgpu, { common, d, std, type TgpuRoot } from 'typegpu';
```
cieplypolar:
```suggestion
import { tgpu, common, d, std, type TgpuRoot } from 'typegpu';
```

apps/typegpu-docs/src/examples/rendering/jelly-slider/index.ts
```diff
@@ -1,47 +1,6 @@
-import * as sdf from '@typegpu/sdf';
-import tgpu, { common, d, std } from 'typegpu';
+import tgpu, { d, std } from 'typegpu';
```
cieplypolar:
```suggestion
import { tgpu, std } from 'typegpu';
```

cieplypolar (commented):
Looks good

cieplypolar (approved):

PR #2732, opened 2026-07-16
feat: Transfer protocol for TypeGPU resources (primarily for worklet support)

reczkok:
- pipeline type guards moved off `in` to `typeof x.foo === 'function'` since JSI HostObjects expose methods through property lookup but their has trap reports false (weird but what can you do)
- transfer-cache.ts declaration order matters, worklets don't hoist
- added some isXYZ checkers that I didn't end up using but maybe they should stay?

Also i'm pretty sure everything in `/serial` can be done in a cleaner fashion (i just didn't come up with it yet - open to suggestions)


apps/typegpu-docs/src/content/docs/integration/react-native/worklets.mdx
```diff
@@ -0,0 +1,108 @@
+---
+title: React Native Worklets
+description: A guide on running TypeGPU render loops on the UI thread with react-native-worklets.
+---
+
+With [react-native-worklets](https://docs.swmansion.com/react-native-worklets/), per-frame GPU work can run on the UI thread, unaffected by a busy JS thread.
```
iwoplaza: A minor suggestion
```suggestion
With [react-native-worklets](https://docs.swmansion.com/react-native-worklets/), per-frame GPU work can be scheduled on the UI thread, unaffected by tasks running on the default React Native thread (or RN thread).
```

apps/typegpu-docs/src/content/docs/integration/react-native/worklets.mdx
```diff
@@ -0,0 +1,108 @@
...
+TypeGPU resources created on the JS thread can be captured by worklets directly, they are transferred between runtimes automatically.
```
iwoplaza:
```suggestion
TypeGPU resources created on the RN thread can be captured by worklets directly, they are transferred between runtimes automatically.
```

apps/typegpu-docs/src/content/docs/integration/react-native/worklets.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+```ts
+import { useRoot, useFrame, useConfigureContext, useUniform } from '@typegpu/react/react-native-worklets';
+```
+
+After changing the babel config, clear the Metro cache with `npx expo start --clear`.
+
+## Example
+
+Create resources on the JS thread, then use them freely inside a `useFrame` worklet:
```
iwoplaza:
```suggestion
Create resources on the RN thread, then use them freely inside a `useFrame` worklet:
```

apps/typegpu-docs/src/content/docs/integration/react-native/worklets.mdx
```diff
@@ -0,0 +1,108 @@
...
+
+The `color` uniform and `pipeline` captured by the worklet are transferred to the UI runtime on first use.
+Both runtimes share the same underlying GPU objects, and transferring the same resource again yields the same object back.
+
+This works for buffers (including `createUniform`/`createMutable`/`createReadonly`), textures, samplers, bind groups and their layouts, vertex layouts, query sets, pipelines, roots, slots, accessors, consts, and vector/matrix instances.
+
+## Rules of transfer
+
+**Definitions are runtime-local.**
+Shader functions (`tgpu.fn`, entry functions), `tgpu.comptime`, and schemas cannot be serialized.
```
aleksanderkatan, line 103: Schemas cannot be serialized?

packages/typegpu/tests/serial.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    expect(restoredSlot.resourceType).toBe('slot');
+    expect(restoredSlot.defaultValue).toBe(42);
+
+    const accessor = tgpu.accessor(d.vec3f, d.vec3f(1, 2, 3));
+    const restoredAccessor = roundTrip(accessor, root);
+    expect(restoredAccessor.resourceType).toBe('accessor');
+    expect(restoredAccessor.schema).toBe(d.vec3f);
+    expect(restoredAccessor.defaultValue).toEqual(d.vec3f(1, 2, 3));
+
+    const constant = tgpu['~unstable'].const(d.arrayOf(d.f32, 3), [1, 2, 3]);
```
aleksanderkatan, line 194:
```suggestion
    const constant = tgpu.const(d.arrayOf(d.f32, 3), [1, 2, 3]);
```

packages/typegpu/tests/serial.test.ts
aleksanderkatan, line 1: Free test
```ts
  it('keeps the initial buffer value', ({ root }) => {
    const buf = root.createBuffer(d.u32, 5);

    const restoredBuf = roundTrip(buf, root);

    const rawBuffer = root.unwrap(restoredBuf);
    const writtenBuffer = vi.mocked(rawBuffer.getMappedRange).mock.results[0]?.value;
    expect(writtenBuffer).toMatchInlineSnapshot(`
      ArrayBuffer [
        5,
        0,
        0,
        0,
      ]
    `);
  });
```

aleksanderkatan, line 1: Free tests
```ts
  it('round-trips buffer accessors', ({ root }) => {
    const getter = () => {
      'use gpu';
      return d.u32(1);
    };

    const fnAccessor = tgpu.accessor(d.u32, getter);
    const restoredFnAccessor = roundTrip(fnAccessor, root);

    expect(restoredFnAccessor.resourceType).toBe('accessor');
    expect(restoredFnAccessor.schema).toBe(d.u32);
    expect(restoredFnAccessor.defaultValue).toBe(getter);
  });

  it('round-trips function accessors', ({ root }) => {
    const getter = () => {
      'use gpu';
      return d.u32(1);
    };

    const fnAccessor = tgpu.accessor(d.u32, getter);
    const restoredFnAccessor = roundTrip(fnAccessor, root);

    expect(restoredFnAccessor.resourceType).toBe('accessor');
    expect(restoredFnAccessor.schema).toBe(d.u32);
    expect(restoredFnAccessor.defaultValue).toBe(getter);
  });
```

aleksanderkatan, line 1: Free tests
```ts
  it('does not allow snapshotting destroyed buffers', ({ root }) => {
    const buf = root.createBuffer(d.u32, 5);
    buf.destroy();

    expect(() => snapshotResource(buf)).toThrowErrorMatchingInlineSnapshot(`[Error: This buffer has been destroyed]`);
  });

  it('does not allow snapshotting destroyed querysets', ({ root }) => {
    const querySet = root.createQuerySet('timestamp', 2);
    querySet.destroy();

    expect(() => roundTrip(querySet, root)).toThrowErrorMatchingInlineSnapshot(
      `[Error: This QuerySet has been destroyed.]`,
    );
  });

  it('does not allow snapshotting destroyed textures', ({ root }) => {
    const texture = root.createTexture({ size: [2, 2], format: 'rgba8unorm' });
    texture.destroy();

    expect(() => roundTrip(texture, root)).toThrowErrorMatchingInlineSnapshot(`[Error: This texture has been destroyed]`);
  });
```

aleksanderkatan, line 1: All names get lost during the serialization process
```ts
  it('retains names', ({ root }) => {
    const buf1 = root.createBuffer(d.u32, 5);
    const buf2 = root.createBuffer(d.u32, 5).$name('myBuf');

    const restoredBuf1 = roundTrip(buf1, root);
    const restoredBuf2 = roundTrip(buf2, root);

    expect(tgpu.resolve([restoredBuf1.$usage('uniform').as('uniform')])).toMatchInlineSnapshot(
      `"@group(0) @binding(0) var<uniform> buffer: u32;"`,
    );
    expect(tgpu.resolve([restoredBuf2.$usage('uniform').as('uniform')])).toMatchInlineSnapshot(
      `"@group(0) @binding(0) var<uniform> buffer: u32;"`,
    );
  });
```

packages/typegpu/src/serial/registry.ts
```diff
@@ -0,0 +1,306 @@
...
+    snapshot(resource: never): SnapshotFor<K>;
+    restore(snapshot: never, ctx: RestoreContext): unknown;
+  };
+};
+
+export interface TgpuBufferBindingSnapshot {
+  readonly type: 'uniform' | 'mutable' | 'readonly';
+  readonly device: GPUDevice;
+  readonly buffer: TgpuBufferSnapshot;
+}
```
aleksanderkatan, line 124: Why are these here instead of 'bufferBinding.ts'? Circular dependency?

packages/typegpu/src/serial/registry.ts
```diff
@@ -0,0 +1,306 @@
...
+
+/** Whether the value is a TypeGPU object that {@link snapshotResource} does not support */
+export function isNonTransferableResource(value: unknown): boolean {
+  return (
+    typeof value === 'object' &&
+    value !== null &&
+    $internal in value &&
+    getSnapshotterFor(value) === undefined
+  );
+}
```
aleksanderkatan, line 284: Unless I don't see something, we use 'transferable' and 'snapshottable' with the same meaning, I'd stick with one name.

apps/typegpu-docs/src/content/docs/integration/react-native/worklets.mdx
```diff
@@ -0,0 +1,117 @@
...
+```
+
+The `color` uniform and `pipeline` captured by the worklet are transferred to the UI runtime on first use.
+Both runtimes share the same underlying GPU objects, and transferring the same resource again yields the same object back.
+
+This works for buffers (including `createUniform`/`createMutable`/`createReadonly`), textures, samplers, bind groups and their layouts, vertex layouts, query sets, pipelines, roots, slots, accessors, consts, and vector/matrix instances.
+
+## Opting out
+
+To keep everything on the JS thread even with `react-native-worklets` installed, pass `disableWorklets` to the `Root` provider:
```
iwoplaza:
```suggestion
To keep everything on the RN thread even with `react-native-worklets` installed, pass `disableWorklets` to the `Root` provider:
```

apps/typegpu-docs/src/content/docs/integration/react-native/worklets.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+```tsx
+<Root disableWorklets>
+  <Pulse />
+</Root>
+```
+
+`useFrame` then runs its callbacks on the JS thread, and `useConfigureContext` returns a plain object of the same shape, so the code above keeps working unchanged.
```
iwoplaza:
```suggestion
`useFrame` then runs its callbacks on the RN thread, and `useConfigureContext` returns a plain object of the same shape, so the code above keeps working unchanged.
```

packages/typegpu-react/src/react-native/use-frame.ts
iwoplaza, line 1: A nitpick, could we move this file to react-native/use-frame.ts instead of react-native/core/use-frame.ts? It might be confusing to know which 'core' we mean when we try to quickly open a file.

packages/typegpu-react/src/react-native/use-frame.ts
```diff
@@ -0,0 +1,78 @@
+import { useEffect, useRef } from 'react';
+
+import { useWorkletsDisabled } from '../../core/root-context.tsx';
+import { type FrameCtx, startFrameLoop } from '../../core/use-frame.ts';
+import { getWorkletsModule } from '../worklets-integration.ts';
+
+type FrameCallback = (ctx: FrameCtx) => void;
+type FrameCallbackRef = { current: FrameCallback };
+type UiValue<T> = {
```
iwoplaza, line 9: Casing nitpick (said the ~TGPU~Tgpu hipocrite)
```suggestion
type UIValue<T> = {
```

packages/typegpu-react/src/react-native/core/use-frame.ts
```diff
@@ -0,0 +1,78 @@
...
+};
+
+/**
+ * Runs the frame loop on the UI runtime when the callback is a worklet and
+ * `react-native-worklets` is available, on the JS thread otherwise
+ */
+export function useFrame(cb: FrameCallback) {
+  const workletsDisabled = useWorkletsDisabled();
+  const worklets = getWorkletsModule();
+  const runOnUI = worklets !== null && !workletsDisabled && worklets.isWorkletFunction(cb);
```
iwoplaza: Would be more efficient for users that opt out of worklets
```suggestion
  const workletsDisabled = useWorkletsDisabled();
  const worklets = workletsDisabled ? null : getWorkletsModule();
  const runOnUI = worklets !== null && worklets.isWorkletFunction(cb);
```

packages/typegpu-react/src/react-native/use-frame.ts
```diff
@@ -0,0 +1,78 @@
...
+ * Runs the frame loop on the UI runtime when the callback is a worklet and
+ * `react-native-worklets` is available, on the JS thread otherwise
+ */
+export function useFrame(cb: FrameCallback) {
+  const workletsDisabled = useWorkletsDisabled();
+  const worklets = getWorkletsModule();
+  const runOnUI = worklets !== null && !workletsDisabled && worklets.isWorkletFunction(cb);
+
+  const latestCb = useRef(cb);
+  const uiCbRef = useRef<UiValue<FrameCallbackRef> | undefined>(undefined);
```
iwoplaza, line 24: Does the UI value in `uiCbRef` need to hold a ref? Is there a reason it couldn't hold just the callback?
```suggestion
  const uiCbRef = useRef<UiValue<FrameCallback> | undefined>(undefined);
```

packages/typegpu/tests/root.test.ts
```diff
@@ -4,6 +4,27 @@ import { tgpu, d } from 'typegpu';
 import { it } from 'typegpu-testing-utility';
 
 describe('TgpuRoot', () => {
+  describe('init', () => {
+    it('does not forward optionalFeatures to requestDevice', async ({ adapter }) => {
```
iwoplaza, line 8: I don't understand the name of this test

packages/typegpu/tests/serial.test.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    const mat = d.mat3x3f(1, 2, 3, 4, 5, 6, 7, 8, 9);
+    expect(roundTrip(mat, root)).toEqual(mat);
+
+    const view = root
+      .createTexture({ size: [2, 2], format: 'rgba8unorm' })
+      .$usage('sampled')
+      .createView();
+    expect(snapshotResource(view)).toBeUndefined();
+    expect(isNonTransferableResource(view)).toBe(true);
```
iwoplaza, line 238: Just curious, why is a view non-transferable? Is it because react-native-webgpu doesn't register serializables for them, or is there a technical limitation?

packages/typegpu/src/serial/schema.ts
```diff
@@ -0,0 +1,193 @@
...
+  if (attrib.type === 'location') {
+    return d.location(attrib.value, schema);
+  }
+  if (attrib.type === 'interpolate') {
+    return d.interpolate(attrib.value as never, schema as never);
+  }
+  if (attrib.type === 'builtin') {
+    return getBuiltinByName(attrib.value);
+  }
+  return d.invariant(schema as Parameters<typeof d.invariant>[0]);
```
iwoplaza: I would rather we hit the `assertExhaustive` to catch if we miss any attrib when we add more.
```suggestion
  if (attrib.type === 'invariant') {
    return d.invariant(schema as Parameters<typeof d.invariant>[0]);
  }
  assertExhaustive(attrib, 'schema.ts#applyAttrib');
```

packages/typegpu/src/core/vertexLayout/vertexLayout.ts
```diff
@@ -1,9 +1,21 @@
...
 import { sizeOf } from '../../data/sizeOf.ts';
-import type { BaseData, WgslArray } from '../../data/wgslTypes.ts';
-import { isDecorated, isWgslStruct } from '../../data/wgslTypes.ts';
+import type { AnyWgslData, BaseData, WgslArray } from '../../data/wgslTypes.ts';
+import { isDecorated, isWgslArray, isWgslStruct } from '../../data/wgslTypes.ts';
+import {
+  deserializeDataSchema,
+  serializeDataSchema,
+  type SerializedDataSchema,
+} from '../../serial/schema.ts';
```
iwoplaza: I feel scared importing from ./serial/* here, since it will include all of `d` here, but it seems like that is being included anyway because the treeshakability didn't change 😆. It's something to investigate, that being said, would it be safer to move the snapshots and restores to the /serial folder and isolate them there (at least those that depend on snapshotters and restorers of other types)?

packages/typegpu/src/core/root/init.ts
```diff
@@ -193,6 +204,66 @@ export class TgpuGuardedComputePipelineImpl<
   }
 }
 
+export function isGuardedComputePipeline(maybe: unknown): maybe is TgpuGuardedComputePipeline {
+  return (
+    (maybe as TgpuGuardedComputePipeline | undefined)?.resourceType ===
+      'guarded-compute-pipeline' && !!(maybe as { [$internal]?: boolean } | undefined)?.[$internal]
```
iwoplaza:
```suggestion
      'guarded-compute-pipeline' && isMarkedInternal(maybe)
```

packages/typegpu/tests/root.test.ts
iwoplaza: Nevermind, I get it 🫡

packages/typegpu/src/core/vertexLayout/vertexLayout.ts
reczkok: I feel like the bundlers are smart enough to not add the internal stuff (at least it was in my bundling tests) - tippy-toeing around this issue is extremely cumbersome but i will try. I would seriously investigate how much this matters for future work

packages/typegpu/tests/serial.test.ts
reczkok: It is transferable, it's just a lot of schemas to write code for when it's easily derived

packages/typegpu-react/src/react-native/use-frame.ts
reczkok: Yes, it is necessary since otherwise `setSync` is unusable (it thinks the function which for us is the value is an updater)

packages/typegpu/tests/serial.test.ts
reczkok: We can probably salvage some logic from layouts but I would leave that for a follow up maybe?

reczkok:
Can't wait what mr. Frog thinks of this one 😆

packages/typegpu/src/core/sampler/sampler.ts
```diff
@@ -78,6 +83,36 @@ export function isComparisonSampler(resource: unknown): resource is TgpuComparis
...
+  value: unknown,
+): value is TgpuSampler | TgpuComparisonSampler {
+  return (isSampler(value) || isComparisonSampler(value)) && value[$internal].props !== undefined;
+}
+
+export function INTERNAL_snapshotSampler(
+  sampler: TgpuSampler | TgpuComparisonSampler,
+): TgpuSamplerSnapshot {
+  const { props, device } = sampler[$internal];
+  invariant(props && device, 'Only samplers created from props can be snapshotted.');
```
aleksanderkatan, line 102: This is never supposed to happen, right? If so, I'd alter the message a little, and if not, I'd use if & throw instead of `invariant`.
```suggestion
  invariant(props && device, 'Internal error, expected sampler to contain props and a device.');
```

packages/typegpu/tests/serial.test.ts
aleksanderkatan, line 1: QuerySet's availability is not carried through snapshots. If we don't want to support that, we can include an `if (!available) throw` in the snapshotting function.

packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -113,12 +119,60 @@ export function INTERNAL_createComputePipeline(
...
+export function INTERNAL_restoreComputePipeline(
+  snapshot: TgpuComputePipelineSnapshot,
+  ctx: RestoreContext,
+): TgpuComputePipeline {
+  const root = ctx.getRoot(snapshot.device) as ExperimentalTgpuRoot;
+  const core = ComputePipelineCore.precompiled(root, {
+    pipeline: snapshot.pipeline,
+    usedBindGroupLayouts: snapshot.usedBindGroupLayouts,
+    catchall: undefined,
+    logResources: undefined,
```
aleksanderkatan, line 162: I think these two require a comment with explanation. I see that catchall becomes one of the regular bind groups, but how do logs work?

packages/typegpu/src/resolutionCtx.ts
```diff
@@ -1123,6 +1123,8 @@ export function resolve(item: Wgsl, options: ResolutionCtxImplOptions): Resoluti
     return [
       catchallIdx,
       new TgpuBindGroupImpl(
+        // Undefined only in rootless `tgpu.resolve()`, where the group is never unwrapped
+        options.root as ExperimentalTgpuRoot,
```
aleksanderkatan, line 1127: Can we avoid this type lie? For example, can we make `readonly root` in the bind group a getter that throws if root is not present?

packages/typegpu/src/tgpuBindGroupLayout.ts
```diff
@@ -233,6 +241,66 @@ export function isBindGroup(value: unknown): value is TgpuBindGroup {
...
+  ctx: RestoreContext,
+): TgpuBindGroup {
+  const root = ctx.getRoot(snapshot.device);
+  const bindGroup = snapshot.bindGroup;
+  return {
+    resourceType: 'bind-group',
+    root,
+    layout: snapshot.layout,
+    unwrap: () => bindGroup,
+  };
```
aleksanderkatan, line 301: Why avoid calling the constructor? This way we'll miss on e.g. toString when we implement it, and instance checks will fail.

Root has a weak memo on it and uses it to know when to recreate resources during unwrap. Is this a workaround for that? Should we add it to the memo here? Should we maybe rewrite root to no longer keep any weak maps like this? This is inconsistent anyway - pipelines keep their own memos while bind groups do not.

packages/typegpu/tests/computePipeline.test.ts
aleksanderkatan, line 1: I'm generally not too fond of tests introduced in this PR, their names are not descriptive (it is not mentioned anywhere that they test snapshots, not pipelines), they are very long and test multiple things at once and they are difficult to read. Let them be, but please try to split future tests into smaller ones and follow the arrange/act/assert where applicable

packages/typegpu/tests/computePipeline.test.ts
```diff
@@ -187,6 +188,91 @@ describe('TgpuComputePipeline', () => {
...
+      .createComputePipeline({
+        compute: tgpu.computeFn({ workgroupSize: [1] })(() => {
+          'use gpu';
+          fixedUniform.$;
+          manualLayout.$.params;
+        }),
+      })
+      .with(manualBindGroup);
+
+    const snapshot = snapshotResource(sourcePipeline);
```
aleksanderkatan, line 208: Internal testing that can be done on built package should be outside of internal test folder (like it is now), right?

packages/typegpu/tests/renderPipeline.test.ts
aleksanderkatan, line 1: `restoredPipeline` acts differently - regular pipeline does not throw on `drawIndexed`, and this one does.
This may be a result of misconfigured mocks or something like this.

I would like you to add a test including index buffers, but I'm not sure how to do that because of #2738

```ts
  it('...', ({ root }) => {
    const indexBuffer = root.createBuffer(d.arrayOf(d.u16, 2)).$usage('index');

    const vertexFn = tgpu.vertexFn({ out: { pos: d.builtin.position } })``;
    const fragmentFn = tgpu.fragmentFn({ out: { color: d.vec4f } })``;

    const pipeline = root
      .createRenderPipeline({
        vertex: vertexFn,
        fragment: fragmentFn,
        targets: { color: { format: 'rgba8unorm' } },
      })
      .withColorAttachment({
        color: {
          view: {} as unknown as GPUTextureView,
          loadOp: 'clear',
          storeOp: 'store',
        },
      })
      .withIndexBuffer(indexBuffer);

    const restoredPipeline = roundTrip(pipeline, root);
    restoredPipeline.drawIndexed(4);
  });
```

packages/typegpu/src/core/root/init.ts
```diff
@@ -193,6 +204,66 @@ export class TgpuGuardedComputePipelineImpl<
...
+}
+
+export interface TgpuRootSnapshot {
+  readonly type: 'root';
+  readonly device: GPUDevice;
+}
+
+export function INTERNAL_snapshotRoot(root: TgpuRoot): TgpuRootSnapshot {
+  return { type: 'root', device: root.device };
+}
```
aleksanderkatan, line 228: `ownDevice` prop gets lost, I suspect it will always be set to false, right? `nameRegistry` and `logOptions` get lost as well

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -442,13 +453,77 @@ export type AnyFragmentColorAttachment = ColorAttachment | Record<string, ColorA
...
+  readonly device: GPUDevice;
+  readonly pipeline: GPURenderPipeline;
+  readonly fragmentOut: SerializedDataSchema | undefined;
+  readonly usedBindGroupLayouts: TgpuBindGroupLayout[];
+  readonly bindGroups: [TgpuBindGroupLayout, TgpuBindGroup | GPUBindGroup][];
+  readonly usedVertexLayouts: TgpuVertexLayout[];
+  readonly vertexBuffers: [TgpuVertexLayout, (TgpuBuffer<BaseData> & VertexFlag) | GPUBuffer][];
+  readonly timestampWrites: TimestampWritesPriors['timestampWrites'];
+  readonly performanceCallback: TimestampWritesPriors['performanceCallback'];
+}
```
aleksanderkatan, line 475: Do `withEncoder`/`withPass` work?

packages/typegpu-react/tsdown.config.ts
```diff
@@ -1,5 +1,19 @@
...
+// restoring the literal call keeps react-native-worklets an optional dependency
+const preserveOptionalRequire = {
+  name: 'preserve-optional-require',
+  renderChunk(code: string) {
+    if (!code.includes('__require("react-native-worklets")')) {
+      return null;
+    }
+    return code
+      .replace('__require("react-native-worklets")', 'require("react-native-worklets")')
+      .replace(/import \{ __require \} from "[^"]+";\n/, '');
```
aleksanderkatan, line 13: nit
```suggestion
      .replaceAll('__require("react-native-worklets")', 'require("react-native-worklets")')
      .replaceAll(/import \{ __require \} from "[^"]+";\n/, '');
```

packages/typegpu-react/src/react-native/serialization/transfer-cache.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export function getOrCreateTransferId(value: object): number {
+  'worklet';
+  const global = globalThis as TypegpuReactTransferGlobals;
+  const ids = (global.__TYPEGPU_REACT_TRANSFER_IDS__ ??= new WeakMap());
+  let id = ids.get(value);
+  if (id === undefined) {
+    id = global.__TYPEGPU_REACT_NEXT_TRANSFER_ID__ ?? 0;
+    global.__TYPEGPU_REACT_NEXT_TRANSFER_ID__ = id + 1;
```
aleksanderkatan, line 53: Will this work when receiving resources from 2 different threads?

aleksanderkatan (commented):
Yay worklets! ⛑️🎨

I didn't have time to properly review the react and react-native part of this PR, I leave some comments regarding the typegpu part.

I guess worklets come at a price. I'm generally a little worried about everything that happens lazily, and about all methods modifying the state like `.with`s on render pipeline, I'm sure we missed some, and it will be impossible to debug. 
Also, I expect many of our future changes to break on worklets. Furthermore, our APIs will be less flexible, every subsequent refactor/API update will be more difficult because it will need to take worklets into account.

But hey, worklets!

Could we also include some internal docs explaining the general idea, intuition behind why we cannot transfer functions, and some weird cases like the `catchall` and `logResources` working?

packages/typegpu/tests/computePipeline.test.ts
reczkok: I don't know - I think they flow in a pretty logical fashion, I could split them up but tbh see no good reason to. I agree they should be named better though

packages/typegpu/tests/serial.test.ts
iwoplaza: I added an issue for it so we don't forget, we don't need to do it in this PR 👌.
https://github.com/software-mansion/TypeGPU/issues/2740

packages/typegpu/tests/computePipeline.test.ts
iwoplaza: That's correct 🫡

aleksanderkatan:
Tgpu functions are tgpu resources, and they are not snapshottable. However, `isNonTransferableResource` returns false for them.

```ts
  it('treats functions as non-snapshottable', ({ root }) => {
    const fn1 = () => {
      'use gpu';
      return 0;
    };

    const fn2 = tgpu.fn(
      [],
      d.u32,
    )(() => {
      'use gpu';
      return 0;
    });

    expect(isNonTransferableResource(fn1)).toMatchInlineSnapshot(`false`);
    expect(isNonTransferableResource(fn2)).toMatchInlineSnapshot(`false`);
  });
```

reczkok:
Replaced by #2797

PR #2737, opened 2026-07-16
feat: CPU implementation of std.transpose

### LINKED ISSUE #2736: Request: CPU fallback for std.transpose()
The matrix transpose is a common and useful operation, e.g. in 3- or 4-channel color processing. For my use case, it would be really nice to have  a CPU fallback like there is for most functions of `std`.

iwoplaza:
In addition, the TS type system is now made aware that matrices have a custom `toString` method, so ignoring the lint rule is no longer necessary.

### DIFF
--- a/packages/typegpu/src/std/numeric.ts
+++ b/packages/typegpu/src/std/numeric.ts
@@ -41,6 +41,7 @@ import {
   type v4i,
   type Vec2f,
   type VecData,
+  WORKAROUND_getSchema,
 } from '../data/wgslTypes.ts';
 import { SignatureNotSupportedError } from '../errors.ts';
 import type { Infer } from '../shared/repr.ts';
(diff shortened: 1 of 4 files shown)


packages/typegpu/src/std/numeric.ts
```diff
@@ -1227,11 +1228,23 @@ export const tanh = dualImpl({
   sideEffects: false,
 });
 
-export const transpose = dualImpl<<T extends AnyMatInstance>(e: T) => T>({
+function cpuTranspose<T extends AnyMatInstance>(value: T): T {
```
cieplypolar, line 1232: I would hardcode it for every size in {2, 3, 4}, for example 2x2 transposition
```ts
const schema = WORKAROUND_getSchema(value);
const transposed = schema() as T;
const src = value.columns;
const dst = transposed.columns;

dst[0]![0] = src[0]![0]!;
dst[0]![1] = src[1]![0]!;

dst[1]![0] = src[0]![1]!;
dst[1]![1] = src[1]![1]!;

return transposed;
```

iwoplaza: For the sake of performance?

cieplypolar: yes

cieplypolar (approved):

PR #2739, opened 2026-07-17
Impr: More restrictive bool constructor and unary operator !

### LINKED ISSUE #2701: [RFC]: impr: bool() and operator! comptime injection vs wgsl generation
```ts
import { tgpu, d } from 'typegpu';

const arr = tgpu.const(d.arrayOf(d.u32, 7), [0, 1, 2, 3, 4, 5, 6]);

const length = () => {
  'use gpu';
  let i = 0;
  while (!!arr.$[i]) { 
    //     ^ number | undefined
    i++;
  }
  return i;
}
```
This pattern is problematic because wgsl does not have an `undefined` entity.

In wgsl:
- `operator!` works only with `bool` and `vecN<bool>`

Proposed behavior in typegpu:
- comptime injection: allow all types and mirror JS
- wgsl generation: only `bool` is allowed. If user tries to use `operator!` on vector, we should advise them to use `std.not`

Still, we are left with array of bools case, but we cannot do much about it :(

As an alternative to `!!`, use might use `bool()` constructor.
```ts
const length = () => {
  'use gpu';
  let i = 0;
  while (d.bool(arr.$[i])) { 
    //          ^ number | undefined
    i++;
  }
  return i;
}
```
In wgsl:
- `bool()` works for every scalar type

Proposed behavior in typegpu:
- we cannot forbid casting numbers
- we cannot make signature that do not accept undefined, because of comptime injection. So...
- comptime injection: allow all types and mirror JS
- wgsl generation: only scalars are allowed

There is one last thing: `vecN<bool>(e: vecN<T>)`. Currently, we don't support that and I think we shouldn't worry about it.

cieplypolar:
Closes #2701.


apps/typegpu-docs/src/content/docs/apis/data-schemas.mdx
```diff
@@ -17,6 +17,31 @@ Schemas can also be called to create instances of the types they represent.
...
 
+:::danger
+WGSL has no `undefined` or `null`.
+
+If you are tempted to write the following (or any code that relies on `undefined` in JS):
+```ts twoslash
+// @noUncheckedIndexedAccess: true
+import { tgpu, d } from 'typegpu';
+
+const arr = tgpu.const(d.arrayOf(d.u32, 7), [0, 1, 2, 3, 4, 5, 6]);
```
aleksanderkatan:
```suggestion
const arr = tgpu.const(d.arrayOf(d.u32, 7), [1, 2, 3, 4, 5, 6, 7]);
```

aleksanderkatan (approved):
Sure!

apps/typegpu-docs/src/content/docs/apis/data-schemas.mdx
cieplypolar: thank you 💙

aleksanderkatan (approved):

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -183,20 +183,17 @@ const unaryOpCodeToCodegen = {
...
+        `Unary operator ! requires boolean operand. Got ${String(argExpr.dataType)}.${
+          wgsl.isVecBool(argExpr.dataType)
+            ? ` For component-wise negation, use 'std.${not.toString()}'.`
+            : ''
+        }`,
+      );
     }
 
-    return snip(false, bool, 'constant', false);
+    return snip(`!${argStr}`, bool, 'runtime', argExpr.possibleSideEffects);
```
iwoplaza: `!` will connect to the first part of a more complex expression

```suggestion
    return snip(`!(${argStr})`, bool, 'runtime', argExpr.possibleSideEffects);
```

packages/typegpu/src/std/boolean.ts
```diff
@@ -191,82 +191,49 @@ export const ge = dualImpl({
...
-    }
-
-    if (isVec(dataType)) {
-      const vecConstructorStr = `vec${dataType.componentCount}<bool>`;
-      return stitch`!(${vecConstructorStr}(${arg}))`;
-    }
-
-    return 'false';
-  },
+  codegenImpl: (_ctx, [arg]) => stitch`!${arg}`,
```
iwoplaza:
```suggestion
  codegenImpl: (_ctx, [arg]) => stitch`!(${arg})`,
```

iwoplaza (approved):
Nice! Just bring back the parentheses

PR #2742, opened 2026-07-19
docs: Format Astro files

iwoplaza:
While we wait for Oxc to add support for .astro files (tracked here)[https://github.com/oxc-project/oxc/issues/19715], I added Prettier to format those files, to keep the consistent across setups.


cieplypolar (approved):
Looks good :shipit:

PR #2744, opened 2026-07-20
fix: Throw when the expectedCalls count does not match

### LINKED ISSUE #2743: fix: Throw when the expectedCalls count does not match

aleksanderkatan:


iwoplaza (approved):
🫡

PR #2747, opened 2026-07-21
feat: Logger

aleksanderkatan:
Since the vast majority of our `console.*()` were `console.warn`s, I decided to only define `logger.warn` and to skip other methods. Let me know if you think this is the right direction for us.


packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -363,19 +364,19 @@ export function convertToCommonType<T extends Snippet[]>(
...
 
-  if (DEV && Array.isArray(restrictTo) && restrictTo.length === 0) {
-    console.warn(
-      'convertToCommonType was called with an empty restrictTo array, which prevents any conversions from being made. If you intend to allow all conversions, pass undefined instead. If this was intended call the function conditionally since the result will always be undefined.',
-    );
-  }
+  invariant(
+    !(Array.isArray(restrictTo) && restrictTo.length === 0),
+    "Internal error, expected 'restrictTo' to not be an empty array.",
+  );
```
aleksanderkatan, line 374:
```suggestion
```
Not sure about this one. This warn was never called before, and I couldn't replicate it in a real scenario. Maybe we should just remove this invariant, the function will just return `undefined`.

packages/typegpu/src/tgpuLogger.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  'precision-loss',
+  'implicit-conversion',
+
+  'webgpu-feature-missing',
+  'webgpu-limits-exceeded',
+  'locations-mismatched',
+  'log-limit-exceeded',
+  'external-omitted',
+] as const;
```
aleksanderkatan, line 16: I've changed this list like 5 times already, any suggestions about category splits and renames are welcome

aleksanderkatan:
@pullfrog review

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -363,19 +363,19 @@ export function convertToCommonType<T extends Snippet[]>(
     return undefined;
   }
 
-  if (DEV && Array.isArray(restrictTo) && restrictTo.length === 0) {
-    console.warn(
-      'convertToCommonType was called with an empty restrictTo array, which prevents any conversions from being made. If you intend to allow all conversions, pass undefined instead. If this was intended call the function conditionally since the result will always be undefined.',
-    );
-  }
+  invariant(
+    !(Array.isArray(restrictTo) && restrictTo.length === 0),
```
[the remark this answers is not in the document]
aleksanderkatan: I already left a comment about it

packages/typegpu/src/data/dataIO.ts
```diff
@@ -861,7 +864,8 @@ export function writeToArrayBuffer<T extends BaseData>(
       compiledWriter(dataView, startOffset, data, isLittleEndian, endOffset);
       return;
     } catch (error) {
-      console.error(
+      logger.warn(
```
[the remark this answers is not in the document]
aleksanderkatan: It does, but I did it for consistency of typegpu messages being silencible. We could add a logger.error method that is impossible to silence.

aleksanderkatan:
I was suggested to use a system of reverse DNS system, like `com.typegpu.deprecated.flush`. Silencing `com.typegpu.deprecated` would get rid of all sub-errors, but you could also just silence the specific warn. What do you think @reczkok @iwoplaza @cieplypolar? Is this overkill for our case?

packages/typegpu/src/tgsl/conversion.ts
reczkok: I think this was more something that prevent future changes from breaking in a werid way or working in an unexpected way

reczkok (approved):
Nice, love the feature

packages/typegpu/src/tgsl/conversion.ts
```diff
@@ -363,19 +363,19 @@ export function convertToCommonType<T extends Snippet[]>(
     return undefined;
   }
 
-  if (DEV && Array.isArray(restrictTo) && restrictTo.length === 0) {
-    console.warn(
-      'convertToCommonType was called with an empty restrictTo array, which prevents any conversions from being made. If you intend to allow all conversions, pass undefined instead. If this was intended call the function conditionally since the result will always be undefined.',
-    );
-  }
+  invariant(
```
iwoplaza, line 370:
```suggestion
  // Calling convertToCommonType with an empty restrictTo array
  // prevents any conversions from being made. If you intend to allow
  // all conversions, pass undefined instead. If this was intended call
  // the function conditionally since the result will always be undefined.
  invariant(
```

iwoplaza (approved):

PR #2756, opened 2026-07-25
feat: Add radix to sort package and improve existing algos

reczkok:


packages/typegpu-sort/src/bitonic/bitonicSort.ts
```diff
@@ -3,291 +3,386 @@ import {
...
-    uniform: copyParamsType,
-  },
-});
+function nextPowerOf2(n: number): number {
+  let p = 1;
+  while (p < n) {
+    p <<= 1;
+  }
+  return p;
+}
```
vende11s, line 51: Can be done in $O(1)$ time without a loop, e.g.:

```typescript
if (n <= 1) return 1;
return 1 << (32 - Math.clz32(n - 1));
```

packages/typegpu-sort/src/radix/schemas.ts
```diff
@@ -0,0 +1,99 @@
...
+function digitOfI32(v: number, shift: number): number {
+  'use gpu';
+  const raw = d.u32((v >> shift) & (RADIX_SIZE - 1));
+  return raw ^ std.select(d.u32(0), d.u32(RADIX_SIZE / 2), shift === 24);
+}
+
+function digitOfF32(v: number, shift: number): number {
+  'use gpu';
+  // -0 and +0 must map to the same bits, otherwise they sort apart
+  const bits = std.select(std.bitcastF32toU32(v), d.u32(0), v === 0);
```
vende11s: `bitcastF32toU32` is deprecated

packages/typegpu-sort/tests/radix.test.ts
vende11s, line 1: Missing correctness tests for the sort itself

reczkok: Well no good way to run the shaders without introducing a lot of testing scaffolding, the sort example is a correctness test

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/calculator.ts
aleksanderkatan, line 1: Not that I'm going to miss this example much, but why are we just yanking it?

packages/typegpu-sort/src/radix/radixSort.ts
```diff
@@ -0,0 +1,162 @@
...
+type KeyBuffer = TgpuBuffer<d.WgslArray<RadixKeyType>> & StorageFlag;
+type ValueBuffer = TgpuBuffer<d.WgslArray<d.AnyWgslData>> & StorageFlag;
+
+/**
+ * Creates a stable LSD radix sorter for a `u32`, `i32` or `f32` key buffer, optionally
+ * reordering a payload buffer alongside the keys. Keys are ordered by the natural order
+ * of their type. All GPU resources are created up front, so `run` only records dispatches.
+ *
+ * For `f32` keys sorted ascending, NaNs with a cleared sign bit sort after +Infinity and
+ * NaNs with a set sign bit sort before -Infinity. -0 and +0 compare equal.
```
aleksanderkatan: I'd skip this note. We don't throw on writing these values to buffers (only when casting with a schema), but the spec still says "Implementations may assume that overflow, infinities, and NaNs are not present during [shader execution](https://www.w3.org/TR/WGSL/#shader-execution-start)."

```suggestion
 * For `f32` keys sorted ascending, -0 and +0 compare equal.
```

packages/typegpu-sort/src/bitonic/bitonicSort.ts
```diff
@@ -3,291 +3,390 @@ import {
...
+interface SortStep {
+  pipeline: TgpuComputePipeline;
+  workgroups: [number, number, number];
+}
 
-  const copyBackPipeline = root.createComputePipeline({ compute: copyBackKernel });
+/**
+ * Creates a bitonic sorter for a `u32`, `i32` or `f32` key buffer, optionally reordering
+ * a payload buffer alongside the keys. The order is defined by an arbitrary comparator.
+ * All GPU resources are created up front, so `run` only records dispatches.
```
aleksanderkatan, line 270: Should I always choose radix over bitonic unless I have a custom comparator?

packages/typegpu-sort/src/bitonic/slots.ts
```diff
@@ -1,8 +1,14 @@
-import { tgpu, d } from 'typegpu';
+import { tgpu } from 'typegpu';
 
-/** Default comparison function: ascending order (a < b means a comes before b) */
-export const defaultCompare = tgpu.fn([d.u32, d.u32], d.bool)((a, b) => a < b);
+export function defaultCompare(a: number, b: number): boolean {
+  'use gpu';
+  return a < b;
+}
```
aleksanderkatan: Do we plan on updating & allowing std functions here?

packages/typegpu-sort/src/runPass.ts
```diff
@@ -0,0 +1,42 @@
...
+
+export type RunPass = GPUComputePassEncoder | TgpuComputePass;
+
+export interface RunRecording {
+  pass: RunPass;
+  finish(): void;
+}
+
+export function bindPass(pipeline: TgpuComputePipeline, pass: RunPass): TgpuComputePipeline {
+  return pipeline.with(pass as TgpuComputePass);
```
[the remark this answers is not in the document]
aleksanderkatan: It's just TS not understanding overloads, we don't care

aleksanderkatan (commented):
What I've ween looks good! 👍 

Please, split such PRs into smaller ones. It is unreasonable to expect a proper review of one PR of this size and we keep doing this over and over again. It's exhausting, and probably introduces many hidden errors to our codebase.

apps/typegpu-docs/src/examples/algorithms/concurrent-chart/calculator.ts
reczkok: I tried to remake it but it's doomed IMO - it just does not show anything useful

packages/typegpu-sort/src/bitonic/bitonicSort.ts
reczkok: Yes

packages/typegpu-sort/src/bitonic/slots.ts
reczkok: Yeah, I can try

reczkok:
> What I've ween looks good! 👍
> 
> Please, split such PRs into smaller ones. It is unreasonable to expect a proper review of one PR of this size and we keep doing this over and over again. It's exhausting, and probably introduces many hidden errors to our codebase.

I actually don't quite get why separating this one would be any easier on reviewers. The 2500 lines of additions are pretty neatly split into: examples + docs, logic, and tests and even the files are in this order. I can apply a stacked PR pattern but the dependancies between the changes are significant (so the core logic would land as one PR unless I artificially split which is more work). What that would probably do (and have already done in the soul + worklet stuff) is double maintenance and extend the review time. I will apply this pattern from now on (since that has become a repo pattern) but I see no point in shaming a PR that happened before stacked PRs were even a thing

packages/typegpu-sort/src/bitonic/slots.ts
aleksanderkatan: Another PR! This needs changes in std, and I'd like to talk through the signatures of std functions (we had a soft convention that when JS supports it, std shouldn't, this would break it)

aleksanderkatan:
> I actually don't quite get why separating this one would be any easier on reviewers. 

Maybe we just are built different. I spent 2.5 hours reviewing the refactor, I am 29 out of 44 files through, and I already am so tired that I don't want to do reviews for the rest of the day. And I didn't even read the radix sort code yet.
If this was split, I could give you approves granularly. Now, since I didn't review everything, I cannot do that with clear conscience. 

Unless we don't care about reviews and a quick skim is good enough, but then why review at all? 
Also, regarding maintenance, the longer a PR is up, the more maintenance it needs. And smaller PRs just are merged faster.

aleksanderkatan:
> I will apply this pattern from now on (since that has become a repo pattern) but I see no point in shaming a PR that happened before stacked PRs were even a thing

Nice!
I thought this was already after stacked PRs, three weeks ago, sorry

PR #2763, opened 2026-07-27
feat: Treat tgpu.lazy as resolvable

iwoplaza:
This already worked at runtime, it's just that the types were protesting


reczkok (approved):
🔢

PR #2769, opened 2026-07-29
fix: Lazy buffer views in bitcast to allow for polyfill

iwoplaza:


iwoplaza:
Superseded by https://github.com/software-mansion/TypeGPU/pull/2773

PR #2770, opened 2026-07-29
docs: "Trippy Raymarching" example

vende11s:
Adds trippy interactive 3d sdf raymarching example.

download (5)
download (6)
download (7)
download (8)
download (9)
download (10)
download (2)
download (4)


apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+import { defineControls } from '../../common/defineControls.ts';
+
+const root = await tgpu.init();
+const canvas = document.querySelector('canvas') as HTMLCanvasElement;
+const context = root.configureContext({ canvas, alphaMode: 'premultiplied' });
+
+const cleanupController = new AbortController();
+const { signal } = cleanupController;
+
+canvas.style.touchAction = 'none';
```
reczkok: You could also do this in the html file since it's not changed dynamically

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
+import { tgpu, d, std, common } from 'typegpu';
+import { defineControls } from '../../common/defineControls.ts';
+
+const root = await tgpu.init();
+const canvas = document.querySelector('canvas') as HTMLCanvasElement;
+const context = root.configureContext({ canvas, alphaMode: 'premultiplied' });
```
reczkok: As far as I can see the fragment shader always sets alpha to 1.0 so this is not needed
```suggestion
const context = root.configureContext({ canvas });
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  sphereRadius: d.f32,
+  waveFreqX: d.f32,
+  waveFreqY: d.f32,
+  waveAmplitude: d.f32,
+  twistFactor: d.f32,
+});
+
+const timeBuf = root.createUniform(d.f32, 0);
+const resolutionBuf = root.createUniform(d.vec2f, d.vec2f(canvas.width, canvas.height));
+const mousePosBuf = root.createUniform(d.vec2f, d.vec2f(0));
```
reczkok: Buffers are zero-initialized by default
```suggestion
const timeBuf = root.createUniform(d.f32);
const resolutionBuf = root.createUniform(d.vec2f, d.vec2f(canvas.width, canvas.height));
const mousePosBuf = root.createUniform(d.vec2f);
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  maxSteps: d.i32,
+  sphereRadius: d.f32,
+  waveFreqX: d.f32,
+  waveFreqY: d.f32,
+  waveAmplitude: d.f32,
+  twistFactor: d.f32,
+});
+
+const timeBuf = root.createUniform(d.f32, 0);
+const resolutionBuf = root.createUniform(d.vec2f, d.vec2f(canvas.width, canvas.height));
```
reczkok: This also an alternative way you can write that is more performant, insignificant here but just dropping it as a fun fact
```suggestion
const resolutionBuf = root.createUniform(d.vec2f, [canvas.width, canvas.height]);
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  const v = (clientY - r.top) / r.height;
+  const aspect = r.width / r.height;
+
+  const uvX = (u - 0.5) * 2.0 * aspect;
+  const uvY = (v - 0.5) * -2.0;
+
+  const worldX = uvX * (7.0 / 3.0);
+  const worldY = uvY * (7.0 / 3.0);
+
+  mousePosBuf.write(d.vec2f(worldX, worldY));
```
reczkok: This one is still probably fine but just as good practice
```suggestion
  mousePosBuf.write([worldX, worldY]);
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+canvas.addEventListener('touchmove', onTouchMove, { passive: false, signal });
+
+const palette = (t: number) => {
+  'use gpu';
+  const a = d.vec3f(0.5, 0.5, 0.5);
+  const b = d.vec3f(0.5, 0.5, 0.5);
+  const c = d.vec3f(1.0, 1.0, 1.0);
+  const dCol = d.vec3f(0.0, 0.33, 0.67);
+  return a + b * std.cos((c * t + dCol) * 6.28318);
+};
```
reczkok, line 67: `a` and `b` are identical and all we do with `c` is multiply by it so it has no effect (threw in `Math.PI` cause why not). Also i don't quite get the naming for `dCol` (seems arbitrary in an inconsistent way to a, b and c), this one is a full nit tho, up to you
```suggestion
const palette = (t: number) => {
  'use gpu';
  const a = d.vec3f(0.5);
  const dCol = d.vec3f(0.0, 0.33, 0.67);
  return a + a * std.cos((t + dCol) * Math.PI * 2);
};
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  const c = std.cos(angle);
+  const s = std.sin(angle);
+  const twistedP = d.vec3f(c * p.x - s * p.y, s * p.x + c * p.y, p.z);
+
+  const spacing = params.sphereSpacing;
+  const cell = d.vec3f(
+    std.fract(twistedP.x * (1 / spacing) + -mouse.x * 2) * spacing - spacing / 2.0,
+    std.fract(twistedP.y * (1 / spacing) + mouse.y * 2) * spacing - spacing / 2.0,
+    std.fract(twistedP.z * (1 / spacing)) * spacing - spacing / 2.0,
+  );
```
reczkok: Should be identical and reads much nicer to me 
```suggestion
  const spacing = params.sphereSpacing;
  const scroll = d.vec3f(-mouse.x, mouse.y, 0) * 2;
  const cell = (std.fract(twistedP / spacing + scroll) - 0.5) * spacing;
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  pipeline.withColorAttachment({ view }).draw(3);
+  animationFrameId = requestAnimationFrame(frame);
+}
+
+animationFrameId = requestAnimationFrame(frame);
+
+// #region Example controls and cleanup
+
+export const controls = defineControls({
```
reczkok, line 208: You could collapse that a lot in fun ways, either a helper function or spreading or both (fully optional but nice and satisfying)

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/meta.json
```diff
@@ -0,0 +1,6 @@
+{
+  "title": "Trippy Raymarching",
+  "category": "rendering",
+  "tags": ["raymarching", "visuals", "trippy", "sdf"],
+  "coolFactor": 6.7
```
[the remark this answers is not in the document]
reczkok: Yup, make it either 6 or 7

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
reczkok: And i think you never use the values apart from calculating aspect ratio so might as well collapse this to a single `f32`

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  const uv = d.vec2f((position.x / res.x - 0.5) * 2.0 * aspect, (position.y / res.y - 0.5) * -2.0);
+
+  const ro = d.vec3f(0, 0, time * 3.0);
+  const rd = std.normalize(d.vec3f(uv, 2.8));
+
+  let t = d.f32(0);
+  let glow = d.f32(0);
+  let hit = false;
+  let hitP = d.vec3f(0);
+  let hitColor = d.vec3f(0);
```
reczkok: (nit) You can actually calculate them out of the loop after the marching step, equivalent functionally but I think it will simplify the code a little

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+  const targetHeight = Math.max(1, Math.floor(canvas.clientHeight * dpr * resolutionScale));
+
+  if (canvas.width !== targetWidth || canvas.height !== targetHeight) {
+    canvas.width = targetWidth;
+    canvas.height = targetHeight;
+    resolutionBuf.write(d.vec2f(targetWidth, targetHeight));
+  }
+}
+
+const resizeObserver = new ResizeObserver(() => resizeCanvas());
```
reczkok, line 178:
```suggestion
const resizeObserver = new ResizeObserver(resizeCanvas);
```

apps/typegpu-docs/src/examples/rendering/trippy-raymarching/index.ts
```diff
@@ -0,0 +1,319 @@
...
+    updatePointerPosition(e.touches[0].clientX, e.touches[0].clientY);
+  }
+}
+
+function onTouchStart(e: TouchEvent) {
+  if (e.touches.length > 0) {
+    e.preventDefault();
+    updatePointerPosition(e.touches[0].clientX, e.touches[0].clientY);
+  }
+}
```
reczkok: Did you check if we actually need all this? I think the pointer move event covers all the interactions we need

reczkok (approved):

PR #2776, opened 2026-07-31
docs: Add "Fire Text" example

vende11s:
image


apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
aleksanderkatan, line 1: On my phone on default settings, it runs in 7fps (it's an issue when clicking through the examples, I cannot easily go back since the phone lags so much). 
Can we try to do something with it? I think the easiest thing would be reducing initial solver steps

apps/typegpu-docs/src/examples/simulation/fire-text/index.html
```diff
@@ -0,0 +1,44 @@
...
+  }
+</style>
+<div id="help">
+  <h3 class="controls-header">Controls (click to dismiss)</h3>
+
+  <p id="description">
+    This is an interactive sandbox.<br />
+    Feel free to tweak the simulation settings and brushes to create various VFX.
+  </p>
+
```
aleksanderkatan, line 38: Nit: I think the fewer words are here, the bigger chance people will read it without dismissing immediately
```suggestion
```

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
aleksanderkatan, line 1: nit: Click & type does not work on mobile, and it desynchronises with the example controls text input. 
I think the text input is easily discoverable, have you considered removing the click & type functionality and updating text input to handle newlines instead?

apps/typegpu-docs/src/examples/simulation/fire-text/config.ts
```diff
@@ -0,0 +1,49 @@
...
+  thermalStrength: 50.0,
+  brushMode: 'Velocity Brush' as const,
+  renderMode: 'Fire' as const,
+  text: 'TypeGPU.',
+  cursorBlink: true,
+  maxParticles: 20000,
+} as const;
+
+export const textureSizeOptions = ['128', '256', '512', '1024', '2048'] as const;
+export const brushModes = ['Instant Brush', 'Constant Source Brush', 'Velocity Brush'] as const;
```
aleksanderkatan, line 28: nit: `Constant Source Brush` does not fit in the selection button, how about `Source Brush`?
Also, it is already named `Brush Mode`, I think we could drop the word `Brush` from these modes

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,598 @@
...
+import { createEmitterBindGroup, createStampPipeline } from './emitter.ts';
+import { createParticleRenderBindGroup, createParticles } from './particles.ts';
+import { createDisplayBindGroups, createRenderPipelines } from './render.ts';
+import { createTextMask } from './text.ts';
+import { defineControls, section } from '../../common/defineControls.ts';
+
+type Rgba16Texture = TgpuTexture & SampledFlag & StorageFlag;
+type R32Texture = TgpuTexture & SampledFlag & StorageFlag;
+
+// #region Setup
```
aleksanderkatan, line 21: nit:
Regions are hidden by default 

Image

That's why in most other examples, we only have one region for example controls

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
aleksanderkatan, line 1: Did you consider adding a resolution test analogous to e.g. `apps/typegpu-docs/tests/individual-example-tests/smoky-triangle.test.ts`?

apps/typegpu-docs/src/examples/simulation/fire-text/config.ts
```diff
@@ -0,0 +1,49 @@
...
+  densityDecay: 0.999,
+  tempDecay: 0.996,
+  vorticityStrength: 40.0,
+  thermalStrength: 50.0,
+  brushMode: 'Velocity Brush' as const,
+  renderMode: 'Fire' as const,
+  text: 'TypeGPU.',
+  cursorBlink: true,
+  maxParticles: 20000,
+} as const;
```
aleksanderkatan, line 25: It's this `as const` that forces you to type all runtime values explicitly, you can just remove it and as well as `: number` in `let tempPower: number = defaults.tempPower;
` etc.

apps/typegpu-docs/src/examples/simulation/fire-text/config.ts
```diff
@@ -0,0 +1,49 @@
...
+  buoyancy: 140,
+  timestep: 0.7,
+  tempPower: 8.0,
+  particleSize: 1.8,
+  densityDecay: 0.999,
+  tempDecay: 0.996,
+  vorticityStrength: 40.0,
+  thermalStrength: 50.0,
+  brushMode: 'Velocity Brush' as const,
+  renderMode: 'Fire' as const,
```
aleksanderkatan, line 21: nit:
If you do this, and switch to keeping these values directly instead of indexes, you'll be able to use deconstruction for runtime state initialization

Or you could just keep the state in this object
```suggestion
  brushMode: 'Velocity Brush' as keyof typeof brushModes,
  renderMode: 'Fire' as keyof typeof renderModes,
```
```ts
let { textureSize: currentTextureSize, solverIterations, ... } = defaults;
```

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
aleksanderkatan, line 1: I experimented a little with drawing segments instead of stamps, what do you think about that?

old:

https://github.com/user-attachments/assets/513d4f20-abb0-4743-ac17-3b77d7f9cc13

new:

https://github.com/user-attachments/assets/8b92ef38-1d80-4c0b-b5b7-00816065815b

Here's my shader, though if you wanted to incorporate it, you'd need to tweak dispatching (this works on all pixels instead of just those in the radius), and config & mouse pos catching (naively implemented, this brush will jump when moving cursor outside of the canvas).

```ts
import * as sdf from '@typegpu/sdf';

// ...

export function createStampPipeline(root: TgpuRoot, configUniform: TgpuUniform<typeof Config>) {
  return root.createGuardedComputePipeline((x, y) => {
    'use gpu';
    const radius = configUniform.$.radius;

    const dist = sdf.sdLine(
      d.vec2f(x, y),
      configUniform.$.oldStampPos,
      configUniform.$.newStampPos,
    );

    if (dist <= radius) {
      let weight = d.f32(1);
      if (configUniform.$.isSoft === 1) {
        weight = d.f32(1) - std.smoothstep(radius * 0.05, radius, dist);
      }
      const old = std.textureLoad(constantSourceLayout.$.tex, d.vec2u(x, y)).x;
      std.textureStore(
        constantSourceLayout.$.tex,
        d.vec2u(x, y),
        d.vec4f(std.max(old, weight), 0, 0, 0),
      );
    }
  });
}
```

apps/typegpu-docs/src/examples/simulation/fire-text/config.ts
```diff
@@ -0,0 +1,49 @@
...
+  softBrush: true,
+  buoyancy: 140,
+  timestep: 0.7,
+  tempPower: 8.0,
+  particleSize: 1.8,
+  densityDecay: 0.999,
+  tempDecay: 0.996,
+  vorticityStrength: 40.0,
+  thermalStrength: 50.0,
+  brushMode: 'Velocity Brush' as const,
```
cieplypolar:
```suggestion
  brushMode: 'Velocity Brush',
```
Redundant, because you have outer `as const` assertion.

apps/typegpu-docs/src/examples/simulation/fire-text/config.ts
cieplypolar:
```suggestion
  renderMode: 'Fire',
```
Same as above

cieplypolar:
```suggestion
  renderMode: renderModes[0],
```
For future reference, this prevents values from drifting.

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
```diff
@@ -0,0 +1,60 @@
...
+  type SampledFlag,
+  type StorageFlag,
+  type TgpuBindGroup,
+  type TgpuRoot,
+  type TgpuTexture,
+  type TgpuUniform,
+} from 'typegpu';
+import { Config } from './config.ts';
+
+type R32Texture = TgpuTexture & SampledFlag & StorageFlag;
```
cieplypolar:
```suggestion
type R32Texture = TgpuTexture<{ size: [number, number], format: 'r32float' }> & SampledFlag & StorageFlag;
```

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
```diff
@@ -0,0 +1,60 @@
...
+    const pos = configUniform.$.stampPos;
+    const radius = configUniform.$.radius;
+    const softBrush = configUniform.$.isSoft;
+
+    const dx = d.i32(x) - d.i32(radius);
+    const dy = d.i32(y) - d.i32(radius);
+    const px = d.i32(pos.x) + dx;
+    const py = d.i32(pos.y) + dy;
+
+    const dist = std.length(d.vec2f(d.f32(dx), d.f32(dy)));
```
cieplypolar:
```suggestion
    const dist = std.length(d.vec2f(dx, dy));
```
Vectors automatically perform conversion.

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
```diff
@@ -0,0 +1,60 @@
...
+    const fRadius = d.f32(radius);
```
cieplypolar:
```suggestion
    const fRadius = radius;
```
Already f32

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
```diff
@@ -0,0 +1,60 @@
...
+    const texSize = d.i32(configUniform.$.textureSize);
+
+    if (px >= 0 && px < texSize && py >= 0 && py < texSize) {
+      if (dist <= fRadius) {
+        let weight = d.f32(1.0);
+        if (softBrush === 1) {
+          weight = d.f32(1.0) - std.smoothstep(fRadius * 0.05, fRadius, dist);
+        }
+        const old = std.textureLoad(constantSourceLayout.$.tex, d.vec2u(d.u32(px), d.u32(py))).x;
```
cieplypolar:
```suggestion
        const old = std.textureLoad(constantSourceLayout.$.tex, d.vec2u(px, py)).x;
```

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
```diff
@@ -0,0 +1,60 @@
...
+        std.textureStore(
+          constantSourceLayout.$.tex,
+          d.vec2u(d.u32(px), d.u32(py)),
```
cieplypolar:
```suggestion
          d.vec2u(px, py),
```

apps/typegpu-docs/src/examples/simulation/fire-text/fluid.ts
```diff
@@ -0,0 +1,490 @@
...
+  type TgpuBindGroup,
+  type TgpuRoot,
+  type TgpuSampler,
+  type TgpuTexture,
+  type TgpuUniform,
+} from 'typegpu';
+import { perlin3d } from '@typegpu/noise';
+import { Config, defaults } from './config.ts';
+
+type Rgba16Texture = TgpuTexture & SampledFlag & StorageFlag;
```
cieplypolar:
```suggestion
type Rgba16Texture = TgpuTexture<{ size: [number, number]; format: 'rgba16float' }> & SampledFlag & StorageFlag;
```

apps/typegpu-docs/src/examples/simulation/fire-text/fluid.ts
```diff
@@ -0,0 +1,490 @@
...
+type R32Texture = TgpuTexture & SampledFlag & StorageFlag;
```
cieplypolar:
```suggestion
type R32Texture = TgpuTexture<{ size: [number, number]; format: 'r32float' }> & SampledFlag & StorageFlag;
```

apps/typegpu-docs/src/examples/simulation/fire-text/emitter.ts
aleksanderkatan, line 33: I see some unnecessary casts, like this one (radius is already f32, this variable is redundant), or `d.f32(1.0) - std.smoothstep(...)` (same as `1 - std.smoothstep(...)`), or `d.vec2f(d.f32(dx), d.f32(dy))` (same as `d.vec2f(dx, dy)`)

apps/typegpu-docs/src/examples/simulation/fire-text/particles.ts
```diff
@@ -0,0 +1,195 @@
...
+      const spawnChance = randf.sample();
+
+      const maskVal = std.textureLoad(
+        smokeLayout.$.textSourceTex,
+        d.vec2u(testUv * configUniform.$.textureSize),
+      ).x;
+
+      if (temperature > 0.8 && spawnChance < 0.4 && maskVal < 0.05) {
+        p.pos = testUv * configUniform.$.textureSize;
+        p.vel = d.vec2f((randU * 2.0 - 1.0) * 160.0, -50.0 - randf.sample() * 50.0);
```
aleksanderkatan, line 72: 2.0 is equivalent to 2, and is treated as abstract int, can you remove the `.0` everywhere?

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,628 @@
...
+
+  let velocity = d.vec2f(0);
+  if (brushMode === 2 && isMouseDown) {
+    let dx = mouseTexX - prevMouseTexX;
+    let dy = mouseTexY - prevMouseTexY;
+
+    // Clamp dx to [-3, 3] so max velocity (90) is naturally reached on fast swipes,
+    // while slow movement (e.g. 0.3px) produces small velocity (9).
+    dx = Math.max(-3, Math.min(3, dx));
+    dy = Math.max(-3, Math.min(3, dy));
```
aleksanderkatan, line 280: nit, you could use std.clamp here

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,628 @@
...
+
+    velocity = d.vec2f(dx * 30, dy * 30);
+  }
+
+  prevMouseTexX = mouseTexX;
+  prevMouseTexY = mouseTexY;
+
+  configUniform.patch({
```
aleksanderkatan, line 288: nit: we update all props, we may as well use write
```suggestion
  configUniform.write({
```
Also nit, I feel like this uniform could be split into two on the basis of when it's updated: one that is patched each time an example control is changed, and one with the time, updated each frame (are these params ever used on the CPU side?)

aleksanderkatan (commented):
Some nits and stylistic suggestions

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
vende11s: i will address this issue as soon as dynamically setting the UI slider's value is implemented

apps/typegpu-docs/src/examples/simulation/fire-text/config.ts
vende11s: yeah thats clean imma do that

apps/typegpu-docs/src/examples/simulation/fire-text/events.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  cleanup() {
+    this.canvas.removeEventListener('pointerdown', this.onPointerDown);
+    this.canvas.removeEventListener('pointermove', this.onPointerMove);
+    this.canvas.removeEventListener('pointerup', this.onPointerRelease);
+    this.canvas.removeEventListener('pointercancel', this.onPointerRelease);
+    this.canvas.removeEventListener('pointerenter', this.onPointerEnter);
+    this.canvas.removeEventListener('pointerleave', this.onPointerLeave);
+  }
```
reczkok: Not needed afaik, canvas get's destroyed after you leave the example, only things registered on window need explicit cleanup

apps/typegpu-docs/src/examples/simulation/fire-text/events.ts
```diff
@@ -0,0 +1,162 @@
...
+  if (x0 > x1 || y0 > y1) {
+    return undefined;
+  }
+  return { originX: x0, originY: y0, width: x1 - x0 + 1, height: y1 - y0 + 1 };
+}
+
+export class EventHandler {
+  private _isMouseDown = false;
+  private _inside = false;
+  private _breakStroke = false;
```
reczkok: I feel like we generally prefer [private elements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_elements)

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+  return [make(0, 1), make(1, 0)];
+}
+
+let smokeGrid: [Rgba16Texture, Rgba16Texture];
+let constantSourceGrid: R32Texture;
+let textSourceGrid: R32Texture;
+let textFillGrid: R32Texture;
+let pressureGrid: [R32Texture, R32Texture];
+let divergenceGrid: R32Texture;
+let gridsReady = false;
```
reczkok, line 108: All of this state (+ the type aliases, overloads, `pingPong`, `gridTextures`, the bind group `let`s and `rebuildBindGroups`) only exists because we recreate things in place. I'd do a single `createGrids(size)` that returns textures and bind groups in one object, types get inferred and resize is just destroy + recreate

```ts
function createGrids(size: number) {
  const smoke = () =>
    root.createTexture({ size: [size, size], format: 'rgba16float' }).$usage('storage', 'sampled');
  const field = () =>
    root.createTexture({ size: [size, size], format: 'r32float' }).$usage('storage', 'sampled');
  const textures = { smoke: [smoke(), smoke()], pressure: [field(), field()], divergence: field(), ... };
  const smokeGroup = (i: number) =>
    root.createBindGroup(smokeLayout, { inTex: textures.smoke[i], outTex: textures.smoke[1 - i], ... });
  return { textures, smoke: [smokeGroup(0), smokeGroup(1)], ... };
}

let grids = createGrids(defaults.textureSize);
```

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+});
+textMask.start();
+
+function updateTextureSize(newSize: number) {
+  currentTextureSize = newSize;
+  recreateGridTextures(newSize);
+  rebuildBindGroups();
+  particleBuffer.clear();
+  textMask.setTextureSize(newSize);
+}
```
reczkok, line 250: Needs an early return when the size doesn't change, the runner calls every control handler with its initial value so right now we build all grids twice on startup (and rasterize the text again)

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+  isMouseDown: 0,
+  mouseVelocity: [0, 0],
+  densityDecay: defaults.densityDecay,
+  tempDecay: defaults.tempDecay,
+});
+const forceUniform = root.createUniform(ForceParams, {
+  buoyancy: defaults.buoyancy,
+  vorticityStrength: defaults.vorticityStrength,
+  thermalStrength: defaults.thermalStrength,
+});
```
reczkok, line 88: Too many separate uniforms imo (8 of them, accessors in 4 files and 5 different `root.with` chains below). I'd group them into 3 structs (sim, brush, shading) with the accessors in params.ts and create all pipelines from one configured root

```ts
const configured = root
  .with(sim, simUniform)
  .with(brush, brushUniform)
  .with(shading, shadingUniform)
  .pipe(noiseCache.inject());
```

Initial values aren't needed either, control handlers write them on startup anyway

apps/typegpu-docs/src/examples/simulation/fire-text/events.ts
```diff
@@ -0,0 +1,153 @@
...
+  const y0 = Math.max(0, Math.floor(Math.min(a.y, b.y) - pad));
+  const x1 = Math.min(texSize - 1, Math.ceil(Math.max(a.x, b.x) + pad));
+  const y1 = Math.min(texSize - 1, Math.ceil(Math.max(a.y, b.y) + pad));
+  if (x0 > x1 || y0 > y1) {
+    return undefined;
+  }
+  return { originX: x0, originY: y0, width: x1 - x0 + 1, height: y1 - y0 + 1 };
+}
+
+export class EventHandler {
```
reczkok, line 18: I'd replace the whole class with a small closure, most of this doesn't do anything:
- no cleanup needed, all listeners are on the canvas which gets destroyed with the example
- `setPointerCapture` can't throw in `pointerdown` and capture is released implicitly on `pointerup`/`pointercancel`
- with capture set `hasCapture()` is true for the whole stroke, so `isPainting` is just `_isMouseDown` and the enter/leave/`_breakStroke` logic never kicks in
- `keydown` never fires on the canvas (not focusable), help can be a single `pointerdown` listener with `{ once: true }` in index.ts
- `strokeAabb` is brush math, not events

```ts
export function trackPointer(canvas: HTMLCanvasElement, gridSize: () => number) {
  const pointer = { down: false, x: 0, y: 0, prevX: 0, prevY: 0 };
  const track = (e: PointerEvent) => {
    const rect = canvas.getBoundingClientRect();
    pointer.x = ((e.clientX - rect.left) / rect.width) * gridSize();
    pointer.y = ((e.clientY - rect.top) / rect.height) * gridSize();
  };
  canvas.addEventListener('pointerdown', (e) => {
    pointer.down = true;
    canvas.setPointerCapture(e.pointerId);
    track(e);
    pointer.prevX = pointer.x;
    pointer.prevY = pointer.y;
  });
  canvas.addEventListener('pointermove', (e) => {
    if (pointer.down) {
      track(e);
    }
  });
  canvas.addEventListener('pointerup', () => {
    pointer.down = false;
  });
  canvas.addEventListener('pointercancel', () => {
    pointer.down = false;
  });
  return pointer;
}
```

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+
+  brushUniform.patch({
+    oldStampPos: from,
+    newStampPos: to,
+    origin: stampBounds ? [stampBounds.originX, stampBounds.originY] : [0, 0],
+  });
+  advectionUniform.patch({
+    isMouseDown: painting ? 1 : 0,
+    mouseVelocity: brushModes[brushMode] === 'Velocity' ? events.pointerVelocity() : [0, 0],
+  });
```
reczkok, line 317: The mode gets decided twice, here and again in the shader (`brushMode !== CONSTANT_BRUSH` in fluid.ts). I'd write a single `painting` flag from here (`down && mode !== 'Constant Source'`) and drop `brushMode` from `AdvectionParams`. Also no need for the `indexOf`s, keep the modes as strings, `as const` already types them

Not sure the AABB is worth it either, it's 1 dispatch out of ~55 per frame and costs `strokeAabb`, `origin` and mirroring `brushRadius` in JS

apps/typegpu-docs/src/examples/simulation/fire-text/text.ts
```diff
@@ -0,0 +1,185 @@
...
+    clearInterval(blinkTimer);
+    if (!blinkEnabled) {
+      caretVisible = false;
+      return;
+    }
+    blinkTimer = setInterval(() => {
+      caretVisible = !caretVisible;
+      uploadMask();
+    }, 800);
+  }
```
reczkok, line 115: Every blink re-rasterizes the whole mask (2x `getImageData` + 2 texture writes), at 2048^2 that's ~64 MB of copies every 800ms. Caret is just a rect, I'd put it in a uniform and blink it in the shader based on `time` (and max it into the mask where we read it so sparks still respect it). You can pass the last keystroke time too if you want the blink to reset while typing

apps/typegpu-docs/src/examples/simulation/fire-text/text.ts
```diff
@@ -0,0 +1,185 @@
...
+
+type WritableTexture = { write(data: Float32Array): void };
+
+export interface CreateTextMaskOptions {
+  getTextSourceGrid: () => WritableTexture;
+  getTextFillGrid: () => WritableTexture;
+  getTextureSize: () => number;
+  outlineWidth?: number;
+  initialText?: string;
+}
```
reczkok, line 11: Most of these aren't needed, `getTextureSize` is only read once (line 17, there's `setTextureSize` anyway), `outlineWidth` is never passed and `initialText` is the default. Could be `createTextMask(size, write)` and `start()` can just happen on creation

apps/typegpu-docs/src/examples/simulation/fire-text/text.ts
```diff
@@ -0,0 +1,185 @@
...
+    getTextSourceGrid().write(floatData);
+
+    maskCtx.clearRect(0, 0, currentTextureSize, currentTextureSize);
+    maskCtx.globalCompositeOperation = 'source-over';
+    maskCtx.fillStyle = 'rgba(255, 255, 255, 1)';
+    for (const { line, x, y } of layout) {
+      if (line.length > 0) maskCtx.fillText(line, x, y);
+    }
+    readAlphaInto(fillData);
+    getTextFillGrid().write(fillData);
```
reczkok, line 102: Outline and fill could share one `rg32f` texture (outline in r, fill in g), one write instead of two

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+  },
+
+  'Clear All Grids': {
+    onButtonClick: () => {
+      for (const tex of gridTextures()) {
+        tex.clear();
+      }
+      particleBuffer.clear();
+      textMask.uploadMask();
+    },
```
reczkok, line 438: No need to clear the text textures here, then `uploadMask` doesn't need to be exported

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+
+  if (renderModes[renderMode] === 'Fire') {
+    particlePipeline
+      .with(particleRenderBg)
+      .withColorAttachment({ view: context, loadOp: 'load' })
+      .with(encoder)
+      .draw(4, numParticles);
+  }
+
+  root.device.queue.submit([encoder.finish()]);
```
reczkok, line 404: I'd use `root['~unstable'].createCommandEncoder()` and do display + particles in one render pass, the second one only exists for `loadOp: 'load'`

```ts
const render = encoder.beginRenderPass({ colorAttachments: [{ view: context }] });
displayPipelines[renderMode].with(displayBgs[even]).with(render).draw(3);
if (renderMode === 'Fire') {
  particlePipeline.with(particleRenderBg).with(render).draw(4, numParticles);
}
render.end();
encoder.submit();
```

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+  .with(fireColorHsvAccess, fireColorHsvUniform);
+const displayPipelines = [smokeFragment, densityFragment, velocityFragment].map((fragment) =>
+  displayRoot.createRenderPipeline({
+    vertex: common.fullScreenTriangle,
+    fragment,
+  }),
+);
+
+function simulate(pass: GPUComputePassEncoder) {
+  even = 1 - even;
```
reczkok, line 299: (nit) I'd flip this at the top of `frame`, it's read there (line 392)

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+    .dispatchWorkgroups(gridWg, gridWg);
+
+  if (stampBounds) {
+    stampPipeline
+      .with(pass)
+      .with(stampSourceBg)
+      .dispatchWorkgroups(Math.ceil(stampBounds.width / 16), Math.ceil(stampBounds.height / 16));
+  }
+
+  events.commitStroke();
```
reczkok, line 334: (nit) Looks like it affects the GPU work, I'd move it next to the uniform patches above

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+function frame() {
+  if (canvas.width === 0 || canvas.height === 0) {
+    animationFrameId = requestAnimationFrame(frame);
+    return;
+  }
+
+  const now = performance.now();
+  const deltaSeconds =
+    lastFrameTime === undefined ? 1 / 60 : Math.min((now - lastFrameTime) / 1000, 1 / 30);
+  lastFrameTime = now;
```
reczkok, line 377: Do we need the zero size guard? Also rAF gives you the timestamp so `lastFrameTime` can just start at `performance.now()`

apps/typegpu-docs/src/examples/simulation/fire-text/fluid.ts
```diff
@@ -0,0 +1,245 @@
...
+    smokeLayout.$.linearSampler,
+    uv - (flow * c.dt) / size,
+    0,
+  );
+
+  const emitted = std.textureLoad(sourceLayout.$.tex, pos, 0).x;
+  const printed = std.textureLoad(smokeLayout.$.textTex, pos, 0).x;
+  let velocity = state.xy;
+  let density = std.max(state.z, std.max(emitted, printed));
+  let heat = std.max(state.w, std.max(emitted, printed * defaults.textStartTemperature));
```
reczkok, line 88: `textStartTemperature` is always 1 (can go from defaults and the import too)
```suggestion
  let heat = std.max(state.w, std.max(emitted, printed));
```

apps/typegpu-docs/src/examples/simulation/fire-text/fluid.ts
```diff
@@ -0,0 +1,245 @@
...
+
+  if (s.isMouseDown === 1 && s.brushMode !== CONSTANT_BRUSH) {
+    const weight = brushFalloff(
+      brushDistance(d.vec2f(pos) + 0.5, b.oldStampPos, b.newStampPos),
+      b.radius,
+      b.isSoft,
+      0.1,
```
reczkok, line 95: (nit) magic `0.1` here and `0.05` in `stamp`, I'd name them

apps/typegpu-docs/src/examples/simulation/fire-text/particles.ts
```diff
@@ -0,0 +1,112 @@
...
+    return {
+      pos: d.vec4f(-2000, -2000, 0, 1),
+      life: 0,
+      offset: d.vec2f(),
+      texUv: d.vec2f(),
+    };
+  }
+
+  const size = textureWidth(particleRenderLayout.$.textTex);
+  const offset = [d.vec2f(-1, -1), d.vec2f(1, -1), d.vec2f(-1, 1), d.vec2f(1, 1)][input.vIdx];
```
reczkok, line 85: (nit) could be a `tgpu.const` like `CROSS`

apps/typegpu-docs/src/examples/simulation/fire-text/render.ts
```diff
@@ -0,0 +1,76 @@
...
+const BASE_FIRE_HSV = rgbToHsv(defaults.fireColor);
+
+export const tintByFireColor = (rgb: d.v3f) => {
+  'use gpu';
+  const hsv = rgbToHsv(rgb);
+  const user = fireColorHsvAccess.$;
+  return hsvToRgb(
+    d.vec3f(
+      std.fract(hsv.x + user.x - BASE_FIRE_HSV.x + 1),
+      hsv.y * (user.y / BASE_FIRE_HSV.y),
```
reczkok, line 19: Hue shift and saturation scale are the same for every fragment, I'd compute them on the CPU when the color changes and pass them in the uniform

apps/typegpu-docs/src/examples/simulation/fire-text/layouts.ts
```diff
@@ -0,0 +1,46 @@
...
+export const constantSourceLayout = tgpu.bindGroupLayout({
+  tex: { storageTexture: d.textureStorage2d('r32float', 'read-write') },
+});
+
+export const particleComputeLayout = tgpu.bindGroupLayout({
+  particles: { storage: d.arrayOf(Particle), access: 'mutable' },
+  linearSampler: { sampler: 'filtering' },
+  inTex: { texture: d.texture2d() },
+  textTex: { texture: d.texture2d(), sampleType: 'unfilterable-float' },
+});
```
reczkok, line 36: Duplicates the smoke texture + sampler, particle update could use `smokeLayout` + a layout with just the particles, then this bind group doesn't need to be recreated on resize. I'd also keep the layouts next to the shaders that use them (matter of taste, you can also just leave it like that, we do that sometimes)

apps/typegpu-docs/src/examples/simulation/fire-text/utils.ts
```diff
@@ -0,0 +1,26 @@
+import { sdLine } from '@typegpu/sdf';
```
reczkok, line 1: (nit) `brushDistance`/`brushFalloff` are only used in fluid.ts and `textureWidth` is a one liner, I'd drop this file

apps/typegpu-docs/src/examples/simulation/fire-text/text.ts
```diff
@@ -0,0 +1,185 @@
...
+    const layout = lines.map((line, i) => {
+      const width = maskCtx.measureText(line).width;
+      const x = (currentTextureSize - width) / 2;
+      const y = firstBaseline + i * lineHeight;
+      return { line, x, y, width };
+    });
+
+    maskCtx.globalCompositeOperation = 'source-over';
+    for (const { line, x, y } of layout) {
+      if (line.length > 0) maskCtx.strokeText(line, x, y);
```
reczkok, line 78: (nit) braces (also in some other places)

apps/typegpu-docs/src/examples/simulation/fire-text/meta.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Fire Text",
+  "description": "A stable-fluids playground. By default it simulates fire from typed text and brush strokes.",
+  "category": "simulation",
+  "tags": ["stable fluids", "fire", "sandbox"],
```
reczkok, line 5: We already have tags for these
```suggestion
  "tags": ["fluid", "particles", "interaction"],
```

apps/typegpu-docs/src/examples/simulation/fire-text/index.ts
```diff
@@ -0,0 +1,600 @@
...
+];
+let particleComputeBgs: [
+  TgpuBindGroup<typeof particleComputeLayout.entries>,
+  TgpuBindGroup<typeof particleComputeLayout.entries>,
+];
+let particleRenderBg: TgpuBindGroup<typeof particleRenderLayout.entries>;
+let displayBgs: [
+  TgpuBindGroup<typeof displayLayout.entries>,
+  TgpuBindGroup<typeof displayLayout.entries>,
+];
```
reczkok, line 165: These all go away with `createGrids`, bind group types are inferred from `root.createBindGroup` so none of this needs to be spelled out

reczkok (changes requested):

PR #2781, opened 2026-08-03
fix: Reject decorated element types in arrayOf

### LINKED ISSUE #2754: bug: `d.align` and `d.size` can be passed to `d.arrayOf` but do not have any effect
```ts
it('align test', ({ root }) => {
  const buf = root.createMutable(d.arrayOf(d.size(16, d.u32), 3));
  console.log(tgpu.resolve([buf]));
});
```

Generates:
```wgsl
@group(0) @binding(0) var<storage, read_write> buf: array<u32, 3>;
```

TODO:
- check buffers

vende11s:
Fixes #2754

Throws when a non-`@location` decorated element type is passed to `arrayOf`.

`@location` on array elements remains allowed, since it's used effectively with `vertexLayout`.


packages/typegpu/tests/buffer.test.ts
```diff
@@ -861,7 +861,7 @@ describe('TgpuBuffer', () => {
   });
 
   it('should ignore decorated types when determining validity usage', ({ root }) => {
-    const validSchema = d.size(1024, d.arrayOf(d.align(16, d.u32), 32));
+    const validSchema = d.size(1024, d.arrayOf(d.u32, 32));
```
reczkok: This change defeats the purpose of this test (either remove it or repurpose)

packages/typegpu/tests/array.test.ts
```diff
@@ -551,6 +551,33 @@ describe('array', () => {
       - fn*:main(): Value [1, 2, 3] is not resolvable]
     `);
   });
+
+  it('allows @location-decorated element types', () => {
+    const located = d.arrayOf(d.location(0, d.u32), 4);
```
aleksanderkatan, line 551: Can arrays hold location? From location docs:

[Must](https://www.w3.org/TR/WGSL/#shader-creation-error) only be applied to an entry point function parameter, entry point return type, or member of a [structure](https://www.w3.org/TR/WGSL/#structure) type.

packages/typegpu/tests/array.test.ts
```diff
@@ -551,6 +551,33 @@ describe('array', () => {
...
+    expectTypeOf(located).toEqualTypeOf<d.WgslArray<d.Decorated<d.U32, [d.Location<0>]>>>();
+  });
+
+  it('should allow calling arrayOf inside generic helpers', () => {
+    function arrayOf32<T extends d.AnyWgslData>(schema: T) {
+      return d.arrayOf(schema, 32);
+    }
+
+    arrayOf32(d.f32);
+  });
```
aleksanderkatan, line 562: Is this test relevant to changes made in this PR?

vende11s: Kind of - when restricting decorator usage inside arrayOf, we had to make sure we didn't break generic helpers like this

packages/typegpu/tests/array.test.ts
vende11s: In the current implementation of vertexLayout, `d.location` in an array isn’t treated like a decorator the way it is for a struct member.

In `vertexLayout `the behaviour of `d.location` is setting the pipeline’s shaderLocation attribute.

This can be misleading, and maybe a better approach would be to make decorators single-purpose. We could find another way to pass shaderLocation to the vertex layout.

What do you think?

reczkok (approved):

PR #2789, opened 2026-08-05
fix: Global identifier inside nested function call

### LINKED ISSUE #2788: fix: Name collisions
```ts
  it('...', () => {
    const myConst = tgpu.const(d.u32, 1).$name('COLLISION');

    const helper = () => {
      'use gpu';
      myConst.$;
    };

    const test = (COLLISION: number) => {
      'use gpu';
      helper();
      const x = COLLISION;
      const y = myConst.$;
    };

    const main = () => {
      'use gpu';
      test(1);
    };

      "const COLLISION: u32 = 1u;

      fn helper() {
        COLLISION;
      }

      fn test(COLLISION: i32) {
        helper();
        let x = COLLISION;
        const y = COLLISION;
      }

      fn main() {
        test(1i);
      }"
    `);
  });
```

cieplypolar:
Closes #2788.


packages/typegpu/src/resolutionCtx.ts
```diff
@@ -250,6 +250,19 @@ class ItemStateStackImpl implements ItemStateStack {
     return false;
   }
 
+  isIdentifierTakenGlobally(id: string): boolean {
```
iwoplaza: I find this name a bit misleading, as it's not checking whether the identifier has been taken globally. It only checks whether an identifier is taken in the current function scope, or any of the scopes of the callers up the stack until the entry function.

Maybe something like:
```suggestion
  isIdentifierTakenInCallStack(id: string): boolean {
```

Some documentation would also be nice why this function is helpful 👌

packages/typegpu/src/types.ts
```diff
@@ -360,7 +360,7 @@ export interface ResolutionCtx {
   makeUniqueIdentifier(primer: string | undefined, scope: 'global' | 'block'): string;
 
   isIdentifierBanned(name: string): boolean;
-  isIdentifierTaken(name: string): boolean;
+  isIdentifierTaken(name: string, scope: 'global' | 'block'): boolean;
```
iwoplaza, line 367: It would be nice to document that `scope` doesn't refer to the scope we want to check, but the scope we want to place the identifier in and check whether it could cause a potential collision.

packages/typegpu/tests/tgsl/nameClashes.test.ts
aleksanderkatan, line 1: Free test

```ts
test('should give different names to global declaration accessed inside nested function and local variable of wrapping function', () => {
  const myConst = tgpu.const(d.u32, 1).$name('COLLISION');

  const helper = () => {
    'use gpu';
    return myConst.$;
  };

  const test = () => {
    'use gpu';
    const COLLISION = 1;
    helper();
    const x = COLLISION;
    const y = myConst.$;
  };

  expect(tgpu.resolve([test])).toMatchInlineSnapshot(`
    "const COLLISION_1: u32 = 1u;

    fn helper() -> u32 {
      return COLLISION_1;
    }

    fn test() {
      const COLLISION = 1;
      helper();
      let x = COLLISION;
      const y = COLLISION_1;
    }"
  `);
});
```

packages/typegpu/src/resolutionCtx.ts
cieplypolar: sure

packages/typegpu/src/types.ts
cieplypolar: will be done

aleksanderkatan (approved):
Great! 
Just address Iwo's comments before merging

iwoplaza (approved):

PR #2792, opened 2026-08-05
feat: Allow shader generators to report different language keys

iwoplaza:


packages/typegpu-gl/src/glslGenerator.ts
```diff
@@ -84,6 +84,10 @@ export class GlslGenerator extends WgslGenerator {
   #functionType: TgpuShaderStage | 'normal' | undefined;
   #entryFnState: EntryFnState | undefined;
 
+  static {
```
cieplypolar, line 87: is there a reason why declare is omitted in glsl generator?
```ts
// prototype properties
declare languageKey: string;
```
The above code snippet comes from wgsl generator.

iwoplaza: Because it inherits that property from the `WgslGenerator` class

cieplypolar (approved):

reczkok (approved):

PR #2794, opened 2026-08-05
feat: Use @typegpu/gl as a fallback when @typegpu/three is made to generate GLSL

iwoplaza:


packages/typegpu-three/src/typegpu-node.ts
```diff
@@ -142,6 +158,11 @@ function forceExplicitVoidReturn(codeIn: string) {
   return codeIn.substring(0, closingParen + 1) + '-> void' + codeIn.substring(closingParen + 1);
 }
 
+function isWebGL(builder: THREE.NodeBuilder): boolean {
+  const backend = builder.renderer?.backend;
+  return !!backend && 'isWebGLBackend' in backend && !!backend.isWebGLBackend;
```
cieplypolar:
```suggestion
  return !!backend && (backend as unknown as { isWebGLBackend}).isWebGLBackend === true;
```

cieplypolar (approved):

reczkok (approved):

PR #2796, opened 2026-08-06
feat: Souls

reczkok:
Groundwork for the transfer protocol (PR on top of this one). To move a resource between runtimes we need its complete definitional state as a plain serializable record, and previously that state was scattered across private fields, closures and getters per resource. This PR gathers it into one place without changing any behavior.

- every resource now carries a `[$soul]` - a plain record of everything that survives a runtime boundary (own data properties only, no methods, no prototype, no symbol keys)
- `[$internal]` changed from `true` to an object exposing `materialize()`, which fills the soul's `raw` with the device object for resources that own one (buffer, texture, query set, pipelines, ...)
- resource impls read their state through the soul instead of private fields, so the soul is the single source of truth rather than a copy that could drift

### DIFF
--- a/packages/typegpu/src/core/buffer/buffer.ts
+++ b/packages/typegpu/src/core/buffer/buffer.ts
@@ -18,7 +18,8 @@ import type {
   IsValidVertexSchema,
   MemIdentity,
 } from '../../shared/repr.ts';
-import { $internal } from '../../shared/symbols.ts';
+import type { TgpuDeviceOwningSoul } from '../../shared/soul.ts';
+import { $internal, $soul } from '../../shared/symbols.ts';
 import type { Prettify, UnionToIntersection } from '../../shared/utilityTypes.ts';
 import { isGPUBuffer } from '../../types.ts';
 import type { TgpuCommandEncoder } from '../commandEncoder/commandEncoder.ts';
@@ -64,7 +65,16 @@ export interface IndirectFlag {
  */
 export type Vertex = VertexFlag;
 
-type UsageLiteral = 'uniform' | 'storage' | 'vertex' | 'index' | 'indirect';
+export type UsageLiteral = 'uniform' | 'storage' | 'vertex' | 'index' | 'indirect';
+
+export interface TgpuBufferSoul<TData extends BaseData = BaseData> extends TgpuDeviceOwningSoul<
+  'buffer',
+  GPUBuffer
+> {
+  readonly dataType: TData;
+  flags: GPUBufferUsageFlags;
+  readonly usages: UsageLiteral[];
+}
 
 type LiteralToUsageType<T extends UsageLiteral> = T extends 'uniform'
   ? UniformFlag
@@ -116,7 +126,10 @@ export type BufferInitialData<TData extends BaseData> =
   | BufferInitCallback<TData>;
 
 export interface TgpuBuffer<TData extends BaseData> extends TgpuNamable {
-  readonly [$internal]: true;
+  readonly [$internal]: {
+    readonly materialize: () => GPUBuffer;
+  };
+  readonly [$soul]: TgpuBufferSoul<TData>;
   readonly resourceType: 'buffer';
   readonly dataType: TData;
   readonly initial?: InferInput<TData> | undefined;
@@ -172,13 +185,12 @@ export function INTERNAL_createBuffer<TData extends AnyData>(
 // Implementation
 // --------------
 class TgpuBufferImpl<TData extends BaseData> implements TgpuBuffer<TData> {
-  readonly [$internal] = true;
+  readonly [$internal]: {
+    readonly materialize: () => GPUBuffer;
+  };
+  readonly [$soul]: TgpuBufferSoul<TData>;
   readonly resourceType = 'buffer';
-  flags: GPUBufferUsageFlags = GPUBufferUsage.COPY_DST | GPUBufferUsage.COPY_SRC;
-  readonly dataType: TData;
 
-  readonly #device: GPUDevice;
-  #buffer: GPUBuffer | null = null;
   #ownBuffer: boolean;
   #destroyed = false;
   #internalBuffer: ArrayBuffer | undefined;
@@ -204,12 +216,19 @@ class TgpuBufferImpl<TData extends BaseData> implements TgpuBuffer<TData> {
     initialOrBuffer?: BufferInitialData<TData> | GPUBuffer,
     disallowedUsages?: UsageLiteral[],
   ) {
-    this.dataType = dataType;
+    this[$soul] = {
+      type: 'buffer',
+      device: root.device,
+      dataType,
+      flags: GPUBufferUsage.COPY_DST | GPUBufferUsage.COPY_SRC,
+      usages: [],
+      raw: undefined,
+      label: undefined,
+    };
     this.#disallowedUsages = disallowedUsages;
-    this.#device = root.device;
     if (isGPUBuffer(initialOrBuffer)) {
       this.#ownBuffer = false;
-      this.#buffer = initialOrBuffer;
+      this[$soul].raw = initialOrBuffer;
     } else {
       this.#ownBuffer = true;
       if (typeof initialOrBuffer === 'function') {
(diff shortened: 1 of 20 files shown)


packages/typegpu/src/shared/symbols.ts
```diff
@@ -1,6 +1,8 @@
 import { version } from 'typegpu/package.json';
 
 export const $internal = Symbol(`typegpu:${version}:$internal`);
+/** A plain record of all definitional state of a resource, surviving a device boundary */
```
iwoplaza: What's a "device boundary"?

packages/typegpu/src/core/buffer/buffer.ts
```diff
@@ -204,12 +216,19 @@ class TgpuBufferImpl<TData extends BaseData> implements TgpuBuffer<TData> {
...
   ) {
-    this.dataType = dataType;
+    this[$soul] = {
+      type: 'buffer',
+      device: root.device,
+      dataType,
+      flags: GPUBufferUsage.COPY_DST | GPUBufferUsage.COPY_SRC,
+      usages: [],
+      raw: undefined,
+      label: undefined,
```
iwoplaza, line 226: Is this label ever updated?

packages/typegpu/src/core/root/init.ts
iwoplaza: This documentation is no longer attached to the root

packages/typegpu/src/shared/symbols.ts
reczkok: My weird mental identity for runtime, will change, forgot that GPUDevice is indeed a device

packages/typegpu/src/core/buffer/buffer.ts
reczkok: image

reczkok: should be now

packages/typegpu/src/shared/soul.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * The base shape of a soul: a plain record of all definitional state of a resource,
+ * holding everything that survives a device boundary. Souls carry own data properties
+ * only, no methods, no prototype and no symbol keys
+ */
```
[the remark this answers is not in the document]
reczkok: yes yes my wording was confusing i get it

packages/typegpu/src/core/buffer/buffer.ts
cieplypolar: where?

packages/typegpu/src/core/buffer/bufferBinding.ts
```diff
@@ -16,8 +17,15 @@ import { type BufferWriteOptions, type TgpuBuffer, type UniformFlag } from './bu
...
 // ----------
 
+export interface TgpuBufferBindingSoul<
+  TBuffer extends TgpuBuffer<BaseData> = TgpuBuffer<BaseData>,
+> extends TgpuSoul<'uniform' | 'mutable' | 'readonly'> {
+  readonly buffer: TBuffer;
+}
+
 interface TgpuBufferBindingBase<TData extends BaseData> extends TgpuNamable {
   readonly [$internal]: true;
```
cieplypolar, line 27: no `materialize`?

packages/typegpu/src/core/root/init.ts
```diff
@@ -116,11 +115,10 @@ const workgroupSizeConfigs = [
 export class TgpuGuardedComputePipelineImpl<
   TArgs extends number[],
 > implements TgpuGuardedComputePipeline<TArgs> {
-  #root: ExperimentalTgpuRoot;
-  #pipeline: TgpuComputePipeline;
-  #sizeUniform: TgpuUniform<Vec3u>;
-  #workgroupSize: v3u;
+  readonly resourceType = 'guarded-compute-pipeline';
```
cieplypolar, line 118: why duplicate it if it exists in `$soul`?

this applies all resources.

packages/typegpu/src/core/buffer/buffer.ts
cieplypolar: nevermind

packages/typegpu/src/shared/meta.ts
```diff
@@ -51,12 +52,18 @@ function isForwarded(value: unknown): value is { [$getNameForward]: unknown } {
...
+}
+
 export function getName(definition: unknown): string | undefined {
   if (isForwarded(definition)) {
     return getName(definition[$getNameForward]);
   }
   return (
-    nameMap.get(definition as object) ?? globalExt.__TYPEGPU_META__?.get(definition as object)?.name
+    nameMap.get(definition as object) ??
+    soulOf(definition)?.label ??
```
cieplypolar, line 66:
```suggestion
```
it is a subset of `nameMap`

packages/typegpu/src/core/texture/texture.ts
```diff
@@ -224,61 +244,70 @@ export function isTextureView(value: unknown): value is TgpuTextureView {
...
   usableAsRender = false;
 
   #formatInfo: TextureFormatInfo;
   #destroyed = false;
-  #flags = GPUTextureUsage.COPY_DST | GPUTextureUsage.COPY_SRC;
-  #flagsOverridden = false;
-  #texture: GPUTexture | null = null;
-  #branch: ExperimentalTgpuRoot;
 
   constructor(props: TProps, branch: ExperimentalTgpuRoot) {
```
cieplypolar: could you rename `branch` to `root`?
- computePipeline.ts
- sampler.ts
- texture.ts

cieplypolar (commented):
I think you left the tgpuVariable soulless 😢

packages/typegpu/src/core/buffer/buffer.ts
aleksanderkatan, line 226: If labels are meant to never be updated by soul bearing objects, can we just make them hidden? (not declare it in the type nor put it in these objects manually, and still assign it in setName)

Otherwise, I'd try to document it well in soul.ts that we should not update it manually.

packages/typegpu/src/shared/meta.ts
iwoplaza: Not necessarily, as `nameMap` is runtime-local, while the `label` stored on the soul can travel between runtimes.

iwoplaza: But a comment would be nice 👍

cieplypolar: I see

packages/typegpu/src/core/root/init.ts
iwoplaza: We can consider deprecating this field and removing it in a future minor

iwoplaza: But currently, it poses no harm

packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -155,11 +170,35 @@ type Memo = {
...
+      label: undefined,
+    };
+    this[$internal] = {
+      core,
+      priors,
+      root: core.root,
+      materialize: () => {
+        const soul = this[$soul];
+        if (!soul.raw) {
+          const memo = core.unwrap();
```
aleksanderkatan, line 191: I think I asked about this somewhere, but I don't remember the answer. What happens to catchall and logResources, is it documented somewhere?

packages/typegpu/src/core/pipeline/priors.ts
```diff
@@ -0,0 +1,15 @@
+const TRANSFERABLE_PRIORS = new Set([
+  'bindGroupLayoutMap',
+  'vertexLayoutMap',
+  'indexBuffer',
+  'stencilReference',
+  'timestampWrites',
+  'performanceCallback',
+]);
```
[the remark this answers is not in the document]
aleksanderkatan: Why is the clanker wrong?

packages/typegpu/src/core/root/init.ts
```diff
@@ -328,6 +353,11 @@ class TgpuRootImpl extends WithBindingImpl implements TgpuRoot, ExperimentalTgpu
     this.shaderGenerator = shaderGenerator;
 
     this['~unstable'] = this;
+    this[$soul] = {
+      type: 'root',
+      device,
+      label: undefined,
+    };
```
aleksanderkatan, line 363: Settings are not included (`unstable_names`, `unstable_logOptions`,  and I guess shaderGenerator as well)

aleksanderkatan (approved):
Great job! 🇰🇷👲

iwoplaza (approved):
Very extra awesome! 🕺🪩
Left one nit, mostly just to document how labels can travel between runtimes, and that name maps are runtime-local.

PR #2797, opened 2026-08-06
feat: Transfer protocol for TypeGPU resources (on souls)

### LINKED ISSUE #2690: feat(@typegpu/react): Add ability to pass device and adapter options to <Root> component

reczkok:
Replaces #2732, rebuilt on top of #2796. With souls in place the protocol got much simpler: a snapshot is just a structured copy of the resource's soul

- transferred pipelines arrive precompiled and are never resolved again, so definitions (`tgpu.fn`, entry functions, `tgpu.comptime`) stay runtime-local
- pipelines carrying attachments or a bound pass/encoder throw on transfer (those hold views and encoders local to their runtime)
- data schemas travel as descriptions of themselves - every schema is callable and host serializers claim functions before asking whether a value is transferable, so there's no way to hand them the schema directly
- transfer cache keeps identity: sending the same resource twice yields the same object back


reczkok:
Replaced by #2797 (i'm dumb it's not this one 🧌)

apps/typegpu-docs/astro.config.mjs
```diff
@@ -243,6 +243,11 @@ export default defineConfig({
               label: 'React Native',
               slug: 'integration/react-native',
             },
+            {
+              label: 'React Native Worklets',
+              slug: 'integration/react-native/worklets',
+              badge: { text: 'experimental', variant: 'caution' },
```
iwoplaza: The 'caution' badge color doesn't look that good. Imo lets use the neutral color until we can style this part of the docs better.
```suggestion
              badge: { text: 'experimental' },
```

aleksanderkatan (commented):
Some of my comments from #2732 still apply

iwoplaza (approved):

PR #2800, opened 2026-08-06
feat: Route function calls through the shader generator

iwoplaza:


packages/typegpu-gl/src/glslGenerator.ts
```diff
@@ -105,6 +116,71 @@ export class GlslGenerator extends WgslGenerator {
...
+      throw new Error(`Cannot bitcast from ${String(sourceSchema)} to ${String(targetSchema)}`);
+    }
+
+    if (name === 'select') {
+      const [falsy, truthy, cond] = args;
+      if (!falsy || !truthy || !cond) {
+        throw new Error(`Invalid number of arguments for 'select'`);
+      }
+
+      if (falsy.dataType !== UnknownData && falsy.dataType.type.startsWith('vec')) {
```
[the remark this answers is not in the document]
cieplypolar: ???

Image

cieplypolar (approved):
Looks good. However, if someone ignores the `select` types, it will fail in the browser instead of during code generation.

iwoplaza:
> However, if someone ignores the `select` types, it will fail in the browser instead of during code generation.

@cieplypolar How so? 👀

cieplypolar:
> > However, if someone ignores the `select` types, it will fail in the browser instead of during code generation.
> 
> @cieplypolar How so? 👀

```ts
it('...', () => {
  function foo() {
    'use gpu';
    const cond = d.vec3b(false, true, false);
    // @ts-ignore
    return std.select(0, 1, cond);
  }

  expect(tgpu.resolve([foo], glOptions())).toMatchInlineSnapshot(`
    "int foo() {
      bvec3 cond = bvec3(false, true, false);
      return (cond ? 1i : 0i);
    }"
  `);
});
```

reczkok (approved):

packages/typegpu-gl/src/glslGenerator.ts
```diff
@@ -68,6 +70,16 @@ ${Object.entries(struct.propTypes)
...
 }
 
+function correspondingBooleanVectorSchema(dataType: d.BaseData) {
+  if (dataType.type.includes('2')) {
+    return d.vec2b;
+  }
+  if (dataType.type.includes('3')) {
+    return d.vec3b;
+  }
+  return d.vec4b;
```
aleksanderkatan: nit (just in case we accidentally use it in a function that accepts numbers)
```suggestion
  if (dataType.type.includes('4')) {
    return vec4b;
  }
  throw new Error(
    `Internal error: schema of type '${dataType.type}' does not have a corresponding boolean vector.`,
  );
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1070,6 +1070,18 @@ ${this.ctx.pre}}`;
     return snip(base, schema, /* origin */ 'constant', false);
   }
 
+  public call(name: string, templateParams: readonly Snippet[], args: readonly Snippet[]): string {
```
aleksanderkatan, line 1092: Do we plan on routing all calls through here? Is there an issue for tracking that?

packages/typegpu/src/std/boolean.ts
```diff
@@ -443,7 +443,7 @@ export const select = dualImpl({
   },
   normalImpl: cpuSelect,
   codegenImpl: (ctx, [f, t, cond]) => {
-    const result = stitch`select(${f}, ${t}, ${cond})`;
+    const result = ctx.gen.call('select', [], [f, t, cond]);
```
aleksanderkatan, line 446: I wonder if this is the best approach. 

How about we aim for an api like this?:
```ts
// numeric.ts
export const dot = 1//...;

// numericGl.ts
polyfillGl(dot, (fn) => ({
  signature: (...args) => { 
    if (isNumeric(args[0])) { throw new Error(...) } 
    return fn.signature(args);
  },
  codegenImpl: (_ctx, [lhs, rhs]) => stitch`dot(${lhs}, ${rhs})`, // or a ctx.gen.call
}))
```
Or we could also try to extract the logic straight from glslgenerator to a list/map of polyfills.

packages/typegpu-gl/src/glslGenerator.ts
```diff
@@ -105,6 +117,79 @@ export class GlslGenerator extends WgslGenerator {
...
+      // Generating a ternary expression, which is supported in GLSL
+      return `(${this.ctx.resolveSnippet(cond).value} ? ${this.ctx.resolveSnippet(truthy).value} : ${this.ctx.resolveSnippet(falsy).value})`;
+    }
+
+    if (name === 'saturate') {
+      const [arg] = args;
+      if (!arg) {
+        throw new Error(`Invalid number of arguments for 'saturate'`);
+      }
+      return super.call('clamp', [], [arg, snip(0, d.f32, 'constant'), snip(1, d.f32, 'constant')]);
```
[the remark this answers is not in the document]
cieplypolar: frog is right about `0f` and `1f` constants, but sadly its fix won't work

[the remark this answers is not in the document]
iwoplaza: The fix for proper GLSL constants is in a PR that's stacked on top of this one

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: Just made one https://github.com/software-mansion/TypeGPU/issues/2829

PR #2809, opened 2026-08-08
fix: Simpler ShaderGenerator lifetime

iwoplaza:
The main idea behind this change is to create a generator per resolution pass (tgpu.resolve call, root.unwrap(pipeline), ...). This makes it very easy to not have to think about nested resolves (which we want to support because of @typegpu/three)

### DIFF
--- a/packages/typegpu/src/core/root/init.ts
+++ b/packages/typegpu/src/core/root/init.ts
@@ -14,7 +14,7 @@ import type {
 } from '../../tgpuBindGroupLayout.ts';
 import { isBindGroup, isBindGroupLayout, TgpuBindGroupImpl } from '../../tgpuBindGroupLayout.ts';
 import type { LogGeneratorOptions } from '../../tgsl/consoleLog/types.ts';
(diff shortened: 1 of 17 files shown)


packages/typegpu/src/core/root/init.ts
```diff
@@ -602,7 +602,7 @@ export type InitFromDeviceOptions = {
    * A custom shader code generator, used when resolving TypeGPU functions.
    * If not provided, the default WGSL generator will be used.
    */
-  unstable_shaderGenerator?: ShaderGenerator | undefined;
+  unstable_shaderGenerator?: ShaderGeneratorClass | undefined;
```
aleksanderkatan: Can we include in this type that this should be a wgsl generator?

aleksanderkatan (approved):

aleksanderkatan (approved):

cieplypolar (approved):

reczkok (approved):

PR #2812, opened 2026-08-08
feat: Immediates

### LINKED ISSUE #2688: [RFC] feat: Immediates
[Intro to immediates](https://webgpufundamentals.org/webgpu/lessons/webgpu-immediates.html)

**API sketch:**
```ts
const Props = d.struct({
  color: d.vec3f,
});

// DEFINITION
const props = tgpu.immediateVar(Props);

// USAGE
const renderPipeline = root.createRenderPipeline({
  vertex: common.fullScreenTriangle,
  fragment: () => {
    'use gpu';
    return d.vec4f(props.$.color);
  },
});

// SUPPLYING
renderPipeline
  .withImmediates(props, { color: d.vec3f(1, 0, 0) })
  .draw(3);
```

Additional TypeGPU niceities:
- Our immediates can be used to provide an accessor, meaning a library can indirectly use an immediate through the user's decision.

reczkok:


packages/typegpu/src/core/pipeline/computePipeline.ts
```diff
@@ -542,11 +587,14 @@ class ComputePipelineCore implements SelfResolvable {
 
     warnIfOverflow(usedBindGroupLayouts, device.limits);
 
+    const immediateSize =
+      usedImmediate !== undefined ? validateImmediateUsage(usedImmediate, this.root) : undefined;
+
     const module = device.createShaderModule({
       label: `${getName(this) ?? '<unnamed>'} - Shader`,
       code,
     });
```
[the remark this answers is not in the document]
reczkok: not needed

packages/typegpu/src/core/immediate/immediateVar.ts
```diff
@@ -0,0 +1,291 @@
...
+  if (!root.enabledWgslLanguageFeatures.has(wgslLanguageExtensions.immediateAddressSpace)) {
+    throw new Error(
+      `Immediate variable '${
+        getName(immediate) ?? '<unnamed>'
+      }' cannot be used, because the '${wgslLanguageExtensions.immediateAddressSpace}' WGSL language extension is not supported in this environment. Check support with root.enabledWgslLanguageFeatures and fall back to a uniform buffer (e.g. via tgpu.accessor) when unavailable.`,
+    );
+  }
+
+  return sizeOf(immediate.dataType);
+}
```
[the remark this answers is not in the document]
reczkok: it's fine, clear webgpu error

packages/typegpu/src/core/pipeline/renderPipeline.ts
```diff
@@ -943,6 +988,9 @@ class RenderPipelineCore implements SelfResolvable {
 
     warnIfOverflow(usedBindGroupLayouts, device.limits);
 
+    const immediateSize =
+      usedImmediate !== undefined ? validateImmediateUsage(usedImmediate, root) : undefined;
+
```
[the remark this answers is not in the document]
reczkok: It will be an error either way, declaration is not required

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -677,6 +677,8 @@ Both styles operate on the same pass state and follow the WebGPU ordering rules:
 `pipeline.with(pass).draw(...)` sets the pass's current pipeline, so a subsequent `pass.draw(...)` executes that pipeline, not one set earlier via `pass.setPipeline(...)`.
 :::
 
+Note that `pipeline.with(pass)` returns a new pipeline wrapper on every call, so hoist it out of draw loops (`const bound = pipeline.with(pass)`) to let deduplication kick in. For values that change between draws, prefer pass-level state (`pass.setBindGroup`, `pass.setImmediates`) over the allocating `with*` methods.
+
```
aleksanderkatan, line 681: I need to create a new pass in each frame, right?
Did you mean something like "hoist it to the top of draw loops"? Also, I don't understand what deduplication means in this context.

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -726,3 +728,86 @@ const pipeline = root.createRenderPipeline({
...
+Immediates are small pieces of data set directly on a pass, on a per-draw (or per-dispatch) basis, without the overhead of a buffer. They are WebGPU's take on push constants.
+An immediate variable is defined with `tgpu['~unstable'].immediateVar` and used in shaders through its `.$` property, like other TypeGPU resources.
+
+```ts
+const tint = tgpu['~unstable'].immediateVar(d.vec4f, d.vec4f(1)); // schema + optional default
+
+const mainFragment = tgpu.fragmentFn({ out: d.vec4f })(() => tint.$);
+```
+
+The value is provided at execution time, either per-pipeline (`pipeline.with(immediate, value)`) or per-pass (`pass.setImmediates`).
```
aleksanderkatan, line 747: I think that "per-pass" usually means that it does not change during a pass, this is not the case, correct?

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -726,3 +728,86 @@ const pipeline = root.createRenderPipeline({
...
+
+// per frame:
+if (isImmediateVar(viewProjSource)) {
+  pass.setImmediates(viewProjSource, viewProjMatrix);
+} else {
+  viewProjSource.write(viewProjMatrix);
+}
+```
+
+A uniform buffer holds one value per submission, so values that change between draws need one uniform (and bind group) per value on the fallback path.
```
aleksanderkatan, line 813: Could you elaborate? I feel like I only know what this sentence means because I already knew that before. Something like this?

```suggestion
A uniform buffer's value cannot be changed by a command encoder, so immediate values that change between draws included in a single pass need one uniform (and bind group) per draw.
```

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
```diff
@@ -726,3 +728,86 @@ const pipeline = root.createRenderPipeline({
...
+pipeline.with(pass).draw(3);
+pass.end();
+encoder.submit();
+```
+
+Immediates come with a few WebGPU-imposed restrictions:
+
+- Only one immediate variable can be used in a single shader. To pass multiple values, use a struct schema.
+- The schema cannot contain arrays, atomics or booleans (use `d.u32` or `d.i32` instead).
+- The size of the schema must be a multiple of 4 bytes (pad with `d.size` if needed, e.g. `d.size(4, d.f16)`) and is limited by the device's `maxImmediateSize` limit (64 bytes by default where supported). A higher limit can be requested at initialization via `tgpu.init({ device: { requiredLimits: { maxImmediateSize: 128 } } })`.
```
aleksanderkatan, line 779: Is this actually true? 
I don't see this limitation anywhere, and I even noticed this:

Image

Either way, this is not asserted properly, as the following test fails:

```ts
    it('rejects schemas of invalid size', () => {
      expect(() => tgpu['~unstable'].immediateVar(d.size(4, d.f16))).not.toThrow();
      expect(() =>
        tgpu['~unstable'].immediateVar(d.struct({ a: d.f16 })),
      ).toThrowErrorMatchingInlineSnapshot();
      expect(() => tgpu['~unstable'].immediateVar(d.f16)).toThrowErrorMatchingInlineSnapshot();
    });
```

packages/typegpu/src/core/immediate/immediateVar.ts
```diff
@@ -0,0 +1,291 @@
...
+
+export function isImmediateVar(value: unknown): value is TgpuImmediateVar {
+  return value instanceof TgpuImmediateVarImpl;
+}
+
+// --------------
+// Implementation
+// --------------
+
+function assertValidImmediateSchema(schema: BaseData, rootSchema: BaseData = schema): void {
```
aleksanderkatan, line 73: This list would work if we assumed that the schema is already a storable type, but it is not. The following is allowed by the types, and the test fails:
```ts
    it('rejects other non-host-shareable values', () => {
      expect(() =>
        tgpu['~unstable'].immediateVar(d.texture2d()),
      ).toThrowErrorMatchingInlineSnapshot();
      expect(() =>
        tgpu['~unstable'].immediateVar(d.sampler()),
      ).toThrowErrorMatchingInlineSnapshot();
      expect(() =>
        tgpu['~unstable'].immediateVar(d.ptrFn(d.u32)),
      ).toThrowErrorMatchingInlineSnapshot();
    });
```

There already is a `StorableData` type we could use. Or we could add a new type:
`export type ImmediateData = ScalarData | VecData | MatData | WgslStruct;`

Also nit: I think it is generally cleaner to case out what is allowed, instead of what is not allowed (small chance of disallowing something that should be allowed > big chance of allowing something that should be disallowed)

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+  }
+
+  describe('resolution', () => {
+    it('resolves to a var<immediate> declaration without bindings', () => {
+      const level = tgpu['~unstable'].immediateVar(d.f32);
+      const fn1 = tgpu.fn([], d.f32)(() => level.$);
+
+      const resolved = tgpu.resolve([fn1]);
+      expect(resolved).toContain('var<immediate> level: f32;');
+      expect(resolved).not.toContain('@group');
```
aleksanderkatan, line 32: Are you intentionally avoiding inline snapshots in these tests? I think that an additional `expect(resolved).toMatchInlineSnapthot()` here would make the test result more readable at a glance, and it would further reduce the chance of us accidentally making changes to the generated code

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+    it('throws when two different immediate variables are used in one shader', () => {
+      const first = tgpu['~unstable'].immediateVar(d.f32);
+      const second = tgpu['~unstable'].immediateVar(d.f32);
+      const fn1 = tgpu.fn([], d.f32)(() => first.$ + second.$);
+
+      expect(() => tgpu.resolve([fn1])).toThrow(
+        /Cannot use both immediate variables 'first' and 'second' in a single shader/,
+      );
+    });
+
```
aleksanderkatan, line 99: Free test (probably redundant, but free)

```ts
    it('allows multiple immediates as long as they are pruned in comptime', () => {
      const slot = tgpu.slot<boolean>();
      const imm1 = tgpu['~unstable'].immediateVar(d.f32);
      const imm2 = tgpu['~unstable'].immediateVar(d.f32);
      const fn = tgpu.fn(
        [],
        d.f32,
      )(() => {
        'use gpu';
        return slot.$ ? imm1.$ : imm2.$;
      });

      expect(tgpu.resolve([fn.with(slot, true)])).toMatchInlineSnapshot(`
        "var<immediate> imm1: f32;

        fn fn_1() -> f32 {
          return imm1;
        }"
      `);
      expect(tgpu.resolve([fn.with(slot, false)])).toMatchInlineSnapshot(`
        "var<immediate> imm2: f32;

        fn fn_1() -> f32 {
          return imm2;
        }"
      `);
    });

```

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+      pass.finish();
+
+      expect(writes).toEqual([[0.25]]);
+    });
+  });
+
+  describe('accessors', () => {
+    it('resolves an accessor fulfilled by an immediate variable', () => {
+      const level = tgpu.accessor(d.f32);
+      const levelImmediate = tgpu['~unstable'].immediateVar(d.f32).$name('levelImmediate');
```
aleksanderkatan, line 666: Can we just add `immediateVar` to the list in 'packages/unplugin-typegpu/src/core/common.ts:320'?

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+      expect(() => pipeline.dispatchWorkgroups(1)).toThrow(
+        /'immediate_address_space' WGSL language extension is not supported/,
+      );
+    });
+  });
+
+  describe('render pipelines', () => {
+    it('rejects distinct immediates across stages of one shader module', ({ root }) => {
+      const vertexValue = tgpu['~unstable'].immediateVar(d.f32, 0).$name('vertexValue');
+      const fragmentValue = tgpu['~unstable'].immediateVar(d.f32, 0).$name('fragmentValue');
```
aleksanderkatan, line 270: nit: irrelevant names
```suggestion
      const vertexValue = tgpu['~unstable'].immediateVar(d.f32, 0);
      const fragmentValue = tgpu['~unstable'].immediateVar(d.f32, 0);
```

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+      const level = tgpu['~unstable'].immediateVar(d.f32, 0.5);
+      const entry = tgpu.computeFn({ workgroupSize: [1] })(() => {
+        level.$;
+      });
+
+      root.createComputePipeline({ compute: entry }).with(level, 0.25).dispatchWorkgroups(1);
+
+      expect(writes).toEqual([[0.25]]);
+    });
+
```
aleksanderkatan, line 241: This is what `immediatesMap` is for, right? I don't see this tested.

```ts
    it('allows providing multiple immediates', ({ root, computePassEncoder }) => {
      const writes = trackWrites(computePassEncoder.mock.setImmediates);

      const imm1 = tgpu['~unstable'].immediateVar(d.f32);
      const imm2 = tgpu['~unstable'].immediateVar(d.f32);
      const fn = tgpu.computeFn({ workgroupSize: [1] })(() => {
        const a = imm1.$;
      });
      const pipeline = root.createComputePipeline({ compute: fn });

      pipeline.with(imm1, 1).with(imm2, 2).dispatchWorkgroups(1);

      expect(writes).toEqual([[1]]);
    });
```

packages/typegpu/tests/tgsl/letDeclaration.test.ts
```diff
@@ -55,4 +55,21 @@ describe('let declarations', () => {
       -----]
     `);
   });
+
+  it('throws when aliasing a value from the handle address space', () => {
```
aleksanderkatan, line 59: Why is this test in the `letDeclaration.test.ts`?

packages/typegpu/src/core/immediate/immediateVar.ts
```diff
@@ -0,0 +1,291 @@
...
+
+/**
+ * Tracks the last immediate snapshot written to a pass, so that repeated
+ * writes of the same captured value can be skipped. Only usable where the
+ * pass's immediate state cannot be reset outside of TypeGPU's control.
+ */
+export interface ImmediatesCache {
+  lastWrittenSnapshot: ImmediateSnapshot | undefined;
+  lastWrittenGeneration: number;
+}
```
aleksanderkatan: Is this purely for optimization cases like this?
```ts
pass.setImmediates(imm, m1);
pass.draw();
pass.setImmeditates(imm, m1);
pass.draw();
```

packages/typegpu/src/core/pipeline/drawState.ts
```diff
@@ -46,8 +55,11 @@ export class RenderDrawState {
   rawAccessed = false;
 }
 
-export class ComputeDrawState {
+export class ComputeDrawState implements ImmediatesCache {
   readonly bindGroups = new Map<TgpuBindGroupLayout, TgpuBindGroup | GPUBindGroup>();
+  readonly immediates: ImmediateSnapshotMap = new Map();
+  lastWrittenSnapshot: ImmediateSnapshot | undefined;
+  lastWrittenGeneration = 0;
```
aleksanderkatan, line 62: These three fields relate to immediates, and their state is correlated (when updating one, we need to remember to update another).
Also, they all appear in both draw states... why is compute state named draw anyway? 
I think that this immediate managing functionality should be delegated to another class for clarity

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+        .with(tint, 0.5)
+        .with(renderPassEncoder);
+
+      bound.draw(3);
+      bound.draw(3);
+
+      expect(renderPassEncoder.mock.setImmediates).toBeCalledTimes(2);
+    });
+
+    it('distinguishes immediate variables with identical values', ({ root, renderPassEncoder }) => {
```
aleksanderkatan, line 582: Is there a reason to distinguish them or is this test just pinning the current behavior?

packages/typegpu/tests/immediates.test.ts
```diff
@@ -0,0 +1,701 @@
...
+      bytes[0] = 0.75;
+      pass.setImmediates(tint, bytes);
+      pipeline.with(pass).draw(3);
+      pass.end();
+      encoder.submit();
+
+      expect(writes).toEqual([[0.5], [0.75]]);
+    });
+
+    it('applies pipeline-level and pass-level immediates in call order', ({
```
aleksanderkatan, line 399: This test's result looks really unintuitive to me.
I'd expect that `pass.setImmediates` > `pipeline.with` > default. In this case, the line 409 does absolutely nothing. Is this documented somewhere? What is an intuitive explanation for that?

aleksanderkatan (changes requested):
Looks good! 🏊🔉
I leave some requests, mostly minor ones.

I just wonder, caching immediates intuitively looks to me like it could be vastly simplified, couldn't it be something like this? 
```ts
const imm = pipeline.usedImmediate;
const value = pass.valueOf(imm) ?? pipeline.valueOf(imm) ?? imm.default
// + some mechanism for cheking if this value was already written,
// if we always copy immediates to new ArrayBuffers,
// then a reference check should suffice
```

Some parts look like they could have been done in separate PRs, this review took me over 5 hours and I still didn't properly parse everything.

reczkok:
> Some parts look like they could have been done in separate PRs, this review took me over 5 hours and I still didn't properly parse everything.

If you share what parts you mean I'm happy to split

packages/typegpu/tests/immediates.test.ts
reczkok: Buuuut the inline snapshots tend to increase the line count drastically across a PR and I got a lot of LC based complaints in the past

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
reczkok: We do not enforce that for anything so I did not start here. Try creating a buffer of type `f16`, TypeGPU will allow it and it will get rejected by WebGPU. As for the rule it is true, enforced at pipeline layout stage 
image

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
reczkok: This will change after the precedence update I'm planning but deduplication refers to `.setXYZ` commands

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
reczkok: I don't think I understand the question

packages/typegpu/tests/immediates.test.ts
reczkok: I agree but it would require a bigger change for consistency. I have proposed it and would like it to land before this PR, then we can clean this up

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
aleksanderkatan: I just think that the "per-" wording may be misleading, I interpret "X is set per-Y" as "X doesn't change during the scope of Y". For example, color attachment is proivided per pass, but e.g. pipeline and immediates are not provided per-pass, because these can change. I'm not sure if this makes sense

apps/typegpu-docs/src/content/docs/apis/pipelines.mdx
aleksanderkatan: Fair

packages/typegpu/tests/immediates.test.ts
aleksanderkatan: I guess that's true, but reducing LC in this way does not reduce the actual complexity & scope of a PR (LC is just a vague measure of these two)

You can always mention how much tests contribute to LC in the description, we already do this from time to time, and I think that this makes people more likely to review big PRs

aleksanderkatan:
> If you share what parts you mean I'm happy to split

I thought about changes like allowing immediates in accessors, implementing glsl side, workletization, the caching mechanism, or maybe even schema validation.
At a glance, they look like small-medium sized incremental changes that could just iterate on a basic implementation of immediates with very little placeholder logic, not sure how many of these are actually like that though.

reczkok:
Moved to a stacked pair: #3031 (language feature) and #3032 (pipelines and passes).

PR #2813, opened 2026-08-08
impr: Use Immediates in examples

reczkok:


reczkok:
Moved to #3034, stacked on #3032.

PR #2815, opened 2026-08-08
fix(@typegpu/three): Allow TgpuFnNode to transitively nest

iwoplaza:
This can happen when a TSL node calls a TypeGPU function which references a TSL node that calls another TypeGPU function.

### DIFF
--- a/packages/typegpu-three/src/typegpu-node.ts
+++ b/packages/typegpu-three/src/typegpu-node.ts
@@ -3,8 +3,7 @@ import type NodeVarying from 'three/src/nodes/core/NodeVarying.js';
 import type VaryingNode from 'three/src/nodes/core/VaryingNode.js';
 import * as THREE from 'three/webgpu';
 import * as TSL from 'three/tsl';
-import { tgpu, type Namespace, type TgpuVar } from 'typegpu';
-import * as d from 'typegpu/data';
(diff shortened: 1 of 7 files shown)


reczkok (approved):

packages/typegpu-three/src/typegpu-node.ts
```diff
@@ -191,16 +211,14 @@ class TgpuFnNode<T> extends THREE.Node {
       stageData,
       dependencies: [],
     };
-    currentlyGeneratingFnNodeCtx = ctx;
-    try {
+
+    withGeneratingFnNodeCtx(ctx, () =>
       tgpu.resolve({
         names: stageData.namespace,
         template: '___ID___ fnName',
```
cieplypolar:
```suggestion
        template: 'impl',
```

packages/typegpu-three/src/typegpu-node.ts
```diff
@@ -191,16 +211,14 @@ class TgpuFnNode<T> extends THREE.Node {
...
         externals: { fnName: this.#impl },
```
cieplypolar:
```suggestion
        externals: { impl: this.#impl },
```

cieplypolar (approved):
Looks good. Left one nit.

PR #2824, opened 2026-08-10
impr(@typegpu/three): More robust `varyings` handling

cieplypolar:
`builtNode.includes('varyings.')` check was not ideal.


iwoplaza (approved):
Yes yes! :shipit:

PR #2830, opened 2026-08-11
feat(@typegpu/noise): new default PRNG - xoroshiro

### LINKED ISSUE #2484: feat(@typegpu/noise): Make XOROSHIRO the default PRNG
**Why?**
1. It's faster
2. It's valid in the whole seed domain (unlike BPETER which breaks for large seeds)
3. It's more numerically stable, because it doesn't use trig functions
4. It has more matching behavior between JS and WGSL (why not fully matching?)

cieplypolar:
Closes #2484.


reczkok (approved):
Yoohoo

iwoplaza (approved):

PR #2835, opened 2026-08-11
chore(@typegpu/three): Remove done TODO

### LINKED ISSUE #2826: fix: Investigate typegpu/three minification
It currently does not work, and is forcibly disabled.

cieplypolar:
Closes #2826.

### DIFF
--- a/packages/typegpu-three/src/typegpu-node.ts
+++ b/packages/typegpu-three/src/typegpu-node.ts
@@ -214,7 +214,6 @@ class TgpuFnNode<T> extends THREE.Node {
           names: stageData.namespace,
           template: 'impl',
           externals: { impl: this.#impl },
-          unstable_minify: false, // TODO(#2826): investigate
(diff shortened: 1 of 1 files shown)


iwoplaza:
How is the minification getting enabled automatically here? From what I understand there are no pipelines passed into the resolve, and therefore no root from where the config could be extracted. To test it, did you just pass `unstable_minify: true` to all of the resolves in typegpu-node.ts?

cieplypolar:
> How is the minification getting enabled automatically here? From what I understand there are no pipelines passed into the resolve, and therefore no root from where the config could be extracted. To test it, did you just pass `unstable_minify: true` to all of the resolves in typegpu-node.ts?

exactly, I forced minification on these resolves

cieplypolar:
@iwoplaza you can find more details in the comments #2826

iwoplaza (approved):

reczkok (approved):

PR #2848, opened 2026-08-13
impr: Clean up `vectorOps`

### LINKED ISSUE #1336: impr: Clean up VectorOps

aleksanderkatan:


packages/typegpu/src/std/operators.ts
```diff
@@ -205,21 +206,8 @@ function cpuDiv<T extends NumVec>(lhs: T, rhs: T): T; // component-wise division
...
-  if (isVecInstance(lhs) && typeof rhs === 'number') {
-    const schema = vecTypeToConstructor[lhs.kind];
-    return VectorOps.div[lhs.kind](lhs, schema(rhs));
-  }
-  if (isVecInstance(lhs) && isVecInstance(rhs)) {
-    return VectorOps.div[lhs.kind](lhs, rhs);
-  }
-  throw new Error('Div called with invalid arguments.');
+  verifyKind([lhs, rhs], numericKind);
+  return generalizeFn((a, b) => a / b, upCast([lhs, rhs]));
```
aleksanderkatan, line 209: Previously integer vectors used a separate function for integer division. This works because vector constructors cast their elements.

aleksanderkatan:
@pullfrog review

packages/typegpu/tests/std/boolean/isCloseTo.test.ts
```diff
@@ -43,16 +43,22 @@ describe('isCloseTo', () => {
...
+    expect(isCloseTo(d.vec2h(0, 0), d.vec2h(0, 0.0011), 0.001)).toBe(false);
+
+    expect(isCloseTo(d.vec2h(0, 0), d.vec2h(0, 9), 10)).toBe(true);
+    expect(isCloseTo(d.vec2h(0, 0), d.vec2h(0, 11), 10)).toBe(false);
+  });
 
-    expect(isCloseTo(vec2h(0, 0), vec2h(0, 9), 10)).toBe(true);
-    expect(isCloseTo(vec2h(0, 0), vec2h(0, 11), 10)).toBe(false);
+  it('works for mixed types', () => {
+    expect(isCloseTo(d.vec2f(0, 0) as d.v2f | number, 0)).toBe(true);
```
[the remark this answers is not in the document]
aleksanderkatan: Ok, back to uniform types. We'll throw in the next PR in the stack if the types mismatch.

packages/typegpu/src/std/boolean.ts
```diff
@@ -326,13 +328,11 @@ export const isCloseTo = dualImpl({
...
-    if (typeof lhs === 'number' && typeof rhs === 'number') {
-      return Math.abs(lhs - rhs) < precision;
-    }
-    if (isVecInstance(lhs) && isVecInstance(rhs)) {
-      return VectorOps.isCloseToZero[lhs.kind](sub(lhs, rhs), precision);
-    }
-    return false;
+    const componentResult = generalizeBoolFn(
+      (lhs, rhs) => Math.abs(lhs - rhs) < precision,
+      [lhs, rhs],
```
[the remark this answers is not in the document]
aleksanderkatan: What upCast?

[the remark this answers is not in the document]
aleksanderkatan: The next pr in stack will fix this

packages/typegpu/src/data/generalizeFn.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+function applyArgs(
+  fn: (...args: never[]) => number | boolean,
+  args: Algebraic[],
+  mode: Mode,
+): Algebraic {
+  // I'm sorry, TypeScript, I swear I won't lie to you no more ;-;
+  const kinds = args.map(kindOf);
+  if (kinds.every((type) => ['boolean', 'number'].includes(type))) {
```
iwoplaza: Tiny nit, doesn't create an array
```suggestion
  if (kinds.every((type) => type === 'boolean' || type === 'number')) {
```

apps/typegpu-docs/tests/individual-example-tests/3d-fish.test.ts
```diff
@@ -462,10 +462,10 @@ describe('3d fish example', () => {
         let textureColorWithAlpha = textureSample(modelTexture, sampler_1, _arg_0.textureUV);
         let textureColor = textureColorWithAlpha.rgb;
         let ambient = ((0.5f * textureColor) * vec3f(0.800000011920929, 0.800000011920929, 1));
-        let cosTheta = dot(_arg_0.worldNormal, vec3f(-0.2357022613286972, 0.9428090453147888, -0.2357022613286972));
+        let cosTheta = dot(_arg_0.worldNormal, vec3f(-0.2357022762298584, 0.9428091049194336, -0.2357022762298584));
```
iwoplaza: Do we know why this changed? 👀

iwoplaza (approved):
EXTRA SMOL 🐹

apps/typegpu-docs/tests/individual-example-tests/3d-fish.test.ts
aleksanderkatan: It's the difference between `div(v, len)` and `generalizeFn((e) => e / len, [v])`.
I thought that this actually introduced the correct rounding, but it is an error, fixed.

cieplypolar (approved):

PR #2849, opened 2026-08-14
feat: std.intdiv

iwoplaza:


packages/typegpu/src/std/numeric.ts
```diff
@@ -1309,3 +1309,30 @@ export const trunc = dualImpl<typeof cpuTrunc>({
...
+  if (typeof lhs === 'number' && typeof rhs === 'number') {
+    return Math.trunc(Math.trunc(lhs) / Math.trunc(rhs));
+  }
+  throw new Error('std.intdiv called with invalid arguments.');
+}
+
+/**
+ * Performs integer division on the passed in scalars.
+ * Equivalent to `Math.trunc(lhs / rhs)`.
+ */
```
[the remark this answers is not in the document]
aleksanderkatan: The result is different for e.g. 3/1.5

reczkok (approved):
Yes please

packages/typegpu/src/data/numberOps.ts
```diff
@@ -9,12 +9,14 @@ export const smoothstepScalar = (edge0: number, edge1: number, x: number): numbe
 export const clamp = (value: number, low: number, high: number) =>
   Math.min(Math.max(low, value), high);
 
-export const divInteger = (lhs: number, rhs: number) => {
+export function divInteger(lhs: number, rhs: number) {
   if (rhs === 0) {
+    // WGSL defines integer division as returning
+    // `lhs` when `rhs` is 0
```
aleksanderkatan: This is not entirely true.
> If e2 is zero:
> It is a [shader-creation error](https://www.w3.org/TR/WGSL/#shader-creation-error) if e2 is a [const-expression](https://www.w3.org/TR/WGSL/#const-expressions).
> It is a [pipeline-creation error](https://www.w3.org/TR/WGSL/#pipeline-creation-error) if e2 is an [override-expression](https://www.w3.org/TR/WGSL/#override-expressions).
> Otherwise, e1.

So, `1 / 0` in WGSL throws, and in TGSL `intDiv(1, 0)` returns 1.
Also, `std.div(d.vec2u(), d.vec2u())` will throw due to the zero division after #2848, should I make this function throw in that PR?

packages/typegpu/src/std/numeric.ts
```diff
@@ -1309,3 +1309,30 @@ export const trunc = dualImpl<typeof cpuTrunc>({
   codegenImpl: (_ctx, [value]) => stitch`trunc(${value})`,
   sideEffects: false,
 });
+
+function cpuIntdiv(lhs: number, rhs: number): number {
```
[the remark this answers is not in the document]
aleksanderkatan: nah, it's removed in the vectorops ticket

aleksanderkatan (approved):
yeppers

packages/typegpu/src/std/numeric.ts
```diff
@@ -1309,3 +1309,29 @@ export const trunc = dualImpl<typeof cpuTrunc>({
...
+  if (typeof lhs !== 'number' || typeof rhs !== 'number') {
+    throw new Error('std.intdiv called with invalid arguments.');
+  }
```
cieplypolar, line 1316: why did you omit case when `Math.trunc(rhs) === 0`?

packages/typegpu/src/std/numeric.ts
```diff
@@ -1309,3 +1309,29 @@ export const trunc = dualImpl<typeof cpuTrunc>({
...
+  return Math.trunc(Math.trunc(lhs) / Math.trunc(rhs));
+}
+
+/**
+ * Performs integer division on the passed in scalars.
+ * Equivalent to `Math.trunc(lhs / rhs)`. Coerces both
```
cieplypolar:
```suggestion
 * Equivalent to `Math.trunc(Math.trunc(lhs) / Math.trunc(rhs))`. Coerces both
```
For example `Math.trunc(1 / 0.5) === 2`, but `Math.trunc(Math.trunc(1) / Math.trunc(0.5))` returns NaN

packages/typegpu/src/std/numeric.ts
iwoplaza: Because we agreed to let these cases throw (which they will as FiniteMathAssumption errors)

PR #2858, opened 2026-08-17
feat(@typegpu/gl): Limited texture support

iwoplaza:


aleksanderkatan (approved):
I didn't review but I believe you 🙆‍♂️

PR #2860, opened 2026-08-17
fix(@typegpu/gl): Emit ternary expression from std.select when condition is scalar

iwoplaza:


cieplypolar (approved):

aleksanderkatan (changes requested):
Doesn't this actually change behavior?
WGSL:
```ts
it('...', () => {
  const counter = tgpu.privateVar(d.u32);
  const getNext = () => {
    'use gpu';
    counter.$++;
    return counter.$;
  };

  const fn = () => {
    'use gpu';
    const cond = true;
    const a = std.select(getNext(), getNext(), cond); // select(1, 2, true)

    return a + getNext(); // 2 + 3
  };

    "var<private> counter: u32;

    fn getNext() -> u32 {
      counter++;
      return counter;
    }

    fn fn_1() -> u32 {
      const cond = true;
      let a = select(getNext(), getNext(), cond);
      return (a + getNext());
    }"
  `);
});
```
function returns 5

GLSL:
```ts
// ...
uint counter;

uint getNext() {
  counter++;
  return counter;
}

uint fn_1() {
  bool cond = true;
  uint a = (cond ? getNext() : getNext());
  return (a + getNext());
}
```
function returns 3

iwoplaza:
Let's generate a ternary operator when all branches and the condition don't have side effects, and the current behavior otherwise.

PR #2895, opened 2026-08-21
fix: Declare side effects in typegpu/package.json

aleksanderkatan:
Before, this wasn't an issue because our package wasn't really tree-shakable. Now, it started pruning code and broke infix operators, since they are assigned via a side effect:
```ts
assignInfixOperator(VecBase, 'mul', Operator.star);
```
This code is still present in the built package, but it may be stripped when building a project with typegpu as a dependency.
Infix operators still work when transpiling 'use gpu' functions since they don't actually ever execute, we just invoke correct operations in `wgslGenerator.ts` based on the AST.

### DIFF
--- a/packages/typegpu/package.json
+++ b/packages/typegpu/package.json
@@ -24,7 +24,7 @@
   },
   "bin": "./bin.mjs",
   "type": "module",
-  "sideEffects": false,
+  "sideEffects": true,
   "exports": {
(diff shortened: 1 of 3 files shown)


aleksanderkatan:
I'll try to select only the files with actual side effects

cieplypolar (approved):

reczkok (approved):

PR #2901, opened 2026-08-21
fix: Counteract sideeffect based code pruning

iwoplaza:
This PR makes it so that infix operator assignment is a side-effect of using any of the vector or matrix schemas, not a side-effect of importing.

The increased bundle size for the vector and matrix schemas is the code that makes infix operators possible, the diff can be inspected here:
https://gist.github.com/iwoplaza/e562b896912c280a254787851b0a8a9b

### DIFF
--- a/packages/typegpu/package.json
+++ b/packages/typegpu/package.json
@@ -24,9 +24,7 @@
   },
   "bin": "./bin.mjs",
   "type": "module",
-  "sideEffects": [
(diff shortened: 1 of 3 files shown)


aleksanderkatan (commented):
This looks fragile. If I was creating my own tree-shaker, and saw `return A(), B` knowing `A()` has no side effects, I'd simplify it to `return B`.
But as long as it works 🤷‍♀️

PR #2902, opened 2026-08-21
impr: Move sort package to tsdown

reczkok:


packages/typegpu-sort/package.json
aleksanderkatan, line 4: How far away are we from releasing this package? Will it be ready after this stack is merged?

aleksanderkatan (approved):
Sure!

packages/typegpu-sort/package.json
reczkok: I guess so

PR #2908, opened 2026-08-24
fix(@typegpu/cli): Production builds

cieplypolar:
Fixes:
- `nextjs`: `output: 'export'` is incompatible with `next start`


packages/typegpu-cli/templates/template-waku-bare/_package.json
```diff
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "waku dev",
     "build": "waku build",
-    "start": "waku start",
+    "start": "waku start --host localhost",
```
iwoplaza: I assume that running the production server usually means you're not running it on your own machine. What motivated this change?

cieplypolar: Without `--host localhost`, Waku prints `0.0.0.0:8080` address to the console which is not considered secure. As a result, WebGPU is disabled.

cieplypolar: As we discussed in person, server shouldn't use `localhost`.

iwoplaza (approved):

PR #2912, opened 2026-08-24
feat: `switch` in 'use gpu' functions

### LINKED ISSUE #2881: feat: Support `switch` in 'use gpu' functions

aleksanderkatan:
Changes:
- parse/obfuscate/tseynit/generate `switch`,
- update some examples to use `switch`.

More than half of this PR is tests (~650 lines).
Maybe I should split this into smaller PRs, ping me if this is an issue.


aleksanderkatan:
@pullfrog review

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1796,6 +1828,87 @@ ${this.ctx.pre}else ${alternate}`,
...
+      const switchType = discriminantExpr.dataType;
+      invariant(switchType !== UnknownData);
+
+      const caseExprs: [test: Snippet, consequent: ResolvedStatement[]][] = cases.map(
+        ([test, consequent]) => {
+          const testExpr =
+            test === null ? switchDefault : this._typedExpression(test, [switchType]);
+          const consequentStmts = consequent.map((s) => {
+            // In WGSL, each case is a different block. This block scope forbids scope leaking.
+            this.ctx.pushBlockScope();
```
cieplypolar: this causes each statement to have separate block so 
```ts
  it('test', () => {
    const fn = () => {
      'use gpu';
      let value = d.u32(1);
      switch (value) {
        case 1:
          let temp = 3;
          temp += 1;
          break;
      }
    };

    expect(tgpu.resolve([fn])).toMatchInlineSnapshot();
  });
```
fails

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1796,6 +1828,87 @@ ${this.ctx.pre}else ${alternate}`,
...
+      const discriminantExpr = this._typedExpression(discriminant, [i32, u32]);
+
+      const switchType = discriminantExpr.dataType;
+      invariant(switchType !== UnknownData);
+
+      const caseExprs: [test: Snippet, consequent: ResolvedStatement[]][] = cases.map(
+        ([test, consequent]) => {
+          const testExpr =
+            test === null ? switchDefault : this._typedExpression(test, [switchType]);
+          const consequentStmts = consequent.map((s) => {
```
cieplypolar: this map has additional problem
```ts
  it('test', () => {
    const fn = () => {
      'use gpu';
      let value = d.u32(1);
      switch (value) {
        case 1:
          return 1;
          return 2;
      }
    };

      "fn fn_1() -> i32 {
        let temp = 1;
        let value = 1u;
        switch value {
          case 1u: {
            return 1;
            return 2;
          }
          case default: {

          }
        }
        return temp;
      }"
    `);
  });
```
it does not prune unreachable code. I think you can use our block machinery to obtain that.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1796,6 +1828,87 @@ ${this.ctx.pre}else ${alternate}`,
       }
     }
 
+    if (statement[0] === NODE.switch) {
```
cieplypolar, line 1838: switch should mark that `break` is now safe for unroll
```ts
  it('test', () => {
    const fn = () => {
      'use gpu';
      let value = d.u32(1);
      for (const i of tgpu.unroll([1, 2])) {
        switch (value) {
          case 1:
            break;
        }
      }
    };

    expect(tgpu.resolve([fn])).toMatchInlineSnapshot();
  });
```

packages/typegpu/tests/switch.test.ts
cieplypolar, line 1: free test that currently fails
```ts
  it('test', () => {
    const temp = 1;
    const fn = () => {
      'use gpu';
      let value = d.u32(1);
      switch (value) {
        case 1:
          let temp = 2;
          break;
      }
      return temp;
    };

    expect(tgpu.resolve([fn])).toMatchInlineSnapshot();
  });
```

packages/typegpu-gl/src/glslGenerator.ts
```diff
@@ -822,6 +822,25 @@ export class GlslGenerator extends WgslGenerator {
     return `${this.ctx.pre}${glslTypeName} ${name}${resolveArraySizeSuffix(this.ctx, dataType)} = ${rhsStr};`;
   }
 
+  override _emitSwitchStatement(
+    discriminantExpr: Snippet,
+    groupedCaseExprs: [tests: Snippet[], consequent: ResolvedStatement[]][],
+  ): string {
+    this.ctx.indent();
+    const cases = groupedCaseExprs.map(([tests, consequent]) => {
```
cieplypolar, line 849: if the last case is empty, glsl fails to compile
```ts
  it('test', () => {
    const fn = () => {
      'use gpu';
      let value = d.u32(1);
      switch (value) {
        case 1:
          let temp = 2;
          break;
        case 2:
      }
    };

    expect(tgpu.resolve([fn], glOptions())).toMatchInlineSnapshot(`
      "void fn_1() {
        uint value = 1u;
        switch (value) {
          case 1u:
            int temp = 2;
            break;
          case 2u:

        }
      }"
    `);
  });
```
Appending `break` to it should fix this.

cieplypolar (changes requested):
Very clean impl, however found some bugs that need to be addressed. Sorry 🙏

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: See #3001, it's a tradeoff

packages/typegpu/src/tgsl/wgslGenerator.ts
aleksanderkatan: I cannot do this with
```ts
      const prevUnrollingChain = this.#unrollingChain;
      this.#unrollingChain = [];
```
because this would allow for `continue` as well. Do you see a simple way to do this? If this requires putting new variables in context, then I don't want it to be a part of this PR.

packages/typegpu/tests/switch.test.ts
aleksanderkatan: Good catch! This is actually transpiler's fault

packages/typegpu-gl/src/glslGenerator.ts
aleksanderkatan: nightmare

cieplypolar (approved):
LGTM

packages/typegpu/src/tgsl/wgslGenerator.ts
cieplypolar: This is harder than I thought. Create issue with this test and leave TODO.

There are 2 options:
- new `wgslGenerator` private variable
- easy traversal of switch's children

packages/tinyest-for-wgsl/tests/parsers.test.ts
```diff
@@ -485,4 +485,59 @@ describe('legacy transpileFn', () => {
       `[Error: Object method elements are not supported in TGSL.]`,
     );
   });
+
+  it(
+    'handles switch cases',
+    dualTest((p) => {
```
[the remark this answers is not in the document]
cieplypolar: Right, these tests shouldn't be placed inside legacy block

packages/unplugin-typegpu/test/obfuscation.test.ts
```diff
@@ -633,4 +633,51 @@ describe('obfuscate', () => {
...
+      "{
+        const b = 1;
+        switch (b) {
+          case 1:
+            return a;
+          case "error":
+          default:
+            return 3;
+          case 2:
+      {
```
iwoplaza, line 674: Is this whitespace mismatch hard to solve?

aleksanderkatan: No, there even is a PR that fixes that: 
https://github.com/software-mansion/TypeGPU/pull/2933/changes

packages/typegpu/tests/switch.test.ts
```diff
@@ -0,0 +1,716 @@
...
+    const helper = () => {
+      'use gpu';
+      return 1;
+    };
+
+    const fn = () => {
+      'use gpu';
+      let value = 1;
+      switch (value) {
+        case helper():
```
iwoplaza, line 709: Snippets with origins `constant` and constant-related should work as well, check if that's true, and adjust behavior if so.

aleksanderkatan: It is allowed, I allowed it in tgsl, here's a small issue with it: #3050

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1834,6 +1867,87 @@ ${this.ctx.pre}else ${alternate}`,
...
+            throw new Error(`All of switch tests must be constant.
+Test '${stringifyNode(testNode)}' is not known at comptime, making the following switch statement invalid. 
+This error may be caused by an implicit conversion.
+${stringifyNode(statement)}`);
+          }
+          return testExpr.value as number | 'default';
+        });
+
+        // Tests should not have duplicates
+        const present = new Set<number | 'default'>();
```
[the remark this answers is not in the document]
aleksanderkatan: Good catch

aleksanderkatan:
I don't check for collisions with non-comptime known values (tgpu.const, possibly more like tgpu.const prop access?). The WGSL error should be readable anyway.

packages/typegpu/tests/internal/tseynit.test.ts
cieplypolar: I don't think this belongs to this PR.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1834,6 +1867,90 @@ ${this.ctx.pre}else ${alternate}`,
...
+This error may be caused by an implicit conversion.
+${stringifyNode(statement)}`);
+          }
+          return testExpr;
+        });
+
+        // Tests should not have duplicates (we only check for comptime known collisions)
+        const present = new Set();
+        tests
+          .filter((test) => isKnownAtComptime(test))
```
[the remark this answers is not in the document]
cieplypolar: froggy may be right

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1834,6 +1867,90 @@ ${this.ctx.pre}else ${alternate}`,
...
+
+      // Validation
+      {
+        // Tests should be constant
+        const tests = caseExprs.map(([testExpr], i) => {
+          if (!isConstant(testExpr)) {
+            const testNode = cases[i]?.[0];
+            invariant(testNode, `Expected node to be not nullish.`);
+            throw new Error(`All of switch tests must be constant.
+Test '${stringifyNode(testNode)}' is not known at constant, making the following switch statement invalid.
```
```suggestion
Test '${stringifyNode(testNode)}' is not constant, making the following switch statement invalid. 
```

packages/typegpu/src/types.ts
```diff
@@ -472,6 +472,14 @@ export function isKnownAtComptime(snippet: Snippet): boolean {
   );
 }
 
+export function isConstant(snippet: Snippet): boolean {
+  return (
+    isKnownAtComptime(snippet) ||
+    snippet.origin === 'constant' ||
```
cieplypolar, line 478: I was wondering when we reach the 'constant' branch in switch statement validation. I only managed to do it with `rawCodeSnippet`. It is not covered by any test, could you please add one (preferably without `rawCodeSnippet` ;) )

packages/typegpu/src/tgsl/wgslGenerator.ts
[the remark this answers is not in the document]
aleksanderkatan: Frog is right, I already said that.

> I don't check for collisions with non-comptime known values (tgpu.const, possibly more like tgpu.const prop access?). The WGSL error should be readable anyway.
```ts
        // Tests should not have duplicates (we only check for comptime known collisions)
```

Is there even a way to get the const's value? I could get it from a tgpu const instance, but in this case I get a snippet. The snippet looks like this:
```ts
  SnippetImpl {
    value: {
      toString: [Function: toString],
      Symbol(typegpu:0.12.5:$internal): true,
      Symbol(typegpu:0.12.5:$ownSnippet): [Getter],
      Symbol(typegpu:0.12.5:$resolve): [Function: [typegpu:0.12.5:$resolve]]
    },
    dataType: [Function: impl] {
      toString: [Function (anonymous)],
      type: 'i32',
      Symbol(typegpu:0.12.5:$gpuCallable): [Object],
      Symbol(typegpu:0.12.5:$internal): {}
    },
    origin: 'constant-immutable-def',
    possibleSideEffects: false
  }
```

I'd be way more inclined to remove this check altogether than to try and make it work with consts.

packages/typegpu/src/types.ts
aleksanderkatan: In a little cursed way, but done

packages/typegpu/src/tgsl/wgslGenerator.ts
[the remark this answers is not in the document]
aleksanderkatan: This check was dropped

cieplypolar (approved):

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: Added issue for it https://github.com/software-mansion/TypeGPU/issues/3068

iwoplaza (approved):

Issue #2318, opened 2026-03-26
feat: Spreading objects in 'use gpu' functions


iwoplaza:
Currently we have to do this:
```ts
const padWLog2 = filterLayout.$.params.padWLog2;
const padWMask = filterLayout.$.params.padWMask;
const padHLog2 = filterLayout.$.params.padHLog2;
const padHMask = filterLayout.$.params.padHMask;
```

We want to be able to do this:
```ts
const { padWLog2, padWMask, padHLog2, padHMask } = filterLayout.$.params;
```

Which should generate:

```wgsl
let padWLog2 = &params.padWLog2;
let padWMask = &params.padWMask;
let padHLog2 = &params.padHLog2;
let padHMask = &params.padHMask;
```

## `let` support

We want to be able to do this:
```ts
let { padWLog2, padWMask, padHLog2, padHMask } = filterLayout.$.params;
```

Which should verify that all spread properties are ephemeral, and generate:

```wgsl
var padWLog2 = params.padWLog2;
var padWMask = params.padWMask;
var padHLog2 = params.padHLog2;
var padHMask = params.padHMask;
```

PR #2913, opened 2026-08-24
feat: Simple Object destructuring in GPU functions

vende11s:
Adds #2318 

Allows simple object destructuring like:

```typescript
 const { a, b: renamed } = createPair();
```
or
```typescript
const pair = Pair({ a: 2, b: 3 });
const { a, b: c } = pair;
```

which generates:
```wgsl
let destructured = createPair();
let a = destructured.a;
let renamed = destructured.b;
```
and
```wgsl
let pair = Pair(2i, 3i);
let a = pair.a;
let c = pair.b;
```

### DIFF
--- a/packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
+++ b/packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
@@ -64,7 +64,7 @@ export const noUnsupportedSyntax = createRule({
         if (!directives.getEnclosingTypegpuFunction()) {
           return;
         }
-        
+
         if (node.left.type === 'ObjectPattern' || node.left.type === 'ArrayPattern') {
           report(node.left, 'destructuring assignment');
           return;
--- a/packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
+++ b/packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
@@ -41,14 +41,12 @@ export const noUnsupportedSyntax = createRule({
 
       for (const parameter of node.params) {
         if (
-          parameter.type === 'Identifier' ||
-          parameter.type === 'AssignmentPattern' ||
-          (parameter.type === 'ObjectPattern' && isSupportedObjectBindingPattern(parameter))
+          parameter.type !== 'Identifier' &&
+          parameter.type !== 'AssignmentPattern' &&
+          (parameter.type !== 'ObjectPattern' || !isSupportedObjectBindingPattern(parameter))
         ) {
-          continue;
+          report(parameter, 'unsupported function parameter binding pattern');
         }
-
-        report(parameter, 'unsupported function parameter binding pattern');
       }
     }
 
--- a/packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
+++ b/packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
@@ -29,8 +29,32 @@ export const noUnsupportedSyntax = createRule({
       });
     }
 
+    function validateFunctionParameters(
+      node:
+        | TSESTree.ArrowFunctionExpression
+        | TSESTree.FunctionExpression
+        | TSESTree.FunctionDeclaration,
+    ) {
+      if (directives.getEnclosingTypegpuFunction() !== node) {
+        return;
+      }
+
+      for (const parameter of node.params) {
+        if (
+          parameter.type === 'Identifier' ||
+          parameter.type === 'AssignmentPattern' ||
+          (parameter.type === 'ObjectPattern' && isSupportedObjectBindingPattern(parameter))
+        ) {
+          continue;
+        }
+
+        report(parameter, 'unsupported function parameter binding pattern');
+      }
+    }
+
     return {
       ArrowFunctionExpression(node) {
+        validateFunctionParameters(node);
         if (directives.getDirectiveStack().at(-2)?.directives.includes('use gpu')) {
           report(node, 'arrow function');
         }
@@ -40,6 +64,12 @@ export const noUnsupportedSyntax = createRule({
         if (!directives.getEnclosingTypegpuFunction()) {
           return;
         }
+        
+        if (node.left.type === 'ObjectPattern' || node.left.type === 'ArrayPattern') {
+          report(node.left, 'destructuring assignment');
+          return;
+        }
+
         if (unsupportedAssignmentOps.includes(node.operator)) {
           report(node, `assignment expression '${node.operator}'`);
         }
@@ -104,12 +134,14 @@ export const noUnsupportedSyntax = createRule({
       },
 
       FunctionDeclaration(node) {
+        validateFunctionParameters(node);
         if (directives.getDirectiveStack().at(-2)?.directives.includes('use gpu')) {
           report(node, 'function declaration');
         }
       },
 
       FunctionExpression(node) {
+        validateFunctionParameters(node);
         if (directives.getDirectiveStack().at(-2)?.directives.includes('use gpu')) {
           report(node, 'function expression');
         }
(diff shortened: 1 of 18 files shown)


packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
```diff
@@ -29,8 +29,32 @@ export const noUnsupportedSyntax = createRule({
...
+      }
+
+      for (const parameter of node.params) {
+        if (
+          parameter.type === 'Identifier' ||
+          parameter.type === 'AssignmentPattern' ||
+          (parameter.type === 'ObjectPattern' && isSupportedObjectBindingPattern(parameter))
+        ) {
+          continue;
+        }
```
vende11s: tf is this reverse guard clause

packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
```diff
@@ -225,8 +257,23 @@ export const noUnsupportedSyntax = createRule({
         if (!directives.getEnclosingTypegpuFunction()) {
           return;
         }
-        if (node.id.type !== 'Identifier') {
-          report(node, 'variable declaration using destructuring');
+
+        const declarationParent = node.parent?.parent;
+        if (
```
vende11s: make this if normal

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -11,6 +11,38 @@ const tsFallthrough = (ctx: Context, node: { expression: babel.Expression }): ti
...
+  }
+
+  if (node.type !== 'ObjectPattern') {
+    throw new Error(`Unsupported binding pattern: ${node.type}`);
+  }
+
+  return {
+    type: BindingPatternType.destructuredObject,
+    props: node.properties.map((prop) => {
+      if (
```
vende11s: weird if

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -245,6 +284,13 @@ const Transpilers: Partial<{
   },
 
   ForStatement(ctx, node) {
+    if (
```
vende11s: weird if

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1617,6 +1672,14 @@ ${this.ctx.pre}else ${alternate}`,
 
     if (statement[0] === NODE.for) {
       const [_, init, condition, update, body] = statement;
+
+      if (
```
vende11s: weird if

packages/eslint-plugin/src/rules/noUnsupportedSyntax.ts
aleksanderkatan, line 1: Have you considered an approach where you start from an object pattern and report by default? Something like this:

```ts
      ObjectPattern(node) {
        if (!directives.getEnclosingTypegpuFunction()) {
          return;
        }

        if (
          node.parent.type === 'VariableDeclaration' &&
          node.parent.parent.type === 'BlockStatement'
        ) {
          // valid, const { a } = obj;
          return;
        }

        // ...
        
        report(node, 'object destructuring');
      },
```

This way, logic related to object pattern would be in one place, instead of spreading it across multiple visitors.

Furthermore, in this case, if you accidentally forget about a supported type of destucturing, you will immediately be notified by the CI.
In the current implementation, where you detect invalid usages, nothing will warn you about usages you missed, like the following: `const fn = () => { 'use gpu'; let a = 0; for ({ a } of source) { } };`

vende11s: Good point, I'll do it this way

packages/tinyest/src/nodes.ts
```diff
@@ -60,15 +60,15 @@ export type Block = readonly [type: NodeTypeCatalog['block'], Statement[]];
...
+  | readonly [type: NodeTypeCatalog['let'], binding: BindingPattern]
+  | readonly [type: NodeTypeCatalog['let'], binding: BindingPattern, value: Expression];
 
 /**
  * Represents a const statement
  */
 export type Const =
-  | readonly [type: NodeTypeCatalog['const'], identifier: string]
-  | readonly [type: NodeTypeCatalog['const'], identifier: string, value: Expression];
+  | readonly [type: NodeTypeCatalog['const'], binding: BindingPattern]
```
aleksanderkatan, line 70: There were changes made to these nodes on main branch, now there's an `Identifier` type possible instead of string.

packages/tinyest/src/nodes.ts
```diff
@@ -266,20 +266,23 @@ export type Expression =
...
   destructuredObject: 'd',
 } as const;
 
-export type FuncParameter =
+export type BindingPattern =
   | {
-      type: typeof FuncParameterType.identifier;
+      type: typeof BindingPatternType.identifier;
       name: string;
     }
```
aleksanderkatan, line 278: I think that we shouldn't replace the original string/Identifier with a struct. This is a change that is not backwards compatible. 

While we can normalize legacy bindings, the normalization required for this cannot be done in a type-safe way (a type-safe way would accept a node of "old" tinyest, and return a node of "new" tinyest, instead of operating on `unknown`, if there are more changes like this, we'll easily get confused, and this is impossible to test without inlining "old" tinyest types).

This will also affect everybody else who's using tinyest (this is a public package), and I think there's just no need for that.

packages/tinyest/src/nodes.ts
```diff
@@ -266,20 +266,23 @@ export type Expression =
...
   | {
-      type: typeof FuncParameterType.destructuredObject;
+      type: typeof BindingPatternType.destructuredObject;
       props: {
         name: string;
         alias: string;
       }[];
     };
```
aleksanderkatan, line 285: I'd avoid structs where possible in tinyest.
- Structs make the generated AST longer (you need to write `{ name: "variable", alias: "value" }` instead of `["variable", "value"]`, and plugins include redundant whitespace for objects).
- Source mapping works out of the box for array-like nodes (though, since this does not go through transpile, it will need a special case in #3044 anyway).

packages/tinyest/src/nodes.ts
```diff
@@ -266,20 +266,23 @@ export type Expression =
...
+
+export type FuncParameter = BindingPattern;
+export const FuncParameterType = BindingPatternType;
```
aleksanderkatan, line 288: Good idea with merging these two types, but I don't think we can do that just yet.

I'm not sure what our policy on non-backwards-compatible changes in tinyest is, it would probably be best to ask @iwoplaza, but I think that it would be fine to update `FuncParameter` to either be as it is now, or `BindingPattern` (updated to be for example `Identifier | [name: number, alias?: number][]`), and then drop the original type in a couple of versions and bump the metadata version.

Actually, not related to this PR, `RawMetadataV1` depending on tinyest may not have been the greatest idea, tinyest updates may change what we think `RawMetadataV1` is... Can you depend on multiple versions of the same package?

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
aleksanderkatan: Free test (while the message is not perfect, I think it is good enough to leave as is)
```ts
  it('throws a descriptive error when destructuring a reference to let', () => {
    const Schema = d.struct({ prop: d.vec3f });
    const testFn = () => {
      'use gpu';
      const struct = Schema();
      let { prop } = struct;
    };

    expect(() => tgpu.resolve([testFn])).toThrowErrorMatchingInlineSnapshot(`
      - <root>
      - fn*:testFn
      - fn*:testFn(): 'let prop = #destructured_source.prop' is invalid, because references cannot be assigned to 'let' variable declarations.
      -----
      - Try 'let prop = vec3f(#destructured_source.prop)' if you need to reassign 'prop' later
      - Try 'const prop = #destructured_source.prop' if you won't reassign 'prop' later.
      -----]
    `);
  });
```

aleksanderkatan: I think that this might be a pretty common usage. Also, it will probably require non-tirvial new infrastructure, so I think that this should be delegated to another PR. Could you create an issue for that?
```ts
  it('allows for destructuring bind group layouts', () => {
      buf: { uniform: d.u32 },
    });

    const fn = () => {
      'use gpu';
      const { buf } = layout.$;
    };

    expect(tgpu.resolve([fn])).toMatchInlineSnapshot();
  });
```

packages/typegpu/tests/tgsl/wgslGenerator.test.ts
```diff
@@ -2392,4 +2392,227 @@ describe('WgslGenerator', () => {
...
+
+        fn fn_1() -> i32 {
+          let destructured = createPair();
+          let a = destructured.a;
+          let renamed = destructured.b;
+          return (a + renamed);
+        }"
+      `);
+    });
+
```
aleksanderkatan: Free test (a little different than the one below)
```ts
 it('does not cause name collisions', () => {
    const Schema = d.struct({ prop: d.vec3f });

    const fn = () => {
      'use gpu';
      const destructured = 1;
      const { prop: a } = Schema();
      const { prop: b } = Schema();
    };

      "struct Schema {
        prop: vec3f,
      }

      fn fn_1() {
        const destructured = 1;
        var destructured_1 = Schema();
        let a = (&destructured_1.prop);
        var destructured_2 = Schema();
        let b = (&destructured_2.prop);
      }"
    `);
  });
```

PR #2914, opened 2026-08-24
impr: Optimize radiance cascade execution

reczkok:
Built on top of #2617.

- shares ray directions through workgroup memory when the cascade layout allows it
- prebuilds jump-flood pipelines instead of updating the offset uniform between dispatches
- records the entire jump-flood run into one compute pass
- allows jump flooding to participate in a caller-owned command encoder


reczkok:
superseded inside #2617

PR #2925, opened 2026-08-26
feat: Add missing `std` functions

reczkok:


packages/typegpu/src/std/packing.ts
```diff
@@ -30,16 +47,51 @@ export const unpack2x16float = dualImpl({
...
-    const buffer = new ArrayBuffer(4);
-    const writer = new TB.BufferWriter(buffer);
-    writeFloat16(writer, e.x);
-    writeFloat16(writer, e.y);
-    const reader = new TB.BufferReader(buffer);
-    return u32(reader.readUint32());
+    const x = Math.floor(0.5 + 32767 * clampScalar(e.x, -1, 1));
+    const y = Math.floor(0.5 + 32767 * clampScalar(e.y, -1, 1));
+    return writePacked((writer) => {
+      writer.writeInt16(x);
```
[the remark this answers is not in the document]
reczkok: image

packages/typegpu/src/std/atomic.ts
```diff
@@ -35,6 +37,22 @@ export const textureBarrier = dualImpl({
   sideEffects: true,
 });
 
+interface WorkgroupUniformLoad {
+  <T extends AnyAtomic>(value: T): number;
+  <T>(value: T): T;
+}
+
+export const workgroupUniformLoad = dualImpl<WorkgroupUniformLoad>({
```
cieplypolar, line 45: It does not work with pointer types and refs. However, it can be easily fixed by:
1. adding `d.ref` overload to interface
2. unwrapping pointer in signature
3. calling `derefSnippet` on value

packages/typegpu/src/std/atomic.ts
```diff
@@ -134,3 +152,41 @@ export const atomicXor = dualImpl<<T extends AnyAtomic>(a: T, value: number) =>
   codegenImpl: (_ctx, [a, value]) => stitch`atomicXor(&${a}, ${value})`,
   sideEffects: true,
 });
+
+export const atomicExchange = dualImpl<<T extends AnyAtomic>(a: T, value: number) => number>({
```
cieplypolar, line 156: typescript signature does not reject `d.ref(atomic.$)` but it could

packages/typegpu/src/std/atomic.ts
```diff
@@ -134,3 +152,41 @@ export const atomicXor = dualImpl<<T extends AnyAtomic>(a: T, value: number) =>
...
+  u32: abstruct({ old_value: u32, exchanged: bool }),
+} as const;
+
+type AtomicCompareExchangeResult = {
+  old_value: number;
+  exchanged: boolean;
+};
+
+export const atomicCompareExchangeWeak = dualImpl<
+  <T extends AnyAtomic>(a: T, compare: number, value: number) => AtomicCompareExchangeResult
```
cieplypolar, line 175: same as above, reject `d.ref` in typescript signature

PR #2929, opened 2026-08-26
fix: evaluate matrix constructor inputs once

### LINKED ISSUE #2859: fix: Repeat call in `d.mat4x4f.translation`
```ts
  it('...', () => {
    const myVar = tgpu.privateVar(d.u32);
    const helper = () => {
      'use gpu';
      myVar.$++;
      return myVar.$;
    };
    const main = () => {
      'use gpu';
      const t = d.mat4x4f.translation(d.vec3f(helper()));
    };

      "var<private> myVar: u32;

      fn helper() -> u32 {
        myVar++;
        return myVar;
      }

      fn main() {
        let t = mat4x4f(1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, vec3f(f32(helper())).x, vec3f(f32(helper())).y, vec3f(f32(helper())).z, 1);
      }"
    `);
  });
```
Either delegate to another function, or find a way to detect this case and throw

huytdps13400:
## Summary

- construct translation matrices from four `vec4f` columns so the translation vector is consumed once
- delegate scaling and X/Y/Z rotation constructors to typed WGSL helpers, so vector/angle expressions are evaluated once at the call boundary
- compute each rotation sine/cosine pair once inside its helper
- use readable inline WGSL snapshots for translation and the four remaining constructors, plus update affected example/rotation snapshots

The CPU matrix layouts and public APIs are unchanged. Generated WGSL now calls a helper only where a constructor must reuse components; the translation constructor retains its compact four-column form.

## TDD evidence

RED: scaling expanded a vector expression three times and each rotation expanded its angle four times.

GREEN: the inline snapshot shows one `nextValue()` argument at each translation/scaling/rotation call, with helper parameters reused inside the generated WGSL.

## Verification

- focused matrix suite: 38/38 passed
- full source unit suite: 275 files; 2,854 passed, 2 skipped
- full monorepo typecheck: 0 errors
- full Oxlint/Prettier/Oxfmt style gate passed
- `git diff --check` passed

Fixes #2859

### DIFF
--- a/packages/typegpu/tests/matrix.test.ts
+++ b/packages/typegpu/tests/matrix.test.ts
@@ -404,21 +404,98 @@ describe('different matrix constructors', () => {
       const translation = d.mat4x4f.translation(d.vec3f(nextValue()));
     };
 
-    const wgsl = tgpu.resolve([main]);
-    const mainStart = wgsl.indexOf('fn main');
-    expect(mainStart).toBeGreaterThanOrEqual(0);
-    const mainSource = wgsl.slice(mainStart);
+    expect(tgpu.resolve([main])).toMatchInlineSnapshot(`
+      "var<private> value: u32;
 
-    expect(mainSource.match(/nextValue\(\)/g) ?? []).toHaveLength(1);
-    expect(mainSource).toMatch(
-      /vec4f\s*\(\s*vec3f\s*\(\s*f32\s*\(\s*nextValue\(\)\s*\)\s*\)\s*,\s*1\s*\)/,
-    );
+      fn nextValue() -> u32 {
+        value++;
+        return value;
+      }
+
+      fn main() {
+        let translation = mat4x4f(vec4f(1, 0, 0, 0), vec4f(0, 1, 0, 0), vec4f(0, 0, 1, 0), vec4f(vec3f(f32(nextValue())), 1));
+      }"
+    `);
   });
 
   it('returns scaling matrix', () => {
     expect(d.mat4x4f.scaling(d.vec3f(3, 4, 5))).toStrictEqual(
       d.mat4x4f(d.vec4f(3, 0, 0, 0), d.vec4f(0, 4, 0, 0), d.vec4f(0, 0, 5, 0), d.vec4f(0, 0, 0, 1)),
     );
+  });
+
+  it('evaluates remaining matrix constructor inputs once during WGSL generation', () => {
+    const value = tgpu.privateVar(d.u32);
+    const nextValue = () => {
+      'use gpu';
+      value.$++;
+      return value.$;
+    };
+    const main = () => {
+      'use gpu';
+      const scaling = d.mat4x4f.scaling(d.vec3f(nextValue()));
+      const rotationX = d.mat4x4f.rotationX(nextValue());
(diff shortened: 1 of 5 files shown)


packages/typegpu/tests/matrix.test.ts
```diff
@@ -392,6 +392,25 @@ describe('different matrix constructors', () => {
...
+      'use gpu';
+      const translation = d.mat4x4f.translation(d.vec3f(nextValue()));
+    };
+
+    const wgsl = tgpu.resolve([main]);
+    const mainSource = wgsl.slice(wgsl.indexOf('fn main'));
+
+    expect(mainSource.match(/nextValue\(\)/g)).toHaveLength(1);
+    expect(mainSource).toContain('vec4f(vec3f(f32(nextValue())), 1)');
+  });
```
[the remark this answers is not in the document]
huytdps13400: Addressed in 4de28e8. The regression now asserts that `fn main` was found before slicing, treats a missing call match as an empty list, and uses a whitespace-tolerant regex for the translation column while preserving both semantic checks. Focused matrix tests pass 37/37, package typecheck passes, and Oxfmt/diff checks are clean.

packages/typegpu/tests/matrix.test.ts
```diff
@@ -392,6 +392,29 @@ describe('different matrix constructors', () => {
...
+
+    const wgsl = tgpu.resolve([main]);
+    const mainStart = wgsl.indexOf('fn main');
+    expect(mainStart).toBeGreaterThanOrEqual(0);
+    const mainSource = wgsl.slice(mainStart);
+
+    expect(mainSource.match(/nextValue\(\)/g) ?? []).toHaveLength(1);
+    expect(mainSource).toMatch(
+      /vec4f\s*\(\s*vec3f\s*\(\s*f32\s*\(\s*nextValue\(\)\s*\)\s*\)\s*,\s*1\s*\)/,
+    );
```
aleksanderkatan: Use inline snapshots, regexes are unreadable

cieplypolar:
This is a partial solution. The issue persists in the four remaining matrix functions.

huytdps13400:
Completed the remaining constructor cases in `26d35fcd`.

- `scaling4` now delegates to a WGSL helper that receives the vector once.
- `rotationX4`, `rotationY4`, and `rotationZ4` each receive the angle once and compute `sin`/`cos` once inside the helper.
- translation keeps the equivalent four-column constructor.
- the regression now uses inline WGSL snapshots for all five constructors; affected example and rotation snapshots were updated.

Focused matrix tests pass 38/38. Full source verification passes 275 files with 2,854 tests / 2 skipped, plus full monorepo typecheck and style gates.

packages/typegpu/tests/matrix.test.ts
huytdps13400: Updated in `26d35fcd`. The regex assertions are replaced with inline WGSL snapshots. The translation snapshot shows the vector expression once, and the new combined snapshot shows a single side-effectful argument for scaling and each X/Y/Z rotation helper. Focused matrix tests pass 38/38.

PR #2933, opened 2026-08-26
fix: indent standalone blocks in tseynit

### LINKED ISSUE #2894: impr: Give correct indent to blocks in tseynit
Currently blocks always start with indent 0. If we change that naively, we introduce whitespaces after `if ()`, `for ()` etc.

huytdps13400:
## Summary

- indent standalone and nested block opening braces using the caller-provided indentation
- omit only the first indentation prefix when a statement is rendered inline after `if`/`else`/`for`/`while`/`for-of`
- avoid `trimStart`, so the stringifier never removes semantic or future formatter-owned whitespace
- keep the focused nested-block regression and affected obfuscation snapshots

Closes #2894.

## Verification

- focused `tseynit` suite: 29/29 passed
- `unplugin-typegpu` obfuscation suite: 24/24 passed
- full source unit suite: 275 files; 2,853 passed, 2 skipped
- full monorepo typecheck: 0 errors
- full Oxlint/Prettier/Oxfmt style gate passed
- `git diff --check` passed

### DIFF
--- a/packages/typegpu/tests/internal/tseynit.test.ts
+++ b/packages/typegpu/tests/internal/tseynit.test.ts
@@ -180,6 +180,28 @@ describe('ast to JS transformation', () => {
       `);
     });
 
+    it('indents standalone nested blocks without adding spaces to control-flow blocks', () => {
+      const fn = () => {
+        'use gpu';
+        {
+          [1, 2];
+        }
+        if (true) {
+          [3, 4];
+        }
+      };
+      expect(stringifyNode(getBodyAst(fn))).toMatchInlineSnapshot(`
+        "{
+          {
+            [1, 2];
+          }
+          if (true) {
+            [3, 4];
+          }
+        }"
+      `);
+    });
+
     it('handles declarations', () => {
       const fn = () => {
         'use gpu';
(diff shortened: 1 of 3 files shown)


aleksanderkatan:
Can we do this without trim?

huytdps13400:
Implemented without trimming in `214b4ebb`.

`stringifyStatement` now accepts an `omitInitialIndent` flag. Control-flow bodies omit only their known first indentation prefix, while nested lines keep the caller-provided indentation unchanged. There is no `trimStart` left in the stringifier.

Focused verification: `tseynit` 29/29 and obfuscation 24/24. Full source verification: 275 files, 2,853 tests passed / 2 skipped; monorepo typecheck and full style gate pass.

aleksanderkatan (approved):
alright

packages/typegpu/tests/internal/tseynit.test.ts
cieplypolar, line 1: add this test
```ts
 it('handles for loop without a block body', () => {
    const fn = () => {
      'use gpu';
      let result = 0;
      for (let i = 0; i < 10; i++) result += i;
    };

    expect(stringifyNode(getBodyAst(fn))).toMatchInlineSnapshot(`
      "{
        let result = 0;
        for (let i = 0; i < 10; i++) result += i;
      }"
    `);
  });
```

PR #2939, opened 2026-08-27
feat: Name collision avoidance in raw-WGSL implemented functions

### LINKED ISSUE #2499: fix: Unique names in raw WGSL code
We provide ways to integrate raw WGSL code into TypeScript shaders, and the other way around. We don't however properly handle name clashes.

```ts
const OFFSET = tgpu.const(d.f32, 10).$name('x');

const hello = tgpu.rawCodeSnippet('buffer[i]', d.f32).$uses({ buffer });
//                                        ^ reference to a local definition
//                                 ^^^^^^ reference to a global definition

// `OFFSET` is resolved before `foo`, therefore by
// the time we start generating `foo`, the identifier
// `x` is already taken, which clashes with the local definition
const foo = tgpu.fn([], d.f32)`() {
  var x = 0f;
  for (var i = 0; i < 10; i++) {
    x += hello;
  }
  return x + OFFSET;
}`.$uses({ OFFSET, hello });

tgpu.resolve([OFFSET, foo]);
```

Let's keep a new `localRenames` map
1. Push a function layer in the ItemStateStack (or an equivalent layer that will allow up to store unique local names)
2. Collect all identifiers used in the function body, excluding external keys (in this case: `x` and `i`)
3. Filter for those that are already taken (in this case: `x`)
4. Generate unique identifiers for those that are taken, save them to the map (`x` => `x_1`), find and replace in the WGSL template.
5. We reserve the non-taken identifiers, so no new global definitions will be generated with those names as long as the function layer is on the stack (in this case just `i`, as `x_1` has already been reserved as part of generating a unique name).
After this process, the WGSL template will look like this:
```wgsl
() {
  var x_1 = 0f;
  for (var i = 0; i < 10; i++) {
    x_1 += hello;
  }
  return x_1;
}
```
Now we can resolve all externals, and replace them in the template:
1. When we resolve tgpu.rawCodeSnippet, we perform basically the same process we did before, where we collect all identifiers used in this now smaller template (excluding external keys), which in this case is just `i`
2. We look up if any local identifiers have been renamed (using the `localRenames` map), and find and replace (in this case, we don't rename anything, as `i` wasn't renamed in the parent function)
```ts
const hello = tgpu.rawCodeSnippet('buffer[i]', d.f32).$uses({ buffer });
//                                        ^ reference to a local definition
//                                 ^^^^^^ reference to a global definition
```

We can determine whether something is an external definition (*external name*), or something referencing a local definition (*local name*) based on if something is passed in `$uses(...)`.

## tgpu.resolve with template

We can treat tgpu.resolve with templates the same way we treat WGSL-implemented functions, we just don't need to find and replace anything in the template as we already know the template is the first thing being resolved, so we can just globally reserve every non-external identifier that
is used in the template, and proceed with regular resolving of externals.
```ts
const FACTOR = tgpu.const(d.u32, 10).$name('factor');

const shader = tgpu.resolve({
  template: `
fn foo() {
  const factor = 1f;
  const hello = FACTOR;
}
`,
  externals: {
    FACTOR,
  },
});
```

iwoplaza:
Changes:
- Reserving local identifiers in raw-WGSL implementations so that subsequent declarations get a different name
- Renaming local definitions is they clash with something already declared.
- Non-external identifiers in raw code snippets should be treated as referring to the nearest scope, and they should get renamed if that's what happened in the raw-WGSL implemented function that used the snippet
- Extend the renaming behavior to uses of raw code snippets in JS-implemented functions
- Moved prop verification to struct resolution
- Removed prop verification from auto-structs, they're already verified by the generated struct

TODO:
- [ ] Register renamed arguments and local definitions in `localRenames` map
- [ ] Consider adding a `globalRenames` map to be used by templated tgpu.resolve
- [ ] Consider making `itemStateStack` public, not a part of the internals of the resolution ctx (perhaps another PR)


packages/typegpu/src/rawShaderCodeUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+
+function swapChars(source: string, offset: number, replacement: string) {
+  return source.slice(0, offset) + replacement + source.slice(offset + replacement.length);
+}
+
+/**
+ * Same as `stripWGSLComments`, but keeps all non-comment code at
+ * the same location where it originally was, and replaces the comments
+ * with whitespace.
+ */
```
aleksanderkatan, line 83: Now there's no need for `stripWgslComments`, as minification removes whitespace anyways. Please remove it to reduce code duplication

packages/typegpu/src/rawShaderCodeUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+        if (code.startsWith('/*', offset)) {
+          depth += 1;
+          result = swapChars(result, offset, '  ');
+          offset += 2;
+        } else if (code.startsWith('*/', offset)) {
+          depth -= 1;
+          result = swapChars(result, offset, '  ');
+          offset += 2;
+        } else {
+          result = swapChars(result, offset, ' ');
```
aleksanderkatan, line 116: Swapping character by character this way puts way too much trust in JS strings IMO (I would be seriously surprised if this wasn't O(n^2)), can we optimize this? We could either reduce the number of replaces to how many comments they are (instead of how many chars they take), or we could use e.g. TextEncoder & TextDecoder to make this truly linear

packages/typegpu/src/rawShaderCodeUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+  while ((result = expr.exec(noCommentsSource)) !== null) {
+    if (result[0][0] === '.') {
+      // Skipping member accesses.
+      continue;
+    }
+
+    if (result[1]) {
+      identifiers.push(result[1]);
+    }
+  }
```
aleksanderkatan, line 158: Wouldn't this be cleaner than a while with an assignment in its test? Am I missing something?

```suggestion
  for (const result of noCommentsSource.matchAll(expr)) {
    if (result[0][0] === '.') {
      // Skipping member accesses.
      continue;
    }

    if (result[1]) {
      identifiers.push(result[1]);
    }
  }
```

packages/typegpu/src/rawShaderCodeUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+}
+
+export function renameIdentifiers(_source: string, renames: Map<string, string>) {
+  // Adding a space at the beginning of `source` so all potential identifiers
+  // have a preceeding character (see the regex for more context).
+  const source = ' ' + _source;
+  const noCommentsSource = blankOutWGSLComments(source);
+  // Capturing the preceeding character (irrespective of whitespace) to make sure that it's not a
+  // chained member access, nor a typed numeric literal (e.g. 1f)
+  const expr = new RegExp(`[^\\d]\\s*${anyIdent.source}`, 'ug');
```
aleksanderkatan, line 170: nit: we don't need to recreate this regex each call, especially when it's used in multiple functions

packages/typegpu/src/rawShaderCodeUtils.ts
```diff
@@ -0,0 +1,197 @@
...
+    const end = result.index + result[0].length;
+    // counting back from the end to keep any extra whitespace that was there
+    const start = end - identifier.length;
+    replaced += source.slice(copiedUpTo, start);
+    replaced += renames.get(identifier) ?? identifier;
+    copiedUpTo = end;
+  }
+
+  // Removing the first space we added at the beginning
+  return (replaced + source.slice(copiedUpTo)).slice(1);
```
aleksanderkatan, line 196: I think this could be done with one singular string.replace.
We already once do something like this in our repo, see the top of `functionDefinition` in wgslGenerator. Can you check this out?

aleksanderkatan (changes requested):
Nice!
I didn't review the entire thing yet. I leave some cosmetic & complexity requests.

iwoplaza:
@aleksanderkatan Thanks for the initial review! I'll put it back into draft, I want to approach some things differently to be more robust, and address your comments.

PR #2946, opened 2026-08-30
fix: Validate unresolved declaration initializers

### LINKED ISSUE #2934: chore: `UnknownData` rhs sometimes cannot be wrapped with appropriate schema
```ts
it('throws when assigning to a variable', () => {
  const myFn = () => {
    'use gpu';
    const a = null;
  };

  expect(() => tgpu.resolve([myFn])).toThrowErrorMatchingInlineSnapshot(`
    - <root>
    - fn*:myFn
    - fn*:myFn(): 'const a = null' is invalid, cannot determine WGSL type of 'null'
    -----
    - Try using or defining a schema that matches your desired value the most, and wrap the value with it: 'const a = Schema(null)'
    -----]
  `);
});
```
Also, applies to strings and `undefined`.

huytdps13400:
## Summary

- reject `let` and `const` declarations whose initializer resolves to `void`, including void-returning function calls and bare `undefined`
- route those declarations through the existing cannot-determine-WGSL-type diagnostic
- show schema-wrapping guidance only when the unresolved runtime value can actually be wrapped by a schema
- retain valid object/array guidance while suppressing invalid `Schema(null)`, `Schema(undefined)`, and string suggestions
- cover `let`/`const`, null/string diagnostics, bare `undefined`, and void-call results

Closes #2934

## Verification

- focused declaration/type-inference tests: 39 passed
- `pnpm --filter typegpu test:types`: passed
- `pnpm test:style`: passed
- `pnpm test:fast-unit`: 216 files passed, 2,815 tests passed, 2 skipped

### DIFF
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -1345,7 +1345,7 @@ Try 'return ${typeStr}(${str});' instead.
 
     const definitionDataType = eq.dataType;
 
-    if (definitionDataType === UnknownData) {
+    if (definitionDataType === UnknownData || wgsl.isVoid(definitionDataType)) {
       const rhsStr = stringifyNode(eqNode);
       const declaration = `let ${rawId}`;
       throw new WgslTypeError(
@@ -1432,7 +1432,7 @@ Try 'return ${typeStr}(${str});' instead.
     let varType: 'var' | 'let' | 'const' | '<deferred>' = '<deferred>';
     let definitionDataType = eq.dataType;
 
-    if (definitionDataType === UnknownData) {
+    if (definitionDataType === UnknownData || wgsl.isVoid(definitionDataType)) {
       const rhsStr = stringifyNode(eqNode);
       const declaration = `const ${rawId}`;
       throw new WgslTypeError(
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -1334,7 +1334,10 @@ Try 'return ${typeStr}(${str});' instead.
 
     const definitionDataType = eq.dataType;
 
-    if (definitionDataType === UnknownData) {
+    if (
+      definitionDataType === UnknownData ||
+      (eq.value === undefined && wgsl.isVoid(definitionDataType))
+    ) {
       const rhsStr = stringifyNode(eqNode);
       throw new WgslTypeError(
         `'let ${rawId} = ${rhsStr}' is invalid, cannot determine WGSL type of '${rhsStr}'
@@ -1423,7 +1426,10 @@ Try 'return ${typeStr}(${str});' instead.
     let varType: 'var' | 'let' | 'const' | '<deferred>' = '<deferred>';
     let definitionDataType = eq.dataType;
 
-    if (definitionDataType === UnknownData) {
+    if (
+      definitionDataType === UnknownData ||
+      (eq.value === undefined && wgsl.isVoid(definitionDataType))
+    ) {
       const rhsStr = stringifyNode(eqNode);
       throw new WgslTypeError(
         `'const ${rawId} = ${rhsStr}' is invalid, cannot determine WGSL type of '${rhsStr}'
--- a/packages/typegpu/src/tgsl/wgslGenerator.ts
+++ b/packages/typegpu/src/tgsl/wgslGenerator.ts
@@ -218,6 +218,17 @@ const usageToVarTemplateMap: Record<VariableScope | BindableBufferUsage, string>
  */
 const functionInitialBlockDepth = 2;
 
+function schemaWrappingSuggestion(declaration: string, rhs: string, value: unknown): string {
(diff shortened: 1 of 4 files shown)


cieplypolar:
Appreciate the effort, but issue is not solved. `WgslGenerator` still advices wrapping `undefined`, `null` or any string with an appropriate schema.

huytdps13400:
@cieplypolar Fixed in `11511f3f`. `WgslGenerator` now keeps the schema-wrapping hint only for values that can actually be wrapped; `null`, `undefined`, and strings report the inference failure without suggesting invalid `Schema(...)` calls. I updated both `let` and `const` regressions, added const-string coverage, and retained the existing hint coverage for wrappable struct/array values. Verification: focused 36 tests, TypeGPU type tests, style gate, and the fast unit suite (216 files, 2,812 passed, 2 skipped) all pass.

cieplypolar:
1. drop `eq.value === undefined` condition
2. add the following test
```ts
it('...', () => {
  const noop = tgpu.fn([])(() => {});

  const f = tgpu.fn([])(() => {
    const a = noop();
  });

    "fn noop() {}

    fn f() {
      let a = noop();
    }"
  `);
});
```
3. add tests that verify `let` behavior

huytdps13400:
Addressed in `182e8c90`.

- Removed the `eq.value === undefined && wgsl.isVoid(...)` declaration guard so calls to void-returning functions remain valid.
- Added the requested `noop()` regression; it resolves to `let a = noop();`.
- Extended `let` coverage for string/null diagnostics without invalid schema hints, plus the bare-`undefined` void resolution error.

Verification:
- focused declaration/type-inference tests: 38 passed
- `pnpm --filter typegpu test:types`: passed
- `pnpm test:style`: passed
- `pnpm test:fast-unit`: 216 files passed, 2,814 tests passed, 2 skipped

huytdps13400:
Updated the PR title/description and corrected my earlier #2934 comment so the record now matches `182e8c90`: invalid schema-wrapping hints are suppressed, valid object/array hints remain, and void-call declaration inference is preserved. No code change was needed.

cieplypolar:
The following test should fail
```ts
it('supports assigning the result of a void function', () => {
    const noop = tgpu.fn([])(() => {});

    const f = tgpu.fn([])(() => {
      const a = noop();
    });

      "fn noop() {}

      fn f() {
        let a = noop();
      }"
    `);
  });
```
Sorry for being unclear.

huytdps13400:
> The following test should fail
> 
> ```ts
> 
> it('supports assigning the result of a void function', () => {
> 
>     const noop = tgpu.fn([])(() => {});
> 
> 
> 
>     const f = tgpu.fn([])(() => {
> 
>       const a = noop();
> 
>     });
> 
> 
> 
>     expect(tgpu.resolve([f])).toMatchInlineSnapshot(`
> 
>       "fn noop() {}
> 
> 
> 
>       fn f() {
> 
>         let a = noop();
> 
>       }"
> 
>     `);
> 
>   });
> 
> ```
> 
> Sorry for being unclear.
> 
> 

I got it

huytdps13400:
Fixed in `080c9cec`. Both `const a = noop()` and `let a = noop()` now fail through the unresolved-declaration diagnostic, and the invalid schema hint remains suppressed. Bare `undefined` follows the same path.

Verification:
- focused declaration/type-inference tests: 39 passed
- TypeGPU type tests: passed
- style: passed
- fast unit suite: 216 files, 2,815 passed, 2 skipped

cieplypolar:
This test shouldn't be in `letDeclaration.test.ts` file
```ts
it('rejects assigning the result of a void function', () => {
  const noop = tgpu.fn([])(() => {});

  const f = tgpu.fn([])(() => {
    const a = noop();
  });

  expect(() => tgpu.resolve([f])).toThrowErrorMatchingInlineSnapshot(`
    - <root>
    - fn:f: 'const a = noop()' is invalid, cannot determine WGSL type of 'noop()']
  `);
});
```

huytdps13400:
Fixed in `c84e186d`. The `const a = noop()` regression now lives with the unresolved `const` initializer cases in `typeInference.test.ts`; the direct `let a = noop()` regression remains in `letDeclaration.test.ts`. Test behavior and snapshots are unchanged.

Verification:
- focused tests: 39 passed
- TypeGPU type tests: passed
- style: passed
- fast unit suite: 216 files, 2,815 passed, 2 skipped

cieplypolar:
Almost there, just delete duplicated test: `tgslFn.test.ts` line 1315

aleksanderkatan (approved):

huytdps13400:
Fixed in `4fa3c88e`. Removed the duplicate null-assignment regression from `tgslFn.test.ts`; the focused `const` coverage remains in `typeInference.test.ts`, and `let` coverage remains in `letDeclaration.test.ts`.

Verification:
- focused three-file suite: 102 passed
- TypeGPU type tests: passed
- style: passed
- fast unit suite: 216 files, 2,814 passed, 2 skipped

aleksanderkatan (approved):
sure

iwoplaza (approved):
Looks good!

Issue #2944, opened 2026-08-28
bug: WGSL generator resolves object properties out of JS order


cieplypolar:
```ts
it('resolves object fields in schema declaration order', () => {
  using consoleLogSpy = vi.spyOn(console, 'log').mockImplementation(() => {});

  const fieldX = tgpu.comptime(() => {
    console.log('fieldX');
    return 6;
  });
  const fieldY = tgpu.comptime(() => {
    console.log('fieldY');
    return 7;
  });

  const f = tgpu.fn(
    [],
    d.struct({ x: d.u32, y: d.u32 }),
  )(() => {
    'use gpu';
    return {
      y: fieldY(),
      x: fieldX(),
    };
  });

  tgpu.resolve([f]);

  expect(consoleLogSpy.mock.calls).toEqual([['fieldX'], ['fieldY']]);
});
```

huytdps13400:
I ran the issue snippet verbatim on current `main` (`2d43507f`, TypeGPU 0.12.4): `vi.spyOn(console, "log")`, direct `tgpu.comptime` `fieldX`/`fieldY`, the reversed literal `{ y: fieldY(), x: fieldX() }`, and the declared output `d.struct({ x: d.u32, y: d.u32 })`. The focused conversion test passes and `consoleLogSpy.mock.calls` is exactly `[["fieldX"], ["fieldY"]]`; no production patch is needed for that assertion.

The title may instead mean preserving JavaScript object-literal evaluation order. If so, the expected calls would be `fieldY` then `fieldX`, which conflicts with the body assertion. Could you clarify which order is intended? I can implement the source-order behavior once that contract is confirmed.

cieplypolar:
Intended order:
` expect(consoleLogSpy.mock.calls).toEqual([['fieldY'], ['fieldX']]);`

aleksanderkatan:
I think that this ordering is a much bigger issue than that. Take a look at the following function:
```ts
    const incrementBy = tgpu.fn(
      [d.ptrFn(d.u32), d.u32],
      d.u32,
    )((p, n) => {
      'use gpu';
      p.$ += n;
      return p.$;
    });

    const f = tgpu.fn(
      [],
      d.struct({ x: d.u32, y: d.u32 }),
    )(() => {
      'use gpu';
      const p = d.ref(0);
      return {
        y: incrementBy(p, 1),
        x: incrementBy(p, 2),
      };
    });
```

[...]

PR #2950, opened 2026-08-31
feat: Support computed object properties

cieplypolar:
In addition to supporting computed object properties, this preserves evaluation order of comptime-known keys and values mentioned in #2944.


packages/tinyest/src/nodes.ts
```diff
@@ -193,6 +195,25 @@ export type ObjectExpression = readonly [
...
+      type: NodeTypeCatalog['objectProperty'],
+      key: Expression,
+      value: Expression,
+      computed: true,
+    ];
+
+export type ObjectExpressionWithComputedProps = readonly [
+  type: NodeTypeCatalog['objectExprWithComputedProps'],
+  ObjectProperty[],
+];
```
aleksanderkatan, line 198: I think the type can be simplified to this (you can always do an `Array.isArray()` check):
```ts
export type ObjectProperty =
  | readonly [
      type: NodeTypeCatalog['objectProperty'],
      key: string,
      value: Expression,
      computed: false,
    ]
  | readonly [
      type: NodeTypeCatalog['objectProperty'],
      key: Expression,
      value: Expression,
      computed: true,
    ];

export type ObjectExpression = readonly [
  type: NodeTypeCatalog['objectExpr'],
  props: Record<string, Expression> | ObjectProperty[],
];
```

Since this node only appears in `ObjectExpression` (therefore, it is neither a standalone statement nor an expression), I'd even consider a type like this:

```ts
// No type, since ObjectProperty can only appear in ObjectExpression
export type ObjectProperty =
  | readonly [key: string, value: Expression, computed: false]
  | readonly [key: Expression, value: Expression];

export type ObjectExpression = readonly [
  type: NodeTypeCatalog['objectExpr'],
  props: Record<string, Expression> | ObjectProperty[],
];
```
If we do this, then in a few versions we could drop the `Record<string, Expression>` since people will need to adapt to the other case anyway.

We already do a trick like this in switch statement:
```ts
export type Switch = readonly [
  type: NodeTypeCatalog['switch'],
  discriminant: Expression,
  cases: readonly (readonly [test: Expression | null, consequent: readonly Statement[]])[],
];
```

cieplypolar: I like it. I will implement that.

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -31,9 +36,7 @@ function createLegacyTraspilers() {
     ...acornTranspilers,
 
     ObjectExpression(ctx, node, transpile) {
-      const properties: Record<string, tinyest.Expression> = {};
-
-      for (const prop of node.properties) {
+      const objectProperties = node.properties.map((prop) => {
```
aleksanderkatan, line 39: This is a legacy parser. I think there is no need for updating it.

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -899,34 +899,53 @@ export class WgslGenerator implements ShaderGenerator {
...
+            ([key, value]) => [key, value, false] satisfies tinyest.ObjectProperty,
+          );
+
+      const seenKeys = new Map<string, tinyest.ObjectProperty>();
+      const resolveUniqueKey = (prop: tinyest.ObjectProperty): string => {
+        const key = this._resolveObjectPropertyKey(prop);
+        const dupProp = seenKeys.get(key);
+        if (dupProp) {
+          throw new WgslTypeError(
+            `Duplicate object property key found: '${stringifyObjectProperty(dupProp)}' and '${stringifyObjectProperty(prop)}'.`,
```
aleksanderkatan, line 915: I'd include the value here as well, in the case both of these are function calls.

aleksanderkatan (approved):
Sure! 🏇🔇

packages/tinyest-for-wgsl/src/transpilers.ts
```diff
@@ -288,6 +312,36 @@ const tsFallthrough = (
   return transpile(ctx, node.expression);
 };
 
+export function parseBabelObjectProperty(
```
iwoplaza: Why is this one named "parse..." while the others are "transpile..."? It even gets imported as `transpileBabelObjectProperty` in parsers.ts 👀

packages/typegpu/src/tgsl/wgslGenerator.ts
iwoplaza: @aleksanderkatan I think `stringifyObjectProperty` returns both the key and the value

packages/tinyest-for-wgsl/src/parsers.ts
cieplypolar: I did it already, it won't hurt.

iwoplaza (approved):
Looks great! Left one nit/question.

packages/tinyest-for-wgsl/src/transpilers.ts
cieplypolar: Yeeea, that was my IDE

PR #2952, opened 2026-08-31
fix: Quote non-identifier object property keys

cieplypolar:
JS identifier rules
image
No need to worry about `UnicodeEscapeSequence` since it is evaluated by parser.


aleksanderkatan (approved):
++

PR #2959, opened 2026-09-02
docs: Examples page visual revamp

iwoplaza:
**Changes:**
- Reworked the layout of the examples page to make it more readable, and have the ability to open the example in fullscreen.
- Reworked the look of the examples page to better match the landing page stylistically.
- Added dark mode the the examples
- The landing page, docs & examples now have a matching header
- There is no longer a toggle for showing "Experimental" examples, they're shown by default, and `~unstable` is not just one of the tags.


apps/typegpu-docs/src/components/ExampleLink.tsx
iwoplaza, line 1: This component has been merged into ExampleCard

PR #2960, opened 2026-09-02
fix: preserve scopes around unrolled raw code

### LINKED ISSUE #2897: fix: Raw code snippet block pruning in unroll
```ts
  it('does not break with raw code snippets', () => {
    const snip = tgpu['~unstable'].rawCodeSnippet('const a = 1', d.Void);

    const fn = () => {
      'use gpu';
      for (const a of tgpu.unroll([1, 2, 3])) {
        snip.$;
      }
    };

      "fn fn_1() {
        // unrolled iteration #0
        const a = 1;
        // unrolled iteration #1
        const a = 1;
        // unrolled iteration #2
        const a = 1;
        // ---
      }"
    `);
  });
```

oxura:
Raw WGSL snippets are opaque to the generator, so the block-pruning pass could not see declarations inside them. During `tgpu.unroll`, that flattened every iteration into the function scope and produced invalid WGSL when a snippet declared the same name more than once.

This marks raw snippets as potentially declaring names, preserves that marker when snippets are copied, and feeds it into the existing `definesInNearestScope` decision. Each affected unrolled iteration now keeps its own brace scope; generated expressions that are not raw snippets retain the current pruning behavior.

Verification:
- focused unroll/raw-snippet suites: 37 passed
- TypeGPU type tests
- changed-file Oxlint and Oxfmt checks
- fast unit suite on Node 24: 2,809 passed, 2 skipped
- `typegpu` package build

Closes #2897.

### DIFF
--- a/packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
+++ b/packages/typegpu/src/core/rawCodeSnippet/tgpuRawCodeSnippet.ts
@@ -112,13 +112,25 @@ class TgpuRawCodeSnippetImpl<TDataType extends BaseData> implements TgpuRawCodeS
             this.#expression,
           );
 
-          return snip(replacedExpression, this.dataType, this.origin, this.possibleSideEffects);
+          return snip(
+            replacedExpression,
+            this.dataType,
+            this.origin,
+            this.possibleSideEffects,
+            /* mayDeclareVariables */ true,
+          );
         },
       }),
       {
         codegenMode: {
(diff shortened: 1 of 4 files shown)


aleksanderkatan:
Not sure if this is worth keeping another prop on snippets.
RawCodeSnippet already mentions that it is intended to hold an expression, not a statement.
Can we just declare this an invalid API use and close this issue & PR? @iwoplaza @cieplypolar @reczkok @vende11s

PR #2969, opened 2026-09-03
fix: Fall back to `@typegpu/cli@latest` only when no matching version exists

reczkok:
The `typegpu` proxy bin treated any non-zero exit from `@typegpu/cli` as "no matching version" and reran the command with `@latest`. Breaks `inspect check`, which exits 1 on findings.

- asks the registry first, falls back only on E404
- propagates the CLI exit code and signals unchanged, a signal re-raises instead of exiting 0
- registry probe uses promisified `execFile` instead of a hand-rolled promise with stdout/stderr accumulators, it already collects both and rejects with them attached
- probe returns one boolean ("no match confirmed") instead of `true | false | undefined`, an unreachable registry just lands in the `catch` and keeps the range
- swapped the version regex for a `split('.')` since we only ever use the first two numbers, which are safe for any valid semver
- async IIFE replaced with top-level await, it's an `.mjs`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Pr1wBWBj6Xj5c13FgrmzGy

### DIFF
--- a/packages/typegpu/bin.mjs
+++ b/packages/typegpu/bin.mjs
@@ -1,79 +1,51 @@
 #!/usr/bin/env node
-import { spawn } from 'node:child_process';
+import { execFile, spawn } from 'node:child_process';
+import { promisify } from 'node:util';
 import pkg from './package.json' with { type: 'json' };
 
-/**
- * Used to extract the version of `typegpu` that was used to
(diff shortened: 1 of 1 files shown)


reczkok:
claude slopped this unprompted

PR #2972, opened 2026-09-03
docs: Better "Why TypeGPU" section

iwoplaza:


apps/typegpu-docs/src/content/docs/why-typegpu/index.mdx
```diff
@@ -64,15 +81,22 @@ They then get destroyed when calling `runtime.dispose()`.
 
 </details> */}
 
-{/* <details>
+<details>
+<summary>🙋 Do I have to use TypeScript?</summary>
+
+No.
```
reczkok, line 87: But you really should

PR #2973, opened 2026-09-04
feat: Add Identifier and Bool nodes

aleksanderkatan:
Changes:
- Allow for identifier and boolean nodes to be wrapped in dedicated nodes instead of being plain boolean/string.
- Add transpile option to use verbose nodes.
- Update tseynit/obfuscate/wgslGenerator accordingly.

Testing: 
I set verboseNodes to true in unplugin, and the only tests that failed included AST snapshots.


aleksanderkatan:
@pullfrog review

aleksanderkatan:
@pullfrog review

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -1991,7 +1998,18 @@ function extractObject(expr: tinyest.Expression): string | undefined {
...
+function isId(expr: unknown): expr is tinyest.Identifier {
+  return typeof expr === 'string' || (Array.isArray(expr) && expr[0] === NODE.identifier);
+}
+
+function extractId(ident: tinyest.Identifier): string {
+  if (typeof ident === 'string') {
+    return ident;
   }
+  return ident[1];
 }
```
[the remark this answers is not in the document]
aleksanderkatan: I think they are simple enough to have them copied, mainly because I don't want to export more things from tinyest

packages/typegpu/tests/internal/tseynit.test.ts
```diff
@@ -371,5 +373,61 @@ describe('ast to JS transformation', () => {
         }"
       `);
     });
+
+    it('handles boolean node', () => {
+      const ast: LogicalExpression = [
```
iwoplaza, line 378: Why not use `tinyest.`, and remove the additional imports?
```suggestion
      const ast: tinyest.LogicalExpression = [
```

iwoplaza (approved):
Let's gooo! They get their own special nodes 🎉🎉

packages/tinyest-for-wgsl/src/parsers.ts
```diff
@@ -93,6 +101,9 @@ function createParser(kind: 'acorn' | 'babel' | 'legacy') {
       const externalChain = tryFindExternalChain(ctx, node);
       if (externalChain) {
         ctx.externalNames.set(externalChain, externalChain);
+        if (ctx.opts.verboseNodes) {
+          return [NODE.identifier, externalChain];
+        }
         return externalChain;
```
cieplypolar, line 107:
```suggestion
return ctx.opts.verboseNodes ? [NODE.identifier, externalChain] : externalChain;
```

packages/tinyest-for-wgsl/tests/verboseNodes.test.ts
```diff
@@ -0,0 +1,166 @@
...
+        Map {
+          "ext" => "ext",
+          "ext.prop" => "ext.prop",
+        }
+      `);
+    }),
+  );
+
+  it(
+    'does not use nodes for object expressions',
```
cieplypolar:
```suggestion
    'does not use nodes for object expression keys',
```

packages/typegpu/src/tgsl/wgslGenerator.ts
```diff
@@ -437,14 +437,18 @@ export class WgslGenerator implements ShaderGenerator {
...
-      return this._identifier(expression);
+    if (isId(expression)) {
+      return this._identifier(extractId(expression));
     }
 
     if (typeof expression === 'boolean') {
       return snip(expression, bool, /* origin */ 'constant', false);
     }
 
+    if (expression[0] === NODE.booleanLiteral) {
```
cieplypolar: I don't like this inconsistency that identifier uses a dedicated `isId` helper, but boolean handles this with two branches. Could you align them to use the same approach?

cieplypolar (approved):
LGTM! Left 3 nits.

PR #2980, opened 2026-09-04
docs: Small example page fixes

iwoplaza:
Changes:
- In fullscreen mode, a select's list is now on top of the control panel, instead of behind it
- The copyright in the footer is now aligned to the center properly


reczkok (approved):
👍

PR #2982, opened 2026-09-06
feat(@typegpu/sdf): Add arc, capped cylinder, rhombus, and triangle shapes

iwoplaza:


packages/typegpu-sdf/src/2d.ts
```diff
@@ -185,3 +185,25 @@ export const sdPie = tgpu.fn(
   const m = distance(p_w, sc * clamp(dot(p_w, sc), 0, radius));
   return max(l, m * sign(sc.y * p_w.x - sc.x * p_w.y));
 });
+
+/**
+ * Signed distance function for a circular arc.
+ * The circle is centered at the origin; the arc is symmetric about the negative Y axis.
```
cieplypolar, line 191: 'symmetric about the negative Y axis' it took me 15 minutes to realize that this means the opening faces upward. Consider adding this intuition in parentheses.

packages/typegpu-sdf/src/3d.ts
```diff
@@ -115,3 +128,110 @@ export const sdCapsule = tgpu.fn(
...
+ * The vertices must be distinct and non-collinear, and may use either winding order.
+ *
+ * @param p Point to evaluate, in the same coordinate space as the vertices
+ * @param a First vertex of the triangle
+ * @param b Second vertex of the triangle
+ * @param c Third vertex of the triangle
+ * @returns Distance to the nearest point on the triangle, including its edges;
+ * zero on the triangle and positive on either side of its plane
+ */
+export const sdTriangle3d = (p: v3f, a: v3f, b: v3f, c: v3f) => {
```
cieplypolar, line 152: I agree with our friends, exported functions should be shelled.

cieplypolar (approved):
Left 2 nits. 🚢 🇮🇹

packages/typegpu-sdf/src/2d.ts
aleksanderkatan: Even after your explanation I still don't get it. Negative Y and positive Y are the same axis, aren't they?

packages/typegpu-sdf/src/2d.ts
```diff
@@ -185,3 +185,25 @@ export const sdPie = tgpu.fn(
...
+ *
```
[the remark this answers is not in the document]
aleksanderkatan: Not exactly agree, I think that this is still an SDF, it's just that there are no points that are inside of the shape. 

Can we be clearer about the fact that this arc is of width 0?
I was surprised that this drew nothing

```ts
const pipeline = root.createRenderPipeline({
  vertex: fullScreenTriangle,
  fragment: ({ uv }) => {
    'use gpu';
    const p = 2 * uv - 1;
    if (sdf.sdArc(p, d.vec2f(std.sin(angle / 2), std.cos(angle / 2)), 0.1) <= 0) {
      return d.vec4f(1, 0, 0, 1);
    }
    return d.vec4f();
  },
});
```

aleksanderkatan (approved):
Sure! more shapez

PR #2987, opened 2026-09-07
impr: Pass state takes precedence over pipeline-held state

reczkok:
`pass.set*` > `pipeline.with*` > WebGPU default (`.with` is per pipeline, does not influence other pipelines)


packages/typegpu-testing-utility/src/extendedIt.ts
```diff
@@ -97,7 +99,10 @@ export const it = base
         return mockDevice;
       },
       features: new Set(['timestamp-query']),
-      createBindGroup: vi.fn((_descriptor: GPUBindGroupDescriptor) => 'mockBindGroup'),
+      createBindGroup: vi.fn(({ label }: GPUBindGroupDescriptor) => ({
+        label,
+        id: nextBindGroupId++,
```
iwoplaza: Since this is not part of the public API, and just for verifying which mock is which, I'd prefer a `__id` field
```suggestion
        __id: nextBindGroupId++,
```

reczkok: fixed via removal 🧌

iwoplaza (approved):

PR #2997, opened 2026-09-08
feat: Add 3D mesh endpoint to `@typegpu/geometry`

reczkok:
Adds `meshes` to `@typegpu/geometry`. Geometry is a vertex schema + `vertexAt` (and optionally `indexAt`), so primitives and transformations compose before anything needs a buffer

- plane, box, sphere, cylinder, torus + custom parametric surfaces
- transform, map, attach extra attributes and concat geometries
- evaluate directly in shaders, bake into buffers or bake just the indices while keeping vertices procedural
- baked meshes can be updated through compute and used as geometry inputs again
- includes an animated primitives example with shadows

Interactive cloth example and docs will follow in separate PRs


deluksic:
Please take a look at my (almost finished) branch, if you would like to incorporate any of it, or steal techniques

https://github.com/deluksic/TypeGPU/tree/de/geometry-spheres (it is more than spheres)

PR #3009, opened 2026-09-09
feat(unplugin-typegpu): Support Node & Deno via synchronous hooks

iwoplaza:
This PR adds the `unplugin-typegpu/node` and `unplugin-typegpu/deno` entrypoints that expose a hook config that can be used by the [`registerHooks` Node API](https://nodejs.org/api/module.html#customization-hooks).

### DIFF
--- a/packages/unplugin-typegpu/src/node.ts
+++ b/packages/unplugin-typegpu/src/node.ts
@@ -1,0 +1,49 @@
+import { registerHooks, type ModuleHooks } from 'node:module';
+import { fileURLToPath } from 'node:url';
+import type { UnpluginBuildContext, UnpluginContext } from 'unplugin';
+import { earlyPruneRegex, type Options } from './core/common.ts';
+import { unpluginFactory } from './core/factory.ts';
+import { createFilterForId } from './core/filter.ts';
+
+export type { Options } from './core/common.ts';
+
+/**
+ * Installs TypeGPU's synchronous module hooks in Node.js 22.15+ and Deno 2.8+.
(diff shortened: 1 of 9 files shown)


packages/unplugin-typegpu/src/node.ts
```diff
@@ -0,0 +1,49 @@
...
+      }
+
+      const code =
+        typeof result.source === 'string' ? result.source : new TextDecoder().decode(result.source);
+      if (rawOptions?.earlyPruning !== false && earlyPruneRegex.every((p) => !p.test(code))) {
+        return result;
+      }
+
+      const transformed = plugin.transform.handler.call(transformContext, code, id);
+      return transformed ? { ...result, source: transformed.code } : result;
```
cieplypolar, line 46: As we discussed in person, source maps are discarded. They live on `transformed.map` property. You can verify this yourself
```ts
import install from './src/node.ts';

if (!import.meta.url.endsWith('?transformed')) {
  install();
  await import(`${import.meta.url}?transformed`);
} else {
  const shader = () => {
    'use gpu';
    return 42;
  };

  throw new Error('This error should point to line 12');
}
```
The good news is there is an easy fix (by my beloved clanker Astra)

```suggestion
      return transformed ? { ...result,   source: `${transformed.code}\n//# sourceMappingURL=${transformed.map.toUrl()}` } : result;
```

This aligns with the approach described here:
https://medium.com/the-node-js-collection/source-maps-in-node-js-5d0abe6ac34b#:~:text=/-,/%23%20sourceMappingURL%3Dtest.js.map,-This%20special%20comment

PR #3023, opened 2026-09-15
feat: Add source mapping functions

aleksanderkatan:
A node is mapped by wrapping it in `[SourceMapNodeType, line, column, node]`. 
I intend this to be an implementation detail, and for users to never embed/strip the source maps manually.


aleksanderkatan:
@pullfrog review

aleksanderkatan:
> packages/tinyest has no test:types script, so pnpm test:types (pnpm -r test:types) skips it entirely,

I don't think that's true

image

packages/tinyest/src/sourceMapping.ts
```diff
@@ -0,0 +1,82 @@
+import { type AnyNode, type SourceMap, type SourceMappedNode } from './nodes.ts';
```
iwoplaza:
```suggestion
import type { AnyNode, SourceMap, SourceMappedNode } from './nodes.ts';
```

packages/tinyest/src/nodes.ts
```diff
@@ -290,3 +290,18 @@ export type FuncParameter =
...
+ * despite them not appearing in NodeTypeCatalog.
+ */
+export type SourceMap = Map<AnyNode | readonly unknown[], [line: number, column: number]>;
+
+const sourceMappedBrand = Symbol();
+/**
+ * The intended way of stripping source maps is via using the `stripSourceMap` function.
+ * Guaranteed to be serializable via `JSON.stringify`.
+ */
+export type SourceMappedNode = { [sourceMappedBrand]: true };
```
iwoplaza, line 311: If I understand correctly, this is only a type-level marker, and the symbol doesn't actually exist as a property on source mapped nodes. If that's true, could we make this more obvious in the JS Doc? 🙏

iwoplaza (approved):
Very awesome implementation, makes total sense 👏

packages/tinyest/src/sourceMapping.ts
```diff
@@ -0,0 +1,82 @@
...
+  return [SourceMapNodeType, line, column, arr] as unknown as T;
+}
+
+/**
+ * Returns a new node with embedded source map.
+ *
+ * Only array nodes will be mapped, this means that
+ * identifier and boolean won't be mapped unless they are in array form.
+ *
+ * Use `stripSourceMap` to restore node and sourceMap.
```
cieplypolar:
```suggestion
 * Use {@link stripSourceMap} to restore node and sourceMap.
```

packages/tinyest/src/nodes.ts
```diff
@@ -290,3 +290,22 @@ export type FuncParameter =
...
+export type SourceMap = Map<AnyNode | readonly unknown[], [line: number, column: number]>;
+
+const sourceMappedBrand = Symbol();
+/**
+ * This type represents a node that has been source mapped.
+ *
+ * The brand is just a type-level marker, and does not exist in runtime.
+ * SourceMappedNode is guaranteed to be serializable via `JSON.stringify`.
+ *
+ * The intended way of stripping source maps is via the `stripSourceMap` function.
```
cieplypolar:
```suggestion
 * The intended way of stripping source maps is via the {@link stripSourceMap} function.
```

packages/tinyest/src/nodes.ts
```diff
@@ -290,3 +290,22 @@ export type FuncParameter =
...
       }[];
     };
+
+/**
+ * Map from an array/node to its source.
+ *
+ * Note that you can also map object properties, switch cases etc.,
+ * despite them not appearing in NodeTypeCatalog.
+ */
+export type SourceMap = Map<AnyNode | readonly unknown[], [line: number, column: number]>;
```
cieplypolar, line 300:
```suggestion
export type SourceMap = WeakMap<AnyNode | readonly unknown[], [line: number, column: number]>;
```

packages/tinyest/src/sourceMapping.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Use this function to split SourceMappedNode into AnyNode and SourceMap.
+ * It is not recommended to strip source maps in any other way.
+ */
+export function stripSourceMap(
+  node: SourceMappedNode | AnyNode,
+): [node: AnyNode, sourceMap: SourceMap] {
+  const sourceMap: SourceMap = new Map();
```
cieplypolar, line 52:
```suggestion
  const sourceMap: SourceMap = new WeakMap();
```

packages/tinyest/src/nodes.ts
aleksanderkatan: I would need to forbid strings and bools from this map.
```ts
export type SourceMap = WeakMap<(AnyNode & unknown[]) | readonly unknown[], [line: number, column: number]>;
```

I want to leave them so that we can later use `map.get` without any extra casts/checks.

packages/tinyest/src/sourceMapping.ts
aleksanderkatan: I cannot apply this without the suggestion above

packages/tinyest/src/nodes.ts
cieplypolar: I see, it's ok

cieplypolar (approved):

PR #3040, opened 2026-09-18
fix: tgpu.unroll removing ranges ending at zero 

vende11s:
before fix:

```typescript
  it('', () => {
    const fn = () => {
      'use gpu';
      let res = 0;
      for (const i of tgpu.unroll(std.range(3, 0, -1))) {
        res += i;
      }
      return res;
    };

      "fn fn_1() -> i32 {
        let res = 0;
        return res;
      }"
    `);
  });
```


cieplypolar (approved):
Nice catch 🚢 🇮🇹

aleksanderkatan (approved):
lesgooo

reczkok (approved):
yup

iwoplaza (approved):

PR #3041, opened 2026-09-19
impr: Rework orbit camera and update examples

reczkok:
`setupOrbitCamera` now passes the full camera state on every update, reusing the same object and computing derived matrices only when read

- pointer events for mouse/touch + cleanup when pointer capture is lost
- examples define their own camera uniforms and upload only what they use
- view-projection and its inverse are cached on the CPU instead of combined in shaders


iwoplaza (approved):
Looks great! Love the derived util 🌠

